PR #1287, opened 2015-06-25
Fix for pub crashing in checked mode

algobardo:
Pub crash in checked mode because spawnUri requires packageRoot to be a Uri, while here it can also be a String.


nex3:
Thanks for the fix!

PR #1291, opened 2015-07-06
fix https://github.com/dart-lang/pub/issues/1290

DisDis:


nex3:
Can you describe (preferable on #1290) what the underlying issue is and how to reproduce it? In order to fix it properly, we'll need to include a regression that that's fixed.

DisDis:
Reproduce steps:
https://github.com/dart-lang/pub/issues/1290#issuecomment-119004689

Issue #1250, opened 2015-06-05
dependency version ranges don't work for github dependencies


DartBot:
_Originally opened as dart-lang/sdk#22481_

*This issue was originally filed by trent.grove...&#064;workiva.com*
_____
**What steps will reproduce the problem?**

pubspec.yaml that lists a git dependency with a version range like this fails:

name: wService
version: 1.0.3
dependencies:
&nbsp;&nbsp;wTransport:
&nbsp;&nbsp;&nbsp;&nbsp;git: git@github.com:Workiva/wTransport.git
&nbsp;&nbsp;&nbsp;&nbsp;version: &quot;^1.0.0&quot;

with this error:

Working dir: /Volumes/CASE/code/wService/examples/dart
/usr/local/Cellar/dart/1.8.5/libexec/bin/pub get
Resolving dependencies...
Package wTransport has no versions that match &gt;=1.0.0 &lt;2.0.0 derived from:
- wService 1.0.3 depends on version ^1.0.0
Process finished with exit code 1

But if version is specified as &quot;&gt;=1.0.0&quot;, it works as expected and installs the 2.0.0 tagged release.

For some reason, Dart doesn't recognize that any of the 1.0.x tagged versions are available for install.
Any idea why that would be?  Is something beyond a git tag with release notes required for Dart to recognize the version?

**What is the expected output? What do you see instead?**

expect install of wTransport library at version 1.0.3

**What version of the product are you using?**

1.8.5

**On what operating system?**

OSX

**What browser (if applicable)?**

N/A

DartBot:
<a href="https://github.com/kasperl"><img src="https://avatars.githubusercontent.com/u/2156198?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [kasperl](https://github.com/kasperl)**

----

*Added [Area-Pub](../labels/Area-Pub), [Triaged](../labels/Triaged) labels.*

DartBot:
*This comment was originally written by john.rya...&#064;workiva.com*
_____
Maybe pub is doing this:

1. see that wService depends on wTransport version ^1.0.0 (shorthand for ">=1.0.0 &lt;2.0.0")
2. get the latest version of github.com:Workiva/wTransport.git
3. see that the latest version in pubspec.yaml is 2.0.0
4. fail because 2.0.0 does not satisfy ">=1.0.0 &lt;2.0.0"

I don't see anything in the documentation that leads me to believe pub is looking at git tags

[...]

Issue #1327, opened 2015-09-02
Pub: support some form of authorization when downloading packages


mit-mit:
_From @maxwellpeterson-wf on June 16, 2015 15:15_

Our organization has setup a fork of the [pub repository](https://github.com/dart-lang/pub-dartlang-dart) for our private/non-OSS packages. 

The current authorization flow for publishing packages to it works, however after perusing though the code, it looks like nothing similar exists for downloading packages with a `pub get`.

It would be great if pub could support authorization on a per-server basis.

_Copied from original issue: dart-lang/sdk#23652_

nex3:
You can set [the `PUB_HOSTED_URL` environment variable](https://www.dartlang.org/tools/pub/installing.html#optional-environment-variables) to configure where pub downloads dependencies from by default. You can also choose which package server they come from on a package-by-package basis by using a [full-form hosted dependency](https://www.dartlang.org/tools/pub/dependencies.html#hosted-packages).

Issue #1241, opened 2015-06-05
the env var PUB_HOSTED_URL seems to be not used by pub get


DartBot:
_Originally opened as dart-lang/sdk#22268_

*This issue was originally filed by benjamin.kim.ngu...&#064;gmail.com*
_____
1. Create a private pub server that can be reached at http://0.0.0.0:30000
2. publish something on it: pub publish --server http://0.0.0.0:30000
3. set the var env: PUB_HOSTED_URL=&quot;http://0.0.0.0:30000&quot;
4. use pub get on a package that need the package we published in the private server 

Expected:
pub get retrieve the package that is host in PUB_HOSTED_URL
Instead: 
using pub get -v there is no trace of the value used in PUB_HOSTED_URL

using: 
&nbsp;- gae 1.9.17 - 2014-12-08 (linux)
&nbsp;- https://github.com/dart-lang/pub-dartlang/releases/tag/v15
&nbsp;- dart sdk 1.8.5

Linux debian 8

DartBot:
<a href="https://github.com/anders-sandholm"><img src="https://avatars.githubusercontent.com/u/3276024?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [anders-sandholm](https://github.com/anders-sandholm)**

----

*Added [Area-Pub](../labels/Area-Pub), [Triaged](../labels/Triaged) labels.*

DartBot:
<a href="https://github.com/nex3"><img src="https://avatars.githubusercontent.com/u/188?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [nex3](https://github.com/nex3)**

----

This works for us; in fact, all our tests use this environment variable for all pub methods to avoid hitting the production server. Are you exporting the variable when you define it in your shell? You have to do that or programs you run won't be able to see it.
______
*Added [CannotReproduce](../labels/CannotReproduce) label.*

DartBot:
*This comment was originally written by benjamin.kim.ng...&#064;gmail.com*
_____
Hi

Thank you very much for your answer. Your resolve my problem.
After exporting, pub get look at the address set in PUB_HOSTED_URL and only this one.
But, I read somewhere that pub get check first the local/private server and then go to the public server is that right ?

Is there additional configuration I have to set in order to have this behaviour ?

Thanks again for your time !!!

DartBot:
*This comment was originally written by benjamin.kim.n...&#064;gmail.com*
_____
[...]

PR #1335, opened 2015-09-21
Add github version range support

computmaxer:
_I know this is not the proper way to submit contributions, but I want to get feedback from dart-lang folks and make sure it is something that could potentially be accepted before I spend any more time on this._

This is a fix for #1250. I work for Workiva (met some of you at the Dart Summit!) and we now have quite a few dart libraries in our Github organization. Some of these are public and OSS, but many of them cannot be. #1250 is a big issue for us for the private libraries as it requires us to update pubspecs all over the place anytime a patch version of some library is released, and consumers end up having a big list of dependency overrides while those transitions are in progress.

This PR implements version ranges for github dependencies with the following in a `pubspec.yaml`:

```yaml
dependencies:
  w_flux:
    version: "^1.0.0"
    git: git@github.com:Workiva/w_flux.git
```

It assumes that versions will exist as git tags on the repo, and uses `git tag -l` to list the tags. It then collects a list of the tags that are considered valid version strings using `Version.parse()`. 

The biggest issue I have had so far is dealing with git tags that include the "v" in the name, i.e. "v1.0.0". I can strip the "v" before giving it to the version solver, but then when it tries to get the ref (`_getRef`) it of course cannot find it without the "v" that was previously removed. At Workiva we do not use the preceding "v" in our tag names, but it seems like many people do so I am assuming this will be a big issue for this feature.

It seems like it would be possible for the Version class to accept "v" versions strings in its `parse` constructor, and keep track of it so that it can put it back when `toString()` is called. This would require a change to pub_semver, though.

I'm hoping to get some feedback from the Dart team on whether or not we should continue down this path. Please let me know your thoughts! 

@nex3 @kevmoo @munificent

### DIFF
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -308,7 +308,7 @@ class GitSource extends CachedSource {
       var result = await git.run(["rev-list", "--max-count=1", ref],
           workingDir: _repoCachePath(id));
       return result.first;
-    } on git.GitException catch (e) {
+    } on git.GitException {
       if (ref == id.version.toString()) {
         // Try again with a "v" before the ref in case this was a version tag
         ref = 'v$ref';
@@ -316,7 +316,7 @@ class GitSource extends CachedSource {
             workingDir: _repoCachePath(id));
         return result.first;
       }
-      throw e;
+      rethrow;
     }
   }
 
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -257,11 +257,11 @@ class GitSource extends CachedSource {
   ///
   /// Returns a future that completes with true if the repo was cloned, and
   /// false if the repo clone already exists.
-  Future<bool> _ensureRepo(PackageRef ref) async {
+  Future<bool> _ensureRepo(PackageRef ref, {mirror: false}) async {
     String cachePath = _repoCachePath(ref);
     if (!entryExists(cachePath)) {
       // Must have the repo cloned in order to list its tags
-      await _clone(_getUrl(ref), cachePath);
+      await _clone(_getUrl(ref), cachePath, mirror: mirror);
       return true;
     }
     return false;
@@ -275,7 +275,7 @@ class GitSource extends CachedSource {
   /// [id].
   Future<String> _ensureRevision(PackageId id) async {
     PackageRef packageRef = id.toRef();
-    if (await _ensureRepo(packageRef)) {
+    if (await _ensureRepo(packageRef, mirror: true)) {
       return _getRev(id);
     }
 
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -54,7 +54,7 @@ class GitSource extends CachedSource {
     await _ensureRepo(ref);
 
     var cachePath = _repoCachePath(ref);
-    List results = await git.run(["tag", "-l"], workingDir: cachePath);
+    List results = await git.run(['tag', '-l', '*.*.*'], workingDir: cachePath);
     List<Pubspec> validVersions = [];
     for (String version in results) {
       // Strip preceding 'v' character so 'v1.0.0' can be parsed into a Version
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -51,12 +51,9 @@ class GitSource extends CachedSource {
     }
 
     var ref = new PackageRef(name, this.name, _getDescription(description));
-    String cachePath = _repoCachePath(ref);
-    if (!entryExists(cachePath)) {
-      // Must have the repo cloned in order to list its tags
-      await _clone(_getUrl(ref), cachePath);
-    }
+    await _ensureRepo(ref);
 
+    var cachePath = _repoCachePath(ref);
     List results = await git.run(["tag", "-l"], workingDir: cachePath);
     List<Pubspec> validVersions = [];
     for (String version in results) {
@@ -255,34 +252,45 @@ class GitSource extends CachedSource {
     return new Pair(successes, failures);
   }
 
+  /// Ensure the canonical clone of the repository referred to by [ref] exists.
+  ///
+  /// Returns a future that completes with true if the repo was cloned, and
+  /// false if the repo clone already exists.
+  Future<bool> _ensureRepo(PackageRef ref) async {
+    String cachePath = _repoCachePath(ref);
+    if (!entryExists(cachePath)) {
+      // Must have the repo cloned in order to list its tags
+      await _clone(_getUrl(ref), cachePath);
+      return true;
+    }
+    return false;
+  }
+
   /// Ensure that the canonical clone of the repository referred to by [id] (the
   /// one in `<system cache>/git/cache`) exists and contains the revision
   /// referred to by [id].
   ///
   /// Returns a future that completes to the hash of the revision identified by
   /// [id].
-  Future<String> _ensureRevision(PackageId id) {
-    return new Future.sync(() {
-      var path = _repoCachePath(id.toRef());
-      if (!entryExists(path)) {
-        return _clone(_getUrl(id), path, mirror: true)
-            .then((_) => _getRev(id));
-      }
+  Future<String> _ensureRevision(PackageId id) async {
+    PackageRef packageRef = id.toRef();
+    if (await _ensureRepo(packageRef)) {
+      return _getRev(id);
+    }
 
-      // If [id] didn't come from a lockfile, it may be using a symbolic
-      // reference. We want to get the latest version of that reference.
-      var description = id.description;
-      if (description is! Map || !description.containsKey('resolved-ref')) {
-        return _updateRepoCache(id).then((_) => _getRev(id));
-      }
+    // If [id] didn't come from a lockfile, it may be using a symbolic
+    // reference. We want to get the latest version of that reference.
+    var description = id.description;
+    if (description is! Map || !description.containsKey('resolved-ref')) {
+      return _updateRepoCache(id).then((_) => _getRev(id));
+    }
 
-      // If [id] did come from a lockfile, then we want to avoid running "git
-      // fetch" if possible to avoid networking time and errors. See if the
-      // revision exists in the repo cache before updating it.
-      return _getRev(id).catchError((error) {
-        if (error is! git.GitException) throw error;
-        return _updateRepoCache(id).then((_) => _getRev(id));
-      });
+    // If [id] did come from a lockfile, then we want to avoid running "git
+    // fetch" if possible to avoid networking time and errors. See if the
+    // revision exists in the repo cache before updating it.
+    return _getRev(id).catchError((error) {
+      if (error is! git.GitException) throw error;
+      return _updateRepoCache(id).then((_) => _getRev(id));
     });
   }
 
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -76,7 +76,7 @@ class GitSource extends CachedSource {
       } on FormatException {}
     }
 
-    if (validVersions.length > 0) {
+    if (validVersions.isNotEmpty) {
       return validVersions;
     }
 
@@ -302,20 +302,22 @@ class GitSource extends CachedSource {
   /// by [id] on the effective ref of [id].
   ///
   /// This assumes that the canonical clone already exists.
-  Future<String> _getRev(PackageId id) {
+  Future<String> _getRev(PackageId id) async {
     var ref = _getEffectiveRef(id);
-    return git
-        .run(["rev-list", "--max-count=1", ref], workingDir: _repoCachePath(id))
-        .then((result) => result.first)
-        .catchError((e) {
+    try {
+      var result = await git.run(["rev-list", "--max-count=1", ref],
+          workingDir: _repoCachePath(id));
+      return result.first;
+    } on git.GitException catch (e) {
       if (ref == id.version.toString()) {
         // Try again with a "v" before the ref in case this was a version tag
         ref = 'v$ref';
-        return git.run(["rev-list", "--max-count=1", ref],
-            workingDir: _repoCachePath(id)).then((result) => result.first);
+        var result = await git.run(["rev-list", "--max-count=1", ref],
+            workingDir: _repoCachePath(id));
+        return result.first;
       }
       throw e;
-    });
+    }
   }
 
   /// Clones the repo at the URI [from] to the path [to] on the local
@@ -344,14 +346,13 @@ class GitSource extends CachedSource {
   }
 
   /// Checks out the reference [ref] in [repoPath].
-  Future _checkOut(String repoPath, String ref) {
-    return git.run(["checkout", ref], workingDir: repoPath)
-        .then((result) => null)
-        .catchError((_) {
+  Future _checkOut(String repoPath, String ref) async {
+    try {
+      await git.run(["checkout", ref], workingDir: repoPath);
+    } on git.GitException {
       // Try again with a "v" before the ref in case this was a version tag
-      return git.run(["checkout", 'v$ref'], workingDir: repoPath)
-          .then((result) => null);
-    });
+      await git.run(["checkout", 'v$ref'], workingDir: repoPath);
+    }
   }
 
   /// Use `git show` to get the pubspec.yaml at a particular ref,
@@ -359,14 +360,14 @@ class GitSource extends CachedSource {
   ///
   /// It is possible that a pubspec didn't always exist, return null if
   /// that is the case.
-  Future<Pubspec> _getPubspec(String repoPath, String ref) {
-    return git.run(['show', '$ref:pubspec.yaml'], workingDir: repoPath)
-        .catchError((_) => null)
-        .then((result) {
-      if (result != null) {
-        return new Pubspec.parse(result.join('\n'), systemCache.sources);
-      }
-    });
+  Future<Pubspec> _getPubspec(String repoPath, String ref) async {
+    try {
+      var result = await git.run(['show', '$ref:pubspec.yaml'],
+          workingDir: repoPath);
+      return new Pubspec.parse(result.join('\n'), systemCache.sources);
+    } on git.GitException {
+      return null;
+    }
   }
 
   /// Returns the path to the canonical clone of the repository referred to by
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -56,16 +56,17 @@ class GitSource extends CachedSource {
     var cachePath = _repoCachePath(ref);
     List results = await git.run(['tag', '-l', '*.*.*'], workingDir: cachePath);
     List<Pubspec> validVersions = [];
-    for (String version in results) {
+    for (String tag in results) {
       // Strip preceding 'v' character so 'v1.0.0' can be parsed into a Version
-      if (version.startsWith('v')) {
-        version = version.substring(1);
+      String versionTag = tag;
+      if (versionTag.startsWith('v')) {
+        versionTag = versionTag.substring(1);
       }
       try {
         // Use Version.parse to determine valid version tags
-        Version validVersion = new Version.parse(version);
+        new Version.parse(versionTag);
         // Fetch the pubspec for this version
-        Pubspec pubspec = await _getPubspec(cachePath, validVersion.toString());
+        Pubspec pubspec = await _getPubspec(cachePath, tag);
         if (pubspec != null) {
           validVersions.add(pubspec);
         }
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -55,7 +55,7 @@ class GitSource extends CachedSource {
 
     var cachePath = _repoCachePath(ref);
     List results = await git.run(['tag', '-l', '*.*.*'], workingDir: cachePath);
-    List<Pubspec> validVersions = [];
+    Map<String, Pubspec> validVersionPubspecs = {};
     for (String tag in results) {
       // Strip preceding 'v' character so 'v1.0.0' can be parsed into a Version
       String versionTag = tag;
@@ -67,14 +67,20 @@ class GitSource extends CachedSource {
         new Version.parse(versionTag);
         // Fetch the pubspec for this version
         Pubspec pubspec = await _getPubspec(cachePath, tag);
+        // Skip this version if a pubspec didn't exist
         if (pubspec != null) {
-          validVersions.add(pubspec);
+          // This logic prevents duplicate versions and prefers the non-"v"
+          // tag in the case of "1.0.0" and "v1.0.0"
+          if (!validVersionPubspecs.containsKey(versionTag)
+              || !tag.startsWith('v')) {
+            validVersionPubspecs[versionTag] = pubspec;
+          }
         }
       } on FormatException {}
     }
 
-    if (validVersions.isNotEmpty) {
-      return validVersions;
+    if (validVersionPubspecs.isNotEmpty) {
+      return validVersionPubspecs.values.toList();
     }
 
     // No valid version tags were found, defer to super
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -50,12 +50,11 @@ class GitSource extends CachedSource {
       return super.getVersions(name, description);
     }
 
-    PackageId id =
-        new PackageId(name, this.name, null, _getDescription(description));
-    String cachePath = _repoCachePath(id);
+    var ref = new PackageRef(name, this.name, _getDescription(description));
+    String cachePath = _repoCachePath(ref);
     if (!entryExists(cachePath)) {
       // Must have the repo cloned in order to list its tags
-      await _clone(_getUrl(id), cachePath);
+      await _clone(_getUrl(ref), cachePath);
     }
 
     List results = await git.run(["tag", "-l"], workingDir: cachePath);
@@ -112,7 +111,8 @@ class GitSource extends CachedSource {
     await _ensureRevision(id);
     var revisionCachePath = getDirectory(await resolveId(id));
     if (!entryExists(revisionCachePath)) {
-      await _clone(_repoCachePath(id), revisionCachePath, mirror: false);
+      var cachePath = _repoCachePath(id.toRef());
+      await _clone(cachePath, revisionCachePath, mirror: false);
     }
 
     var ref = _getEffectiveRef(id);
@@ -263,7 +263,7 @@ class GitSource extends CachedSource {
   /// [id].
   Future<String> _ensureRevision(PackageId id) {
     return new Future.sync(() {
-      var path = _repoCachePath(id);
+      var path = _repoCachePath(id.toRef());
       if (!entryExists(path)) {
         return _clone(_getUrl(id), path, mirror: true)
             .then((_) => _getRev(id));
@@ -291,7 +291,7 @@ class GitSource extends CachedSource {
   ///
   /// This assumes that the canonical clone already exists.
   Future _updateRepoCache(PackageId id) {
-    var path = _repoCachePath(id);
+    var path = _repoCachePath(id.toRef());
     if (_updatedRepos.contains(path)) return new Future.value();
     return git.run(["fetch"], workingDir: path).then((_) {
       _updatedRepos.add(path);
@@ -306,14 +306,14 @@ class GitSource extends CachedSource {
     var ref = _getEffectiveRef(id);
     try {
       var result = await git.run(["rev-list", "--max-count=1", ref],
-          workingDir: _repoCachePath(id));
+          workingDir: _repoCachePath(id.toRef()));
       return result.first;
     } on git.GitException {
       if (ref == id.version.toString()) {
         // Try again with a "v" before the ref in case this was a version tag
         ref = 'v$ref';
         var result = await git.run(["rev-list", "--max-count=1", ref],
-            workingDir: _repoCachePath(id));
+            workingDir: _repoCachePath(id.toRef()));
         return result.first;
       }
       rethrow;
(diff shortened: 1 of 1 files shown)


nex3:
I don't think you need to do that many backflips to get "v" versions working. Just add logic to `GitSource._getEffectiveRef` to check for refs named `"v$version"` and `"$version"`. That way you can have everything else just track the version number as part of the package ID, like it does for hosted dependencies.

computmaxer:
Sorry for the delay! Finally had some time to spend on this yesterday. I've implemented @nex3 's suggestion for dealing with "v" version tags.

I also fixed a major issue with my initial approach. I was creating bare-bones `Pubspec` objects with just `name` and `version` in `getVersions()`, but it turns out those pubspecs need to be complete since one will eventually be used to determine which nested dependencies to get. So unfortunately we'll need to return the entire pubspec for each potential valid version. I have implemented that here by calling `describeUncached()` with each valid version string in a `PackageId`. This has the potential to impact performance since that will checkout each tag and load a pubspec, serially.

To help combat that issue I added a flag to enable this feature per dependency. Looks like this:

```yaml
dependencies:
  w_flux:
    version: "^1.0.0"
    git: 
      url: git@github.com:Workiva/w_flux.git
      use_version_tags: true
```

Pub will only `getVersions()` for a particular git dependency if `use_version_tags` is true. This has the additional benefit of insulating this feature from existing tests and functionality, reducing its impact.

---

All unit tests are passing and I have tested this with a number of repositories, with and without "v"s in the version tags. Let me know what you think and what the next steps would be if you want to move forward with this solution.

greglittlefield-wf:
If checking out each tag is slow, would using [`git show`](https://git-scm.com/docs/git-show) (e.g., `git show TAGNAME:pubspec.yaml`) instead of `git checkout` help to significantly reduce that performance hit?

computmaxer:
@greglittlefield-wf nice! Didn't know about `git show`. I've added a commit that uses that instead of `describeUncached()`.

kevmoo:
@computmaxer If you're fixing your pull request, please rebase against master and do a `--force` push to your staging branch. This makes the pull request "clean".

lib/src/source/git.dart
```diff
@@ -40,6 +41,49 @@ class GitSource extends CachedSource {
...
+        Version validVersion = new Version.parse(version);
+        // Fetch the pubspec for this version
+        Pubspec pubspec = await _getPubspec(cachePath, validVersion.toString());
+        if (pubspec != null) {
+          validVersions.add(pubspec);
+        }
+      } on FormatException {}
+    }
+
+    if (validVersions.length > 0) {
```
kevmoo: use `.isNotEmpty`

lib/src/source/git.dart
```diff
@@ -254,8 +303,19 @@ class GitSource extends CachedSource {
   ///
   /// This assumes that the canonical clone already exists.
   Future<String> _getRev(PackageId id) {
-    return git.run(["rev-list", "--max-count=1", _getEffectiveRef(id)],
-        workingDir: _repoCachePath(id)).then((result) => result.first);
+    var ref = _getEffectiveRef(id);
```
kevmoo: Consider making this method `async`

lib/src/source/git.dart
```diff
@@ -285,8 +345,28 @@ class GitSource extends CachedSource {
 
   /// Checks out the reference [ref] in [repoPath].
   Future _checkOut(String repoPath, String ref) {
-    return git.run(["checkout", ref], workingDir: repoPath).then(
-        (result) => null);
+    return git.run(["checkout", ref], workingDir: repoPath)
```
kevmoo: ditto: consider making async

lib/src/source/git.dart
```diff
@@ -285,8 +345,28 @@ class GitSource extends CachedSource {
...
+    });
+  }
+
+  /// Use `git show` to get the pubspec.yaml at a particular ref,
+  /// then parse it into a Pubspec object
+  ///
+  /// It is possible that a pubspec didn't always exist, return null if
+  /// that is the case.
+  Future<Pubspec> _getPubspec(String repoPath, String ref) {
+    return git.run(['show', '$ref:pubspec.yaml'], workingDir: repoPath)
```
kevmoo: ditto: consider async

computmaxer:
@kevmoo comments addressed. Would you all commits squashed into one as well?

lib/src/source/git.dart
```diff
@@ -253,9 +302,20 @@ class GitSource extends CachedSource {
...
   ///
   /// This assumes that the canonical clone already exists.
-  Future<String> _getRev(PackageId id) {
-    return git.run(["rev-list", "--max-count=1", _getEffectiveRef(id)],
-        workingDir: _repoCachePath(id)).then((result) => result.first);
+  Future<String> _getRev(PackageId id) async {
+    var ref = _getEffectiveRef(id);
+    return git
+        .run(["rev-list", "--max-count=1", ref], workingDir: _repoCachePath(id))
+        .then((result) => result.first)
```
kevmoo: By async, I also meant getting red of `.then` and `.catchError` and using `await`, too. :smile:

lib/src/source/git.dart
```diff
@@ -284,9 +344,29 @@ class GitSource extends CachedSource {
   }
 
   /// Checks out the reference [ref] in [repoPath].
-  Future _checkOut(String repoPath, String ref) {
-    return git.run(["checkout", ref], workingDir: repoPath).then(
-        (result) => null);
+  Future _checkOut(String repoPath, String ref) async {
```
kevmoo: ditto

kevmoo:
@computmaxer I'd wait to squash until reviews are done – but it's good to do once you think you're done

lib/src/source/git.dart
computmaxer: Ok. Do you guys _always_ prefer async/await over future chaining, then?

kevmoo: In general, yes. It's easier to read and our debugger actually has special support for async/await – that we don't get if you use the Future API directly.

@nex3 has found a few cases where it's better to use the API directly, but she'd have to comment.

lib/src/source/git.dart
computmaxer: Fixed.

lib/src/source/git.dart
```diff
@@ -253,9 +302,22 @@ class GitSource extends CachedSource {
...
+      return result.first;
+    } on git.GitException catch (e) {
+      if (ref == id.version.toString()) {
+        // Try again with a "v" before the ref in case this was a version tag
+        ref = 'v$ref';
+        var result = await git.run(["rev-list", "--max-count=1", ref],
+            workingDir: _repoCachePath(id));
+        return result.first;
+      }
+      throw e;
```
kevmoo: Use `rethrow` – it preserves the stack trace

computmaxer: nice :+1:

lib/src/source/git.dart
```diff
@@ -40,6 +41,49 @@ class GitSource extends CachedSource {
     });
   }
 
+  /// Gets the list of all versions from git tags.
+  /// Anything that Version.parse understands is considered a version,
+  /// it will also attempt to strip off a preceding 'v' e.g. v1.0.0
+  Future<List<Pubspec>> getVersions(String name, description) async {
+    if (!_useVersionTags(description)) {
```
nex3: I'm really not a fan of requiring users to explicitly annotate to get this behavior. For the most part, I don't expect people to be using version declarations for Git dependencies at all unless they want some sort of resolution, so requiring them to write more stuff seems annoying. And now that we're using `git show`, it shouldn't be too bad to get all the pubspecs.

lib/src/source/git.dart
```diff
@@ -40,6 +41,49 @@ class GitSource extends CachedSource {
...
+      // No need to get versions if the useVersionTags option is false
+      return super.getVersions(name, description);
+    }
+
+    PackageId id =
+        new PackageId(name, this.name, null, _getDescription(description));
```
nex3: Passing `null` for the version here is a smell. It looks like `_repoCachePath` doesn't actually require a version, so it should be modified to take a `PackageRef` instead of a `PackageId`. You can update other call sites by calling `PackageId.toRef()` to get the ref.

lib/src/source/git.dart
```diff
@@ -40,6 +41,49 @@ class GitSource extends CachedSource {
...
+    String cachePath = _repoCachePath(id);
+    if (!entryExists(cachePath)) {
+      // Must have the repo cloned in order to list its tags
+      await _clone(_getUrl(id), cachePath);
+    }
```
nex3: This is the same logic as in `_ensureRevision`; consider factoring it out into `String _ensureRepo(PackageRef ref)`.

lib/src/source/git.dart
```diff
@@ -40,6 +41,49 @@ class GitSource extends CachedSource {
...
+
+    List results = await git.run(["tag", "-l"], workingDir: cachePath);
```
nex3: If you pass a pattern to `--list`, you can cut down on the number of irrelevant tags you have to look at. The pattern `*.*.*` will match all valid semantic versions with or without the leading `v` (although obviously it will also match a lot of other stuff).

lib/src/source/git.dart
```diff
@@ -40,6 +41,49 @@ class GitSource extends CachedSource {
...
+    List<Pubspec> validVersions = [];
+    for (String version in results) {
```
nex3: Consider using `Iterable.where` here.

lib/src/source/git.dart
```diff
@@ -40,6 +41,49 @@ class GitSource extends CachedSource {
...
+      // Strip preceding 'v' character so 'v1.0.0' can be parsed into a Version
+      if (version.startsWith('v')) {
+        version = version.substring(1);
+      }
+      try {
+        // Use Version.parse to determine valid version tags
+        Version validVersion = new Version.parse(version);
+        // Fetch the pubspec for this version
+        Pubspec pubspec = await _getPubspec(cachePath, validVersion.toString());
```
nex3: `validVersion.toString()` here will break if the actual tag started with a `v`.

lib/src/source/git.dart
```diff
@@ -40,6 +41,49 @@ class GitSource extends CachedSource {
...
+        if (pubspec != null) {
+          validVersions.add(pubspec);
+        }
+      } on FormatException {}
```
nex3: We should definitely catch format exceptions from the version parse, but I don't think we should catch errors caused by an invalid pubpsec. It's a pretty safe assumption that once we're here, the user intentionally opted into version resolution over a given repo, and having some versions silently not appear may be confusing.

lib/src/source/git.dart
```diff
@@ -40,6 +41,49 @@ class GitSource extends CachedSource {
...
+      // Must have the repo cloned in order to list its tags
+      await _clone(_getUrl(id), cachePath);
+    }
+
+    List results = await git.run(["tag", "-l"], workingDir: cachePath);
+    List<Pubspec> validVersions = [];
+    for (String version in results) {
+      // Strip preceding 'v' character so 'v1.0.0' can be parsed into a Version
+      if (version.startsWith('v')) {
+        version = version.substring(1);
```
nex3: This should do some de-duplication to ensure that, if a version exists both with and without `v`, only the canonical one (that is, the one without `v`) is returned.

lib/src/source/git.dart
```diff
@@ -253,9 +302,22 @@ class GitSource extends CachedSource {
   /// by [id] on the effective ref of [id].
   ///
   /// This assumes that the canonical clone already exists.
-  Future<String> _getRev(PackageId id) {
-    return git.run(["rev-list", "--max-count=1", _getEffectiveRef(id)],
-        workingDir: _repoCachePath(id)).then((result) => result.first);
+  Future<String> _getRev(PackageId id) async {
```
nex3: If you're changing `_getRev` rather than `_getEffectiveRef` (which makes sense), you'll need to update `downloadToSystemCache` to use the former so that it behaves properly with `v`-versions.

lib/src/source/git.dart
```diff
@@ -253,9 +302,22 @@ class GitSource extends CachedSource {
...
+    var ref = _getEffectiveRef(id);
```
nex3: `_getEffectiveRef` doesn't really make much sense in the new world. It returns `"HEAD"` if the ID doesn't specify a revision, but that's incorrect if version resolution needs to be done. I think what you want to do here is more like:

```dart
Future<String> _getRev(PackageId id) {
  var description = _getDescription(id);

  // If the id is already resolved, there's no need to rev-list to look for it.
  var resolved = description['resolved-ref'];
  if (description is Map && resolved != null) return resolved;

  // If [id] includes a version and not a ref, try "rev-list $version" and
  // "rev-list v$version". Otherwise, try "rev-list ${_getRef(id) ?? 'HEAD'}".
}
```

lib/src/source/git.dart
nex3: Note that you won't need to modify this if you change `downloadToSystemCache` to use `_getRef`.

nex3:
@computmaxer Done reviewing!

lib/src/source/git.dart
computmaxer: The issue I'm having here is that I can't seem to access the `version` in this method, as it is not part of the description.

```
w_transport:
  version: ^1.0.0
  git: git@github.com:Worvia/w_transport.git
```
> description --> "git@github.com:Workiva/w_transport.git"

```
w_transport:
  version: ^1.0.0
  git: 
    url: git@github.com:Worvia/w_transport.git
    use_version_tags: true
```
> description --> {'url': 'git@github.com:Worvia/w_transport.git', 'use_version_tags': true}

Any idea how to short-circuit based on the `version`?

lib/src/source/git.dart
computmaxer: Addressed this in https://github.com/computmaxer/pub/commit/1a9120c4f67baf65f026c98fd282b4bfb9d0499b

lib/src/source/git.dart
computmaxer: Addressed in https://github.com/dart-lang/pub/commit/3d9c0abe7d8e774e1f7aca1765e9cf6e0ef17891

lib/src/source/git.dart
computmaxer: Addressed in https://github.com/computmaxer/pub/commit/59f2e6f28fdad7bdbe272a72e78f532c9ad27537

lib/src/source/git.dart
computmaxer: Consider the case where a repo initially was tagged 1.0.0 without a pubspec (perhaps it used to be in JS or something ¯\_(ツ)_/¯ ). But all tags after 2.0.0 have a pubspec.
If we don't handle invalid pubspec here, this will cause an exception attempting to load the pubspecs for versions <2.0.0 and they won't be able to use version resolution at all for that given repo.

lib/src/source/git.dart
computmaxer: Addressed in https://github.com/dart-lang/pub/commit/8b1aabdd4396e8f2b7147d4b7f4fc8f4f94eda9d

lib/src/source/git.dart
computmaxer: Addressed in https://github.com/dart-lang/pub/commit/01e6787b1bd9bdc5fec877ee288241810a747da4

kevmoo: Thanks for linking to the commit w/ the fix. Makes these reviews so much better. :smile:

lib/src/source/git.dart
computmaxer: I'm having an issue with this. Here's what the code roughly looks like:

```dart
Future<String> _getRev(PackageId id) {
  print('_getRev called: ${id.toString()}');
  ...

  var ref = description is Map ? description['ref'] : null;
  if (ref == null && id.version != null && id.version != Version.none) {
    print('Trying to use version: ${id.version.toString()}');
   // return the rev-list $version result
  } else {
    print('Returning _getRef(id) or HEAD');
    // return rev-list _getRef(id) ?? 'HEAD';
  }
}
```

For some reason `id.version` ends up getting set to something even if no `version` is specified in the pubspec. So if I put some print statements in this `_getRev` method the output looks like this when I run a simple test like `test/get/git/check_out_test.dart`

```
  | MSG : Resolving dependencies...
  | _getRev called: foo 0.0.0 from git
  | Returning _getRef(id) or HEAD
  | _getRev called: foo 0.0.0 from git
  | Returning _getRef(id) or HEAD
  | _getRev called: foo 0.0.0 from git
  | Returning _getRef(id) or HEAD
  | _getRev called: foo 0.0.0 from git
  | Returning _getRef(id) or HEAD
  | _getRev called: foo 1.0.0 from git
  | Trying to use version: 1.0.0
```
That's with a description of `foo: "../foo.git"` (no version specified). So by the logic in the if condition, it will try to get a revision with `1.0.0` which fails as it should. It should have hit the `else` case instead for this to work properly.

If I leverage `_useVersionTags(id)` in that if condition I can correctly determine whether or not the version should be used as a ref. This is what the existing code in `_getEffectiveRef` is doing. However, you mentioned you don't want to rely on an additional annotation for this to work.

I will keep looking for a solution, but let me know if you have any ideas that will solve this. Then I should be able to get rid of the `_useVersionTags(id)` stuff.

lib/src/source/git.dart
computmaxer: Addressed here: https://github.com/dart-lang/pub/commit/c9353549715b634bd59a9445c462ee612985be5d

lib/src/source/git.dart
computmaxer: This would require an additional loop - the `where` which selects the valid version strings, and another that loops through the valid version strings and awaits the pubspec for each.

I sort of prefer it as one loop, but if you would rather have that separation I can change it.

computmaxer:
@nex3 I've addressed each of your comments, some with further questions.

computmaxer:
@greglittlefield-wf @travissanderson-wf @trentgrover-wf @evanweible-wf Perhaps you guys have some ideas for some of these remaining issues?

lib/src/source/git.dart
greglittlefield-wf: It'd be possible to access `version` if it was in the description under `git`, like you initially had it, right?
```yaml
dependencies:
  w_flux:
    version: "^1.0.0"
    git: git@github.com:Workiva/w_flux.git
```

Is there any reason that version can't be under `git`?

computmaxer: The version solver (which is abstracted from the package source) won't respect it when it is under `git`.

lib/src/source/git.dart
computmaxer: `id.version` ends up being set to the version number in the pubspec.yaml after the solver has selected a pubspec: https://github.com/dart-lang/pub/blob/master/lib/src/solver/version_solver.dart#L257

`resolveId` is called once again after a solution has been reached: https://github.com/dart-lang/pub/blob/master/lib/src/solver/backtracking_solver.dart#L168. `resolveId` uses `_ensureRevision` which uses `_getRev`.

At this point `id.version` has something even if one wasn't specified in the consumer's pubspec.

So this is why `if (ref == null && id.version != null && id.version != Version.none)` is not a sufficient condition to determine whether or not we should do a `rev-list $version`.

Open to suggestions :)

lib/src/source/git.dart
nex3: Yikes, good point. I'd just change `Source.getVersions()` to take a `PackageRef` rather than a name and description, and check whether the version constraint is `any` or not.

lib/src/source/git.dart
nex3: Yeah, I forgot that `PackageId` doesn't include the original version constraint because it's not part of the description. This is going to be trickier than I thought.

The underlying issue is that Git IDs are no longer context-independent in the way they used to be; we don't know whether one was created due to new-style version resolution or old-style looking at `HEAD`. Git does have a means of "resolving" package IDs, but because `getVersions()` returns `Pubspec`s rather than `PackageId`s, those resolved IDs don't make their way into the solver.

This probably calls for a full refactor of `getVersions()` that changes the signature to `List<PackageId> getVersions(PackageRef ref)` and moves some caching logic into the hosted source. This is out of the scope of what you're doing, though, so I'll implement it myself; hold tight until then.

lib/src/source/git.dart
nex3: I think with my refactor, we'll be able to avoid trying to load pubspecs until they're actually required by the version solver, which should mitigate this.

lib/src/source/git.dart
computmaxer: Do you mean `PackageDep` instead of `PackageRef`? `PackageRef` does not have a `VersionConstraint` available.

nex3:
I'm really really sorry for pulling the rug out from this patch at the last minute, but the more I think about its UI and the more I talk to @rnystrom, the less I think we'll be able to get it to work. On one hand, I don't think we're going to be able to trigger this based on a version constraint existing; that information just doesn't get piped where it would need to go, and it doesn't make sense in general to add that piping. But on the other hand, we really don't want to have to hide this behind a cumbersome flag that would have to go on every dependency. So I don't think it's going to make it. Again, I'm sorry I didn't think this through thoroughly earlier on.

trentgrover-wf:
The main problem that prompted this effort is that there doesn't seem to be a complete and secure mechanism for version solving with private Dart packages via pub.

The 2 main mechanisms for consuming dependencies are:

1) Pub server

The default pub server works to solve dependency constraints perfectly for public packages only.  Using this functionality with private dart libraries requires standing up and maintaining a private pub server.  There are also some functional gaps in the existing implementation:
  - There's no auth mechanism for pub get (issue raised and closed here: https://github.com/dart-lang/pub/issues/1327).
  - Setting the PUB_HOSTED_URL environment variable doesn't automatically fall back to the normal pub server if the custom server doesn't contain the package requested (mentioned here: https://github.com/dart-lang/pub/issues/1241).
  - Using the full form hosted dependency syntax would resolve the PUB_HOSTED_URL fallback issue, but requires more cumbersome per dependency syntax.

2) GitHub repos

The existing GitHub repo functionality in pub supports private pub packages well in a secure way, but contains no version constraint solving (hence this attempt at a solution), which makes dependency management across complex dependency trees extremely cumbersome.
  - Given the two avenues available to us, it seems easier and more useful to a wide range of consumers if we can get the constraint solving of GitHub repos implemented (auth already works, GitHub can provide the infrastructure / server, and fallback to existing public pub is transparent and automatic).
  - Even hiding the GitHub range support behind some cumbersome per-dependency flag seems no worse than the existing full form hosted dependency syntax.

If there are significant reasons why GitHub version solving isn't technically feasible, it would be really helpful to have a more straightforward way to accomplish the same goals via private pub packages.  Is there anything we can do to get something workable via 1 of these 2 paths?

travissanderson-wf:
+1, our team is currently in dependency_overrides hell to work around the issue, we'd definitely be happy to add a flag to every dependency for this solution

kevmoo:
@travissanderson-wf @trentgrover-wf I'll chat with @nex3 to see what our options are – reading https://github.com/dart-lang/pub/issues/1327, it sounds like you'd be open to running your own server

computmaxer:
@kevmoo we are open to it, but would rather have support for Git version solving to avoid maintenance headaches and the auth hurdle. 

> I don't think we're going to be able to trigger this based on a version constraint existing; that information just doesn't get piped where it would need to go, and it doesn't make sense in general to add that piping.

Why doesn't it make sense to add that piping? It seems like `getVersions` would just need to be given a `PackageDep` instead of a `PackageRef`. Looking briefly at the code, it seems like that might not be a trivial change (may affect a lot of the version solver code), but wouldn't it make sense to have constraint information when solving the version? How does that become a blocker for this feature? Pub already has the design/architecture to support version solving across different package sources, but only hosted packages are actually able to use it. Doesn't it make sense to refactor that functionality slightly to allow the design to be utilized?

trentgrover-wf:
@kevmoo any further thoughts on this issue?

travissanderson-wf:
@kevmoo @nex3 I already gave @computmaxer and his team breakfast pizza to thank him for getting this fixed for us, as you know that is a morally binding contract so he is feeling a lot of heat right now since we still can't use version ranges on github dependencies :)

nex3:
If you're looking to only allow authenticated downloads, I think your best bet is to either run some sort of VPN or to use HTTP basic auth and have people include their credentials in the `PUB_HOSTED_URL` variable. Either way you'll also need to set up your server to fall back on `pub.dartlang.org` packages.

trentgrover-wf:
So is there no chance of a GitHub based solution (eliminates a bunch of infrastructure requirements)?

If you could point us to your specific areas of concern, we're more than happy to devote some dev time to it, but it's sounding like the NO is hardening...

nex3:
I don't think we're going to find a way to make version range resolution work with GitHub, no.

johncblandii:
That kind of sucks to go through so much PR review/work to say "no" to the whole idea of it later.

The feature is useful. Hope you reconsider @nex3.

kevmoo:
@johncblandii It absolutely sucks. We're very sorry.

We try **very** hard to make sure these types of changes don't have long-term implications or side-effects that can hurt the ecosystem.

@nex3 is especially diligent when it comes to these types of changes – although, as she mentioned, we realized there were gnarly edge cases we didn't initially consider when we first encouraged the work to proceed.

We'd be more than happy to discuss other approaches to this problem – we know path dependencies are a pain.

travissanderson-wf:
Describing it as a pain is a bit of an understatement.  We have a number of in-house libraries written in Dart and a handful of apps consuming those libraries.  Our options today are either move everything in lockstep (dependencies of dependencies included) or to use a horrific set of dependency_overrides.  The other prescribed option of running a private server with VPN access required puts an unnecessary burden of access control and infrastructure on the end-users of the Dart language.  None of the other services we use require restricted network access, we do not generally use a VPN and can work with all our other "cloud" services from anywhere.

It's a little demoralizing as a company that has evangelized Dart and is offering to even do the work ourselves.  I'm sure we're not alone with this being a major difficulty for anyone else trying to use Dart in a medium-large scale company.  We would be elated to even use the "cumbersome" flag on every dependency.  Our team alone has 40 lines (not exaggerating) of dependency_overrides entries in our primary app - it's almost a mirror of the dependencies section.

johnryan-wf:
In hindsight, knowing this was a dead end would have allowed us to work on an internal pub server.  

We want to make sure that whatever solution we end up with is solid, but I'm not sure why this can't be accomplished (in principle.)  It *seems* like an elegant way to support private dependencies.  So close :)

kevmoo:
I'll chat w/ @nex3 a bit more tomorrow. We know git (and GitHub) is a really close fit here.

Maybe we can get creative.

travissanderson-wf:
Thanks, @kevmoo @nex3.  For what it's worth, if you identify some work you need done we'd be happy to provide some bandwidth.

trentgrover-wf:
@kevmoo @nex3 any further update?  sorry to badger you

munificent:
Various holidays and related vacations are kicking in, so @kevmoo and @nex3 might be offline for a while, but they will follow up on this. :)

trentgrover-wf:
:+1:

travissanderson-wf:
Have a good holiday vacation, all!  See you when you get back.

darty-bird-xmas

trentgrover-wf:
Now that the holidays have passed, any further update @kevmoo @nex3?  We're trying to nail down how we want to address this issue internally, so your answer will help us determine what resources to devote to it.

trentgrover-wf:
For posterity, here's where we landed:
- Google will not be merging GitHub version range support because of the issues noted above and because the risk associated with introducing breaking changes this low in the stack is too high.
- We're going with the private pub server route:
  - Google has committed to adding some oauth capability to pub so that `pub get` can be performed in a secure way without requiring the auth credentials to reside in the pubspec (as would be required if using only HTTP basic auth).  Delivery date TBD.
  - Once we have a private pub server up and running with oauth, we'll use the hosted package syntax with a custom hosted url specified for each of our private pubspec dependencies.

munificent:
Sounds right to me. Just for the record (and you probably already know this):

> * Once we have a private pub server up and running with oauth, we'll use the hosted package syntax with a custom hosted url specified for each of our private pubspec dependencies.

This is supported already in pub, so there's no work on our end to do. It looks like:

```yaml
dependencies:
  transmogrify:
    hosted:
      name: transmogrify
      url: http://your-package-server.com
    version: ^1.4.0
```

computmaxer:
@kevmoo @munificent @nex3 
We planned on firewalling off our private pub server from the internet and only allowing our office/VPN IPs until the OAuth2 support for `pub get` is added.

I am having trouble getting this working with the Managed VM GAE deployment. I've added the firewall rule to allow our office IPs in Networking --> Firewall rules, but I can't seem to block access to everything else. I believe this would work if it was just a GCE deployment, but it seems GAE must add some rules or routes to allow traffic to be routed from the internet to the instances.

Do you guys know if it is possible to implement an IP whitelist for a GAE Managed VMs deployment?

nex3:
I'm far from an expert on AppEngine infrastructure, and I think the same is true for Kevin and Bob. Maybe @mkustermann would know; he works on pub.dartlang.org.

kevmoo:
Try https://groups.google.com/forum/#!forum/app-engine-managed-vms

This is purely outside of the Dart team

computmaxer:
@nex3 @kevmoo 
Do you all have any updates on the additional OAuth2 functionality for pub? We are yet to have a functioning private pub server approved for production use (I'll spare you the details :smile:), and since we are approaching the end of Q1 I figured I'd check in to see where this is at. Thanks!

nex3:
I haven't had a chance to work on this yet. I'll try to get it in the works next quarter.

PR #1352, opened 2015-10-27
Clarify that new users need submodules before they run the tests

sethladd:


nex3:
The existing phrasing is correct: most tests don't require archived barback versions. I'd prefer this go between the instructions for running a single test and the instructions for running all tests.

PR #1383, opened 2016-01-15
Add a `pub cache path` command for printing the path

eseidelGoogle:
Without this there is no easy way to get the path
pub is uses for its cache (which seems like a very
reasonable thing for the tool to be able to tell
the user).

Happy to expose this in a different way.

### DIFF
--- a/lib/src/command/cache_path.dart
+++ b/lib/src/command/cache_path.dart
@@ -1,0 +1,18 @@
+// Copyright (c) 2013, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../command.dart';
+
+/// Handles the `cache path` pub command.
+class CachePathCommand extends PubCommand {
+  String get name => "path";
+  String get description => "Prints the location of the pub cache";
+  String get invocation => "pub cache path";
+  bool get hidden => true;
+  bool get takesArguments => false;
+
+  void run() {
+    print(cache.rootDir);
+  }
+}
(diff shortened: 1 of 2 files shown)


lib/src/command/cache_path.dart
```diff
@@ -0,0 +1,18 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import '../command.dart';
+
+/// Handles the `cache path` pub command.
+class CachePathCommand extends PubCommand {
+  String get name => "path";
+  String get description => "Prints the location of the pub cache";
+  String get invocation => "pub cache path";
+  bool get hidden => true;
```
eseidelGoogle: Arguably this should not be hidden, up to you all.

eseidelGoogle:
Other packaging systems expose this in other ways:

```
%npm cache ls
~/.npm
~/.npm/registry.npmjs.org/
~/.npm/registry.npmjs.org/cordova-android/
~/.npm/registry.npmjs.org/cordova-android/.cache.json
```

For `apm` it appears to be part of `apm ls`.

Python would be:

python -c 'import sys; print sys.path;'

eseidelGoogle:
Not at all sure `pub` wants this.  But I feel like this comes up for me that I want to know where pub is caching or grabbing its cache from.  Largely in debugging builders. :)

nex3:
I'm on board with this! I'm fine with having this not be hidden—the point is to give users more visibility into the cache anyway, so they should know the command is there.

In order to land this pull request, we'll need tests. See [deps_test.dart](https://github.com/dart-lang/pub/blob/master/test/deps_test.dart) for an example of testing command output. It's fine to just test that it prints out the same value as `p.join(sandboxDir, cachePath)`.

kevmoo:
Any updates on providing tests, @eseidelGoogle ?

eseidelGoogle:
I don't remember what we wanted this for.  I'm going to close this and we can re-open if we remember. :)

PR #1393, opened 2016-02-12
Suppress permission-related failures in touch().

cbracken:
If unable to update modification time due to a permissions error, such
as the file being read-only, continue without throwing.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -433,7 +433,7 @@ class Entrypoint {
         hasPathDependencies) {
       if (_isLockFileUpToDate() && _arePackagesAvailable()) {
         touchedLockFile = true;
-        touch(lockFilePath);
+        tryTouch(lockFilePath);
       } else {
         dataError('The pubspec.yaml file has changed since the pubspec.lock '
             'file was generated, please run "pub get" again.');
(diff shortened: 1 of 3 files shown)


kevmoo:
@cbracken @nex3 is this still interesting?

cbracken:
I've found an alternate solution, so technically I don't need this change anymore, though I do still think it's likely the Right Thing To Do(TM).

The issue stems from .packages and pubspec.lock being both outputs of and inputs to pub builds. Generally speaking, build systems should avoid modifying their inputs for a variety of reasons but the most obvious case is that they may be read-only. Given that pub build allows the specification of an output directory, it should be legit to, say, cd into a pub project in someone else's home directory and `pub build --output /tmp/mybuild`, particularly if the `.packages` and `pubspec.lock` file exist and are valid.

I was actually finally able to get my hands on a Windows 10 machine to test this on last week and confirmed the current patch doesn't work on Windows; turns out that you get an `errno` of `5` when attempting to append to a readonly file. Generalising to catch and suppress any exception around the append in `touch` is probably the right solution.

In any case, I've got a workaround that covers my case and the above scenario is probably fairly unlikely, so I'm entirely fine closing this out if @nex3 would rather keep the code as is given that I don't need the patch anymore.

nex3:
Yeah, I think this should probably be closed.

PR #1394, opened 2016-02-15
Fix docs typos

sdboyer:
I am reasonably certain these fixes will enhance the performance of your SAT solver by 72%, but YMMV


nex3:
Thanks for the fixes!

PR #1425, opened 2016-06-24
Change duration formatting in niceDuration

ddrone:
niceDuration used to display 61 second as "1:1" instead of "1:01". This patch fixes that and minute duration (which is 60 seconds last time I've checked) and adds couple of unit tests.

### DIFF
--- a/test/utils_test.dart
+++ b/test/utils_test.dart
@@ -82,5 +82,20 @@ true: bool"""));
 a: {}
 b: {}"""));
     });
+
+  });
+
+  group('niceDuration()', () {
+    test('formats duration longer than a minute correctly', () {
+      expect(
+          niceDuration(new Duration(minutes: 3, seconds: 1, milliseconds: 337)),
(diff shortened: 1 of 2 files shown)


test/utils_test.dart
```diff
@@ -82,5 +82,20 @@ true: bool"""));
 a: {}
 b: {}"""));
     });
+
+  });
+
+  group('niceDuration()', () {
+    test('formats duration longer than a minute correctly', () {
```
nex3: Also test a duration that's exactly on the minute boundary.

nex3:
One additional test and I'll merge this in. Thanks for the fix!

test/utils_test.dart
ddrone: Done

ddrone:
Ping

nex3:
Sorry, I was on vacation all last week.

ddrone:
No problem!

Issue #1323, opened 2015-09-01
Way to list all bins available to me in my app?


sethladd:
I was just working with a customer and we were curious what I could run from `pub run ...`. We didn't find a way to discover what's available to us. Is there such a thing? Something like a `pub run --list` ?

nex3:
There isn't, but that's a good idea.

sethladd:
Neat!

antonmoiseev:
Are you still considering adding this command? I would like to work on it, have a couple of questions.

nex3:
Yeah, I'd accept a patch that adds this.

antonmoiseev:
Nice! I'm thinking about the right way to implement this, and have two options in mind:

1. List all executables from `bin` and `.pub/bin` directories.
2. List all executables from `bin` and `bin` directories of the immediate dependencies.

I have a couple of concerns about the first option:

- Looks like it doesn't cover all the cases. For example snapshots are not generated for path dependencies.
- It requires `pub get` to be invoked before running `pub run --list`.

The second option looks good, but taking into account different types of sources, looks like it also requires `pub get` to be executed beforehand. However I'm still exploring this approach.

Could you advise what would be the correct way to get the complete list of executables?

nex3:
The second option is definitely correct, for the reasons you've discovered. You'll always need `pub get` to be run though, since otherwise you won't have concrete dependency versions—and the executables in a dependency can change from version to version.

antonmoiseev:
Could you answer several more questions:

1. `dartfmt` formats the entire file including the lines I haven't modified while addressing the issue. Should I leave or revert these changes?
2. Do you have any preference regarding the output format of the command? Right now I just print each executable on its own line:
  ![image](https://cloud.githubusercontent.com/assets/182853/16398124/36bbbfb2-3cd1-11e6-9a5c-666a16b42aff.png)
3. I believe we won't be able to list all possible executables that can be invoked with `pub run`, for example executables generated by transformers or located outside of the `bin` directory (in `tool`, `lib`, etc.) Can you confirm it's out of the scope of the issue?
4. I noticed the integration tests keep each file descriptor on its own line, however `dartfmt` not always preserve this formatting. Should I prefer `dartfmt` result or manually re-format descriptors?

nex3:
> `dartfmt` formats the entire file including the lines I haven't modified while addressing the issue. Should I leave or revert these changes?

[...]

PR #1427, opened 2016-06-28
Add "pub run --list" command to show all available executables.

antonmoiseev:
See #1323

### DIFF
--- a/lib/src/command/run.dart
+++ b/lib/src/command/run.dart
@@ -10,6 +10,8 @@ import 'package:path/path.dart' as p;
 import '../command.dart';
 import '../executable.dart';
 import '../io.dart';
+import '../log.dart' as log;
+import '../package.dart';
 import '../utils.dart';
 
 /// Handles the `run` pub command.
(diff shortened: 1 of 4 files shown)


nex3:
See comments [here](https://github.com/dart-lang/pub/issues/1323#issuecomment-230898411).

antonmoiseev:
@nex3 Natalie, I've addressed your comments, I think it's ready for review, please take a look.

nex3:
The more I think about it, the more I feel like this doesn't belong under `pub run`. It does something completely different than running an executable, and none of the existing `pub run` options make sense along with `--list`. I think maybe it would make more sense as an option to `pub deps`, especially since it only prints dependency executables now. Maybe `pub deps --executables`?

antonmoiseev:
I don't have a strong opinion on this one, just some considerations:

1. From organizational point of view `pub deps` looks like a better place
2. I think users will more likely discover it under `pub run`
3. `pub deps` is a more natural place for executables from dependencies, but not sure about executables defined in the current app

Anyway, I'm totally fine with any option you decide is the right thing to do.

nex3:
I agree it's more discoverable under `pub run`, and if it could have subcommands that's what I'd recommend. But that would conflict with running executables, and I don't like making flags do the work of subcommand.

For point 3, `pub deps` already lists the current executable when listing deps, so I don't think it's *too* weird.

antonmoiseev:
OK, moving it to `pub deps`. Should it respect the `--dev` flag?

nex3:
Yeah, it should.

nex3:
@antonmoiseev Are you planning on iterating on this?

PR #1437, opened 2016-08-05
Support simple path rewrites to index.html for angular2

### LINKED ISSUE #1374: angular2 routing defaults don't work well with pub serve
If I set up routing in angular2, and run `pub serve` everything works at first, I can click on routes, but when I refresh the page I get a 404.

This doesn't happen when I run the typescript version using `lite-server`

I can run the `pub build` output using `lite-server`, and it works fine as well. 

As a workaround, I can enable the hashlocation routing strategy in angular2, but I think it would be nice if pub serve would support routing without hash as well.

chalin:
Fixes #1374.

This commit adds a serve option:<br>
  `pub serve`**`--rewrite-to-index`**_`<path-regexp>`_.<br>
A request that would yield 404 (not found), but whose path matches the regexp, will be directed to `/index.html`.

This will be useful while developing Angular 2 Dart apps with routing. In the simplest case, the empty string will match any path:

```
pub serve --rewrite-to-index ''
```

**All 1130 tests pass** (including the newly added test).

cc @kwalrath @kasperpeulen @zoechi

### DIFF
--- a/lib/src/barback/barback_server.dart
+++ b/lib/src/barback/barback_server.dart
@@ -22,6 +22,8 @@ import 'asset_environment.dart';
 /// Callback for determining if an asset with [id] should be served or not.
 typedef bool AllowAsset(AssetId id);
 
+typedef bool Filter(Uri uri);
+
 /// A server that serves assets transformed by barback.
 class BarbackServer extends BaseServer<BarbackServerResult> {
   /// The package whose assets are being served.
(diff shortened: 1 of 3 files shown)


adaojunior:
What if the file is not `index.html` ?, I have a angular application that has a index.html, admin.html and partnet-reports.html. if I'm n ot mistaken there's an issue (feature) to add the possibility of using pub serve from a server application, this way a user would be able to configure rewrites, and proxies.

chalin:
As is mentioned on the [pub serve](https://webdev.dartlang.org/tools/pub/pub-serve) tooling page
> This command starts up a _development server_, or dev server, for your Dart web app.

I've purposefully kept this feature simple, as it should be for a _development server_. Current behavior matches that of, e.g., the [lite-server](https://github.com/johnpapa/lite-server) dev server which provides a default fallback to `index.html`.

> if I'm not mistaken there's an issue (feature) to add the possibility of using pub serve from a server application

That will be great if/when it happens and will provide a broader solution encompassing the use case you mention.

kevmoo:
n/a

PR #1451, opened 2016-09-27
Test package APIs rather than resource APIs.

### LINKED ISSUE #1446: Resource class is being removed
The Resource class will be removed in the next Dart release. There are still tests that depend on it (for example, 'test/run/resource_test.dart'. These tests will stop working.

nex3:
Closes #1446


test/global/run/package_api_test.dart
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
munificent: 2016

test/global/run/package_api_test.dart
```diff
@@ -0,0 +1,99 @@
...
+              cachePath,
+              "hosted/localhost%58${await globalPackageServer.port}/foo-1.0.0/"
+                "lib/resource.txt"))
+          .toString());
+      pub.stdout.expect(p.toUri(p.join(
+              sandboxDir,
+              cachePath,
+              "hosted/localhost%58${await globalPackageServer.port}/bar-1.0.0/"
+                "lib/resource.txt"))
+          .toString());
```
munificent: These three statements got wrapped an awful lot. It might be cleaner to declare little local variables for each path.

test/global/run/package_api_test.dart
```diff
@@ -0,0 +1,99 @@
...
+
+main() async {
+  print(await Isolate.packageRoot);
+  print(await Isolate.packageConfig);
+  print(await Isolate.resolvePackageUri(
+      Uri.parse('package:myapp/resource.txt')));
+  print(await Isolate.resolvePackageUri(
+      Uri.parse('package:foo/resource.txt')));
+}
+""")
```
munificent: This is the same string as the above script, right? If so, declare a constant for it. Ideally stick it somewhere that the run/package_api_test can get to it too.

test/global/run/package_api_test.dart
nex3: Done.

test/global/run/package_api_test.dart
nex3: Done.

test/global/run/package_api_test.dart
nex3: Done. I didn't share it across files, though, because I couldn't think of a nice semantic thing to call it.

PR #1456, opened 2016-10-04
strong mode cleanup

kevmoo:


lib/src/error_group.dart
```diff
@@ -188,7 +188,7 @@ class _ErrorGroupFuture<T> implements Future<T> {
     _completer.future.catchError((_) {});
   }
 
-  Future/*<S>*/ then/*<S>*/(/*=S*/ onValue(/*=T*/ value), {Function onError}) {
+  Future/*<S>*/ then/*<S>*/(onValue(/*=T*/ value), {Function onError}) {
```
nex3: The original declaration was intended. The way it's written now will not be runtime-strong-mode clean.

test/cache/repair/reinstalls_git_packages_test.dart
```diff
@@ -28,7 +28,7 @@ main() {
     pubUpgrade();
 
     // Break them.
-    var fooDirs;
+    List fooDirs;
```
nex3: You're welcome to make changes in tests, but be aware that there's no infrastructure in place to make sure they stay strong-mode clean.

lib/src/error_group.dart
kevmoo: I updated the signature to match Future in the SDK – for which I was given a strong-mode error.

So this is a known delta?

test/cache/repair/reinstalls_git_packages_test.dart
kevmoo: Noted.

lib/src/error_group.dart
nex3: Oh I didn't realize this was in a `Future` subclass. That's fine, then.

PR #1463, opened 2016-10-28
A few more strong-mode fixes

kevmoo:


test/pub_uploader_test.dart
```diff
@@ -31,7 +31,7 @@ See http://dartlang.org/tools/pub/cmd/pub-uploader.html for detailed documentati
 ScheduledProcess startPubUploader(ScheduledServer server, List<String> args) {
   var tokenEndpoint = server.url.then((url) =>
       url.resolve('/token').toString());
-  args = ['uploader', '--server', tokenEndpoint]..addAll(args);
+  args = <Object>['uploader', '--server', tokenEndpoint]..addAll(args);
```
nex3: Why does this need to be annotated as `Object`? Isn't that the LUB for `Future<String>` and `String`?

lib/src/barback.dart
```diff
@@ -34,7 +34,7 @@ import 'package:pub_semver/pub_semver.dart';
 /// constraint and barback's version.
 ///
 /// [compat]: https://gist.github.com/nex3/10942218
-final pubConstraints = {
+final pubConstraints = <String, VersionRange>{
```
nex3: There's no guarantee that `new VersionConstraint.parse()` will emit a `VersionRange`, so this annotation will probably end up generating an implicit cast warning at some point. If we rely on these being `VersionRange`s, you should explicitly cast them with `as`.

nex3:
@kevmoo Are you planning on iterating on this?

lib/src/barback.dart
kevmoo: True. But in every case where you call in in this final, it is a `VersionRange`. Moreover, in every usage of `pubConstraints` you assume the type is `VersionRange`.

Seems better to fail here than elsewhere, no?

kevmoo:
I replied to your requests. Happy to discuss.

test/pub_uploader_test.dart
nex3: Go on...

kevmoo:
@nex3 PTAL?

nex3:
There are outstanding code review comments...

test/pub_uploader_test.dart
kevmoo: `String` is not the LUB for `Future<String>` and `String`. Hence, `Object`.

nex3: I'm asking why `Object` isn't the LUB.

lib/src/barback.dart
kevmoo: PTAL

test/pub_uploader_test.dart
kevmoo: Let's talk in person – I'm not sure what you're asking...

Issue #1486, opened 2016-12-07
Set tar flags `--no-same-owner` and `--no-same-permissions`


mehaase:
Pub extracts tarballs on Linux using the system's `tar`. When running `pub` as root (a common scenario inside a Docker container or development VM), `tar` by default preserves the owner and permissions that the tarball was created with. For example:

```
root@7c723f4fdc37:/starbelly/static/dart# pub get --packages-dir
Resolving dependencies... (7.7s)
Downloading angular2 2.0.0...
Downloading analyzer 0.27.6...
Downloading args 0.13.7...
Downloading async 1.11.3...
Downloading barback 0.15.2+9...
Downloading browser 0.10.0+2...
...snip...
```

After getting packages, the pub cache has permissions that are not consistent with my umask:

```
root@7c723f4fdc37:/starbelly/static/dart# umask
0022
root@7c723f4fdc37:/starbelly/static/dart# ll $PUB_CACHE/hosted/pub.dartlang.org/
total 256K
drwxr-xr-x 64 root root 4.0K Dec  7 00:12 .
drwxr-xr-x  3 root root 4.0K Dec  7 00:12 ..
drwx------  8 root root 4.0K Dec  7 00:12 analyzer-0.27.6
drwx------  6 root root 4.0K Dec  7 00:12 angular2-2.0.0
drwx------  5 root root 4.0K Dec  7 00:12 args-0.13.7
drwx------  4 root root 4.0K Dec  7 00:12 async-1.11.3
drwx------  5 root root 4.0K Dec  7 00:12 barback-0.15.2+9
drwx------  5 root root 4.0K Dec  7 00:12 bootstrap_sass-4.0.0-alpha.3+5
drwx------  3 root root 4.0K Dec  7 00:12 browser-0.10.0+2
...snip...
```

Some package files aren't even owned by root at all! They are owned by a non-existent user/group. This prevents any other user inside the container from accessing these files — even root — unless `chmod` or `chown` is used after `pub get`.

```
root@7c723f4fdc37:/starbelly/static/dart# ll $PUB_CACHE/hosted/pub.dartlang.org/html-0.13.0/
total 48K
drwx------  4 root  root 4.0K Dec  7 00:12 .
drwxr-xr-x 64 root  root 4.0K Dec  7 00:12 ..
-rw-r-----  1 90694 5000   30 May 11  2016 .analysis_options
-rw-r-----  1 90694 5000  264 Dec 17  2015 .gitignore
[...]

PR #1487, opened 2016-12-07
Suggested patch for #1486

mehaase:
Add two flags to `tar` invocation that will provide consistent behavior whether `pub` is executed as root or a normal user.


nex3:
Thanks for the patch!

PR #1489, opened 2016-12-08
Fix extractTarGz().

nex3:
3fa5df4d55328628d8c91560f61cfb1f92a003b7 added new flags between the
"--directory" flag and the directory in question, which broke untarring.


nex3:
TBR

PR #1492, opened 2016-12-20
Fix expected output in cache/repair/updates_binstubs test

whesse:
Pub no longer outputs the "Loading source assets..." line in this case.

The test is failing on the buildbots, and locally.


nex3:
Thanks!

Issue #343, opened 2015-06-05
Pub should parse Dart code to do more validation before uploading


DartBot:
<a href="https://github.com/nex3"><img src="https://avatars.githubusercontent.com/u/188?v=3" align="left" width="96" height="96"hspace="10"></img></a> **Issue by [nex3](https://github.com/nex3)**
_Originally opened as dart-lang/sdk#7198_

----

Pub should do various sorts of validation that requires Dart code to be parsed. In particular:

Errors:

\* Part files in &quot;lib&quot; outside of &quot;lib/src&quot;.
\* Non-&quot;package:&quot; imports in &quot;bin&quot;.
\* Examples importing from &quot;lib/src&quot;.
\* Dart code with relative imports to files that don't exist.
\* Dart code with &quot;package:&quot; imports to files in this package that don't exist.
\* Dart code with &quot;package:&quot; imports for packages that aren't in the pubspec.

Warnings:

\* Dart files in &quot;bin&quot; without main() methods.
\* Non-&quot;package:&quot; imports in &quot;test&quot; or &quot;example&quot;.
\* Imports in &quot;test&quot; or &quot;example&quot; that reach outside the package.

DartBot:
<a href="https://github.com/munificent"><img src="https://avatars.githubusercontent.com/u/46275?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [munificent](https://github.com/munificent)**

----

*Added this to the [Later](../milestones/Later) milestone.*

DartBot:
<a href="https://github.com/munificent"><img src="https://avatars.githubusercontent.com/u/46275?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [munificent](https://github.com/munificent)**

----

*Removed [Type-Defect](../labels/Type-Defect) label.*
*Added [Type-Enhancement](../labels/Type-Enhancement) label.*

DartBot:
<a href="https://github.com/kasperl"><img src="https://avatars.githubusercontent.com/u/2156198?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [kasperl](https://github.com/kasperl)**

----

*Removed this from the [Later](../milestones/Later) milestone.*
*Added [Oldschool-Milestone-Later](../labels/Oldschool-Milestone-Later) label.*

DartBot:
[...]

PR #1508, opened 2017-01-24
First pass at strict dependencies

matanlurey:
First step of https://github.com/dart-lang/pub/issues/343.

Implemented `StrictDependenciesValidator`, which crawls all source files, finds all referenced `import` and `export` directives, and compares them against all declared `dependencies` and `dev_dependencies`.

IMO this should be an error, not a warning, but wanted input first to see if I've possibly missed legitimate use cases where people should be able to override and upload anyway. Should this also be added as a flag to `pub` to opt-in or just added by default? I didn't do anything but implement the validator yet.

~~Future work:~~

~~- [ ] Validate you haven't declared dependencies you don't use (?)~~

### DIFF
--- a/lib/src/validator/strict_dependencies.dart
+++ b/lib/src/validator/strict_dependencies.dart
@@ -1,0 +1,50 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../entrypoint.dart';
+import 'dart:async';
+import 'dart:io';
+
+import 'package:analyzer/analyzer.dart';
+import 'package:pub/src/solver/version_solver.dart';
+import 'package:pub/src/validator.dart';
+
+/// Returns a map of files -> package names imported or exported within [files].
+Map<String, Iterable<String>> _findUsedPackages(Iterable<String> files) {
+  var packageNames = <String, Iterable<String>>{};
+  for (var file in files) {
+    var usedPackages = <String>[];
+    var compilationUnit = parseDirectives(new File(file).readAsStringSync());
+    for (final directive in compilationUnit.directives) {
+      if (directive is UriBasedDirective) {
+        usedPackages
+            .add(Uri.parse(directive.uri.stringValue).pathSegments.first);
+      }
+    }
+    packageNames[file] = usedPackages.toSet();
+  }
+  return packageNames;
+}
+
+class StrictDependenciesValidator extends Validator {
+  StrictDependenciesValidator(Entrypoint entrypoint) : super(entrypoint);
+
+  @override
+  Future validate() {
+    return new Future.sync(() async {
+      var declared = new Set<String>()
+        ..addAll(entrypoint.root.dependencies.map((d) => d.name))
+        ..addAll(entrypoint.root.devDependencies.map((d) => d.name));
+      var allUsed = _findUsedPackages(entrypoint.root.listFiles());
+      allUsed.forEach((file, packageNames) {
+        for (var package in packageNames) {
+          if (!declared.contains(package)) {
+            warnings.add(
+                'Referenced "$package" in $file, but no dependency declared');
+          }
+        }
+      });
+    });
+  }
+}
--- a/lib/src/validator/strict_dependencies.dart
+++ b/lib/src/validator/strict_dependencies.dart
@@ -2,49 +2,76 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import '../entrypoint.dart';
 import 'dart:async';
-import 'dart:io';
 
 import 'package:analyzer/analyzer.dart';
-import 'package:pub/src/solver/version_solver.dart';
+import 'package:path/path.dart' as path;
+import 'package:pub/src/dart.dart';
+import 'package:pub/src/entrypoint.dart';
+import 'package:pub/src/io.dart';
 import 'package:pub/src/validator.dart';
-
-/// Returns a map of files -> package names imported or exported within [files].
-Map<String, Iterable<String>> _findUsedPackages(Iterable<String> files) {
-  var packageNames = <String, Iterable<String>>{};
-  for (var file in files) {
-    var usedPackages = <String>[];
-    var compilationUnit = parseDirectives(new File(file).readAsStringSync());
-    for (final directive in compilationUnit.directives) {
-      if (directive is UriBasedDirective) {
-        usedPackages
-            .add(Uri.parse(directive.uri.stringValue).pathSegments.first);
-      }
-    }
-    packageNames[file] = usedPackages.toSet();
-  }
-  return packageNames;
-}
+import 'package:source_span/source_span.dart';
 
 class StrictDependenciesValidator extends Validator {
+  static bool _isDartFile(String file) => path.extension(file) == '.dart';
+
   StrictDependenciesValidator(Entrypoint entrypoint) : super(entrypoint);
 
-  @override
-  Future validate() {
-    return new Future.sync(() async {
-      var declared = new Set<String>()
-        ..addAll(entrypoint.root.dependencies.map((d) => d.name))
-        ..addAll(entrypoint.root.devDependencies.map((d) => d.name));
-      var allUsed = _findUsedPackages(entrypoint.root.listFiles());
-      allUsed.forEach((file, packageNames) {
-        for (var package in packageNames) {
-          if (!declared.contains(package)) {
-            warnings.add(
-                'Referenced "$package" in $file, but no dependency declared');
+  /// Returns all pub packages imported or exported within [files].
+  List<_DependencyUse> _findPackages(Iterable<String> files) {
+    var allUses = <_DependencyUse>[];
+    for (var file in files) {
+      var dependencies = new List<_DependencyUse>();
+      try {
+        var contents = readTextFile(file);
+        var directives = parseImportsAndExports(contents, name: file);
+        for (var directive in directives) {
+          var usage = new _DependencyUse(directive, file, contents);
+          if (usage.isPubPackage) {
+            dependencies.add(usage);
           }
         }
-      });
-    });
+        allUses.addAll(dependencies);
+      } on AnalyzerErrorGroup catch (e) {
+        warnings.add('Failed parsing "$file": $e');
+      }
+    }
+    return allUses;
+  }
+
+  Future validate() async {
+    var declared = new Set<String>()
+      ..addAll(entrypoint.root.dependencies.map((d) => d.name))
+      ..addAll(entrypoint.root.devDependencies.map((d) => d.name))
+      ..add(entrypoint.root.name);
+    var allUsed = _findPackages(entrypoint.root.listFiles().where(_isDartFile));
+    for (var usage in allUsed) {
+      if (!declared.contains(usage.package)) {
+        warnings.add(usage.toErrorMessage());
+      }
+    }
+  }
+}
+
+class _DependencyUse {
+  final String _contents;
+  final UriBasedDirective _directive;
+  final String _file;
+  final Uri _parsedUri;
+
+  _DependencyUse(UriBasedDirective directive, this._file, this._contents)
+      : _parsedUri = Uri.parse(directive.uri.stringValue),
+        _directive = directive;
+
+  bool get isPubPackage => _parsedUri.scheme == 'package';
+
+  String get package => _parsedUri.pathSegments.first;
+
+  String toErrorMessage() {
+    return new SourceFile(_contents, url: _file)
+        .span(_directive.offset, _directive.length)
+        .message(
+            '$_file imports $package, but this package doesn\'t depend '
+            'on $package');
   }
 }
--- a/lib/src/validator/strict_dependencies.dart
+++ b/lib/src/validator/strict_dependencies.dart
@@ -5,73 +5,156 @@
 import 'dart:async';
 
 import 'package:analyzer/analyzer.dart';
-import 'package:path/path.dart' as path;
 import 'package:pub/src/dart.dart';
 import 'package:pub/src/entrypoint.dart';
 import 'package:pub/src/io.dart';
+import 'package:pub/src/log.dart';
 import 'package:pub/src/validator.dart';
 import 'package:source_span/source_span.dart';
 
+/// Validates that Dart source files only import declared dependencies.
 class StrictDependenciesValidator extends Validator {
-  static bool _isDartFile(String file) => path.extension(file) == '.dart';
+  static Iterable<String> _combine(Iterable<String> a, Iterable<String> b) {
+    return a.toList()..addAll(b);
+  }
 
   StrictDependenciesValidator(Entrypoint entrypoint) : super(entrypoint);
 
-  /// Returns all pub packages imported or exported within [files].
-  List<_DependencyUse> _findPackages(Iterable<String> files) {
-    var allUses = <_DependencyUse>[];
+  Set<String> _dependencies;
+
+  /// Lazily returns all dependency uses in [files].
+  ///
+  /// Files that do not parse are skipped.
+  ///
+  /// Directives that do not import package: URIs are skipped.
+  Iterable<_Usage> _findPackages(Iterable<String> files) sync* {
     for (var file in files) {
-      var dependencies = new List<_DependencyUse>();
+      List<UriBasedDirective> directives;
+      var contents = readTextFile(file);
       try {
-        var contents = readTextFile(file);
-        var directives = parseImportsAndExports(contents, name: file);
-        for (var directive in directives) {
-          var usage = new _DependencyUse(directive, file, contents);
-          if (usage.isPubPackage) {
-            dependencies.add(usage);
-          }
+        directives = parseImportsAndExports(contents, name: file);
+      } on AnalyzerErrorGroup catch (e, s) {
+        // Ignore files that do not parse.
+        exception(e, s);
+        continue;
+      }
+      for (var directive in directives) {
+        Uri uri;
+        try {
+          uri = Uri.parse(directive.uri.stringValue);
+        } on FormatException catch (_){}
+        var usage = new _Usage(file, contents, directive, uri);
+        if (usage.isPackageUrl) {
+          yield usage;
         }
-        allUses.addAll(dependencies);
-      } on AnalyzerErrorGroup catch (e) {
-        warnings.add('Failed parsing "$file": $e');
       }
     }
-    return allUses;
+  }
+
+  /// Returns whether [package] is listed under `dependencies: `.
+  ///
+  /// The current package is implicitly a dependency.
+  bool _isDependency(String package) {
+    if (_dependencies == null) {
+      _dependencies = entrypoint.root.dependencies
+          .map((d) => d.name)
+          .toSet();
+    }
+    return entrypoint.root.name == package || _dependencies.contains(package);
+  }
+
+  Set<String> _devDependencies;
+
+  /// Returns whether [package] is listed under `dev_dependencies: `.
+  bool _isDevDependency(String package) {
+    if (_devDependencies == null) {
+      _devDependencies = entrypoint.root.devDependencies
+          .map((d) => d.name)
+          .toSet();
+    }
+    return _devDependencies.contains(package);
   }
 
   Future validate() async {
-    var declared = new Set<String>()
-      ..addAll(entrypoint.root.dependencies.map((d) => d.name))
-      ..addAll(entrypoint.root.devDependencies.map((d) => d.name))
-      ..add(entrypoint.root.name);
-    var allUsed = _findPackages(entrypoint.root.listFiles().where(_isDartFile));
-    for (var usage in allUsed) {
-      if (!declared.contains(usage.package)) {
-        warnings.add(usage.toErrorMessage());
+    _validateLibBin();
+    _validateTestTool();
+  }
+
+  void _validateLibBin() {
+    var libFiles = entrypoint.root.listFiles(beneath: 'lib');
+    var binFiles = entrypoint.root.listFiles(beneath: 'bin');
+    for (var usage in _findPackages(_combine(libFiles, binFiles))) {
+      if (!usage.isUriValid) {
+        warnings.add(usage.uriInvalidMessage());
+      } else if (!_isDependency(usage.package)) {
+        if (_isDevDependency(usage.package)) {
+          warnings.add(usage.dependencyMisplaceMessage());
+        } else {
+          warnings.add(usage.dependencyMissingMessage());
+        }
+      }
+    }
+  }
+
+  void _validateTestTool() {
+    var testFiles = entrypoint.root.listFiles(beneath: 'test');
+    var toolFiles = entrypoint.root.listFiles(beneath: 'tool');
+    for (var usage in _findPackages(_combine(testFiles, toolFiles))) {
+      if (!usage.isUriValid) {
+        warnings.add(usage.uriInvalidMessage());
+      } else if (!_isDependency(usage.package) &&
+                 !_isDevDependency(usage.package)) {
+        warnings.add(usage.dependencyMissingMessage());
       }
     }
   }
 }
 
-class _DependencyUse {
+/// Represents a parsed import or export directive in a dart source file.
+class _Usage {
   final String _contents;
-  final UriBasedDirective _directive;
   final String _file;
-  final Uri _parsedUri;
+  final Uri _uri;
+  final UriBasedDirective _directive;
 
-  _DependencyUse(UriBasedDirective directive, this._file, this._contents)
-      : _parsedUri = Uri.parse(directive.uri.stringValue),
-        _directive = directive;
+  _Usage(this._file, this._contents, this._directive, this._uri);
 
-  bool get isPubPackage => _parsedUri.scheme == 'package';
+  /// Returns the package name if [isPackageUrl], otherwise `null`.
+  String get package => isPackageUrl ? _uri.pathSegments.first : null;
 
-  String get package => _parsedUri.pathSegments.first;
+  /// Returns whether the URI was parsable and correct in the directive.
+  bool get isUriValid => _uri != null && _uri.pathSegments.length >= 2;
 
-  String toErrorMessage() {
-    return new SourceFile(_contents, url: _file)
-        .span(_directive.offset, _directive.length)
-        .message(
-            '$_file imports $package, but this package doesn\'t depend '
-            'on $package');
+  /// Returns whether the directive references a pub package.
+  bool get isPackageUrl => _uri.scheme == 'package';
+
+  // Assumption is that normally all directives are valid and we won't see
+  // an error message - so a SourceFile is created lazily (on demand) to avoid
+  // parsing line endings in the case of only valid directives.
+  String _toMessage(String message) => new SourceFile(_contents, url: _file)
+      .span(_directive.offset, _directive.offset + _directive.length)
+      .message(message);
+
+  /// Returns an error message saying that the URI is invalid.
+  String uriInvalidMessage() {
+    assert(!isUriValid);
+    var uri = _directive.uri.stringValue;
+    return _toMessage('$_file references an invalid URI: $uri');
+  }
+
+  /// Returns an error message saying the package is not listed in dependencies.
+  String dependencyMissingMessage() {
+    return _toMessage(
+      '$_file imports $package, but this package doesn\'t depend on $package.'
+    );
+  }
+
+  /// Returns an error message saying the package should be in `dependencies`.
+  String dependencyMisplaceMessage() {
+    return _toMessage(
+      '$_file imports $package, but is only listed in `devDependencies`. Files '
+      'in the `lib` or `bin` folder must declare in `dependencies` - '
+      'devDependencies is only valid for files in `tool` or `test`.'
+    );
   }
 }
--- a/lib/src/validator/strict_dependencies.dart
+++ b/lib/src/validator/strict_dependencies.dart
@@ -45,10 +45,15 @@ class StrictDependenciesValidator extends Validator {
         // there are no segments OR
         // any segment are empty
         if (uri == null ||
-            uri.scheme == 'package' &&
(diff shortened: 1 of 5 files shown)


kevmoo:
Crazy excited to see this!

lrhn:
One use-case that might be affected is to have a package that is only used through `Isolate.spawnUri`.
It's probably rare - a package intended for such use will likely have its own launcher code that you would include in the main application too, instead of just doing `Isolate.spawnUri(Uri.parse("package:otherwise_unused/main.dart")`

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,50 @@
...
+
+class StrictDependenciesValidator extends Validator {
+  StrictDependenciesValidator(Entrypoint entrypoint) : super(entrypoint);
+
+  @override
+  Future validate() {
+    return new Future.sync(() async {
+      var declared = new Set<String>()
+        ..addAll(entrypoint.root.dependencies.map((d) => d.name))
+        ..addAll(entrypoint.root.devDependencies.map((d) => d.name));
```
nex3: This should also verify that files in lib and bin don't use dev dependencies.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,50 @@
...
+      var allUsed = _findUsedPackages(entrypoint.root.listFiles());
```
nex3: This is going to list non-Dart files as well.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,50 @@
...
+      allUsed.forEach((file, packageNames) {
+        for (var package in packageNames) {
+          if (!declared.contains(package)) {
+            warnings.add(
+                'Referenced "$package" in $file, but no dependency declared');
```
nex3: I think this could be clearer. Maybe track the directive type, and say e.g. `$file imports $package, but this package doesn't depend on $package.`

test/validator/strict_dependencies_test.dart
```diff
@@ -0,0 +1,49 @@
...
+import '../descriptor.dart' as d;
+import '../test_pub.dart';
+import 'utils.dart';
+
+Validator strictDeps(Entrypoint entrypoint) {
+  return new StrictDependenciesValidator(entrypoint);
+}
+
+main() {
+  group('should consider a package valid if', () {
```
nex3: "if it"

Concatenated test names should generally read as sentences.

test/validator/strict_dependencies_test.dart
```diff
@@ -0,0 +1,49 @@
...
+
+    integration('looks normal', () => expectNoValidationError(strictDeps));
+
+    integration('declares an "import" as a dependency', () {
+      d.dir(appPath, [
+        d.libPubspec("test_pkg", "1.0.0", deps: {
+          "not_declared": "^1.2.3"
+        }, sdk: ">=1.8.0 <2.0.0")
+      ]).create();
+      d.file(path.join(appPath, 'lib', 'library.dart'), r'''
```
nex3: Nest this in the previous `d.dir()` call.

test/validator/strict_dependencies_test.dart
```diff
@@ -0,0 +1,49 @@
...
+main() {
+  group('should consider a package valid if', () {
+    setUp(d.validPackage.create);
+
+    integration('looks normal', () => expectNoValidationError(strictDeps));
+
+    integration('declares an "import" as a dependency', () {
+      d.dir(appPath, [
+        d.libPubspec("test_pkg", "1.0.0", deps: {
+          "not_declared": "^1.2.3"
```
nex3: Declaring a dependency on a package named "not_declared" is pretty confusing :p.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,50 @@
...
+/// Returns a map of files -> package names imported or exported within [files].
+Map<String, Iterable<String>> _findUsedPackages(Iterable<String> files) {
+  var packageNames = <String, Iterable<String>>{};
+  for (var file in files) {
+    var usedPackages = <String>[];
+    var compilationUnit = parseDirectives(new File(file).readAsStringSync());
+    for (final directive in compilationUnit.directives) {
+      if (directive is UriBasedDirective) {
+        usedPackages
+            .add(Uri.parse(directive.uri.stringValue).pathSegments.first);
```
nex3: What if it's not a `package:` import?

test/validator/strict_dependencies_test.dart
```diff
@@ -0,0 +1,49 @@
...
+
+    integration('does not declare an "import" as a dependency', () {
+      d.file(path.join(appPath, 'lib', 'library.dart'), r'''
+        import 'package:not_declared/not_declared.dart';
+      ''').create();
+
+      expectValidationWarning(strictDeps);
+    });
+  });
+}
```
nex3: Also test:

* Files outside `lib/`.
* A matching dev dependency.
* A non-Dart file that looks like a Dart file that contains an invalid dependency.
* A Dart file with a parse error early enough that `parseDirectives()` runs into it.
* A Dart file with a relative import.
* A Dart file with an absolute, non-`package:` import.
* Dart files with invalid `package:` URIs like `"package:"`, `"package:foo"`, `"package:/"`, `"package:/]"`.

matanlurey: I did not add a file outside `lib/` as per my comment above, nor invalid packages - I could use some guidance on the latter if there is already something in pub that can do simple validation or if I should just check the parsed `Uri`.

The rest are done!

test/validator/strict_dependencies_test.dart
matanlurey: Done.

test/validator/strict_dependencies_test.dart
matanlurey: Renamed to kevmoo@'s favorite, `silly_monkey`

test/validator/strict_dependencies_test.dart
matanlurey: Done

lib/src/validator/strict_dependencies.dart
matanlurey: Done. Cleaned it up a bit and even added a source span of the usage.

lib/src/validator/strict_dependencies.dart
matanlurey: Done, filtered based on the extension being dart.

lib/src/validator/strict_dependencies.dart
matanlurey: Is that always true? For example I've seen packages that have something like

`lib/src/test_utils.dart`

I'm not saying it's right, but this seems like a much bigger change than just enforcing that you've declared your dependencies so far and I'd rather tackle it in a future improvement with some consensus.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,50 @@
...
+  }
+  return packageNames;
+}
+
+class StrictDependenciesValidator extends Validator {
+  StrictDependenciesValidator(Entrypoint entrypoint) : super(entrypoint);
+
+  @override
+  Future validate() {
+    return new Future.sync(() async {
```
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,50 @@
...
+    }
+    packageNames[file] = usedPackages.toSet();
+  }
+  return packageNames;
+}
+
+class StrictDependenciesValidator extends Validator {
+  StrictDependenciesValidator(Entrypoint entrypoint) : super(entrypoint);
+
+  @override
```
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
matanlurey: Done, thanks.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,50 @@
...
+import 'package:analyzer/analyzer.dart';
+import 'package:pub/src/solver/version_solver.dart';
+import 'package:pub/src/validator.dart';
+
+/// Returns a map of files -> package names imported or exported within [files].
+Map<String, Iterable<String>> _findUsedPackages(Iterable<String> files) {
+  var packageNames = <String, Iterable<String>>{};
+  for (var file in files) {
+    var usedPackages = <String>[];
+    var compilationUnit = parseDirectives(new File(file).readAsStringSync());
```
matanlurey: Done, thanks.

Added a simple try/catch but let me know if you want something more sophisticated...

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,50 @@
...
+import '../entrypoint.dart';
+import 'dart:async';
+import 'dart:io';
+
+import 'package:analyzer/analyzer.dart';
+import 'package:pub/src/solver/version_solver.dart';
+import 'package:pub/src/validator.dart';
+
+/// Returns a map of files -> package names imported or exported within [files].
+Map<String, Iterable<String>> _findUsedPackages(Iterable<String> files) {
```
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
```diff
@@ -2,49 +2,76 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import '../entrypoint.dart';
 import 'dart:async';
-import 'dart:io';
 
 import 'package:analyzer/analyzer.dart';
-import 'package:pub/src/solver/version_solver.dart';
+import 'package:path/path.dart' as path;
```
nex3: `as p`. Pub is kind of inconsistent about this, but that's only because it's old—new files always use `as p`.

lib/src/validator/strict_dependencies.dart
```diff
@@ -2,49 +2,76 @@
...
-  var packageNames = <String, Iterable<String>>{};
-  for (var file in files) {
-    var usedPackages = <String>[];
-    var compilationUnit = parseDirectives(new File(file).readAsStringSync());
-    for (final directive in compilationUnit.directives) {
-      if (directive is UriBasedDirective) {
-        usedPackages
-            .add(Uri.parse(directive.uri.stringValue).pathSegments.first);
+class StrictDependenciesValidator extends Validator {
+  static bool _isDartFile(String file) => path.extension(file) == '.dart';
```
nex3: I'd just inline this.

lib/src/validator/strict_dependencies.dart
```diff
@@ -2,49 +2,76 @@
...
-Map<String, Iterable<String>> _findUsedPackages(Iterable<String> files) {
-  var packageNames = <String, Iterable<String>>{};
-  for (var file in files) {
-    var usedPackages = <String>[];
-    var compilationUnit = parseDirectives(new File(file).readAsStringSync());
-    for (final directive in compilationUnit.directives) {
-      if (directive is UriBasedDirective) {
-        usedPackages
-            .add(Uri.parse(directive.uri.stringValue).pathSegments.first);
+class StrictDependenciesValidator extends Validator {
```
nex3: Document this, and also the class below and its members.

lib/src/validator/strict_dependencies.dart
```diff
@@ -2,49 +2,76 @@
...
+        var directives = parseImportsAndExports(contents, name: file);
+        for (var directive in directives) {
+          var usage = new _DependencyUse(directive, file, contents);
+          if (usage.isPubPackage) {
+            dependencies.add(usage);
+          }
+        }
+        allUses.addAll(dependencies);
+      } on AnalyzerErrorGroup catch (e) {
+        warnings.add('Failed parsing "$file": $e');
```
nex3: I think for now we should just ignore files that don't parse. I don't want to make users think that pub will verify that their files parse when in fact it's only looking at the first few lines.

lib/src/validator/strict_dependencies.dart
```diff
@@ -2,49 +2,76 @@
...
+    var allUses = <_DependencyUse>[];
+    for (var file in files) {
+      var dependencies = new List<_DependencyUse>();
+      try {
+        var contents = readTextFile(file);
+        var directives = parseImportsAndExports(contents, name: file);
+        for (var directive in directives) {
+          var usage = new _DependencyUse(directive, file, contents);
+          if (usage.isPubPackage) {
+            dependencies.add(usage);
```
nex3: Why not add directly to `allUses`?

test/validator/strict_dependencies_test.dart
```diff
@@ -17,30 +17,88 @@ Validator strictDeps(Entrypoint entrypoint) {
...
+    integration('only uses dart: dependencies (not pub packages)', () {
+      d.file(path.join(appPath, 'lib', 'library.dart'), r'''
+        import 'dart:async';
+        import 'dart:collection';
+        import 'dart:typed_data';
+      ''').create();
+      expectNoValidationError(strictDeps);
+    });
+
+    integration('imports itself', () {
```
nex3: Good thinking :smile:

lib/src/validator/strict_dependencies.dart
```diff
@@ -2,49 +2,76 @@
...
-                'Referenced "$package" in $file, but no dependency declared');
-          }
-        }
-      });
-    });
+  _DependencyUse(UriBasedDirective directive, this._file, this._contents)
+      : _parsedUri = Uri.parse(directive.uri.stringValue),
+        _directive = directive;
+
+  bool get isPubPackage => _parsedUri.scheme == 'package';
```
nex3: This is kind of misleading, since everything (including the entrypoint) is a pub package. Maybe `isPackageUrl`?

lib/src/validator/strict_dependencies.dart
```diff
@@ -2,49 +2,76 @@
...
+
+  String get package => _parsedUri.pathSegments.first;
+
+  String toErrorMessage() {
+    return new SourceFile(_contents, url: _file)
```
nex3: I'm guessing that you're creating the `SourceFile` here rather than sharing one between all the directives for a given file so that you avoid parsing the line endings in the common case of a file that has only valid directives? If so, mention that in a comment.

lib/src/validator/strict_dependencies.dart
```diff
@@ -2,49 +2,76 @@
...
+        .span(_directive.offset, _directive.length)
+        .message(
+            '$_file imports $package, but this package doesn\'t depend '
+            'on $package');
```
nex3: Nit: "."

test/validator/strict_dependencies_test.dart
```diff
@@ -17,30 +17,88 @@ Validator strictDeps(Entrypoint entrypoint) {
 }
 
 main() {
-  group('should consider a package valid if', () {
+  group('should consider a package valid if it', () {
     setUp(d.validPackage.create);
 
     integration('looks normal', () => expectNoValidationError(strictDeps));
 
     integration('declares an "import" as a dependency', () {
```
nex3: We don't need to duplicate every test, but we should probably have at least one test for an `export` that *doesn't* produce a warning.

lib/src/validator/strict_dependencies.dart
```diff
@@ -2,49 +2,76 @@
...
+      : _parsedUri = Uri.parse(directive.uri.stringValue),
+        _directive = directive;
+
+  bool get isPubPackage => _parsedUri.scheme == 'package';
+
+  String get package => _parsedUri.pathSegments.first;
+
+  String toErrorMessage() {
+    return new SourceFile(_contents, url: _file)
+        .span(_directive.offset, _directive.length)
```
nex3: `.span()` takes a start and an end, not a start and a length. It's also 0-based, so make sure that the analyzer is as well or add some `- 1`s if not.

test/validator/strict_dependencies_test.dart
matanlurey: > > I did not add a file outside `lib/` as per my comment above
> 
> Which comment?

Got outdated, but I am not sure how to do this. Do you have any pointers?

> 
> > I could use some guidance on the latter if there is already something in pub that can do simple validation or if I should just check the parsed `Uri`.
> 
> All the validators were written before the analyzer library existed, so we currently validate nothing that involves parsing Dart code. For invalid URLs I think we should probably just emit a warning that says "Invalid URL" and points to the source span.

How can I determine an invalid URL? Other than reinventing the world.

lib/src/validator/strict_dependencies.dart
matanlurey: Is there a "verbose" flag where I can conditionally emit something for users?

test/validator/strict_dependencies_test.dart
lrhn: I'm guessing you already parse URIs with `Uri.parse`, or similar. That should ensure that the URI is actually valid URIs.

Package URIs have a more restricted syntax that you might also want to check. To validate those, you can either reuse the code in [package:package_config](https://github.com/dart-lang/package_config/blob/master/lib/src/util.dart) or aim lower (that code is written for `.packages` files which are very specific, actual imports might want to be more permissive).

The base rules for package URIs is:
- Starts with `package:`. 
- Has no authority part.
- The path doesn't start with a `/`,
- but with a path segment containing at least one non-`.` character and no `:` or `%` characters,
- followed by a `/`.
- Has no query part (but that's probably not important).

You can check that at the string level with a RegExp:
```dart
var pchar = r"!$&-\-0-9;=@-Z_a-z~";
const pcharDot = r"!$&-.0-9;=@-Z_a-z~";
var packageUriRE = new RegExp("^package:[$pcharDot]*[$pchar]$[pcharDot]*/");
```
or on a parsed URI:
```dart
const pchar = r"!$&-\-0-9;=@-Z_a-z~";
const pcharDot = r"!$&-.0-9;=@-Z_a-z~";
final packagePathUriRE = new RegExp("^[$pcharDot]*[$pchar]$[pcharDot]*/");
bool isPackageUriValid(Uri uri) {
  return (uri.scheme == "package" &&
      !uri.hasAuthority &&
      uri.path.matches(packageUriPathRE) &&
      !uri.hasQuery);
}
```

nex3: > Got outdated, but I am not sure how to do this. Do you have any pointers?

You could add a `_DependencyUse.inLib` getter, only check against normal dependencies if it's set, then check against dev dependencies when generating a message to provide more information.

> How can I determine an invalid URL? Other than reinventing the world.

I don't think you need to do anything as complex as Lasse is suggesting. `Uri.parse()` will throw a `FormatError` if the URL doesn't parse at all, and if it does I think it's sufficient to check that the package name isn't `null`—any other weird names will fail to match any dependencies.

lib/src/validator/strict_dependencies.dart
matanlurey: I rewrote most of this class, should be more understandable now.

lib/src/validator/strict_dependencies.dart
matanlurey: Found it.

lib/src/validator/strict_dependencies.dart
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
matanlurey: Done.

test/validator/strict_dependencies_test.dart
matanlurey: Sounds good, I've implemented that.

matanlurey:
@nex3 I think I've tackled everything you asked for - PTAL.

If there are only incremental things left I'd be happy to fix them in more follow-up PRs since it's not yet enabled by default or behind a flag or anything. Let me know.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,160 @@
...
+  String dependencyMissingMessage() {
+    return _toMessage(
+      '$_file imports $package, but this package doesn\'t depend on $package.'
+    );
+  }
+
+  /// Returns an error message saying the package should be in `dependencies`.
+  String dependencyMisplaceMessage() {
+    return _toMessage(
+      '$_file imports $package, but is only listed in `devDependencies`. Files '
```
nex3:
```
$package is a dev dependency. Packages used in ${p.split(p.relative(_file)).first}/ must be declared as
normal dependencies.
```

The newline is important—we try to wrap publish messages around 80 characters.

test/validator/strict_dependencies_test.dart
```diff
@@ -0,0 +1,176 @@
...
+        }, sdk: ">=1.8.0 <2.0.0"),
+        d.dir('bin', [
+          d.file('library.dart', r'''
+            import 'package:silly_monkey/silly_monkey.dart';
+          '''),
+        ]),
+      ]).create();
+      expectValidationWarning(strictDeps);
+    });
+  });
```
nex3: Still needs no-warning tests for:

* An `export` of a dependency.
* A top level `.dart` file that contains an invalid dependency.
* A non-Dart file that looks like a Dart file that contains an invalid dependency.

And a warning test for Dart files with URIs like `"package:"`, `"package:foo"`, and `"package:/]"`.

matanlurey: Done. Nice catch on the non-Dart file.

lib/src/validator/strict_dependencies.dart
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,160 @@
...
+  String uriInvalidMessage() {
+    assert(!isUriValid);
+    var uri = _directive.uri.stringValue;
+    return _toMessage('$_file references an invalid URI: $uri');
+  }
+
+  /// Returns an error message saying the package is not listed in dependencies.
+  String dependencyMissingMessage() {
+    return _toMessage(
+      '$_file imports $package, but this package doesn\'t depend on $package.'
```
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,160 @@
...
+  // parsing line endings in the case of only valid directives.
+  String _toMessage(String message) => new SourceFile(_contents, url: _file)
+      .span(_directive.offset, _directive.offset + _directive.length)
+      .message(message);
+
+  /// Returns an error message saying that the URI is invalid.
+  String uriInvalidMessage() {
+    assert(!isUriValid);
+    var uri = _directive.uri.stringValue;
+    return _toMessage('$_file references an invalid URI: $uri');
```
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,160 @@
...
+  _Usage(this._file, this._contents, this._directive, this._uri);
+
+  /// Returns the package name if [isPackageUrl], otherwise `null`.
+  String get package => isPackageUrl ? _uri.pathSegments.first : null;
+
+  /// Returns whether the URI was parsable and correct in the directive.
+  bool get isUriValid => _uri != null && _uri.pathSegments.length >= 2;
+
+  /// Returns whether the directive references a pub package.
+  bool get isPackageUrl => _uri.scheme == 'package';
```
matanlurey: Simplified, so this is gone now.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,160 @@
...
+  ///
+  /// The current package is implicitly a dependency.
+  bool _isDependency(String package) {
+    if (_dependencies == null) {
+      _dependencies = entrypoint.root.dependencies
+          .map((d) => d.name)
+          .toSet();
+    }
+    return entrypoint.root.name == package || _dependencies.contains(package);
+  }
```
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,160 @@
...
+        exception(e, s);
+        continue;
+      }
+      for (var directive in directives) {
+        Uri uri;
+        try {
+          uri = Uri.parse(directive.uri.stringValue);
+        } on FormatException catch (_){}
+        var usage = new _Usage(file, contents, directive, uri);
+        if (usage.isPackageUrl) {
```
matanlurey: Done. Simplified.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,160 @@
...
+  /// Directives that do not import package: URIs are skipped.
+  Iterable<_Usage> _findPackages(Iterable<String> files) sync* {
+    for (var file in files) {
+      List<UriBasedDirective> directives;
+      var contents = readTextFile(file);
+      try {
+        directives = parseImportsAndExports(contents, name: file);
+      } on AnalyzerErrorGroup catch (e, s) {
+        // Ignore files that do not parse.
+        exception(e, s);
```
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,160 @@
...
+
+/// Validates that Dart source files only import declared dependencies.
+class StrictDependenciesValidator extends Validator {
+  static Iterable<String> _combine(Iterable<String> a, Iterable<String> b) {
+    return a.toList()..addAll(b);
+  }
+
+  StrictDependenciesValidator(Entrypoint entrypoint) : super(entrypoint);
+
+  Set<String> _dependencies;
```
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,160 @@
...
+import 'package:pub/src/dart.dart';
+import 'package:pub/src/entrypoint.dart';
+import 'package:pub/src/io.dart';
+import 'package:pub/src/log.dart';
+import 'package:pub/src/validator.dart';
+import 'package:source_span/source_span.dart';
+
+/// Validates that Dart source files only import declared dependencies.
+class StrictDependenciesValidator extends Validator {
+  static Iterable<String> _combine(Iterable<String> a, Iterable<String> b) {
```
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
```diff
@@ -0,0 +1,160 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:analyzer/analyzer.dart';
+import 'package:pub/src/dart.dart';
+import 'package:pub/src/entrypoint.dart';
+import 'package:pub/src/io.dart';
+import 'package:pub/src/log.dart';
```
matanlurey: Done.

lib/src/validator/strict_dependencies.dart
```diff
@@ -35,59 +32,51 @@ class StrictDependenciesValidator extends Validator {
...
-        var usage = new _Usage(file, contents, directive, uri);
-        if (usage.isPackageUrl) {
+        // If the URL could not be parsed or it is a *package* AND
+        // there are no segments OR
+        // any segment are empty
+        if (uri == null ||
+            uri.scheme == 'package' &&
+            uri.pathSegments.length < 2 ||
+            uri.pathSegments.any((s) => s.isEmpty)) {
+          warnings.add('Invalid URL');
```
nex3: When you moved this out here, you lost the source span context.

lib/src/validator/strict_dependencies.dart
```diff
@@ -35,59 +32,51 @@ class StrictDependenciesValidator extends Validator {
...
         } on FormatException catch (_){}
-        var usage = new _Usage(file, contents, directive, uri);
-        if (usage.isPackageUrl) {
+        // If the URL could not be parsed or it is a *package* AND
+        // there are no segments OR
+        // any segment are empty
+        if (uri == null ||
+            uri.scheme == 'package' &&
+            uri.pathSegments.length < 2 ||
+            uri.pathSegments.any((s) => s.isEmpty)) {
```
nex3: Please parenthesize sub-expressions with different boolean operators. I'm sure the order of operation works out here, but it's a lot harder to visually verify that without explicit grouping.

lib/src/validator/strict_dependencies.dart
```diff
@@ -135,26 +116,17 @@ class _Usage {
...
-    var uri = _directive.uri.stringValue;
-    return _toMessage('$_file references an invalid URI: $uri');
-  }
-
   /// Returns an error message saying the package is not listed in dependencies.
   String dependencyMissingMessage() {
-    return _toMessage(
-      '$_file imports $package, but this package doesn\'t depend on $package.'
-    );
+    return _toMessage('This packagee doesn\'t depend on $package.');
```
nex3: "package"

test/validator/strict_dependencies_test.dart
```diff
@@ -172,5 +199,43 @@ main() {
...
+        d.dir(appPath, [
+          d.dir('lib', [
+            d.file('library.dart', r'''
+            import 'package:/';
+          '''),
+          ]),
+        ]).create();
+        expectValidationWarning(strictDeps);
+      });
+    });
```
nex3: We still need to test for `package:/]`. This is meaningfully different because it's invalid URL syntax, whereas the rest of these are invalid only as `package:` URLs in particular.

lib/src/validator/strict_dependencies.dart
matanlurey: Sure, I understand. Done.

lib/src/validator/strict_dependencies.dart
matanlurey: I pulled it out to a static function to re-use here with minimal boiler.

test/validator/strict_dependencies_test.dart
matanlurey: Done.

nex3:
Thanks for working so hard on this, Matan!

PR #1515, opened 2017-01-31
Enable strict dependencies

matanlurey:
Let me know if there is any recommendation on how to test this externally.

I did locally alias it as `pub2` on my machine and tried it on a few packages successfully.


nex3:
I just test locally using `dart path/to/pub.dart ...`.

PR #1516, opened 2017-01-31
Produce a better error when Flutter isn't found.

nex3:
Closes #1514

PR #1517, opened 2017-02-01
Fix reference to ApplicationException

matanlurey:
Side effect of https://github.com/dart-lang/pub/pull/1516.


nex3:
Haha oops, thanks for the fix!

PR #1521, opened 2017-02-27
Add metadata to requests to pub.dartlang.org.

nex3:
This will help pub.dartlang.org better track how packages relate to one
another.


lib/src/package.dart
```diff
@@ -496,3 +507,21 @@ class PackageDep extends PackageName {
       samePackage(other) &&
       other.constraint == constraint;
 }
+
+/// The type of dependency from one package to another.
+class DependencyType {
```
kevmoo: Any reason to *not* use enum here?

lib/src/http.dart
```diff
@@ -37,6 +44,20 @@ class _PubHttpClient extends http.BaseClient {
       : this._inner = inner == null ? new http.Client() : inner;
 
   Future<http.StreamedResponse> send(http.BaseRequest request) async {
+    if (_shouldAddMetadata(request)) {
+      request.headers["X-Pub-OS"] = Platform.operatingSystem;
```
kevmoo: Happy to discuss not following the PRD (re how things are sent to the server) – but please do so before the PR

kevmoo: Sorry. I thought I was more explicit than I was – at least we need to make sure we document how it works for the person handling the server side...

lib/src/package.dart
nex3: Dart's enums are so broken for so many use-cases that I've never bothered to learn how they work. Can you provide different stringifications for each value?

lib/src/http.dart
nex3: Do you mean the header names? I didn't realize you were expecting the implementation to use the exact names in the PRD. You're very much not supposed to use custom headers that don't begin with `X-`.

kevmoo: This way is fine. Will be easy to parse and use...

lib/src/package.dart
kevmoo: Nope. The toString is 'enumName.vauleName'. So if that won't work here – nevermind. 😄

lib/src/http.dart
```diff
@@ -37,6 +44,20 @@ class _PubHttpClient extends http.BaseClient {
...
+      request.headers["X-Pub-Command"] = PubCommandRunner.command.join(" ");
+      request.headers["X-Pub-Session-ID"] = _sessionId;
+
+      var environment = Platform.environment["PUB_ENVIRONMENT"];
+      if (environment != null) {
+        request.headers["X-Pub-Environment"] = environment;
```
mkustermann: Generally speaking I think we should sanitize the data that goes into the headers (e.g. could this environment variable  have PII data in it?) and also what format we use: To avoid any issues we should e.g. restrict ourselves to ASCII-only header values.

If we want to have non-ASCII support we should base64 encode it.

Also the document kevin went out says something about one header with json value whereas this code seems to set different `X-Pub-*` headers. What's the status of this? 

@kevmoo @nex3

nex3: > Generally speaking I think we should sanitize the data that goes into the headers (e.g. could this environment variable have PII data in it?) and also what format we use: To avoid any issues we should e.g. restrict ourselves to ASCII-only header values.

The environment variable is scoped to this application, so I don't think it's likely anyone's already using it for anything, much less for storing PII.

I think allowing non-ASCII characters here is fine. They aren't guaranteed to be well-preserved through proxies, but they won't break the request and we don't expect to need to track them effectively.

kevmoo:
@munificent Could you take a look?

munificent:
This CL is giant and has PII implications, so I don't think I know enough to give it a good review. Can it wait until @nex3 is back?

lib/src/command_runner.dart
```diff
@@ -77,13 +97,14 @@ class PubCommandRunner extends CommandRunner {
   }
 
   Future run(Iterable<String> arguments) async {
-    var options;
+    ArgResults options;
```
munificent: Do you need this local anymore? Why not just use `_options` everywhere in the body instead?

lib/src/http.dart
```diff
@@ -104,6 +125,29 @@ class _PubHttpClient extends http.BaseClient {
...
+    if (command.length > 2) return false;
+    if (command.length == 1) {
+      return ["downgrade", "get", "upgrade"].contains(command.first);
+    }
+
+    switch (command[0]) {
+      case "cache": return ["add", "repair"].contains(command[1]);
+      case "global": return command[1] == "activate";
+      default: return false;
+    }
```
munificent: This chunk of code is a little complex. Could it just be:

```dart
var command = PubCommandRunner.command.join(" ");
return const [
  "downgrade",
  "get",
  "upgrade",
  "cache add",
  "cache repair",
  "global activate"
].contains(command);
```

?

kevmoo:
@munificent I'm tracking down a regression in 6 test w/ this CL

```
pub run test test/validator/dependency_test.dart -N 'should consider a package invalid if it has' -N 'should suggest' | grep [E]
00:00 +0 -1: should consider a package invalid if it has a git dependency where a hosted version exists and should suggest the hosted primary version [E]
  ScheduleError:
  | Expected: completes to a value that (empty, some element contains '  foo: ">=2.0.0 <3.0.0"')
  Error detected when waiting for out-of-band callbacks in queue "tasks".
00:01 +0 -2: should consider a package invalid if it has a git dependency where a hosted version exists and should suggest the hosted prerelease version if it's the only version available [E]
  ScheduleError:
  | Expected: completes to a value that (empty, some element contains '  foo: ">=3.0.0-pre <4.0.0"')
  Error detected when waiting for out-of-band callbacks in queue "tasks".
00:01 +0 -3: should consider a package invalid if it has a git dependency where a hosted version exists and should suggest a tighter constraint if primary is pre-1.0.0 [E]
  ScheduleError:
  | Expected: completes to a value that (empty, some element contains '  foo: ">=0.0.2 <0.1.0"')
  Error detected when waiting for out-of-band callbacks in queue "tasks".
00:01 +0 -4: should consider a package invalid if it has a path dependency where a hosted version exists and should suggest the hosted primary version [E]
  ScheduleError:
  | Expected: completes to a value that (some element contains '  foo: ">=2.0.0 <3.0.0"', empty)
  | test/validator/dependency_test.dart 24:3                   expectDependencyValidationError
  Error detected when waiting for out-of-band callbacks in queue "tasks".
00:01 +0 -5: should consider a package invalid if it has a path dependency where a hosted version exists and should suggest the hosted prerelease version if it's the only version available [E]
  ScheduleError:
  | Expected: completes to a value that (some element contains '  foo: ">=3.0.0-pre <4.0.0"', empty)
  | test/validator/dependency_test.dart 24:3                   expectDependencyValidationError
  Error detected when waiting for out-of-band callbacks in queue "tasks".
00:01 +0 -6: should consider a package invalid if it has a path dependency where a hosted version exists and should suggest a tighter constraint if primary is pre-1.0.0 [E]
  ScheduleError:
  | Expected: completes to a value that (some element contains '  foo: ">=0.0.2 <0.1.0"', empty)
  | test/validator/dependency_test.dart 24:3                   expectDependencyValidationError
  Error detected when waiting for out-of-band callbacks in queue "tasks".
```

PR #1525, opened 2017-03-05
Fixes #1442 'UID too large' on Mac

### LINKED ISSUE #1442: "Invalid upload: no pubspec.yaml file" when trying to publish a package
I am trying to publish a new version of https://bitbucket.org/aholmgren/dart_jwt, a package that I have published many times before. For some reason it keeps failing

```
pub publish
Publishing dart_jwt 0.5.2 to https://pub.dartlang.org:
|-- .gitignore
|-- AUTHORS
|-- CHANGELOG.md
|-- LICENSE
|-- README.md
|-- bitbucket-pipelines.yml
|-- lib
|   |-- dart_jwt.dart
|   '-- src
|       |-- codec.dart
|       |-- jose.dart
|       |-- jwa.dart
|       |-- jws.dart
|       |-- jwt.dart
|       |-- jwt_claimset.dart
|       |-- util.dart
|       '-- validation_constraint.dart
|-- pubspec.yaml
'-- test
    |-- all_tests.dart
    |-- jwa_test.dart
    |-- jwt_test.dart
    '-- run.sh

Looks great! Are you ready to upload your package (y/n)? y
Uploading... (3.7s)
Invalid upload: no pubspec.yaml file.
```

nilsdoehring:
Fixes #1442 'UID too large' on Mac

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -1034,32 +1034,48 @@ ByteStream createTarGz(List contents, {String baseDir}) {
         "--create",
         "--gzip",
         "--directory",
-        baseDir,
-        "--files-from",
-        "/dev/stdin"
+        baseDir
       ];
 
-      // The ustar format doesn't support large UIDs. We don't care about
-      // preserving ownership anyway, so we just set them to "pub".
       if (Platform.isLinux) {
         // GNU tar flags.
         // https://www.gnu.org/software/tar/manual/html_section/tar_33.html
-        args.addAll(["--owner=pub", "--group=pub"]);
-      } else {
-        // BSD tar flags.
-        // https://www.freebsd.org/cgi/man.cgi?query=bsdtar&sektion=1
-        // TODO(rnystrom): These flags are, alas, not supported by the version
-        // of tar on OS X. Passing them causes tar to exit with an error. For
-        // now, we'll just not handle large UIDs on Mac until we can come up
-        // with something.
-        // See: https://github.com/dart-lang/pub/issues/1442
-        // args.addAll(["--uname=pub", "--gname=pub"]);
-      }
 
-      var process = await startProcess("tar", args);
-      process.stdin.add(UTF8.encode(contents.join("\n")));
-      process.stdin.close();
-      return process.stdout;
+        args.addAll(["--files-from", "/dev/stdin"]);
+
+        // The ustar format doesn't support large UIDs. We don't care about
+        // preserving ownership anyway, so we just set them to "pub".
+        args.addAll(["--owner=pub", "--group=pub"]);
+
+        var process = await startProcess("tar", args);
+        process.stdin.add(UTF8.encode(contents.join("\n")));
+        process.stdin.close();
+        return process.stdout;
+      } else {
+        // OSX tar flags, applicable since at least OSX 10.9 (bsdtar 2.8.3)
+        // https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/tar.1.html
+
+        // Create the file containing the list of files to compress.
+        var tempDir = createSystemTempDir();
+
+        try {
+          // Create the file containing the list of files to compress.
+          var contentsPath = path.join(tempDir, "files.mtree");
+
+          // The ustar format doesn't support large UIDs, so we set
+          // UIDs and GIDs to 0 via an mtree input file.
+          var uidGid = " uid=0 gid=0 type=file";
+          writeTextFile(contentsPath, "#mtree\n" + contents.join("$uidGid\n") + uidGid + "\n");
+          args.add("@${contentsPath}");
+
+          return (await startProcess("tar", args, workingDir: baseDir))
+              .stdout
+              .transform(onDoneTransformer(() => deleteEntry(tempDir)));
+        } catch (_) {
+          deleteEntry(tempDir);
+          rethrow;
+        }
+      }
     }
 
     // Don't use [withTempDir] here because we don't want to delete the temp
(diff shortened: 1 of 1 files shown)


kevmoo:
@nilsdoehring Curious: did you do a before/after w/ tests on your mac?

Did they fail before and now succeed?

nilsdoehring:
@kevmoo tests ran fine before and after my changes because they don't seem to validate the generated tar (which was just some bytes before the changes) on the server side. In production, the error appeared only after uploading to pub.dartlang.org, which reported back that no pubspec.yaml is present in the upload. Similarly, `--dry-run` doesn't validate the generated tar and reports that all is fine for uploading.

Regarding the root cause for 'empty' tars:
Standard bsdtar version is 2.8.3 since OSX 10.9, and mtree is the only way to set UIDs. Before changing code, I had tried to get `pub publish` running with gnutar (installed via brew), but that also failed because it doesn't support the gname and uname options that the code uses. 

When reading through the bug thread, I was noticing that you seem to not have any problems that other users reported... that's strange. Are you sure that your tars are non-empty? Did you also try a real, non-mocked, non-dryrun publish?

If you like to investigate this new implementation, here are two commands to test on command line. 
```
# List the lib dir in mtree format
tar -cf libdir.mtree --format=mtree --options='!all, uid, gid, type' lib

# In libdir.mtree, set uid and gid to 0.
# Please note that the mtree created in my PR only contains type=file entries, while this example 
# also includes type=dir ... both ways should work, though. There's also a bug ignoring the last 
# entry if there's no line break at the end. Code in PR adds this line break.

# Create tar.gz with mtree file as input.
# FYI code in PR omits the -f option and outputs to stdout.
tar -f files.tar.gz --format=ustar --create --gzip --directory . @libdir.mtree
```

lib/src/io.dart
```diff
@@ -1034,32 +1034,48 @@ ByteStream createTarGz(List contents, {String baseDir}) {
...
-        // of tar on OS X. Passing them causes tar to exit with an error. For
-        // now, we'll just not handle large UIDs on Mac until we can come up
-        // with something.
-        // See: https://github.com/dart-lang/pub/issues/1442
-        // args.addAll(["--uname=pub", "--gname=pub"]);
+        // OSX tar flags, applicable since at least OSX 10.9 (bsdtar 2.8.3)
+        // https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/tar.1.html
+
+        // Create the file containing the list of files to compress.
+        var tempDir = createSystemTempDir();
```
nex3: Can we pass the mtree on stdin to avoid creating a temp file here, for example using `@/dev/stdin`?

lib/src/io.dart
```diff
@@ -1034,32 +1034,48 @@ ByteStream createTarGz(List contents, {String baseDir}) {
...
+
+          // The ustar format doesn't support large UIDs, so we set
+          // UIDs and GIDs to 0 via an mtree input file.
+          var uidGid = " uid=0 gid=0 type=file";
+          writeTextFile(contentsPath, "#mtree\n" + contents.join("$uidGid\n") + uidGid + "\n");
+          args.add("@${contentsPath}");
+
+          return (await startProcess("tar", args, workingDir: baseDir))
+              .stdout
+              .transform(onDoneTransformer(() => deleteEntry(tempDir)));
```
nex3: Make sure the pubspec constraint on the `async` package ensures that `onDoneTransformer` exists.

nilsdoehring: Not sure if I understand. `onDoneTransformer` is from `lib/src/utils.dart`. Also, it is referenced in L1106 already.

lib/src/io.dart
nilsdoehring: Didn't think so, but it actually works! Will change.

nilsdoehring:
@nex3 Thanks for the review, passing the mtree to stdin actually works. I changed the code, accordingly.

lib/src/io.dart
nex3: Oh, never mind then.

lib/src/io.dart
```diff
@@ -1036,46 +1036,33 @@ ByteStream createTarGz(List contents, {String baseDir}) {
...
-              .stdout
-              .transform(onDoneTransformer(() => deleteEntry(tempDir)));
-        } catch (_) {
-          deleteEntry(tempDir);
-          rethrow;
-        }
+        // The ustar format doesn't support large UIDs, so we set
+        // UIDs and GIDs to 0 via an mtree input file.
+        args.add("@/dev/stdin");
+        var uidGid = " uid=0 gid=0 type=file\n";
```
nex3: From [this page](https://www.freebsd.org/cgi/man.cgi?mtree(8)), it looks like the mtree format supports symbolic names for the user and group rather than numeric values. If we can set the symbolic names to "pub", then we'll be generating the same files on OS X as we do on Linux, which is nice. It also means manually-extracted package archives won't have root permissions, which is a bonus.

lib/src/io.dart
```diff
@@ -1036,46 +1036,33 @@ ByteStream createTarGz(List contents, {String baseDir}) {
...
+        stdin = "#mtree\n" + contents.join(uidGid) + uidGid;
       }
+
+      var process = await startProcess("tar", args);
+      process.stdin.add(UTF8.encode(stdin));
```
nex3: I think this code was originally written before `process.stdin.write()` existed, but it would be good to use it now.

lib/src/io.dart
nilsdoehring: Good catch! And it actually works!

nilsdoehring:
@nex3 thanks for referring me to the symbolic names info! I made the necessary changes. Also, these attributes are now set globally, which is a bit more elegant.

nex3:
Thanks again!

PR #1527, opened 2017-03-07
Fix a runtime type error in the validator.

nex3:
I'm not sure why this wasn't causing us problems before.


nex3:
I think the issue is more than `Iterable<T>` is (currently) assignable to `List<T>`.

PR #1530, opened 2017-03-15
run dartfmt

jakemac53:
This is quite a large change, but allows for standard development workflows like format on save without having crazy diffs when working on this package. Otherwise contributors have to disable format on save when developing on this package or use an alternative editor just for editing this package.

cc @kevmoo


munificent:
I'm sympathetic, but I think we should wait until @nex3 is back before we consider this. She does most of the maintenance on pub these days and she deliberately isn't using the formatter yet.

munificent:
After talking to @dgrove, I'm fine with this. You can go ahead and merge it in, @jakemac53.

jakemac53:
Ok, thanks! This definitely makes it easier for other contributors to work on the project :).

PR #1531, opened 2017-03-15
Exit as failure if any transformer had an error

### LINKED ISSUE #1336: Incorrect exit code if use dart2js transformer
If we have dart files with `main() {}` then dart2js transformer will hide error exit code from another transformers.

For example we have transformer which take `.md` files and throw error while transformation. Transformer correctly log error, but when `pub build` will be finished we got error code 0, but not 65. If we remove dart files then we got correct exit code 65.

Example for reproduce:
https://github.com/Aetet/transformer-issue

natebosch:
Fixes #1336

Record when we log and error and mark the entire build as failed. This
catches the cases where there are errors on assets that did not
transitively impact dart2js output - like transformers on non-Dart
files.

### DIFF
--- a/test/build/reports_failures_test.dart
+++ b/test/build/reports_failures_test.dart
@@ -23,6 +23,7 @@ class FailingTransformer extends Transformer {
 
 main() {
   integration("reports failures in transformers which don't output dart", () {
+    // Test for https://github.com/dart-lang/pub/issues/1336
     serveBarback();
 
     d.dir(appPath, [
--- a/test/build/reports_failures_test.dart
+++ b/test/build/reports_failures_test.dart
@@ -1,0 +1,44 @@
+// Copyright (c) 2013, the Dart project authors.  Please see the AUTHORS d.file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../descriptor.dart' as d;
+import '../test_pub.dart';
+
+const _failingTransformer = """
+import 'dart:async';
+
+import 'package:barback/barback.dart';
+
+class FailingTransformer extends Transformer {
+  FailingTransformer.asPlugin();
+
+  String get allowedExtensions => '.txt';
+
+  Future apply(Transform transform) {
+    throw 'FAIL!';
+  }
+}
+""";
+
+main() {
+  integration("reports failures in transformers which don't output dart", () {
+    serveBarback();
+
+    d.dir(appPath, [
+      d.pubspec({
+        "name": "myapp",
+        "transformers": ["myapp/src/transformer"],
+        "dependencies": {"barback": "any"}
+      }),
+      d.dir("lib", [
+        d.dir("src", [d.file("transformer.dart", _failingTransformer)])
+      ]),
+      d.dir("web",
+          [d.file("foo.txt", "foo"), d.file("main.dart", "void main() {}")])
+    ]).create();
+
+    pubGet();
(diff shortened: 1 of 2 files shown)


natebosch:
Sending for early feedback - I still need to see if this impacted any tests but they weren't passing for me before starting the CL...

We also need to consider whether anyone would have been relying on the old behavior and if that is something we'd like to keep compatible, my gut feeling is that this change should be pretty safe.

kevmoo:
...also would be good if this change *had* tests 😄

munificent:
This change shouldn't be necessary unless something else is going wrong. `getAllAssets()` should throw an exception if any error occurs in any transformer that runs, not just ones that touch dart2js.

natebosch:
In that case something else is wrong... I can take a look in barback to see.

What happens today is that the failure only flows through when the last Transformer (dart2js) has the asset that failed as an input.

natebosch:
@munificent It looks like at one point barback would have thrown in `getAllAssets()` if there was a failure in any Transformer, and that was changed 3 years ago: https://github.com/dart-lang/barback/commit/ea660d60b012134df07493585b87d060a26fee66#diff-cc5479bcfb96f0a899a49dc8a64a3affR141

We went form collecting all of the `BuildResult`s and checking if any had an error, to only checking if the most recently finished one had an error.

Unfortunately the commit is not explicit in either the description, comments, or code review as to why this was changed...

Do you think this was unintentional? Should I try a refactor back to collecting an aggregate `BuildResult` and using that to check for errors?

natebosch:
There also aren't any tests in Barback against the current behavior 😒 

If you think it's safe it's a pretty easy change to make. I'll file a PR on barback and look for feedback there.

munificent:
> It looks like at one point barback would have thrown in getAllAssets() if there was a failure in any Transformer, and that was changed 3 years ago: dart-lang/barback@ea660d6#diff-cc5479bcfb96f0a899a49dc8a64a3affR141

Oof, nice find.

> Unfortunately the commit is not explicit in either the description, comments, or code review as to why this was changed...

Probably something @nex3 and I talked about but didn't write down. This was during the point in time when she was making barback increasingly more stream-based. She really liked that model, but I found (and still find) it impenetrable.

> Do you think this was unintentional? Should I try a refactor back to collecting an aggregate BuildResult and using that to check for errors?

Knowing @nex3, it most likely intentional, but it may be because *not* doing it causes other problems, and not because it was a first-order goal.

You could try changing this, but I don't understand barback well enough anymore to be able to tell you how easy it would be. I find *any* changes to barback to be hard. I'm not smart enough to hack on it.

> There also aren't any tests in Barback against the current behavior 😒

😭  Usually, we're pretty good about testing stuff.

natebosch:
Added a test. PTAL

test/build/reports_failures_test.dart
```diff
@@ -0,0 +1,44 @@
...
+
+  String get allowedExtensions => '.txt';
+
+  Future apply(Transform transform) {
+    throw 'FAIL!';
+  }
+}
+""";
+
+main() {
```
kevmoo: May be worth a comment saying this regresses https://github.com/dart-lang/pub/issues/1336

natebosch: You don't think the `git blame` output will be sufficient?

natebosch: Done

kevmoo: https://github.com/dart-lang/pub/blob/master/test/version_solver_test.dart#L388

PR #1533, opened 2017-03-16
Remove dependency on pkg/uuid

kevmoo:
Use a home-rolled version


lib/src/utils.dart
```diff
@@ -824,3 +824,26 @@ void dataError(String message) => throw new DataException(message);
 Iterable/*<T>*/ combineIterables/*<T>*/(
         Iterable/*<T>*/ iter1, Iterable/*<T>*/ iter2) =>
     iter1.toList()..addAll(iter2);
+
+/// Returns a UUID in v4 format as a `String`.
```
munificent: Nit: "String" doesn't need to be code formatted.

kevmoo: I like to do that with types. Maybe I'm weird...

Issue #1537, opened 2017-03-17
StrictDependenciesValidator: check benchmark and example


kevmoo:
Just like test and tool

PR #1538, opened 2017-03-17
Validate imports/exports in benchmark and example dirs

kevmoo:
Just like test and tool
Also expanded and generalized these tests

Closes https://github.com/dart-lang/pub/issues/1537


nex3:
It was intentional that `example` *not* have this validation—some people use it to contain separate packages with their own pubspecs, which may include dependencies that aren't dependencies of the outer package.

kevmoo:
Given these are all FYI, I have no problem undoing this...

PR #1541, opened 2017-03-23
Add --compiler=dartdevc|dart2js|none flag to pub serve/build.

jakemac53:
This is somewhat of a WIP, but it does enable the basic functionality (this is why I am merging into a separate branch though). I will follow up with some more tests and also the configuration required to pass the remaining tests. 

Mostly I wanted to get this in your hands so you can take a look at the general strategy I am using and provide some early feedback.

At a high level this does a few things:

1. Adds a new transformer `DevCompilerPackageModuleTransformer` that is added to all packages.

    - This runs on each package's `lib` folder and creates a single module containing all the Dart files in that package.
    - Note that I also had to effectively disable the notion of static packages if using `--compiler=dartdevc`, since all packages have a transformer now.
    - I also had to change `getBuiltInTransformers` to return multiple phases of transformers.

2. Adds several new transformers that only apply to the root package:

    - DevCompilerResourceTransformer: Copies the dart_sdk.js script from your sdk into each non-lib directory that contains a dart file, and does the same with the require.js script if available.
    - DevCompilerEntryPointModuleTransformer: For each dart file outside of lib, this creates a module for that file and all its relative imports (`*.module.js`), as well as a bootstrap script that loads that module and calls its main (`*.js`).  

3. Added `Compiler` class and replaced all usages of `enableDart2JS` or similar parameters to use this. This is generally set by the `--compiler=X` flag.


jakemac53:
cc @natebosch @kevmoo

lib/src/barback/asset_environment.dart
```diff
@@ -167,21 +176,42 @@ class AssetEnvironment {
...
 
-  /// Gets the built-in [Transformer]s that should be added to [package].
+  /// Gets the built-in [Transformer]s or [AggregateTransformer]s that should be
+  /// added to [package].
   ///
   /// Returns `null` if there are none.
-  Iterable<Transformer> getBuiltInTransformers(Package package) {
-    // Built-in transformers only apply to the root package.
-    if (package.name != rootPackage.name) return null;
+  Iterable<Set> getBuiltInTransformers(Package package) {
```
natebosch: should we flatten this?

lib/src/barback/asset_environment.dart
```diff
@@ -167,21 +176,42 @@ class AssetEnvironment {
...
+    var transformers = <Set>[];
+    // Transitive built-in transformers apply to all packages.
+    //
+    // Today these transformers are only for dartdevc.
```
natebosch: [nit] this seems like the kind of comment that is very unlikely to be kept up to date if the information goes stale... should we omit it?

lib/src/barback/compiler_mode.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+enum CompilerMode { Dart2Js, DevCompiler, None }
```
natebosch: [nit] is there precedence within this package for capitalized enum values? Typically we use lower camel case

lib/src/barback/asset_environment.dart
jakemac53: Updated to instead describe why these need to be in their own phase (and first)

lib/src/barback/compiler_mode.dart
jakemac53: done

pubspec.yaml
```diff
@@ -14,6 +14,7 @@ dependencies:
   compiler_unsupported: "^1.22.0"
   crypto: ">=1.0.0 <3.0.0"
   glob: "^1.0.0"
+  html: "^0.13.0"
```
jakemac53: Note that once https://github.com/dart-lang/pub/issues/1544 is resolved this can be removed.

lib/src/barback/ddc_transformer.dart
```diff
@@ -340,6 +324,29 @@ require(["$appModuleName", "dart_sdk"], function(app, dart_sdk) {
...
+Future<Set<File>> _createTmpFiles(
+    Set<AssetId> ids,
+    Directory tmpDir,
+    Directory packagesDir,
+    String basePackage,
+    _InputChecker hasInput,
+    _InputReader readInput,
+    TransformLogger logger) async {
+  var files = new Set<File>();
+  await Future.wait(ids.map((id) async {
```
kevmoo: format?

pubspec.yaml
```diff
@@ -14,7 +14,6 @@ dependencies:
   compiler_unsupported: "^1.22.0"
   crypto: ">=1.0.0 <3.0.0"
   glob: "^1.0.0"
-  html: "^0.13.0"
```
kevmoo: Did you find this w/ the new pub hotness?

jakemac53: I added it in an earlier commit, but was able to remove the dependency (no special html handling at all any more - it works out of the box with all the regular dart2js workflows)

lib/src/barback/ddc_transformer.dart
```diff
@@ -0,0 +1,500 @@
...
+  return new Set<AssetId>()
+    ..add(
+        new AssetId(package, p.url.join('lib', '$package.$_summaryExtension')));
+}
+
+Future<bool> _isEntryPoint(
+    AssetId id, TransformLogger logger, _InputAsStringReader readInputAsString,
+    {Set<AssetId> seenIds}) async {
+  seenIds ??= new Set<AssetId>();
+  if (!seenIds.add(id)) return false;
```
munificent: Do you have a test that you handle export cycles?

lib/src/barback/ddc_transformer.dart
```diff
@@ -0,0 +1,500 @@
...
+    var exportId = _urlToAssetId(
+        id, (export as UriBasedDirective).uri.stringValue, logger);
+    if (exportId == null) continue;
+    if (await _isEntryPoint(exportId, logger, readInputAsString,
+        seenIds: seenIds)) return true;
+  }
+  return false;
+}
+
+AssetId _urlToAssetId(AssetId source, String url, TransformLogger logger) {
```
munificent: Likewise, lib/src/barback.dart already has something along these lines.

lib/src/barback/ddc_transformer.dart
```diff
@@ -0,0 +1,500 @@
...
+    }
+  } else {
+    // Relative path.
+    var targetPath =
+        p.url.normalize(p.url.join(p.url.dirname(source.path), uri.path));
+    return new AssetId(source.package, targetPath);
+  }
+}
+
+Future _writeFile(File file, Stream<List<int>> stream) async {
```
munificent: Is there something similar to this already in lib/src/io.dart? If so use it. If not, move this there.

lib/src/barback/ddc_transformer.dart
```diff
@@ -0,0 +1,500 @@
...
+  await file.create(recursive: true);
+  var sink = file.openWrite();
+  await sink.addStream(stream);
+  await sink.close();
+}
+
+const _summaryExtension = 'api.ds';
```
munificent: Nit: I think we generally keep top level variables and constants at the top of files.

lib/src/command/barback.dart
```diff
@@ -7,13 +7,20 @@ import 'dart:async';
...
 import '../log.dart' as log;
 import '../utils.dart';
 
 final _arrow = getSpecial('\u2192', '=>');
 
+final _compilerArgToMode = const {
+  'dart2js': CompilerMode.dart2Js,
+  'dartdevc': CompilerMode.devCompiler,
+  'none': CompilerMode.none,
+};
```
munificent: Can you turn CompilerMode into a class (enums are almost always pointless in Dart...) and put the logic to loop up a mode by name there?

lib/src/command/barback.dart
```diff
@@ -24,6 +31,12 @@ abstract class BarbackCommand extends PubCommand {
   /// The build mode.
   BarbackMode get mode => new BarbackMode(argResults["mode"]);
 
+  // The current compiler mode.
+  CompilerMode get compilerMode =>
+      argResults.options.contains('dart2js') && argResults.wasParsed('dart2js')
+          ? argResults['dart2js'] ? CompilerMode.dart2Js : CompilerMode.none
+          : _compilerArgToMode[argResults["compiler"]];
```
munificent: A nested conditional operator is not exactly reader friendly. :)

I think this chunk of code is big enough to break into a statement or two.

lib/src/command/barback.dart
```diff
@@ -44,6 +57,11 @@ abstract class BarbackCommand extends PubCommand {
         help: "Use all default source directories.",
         defaultsTo: false,
         negatable: false);
+
+    argParser.addOption("compiler",
+        allowed: const ['dartdevc', 'dart2js', 'none'],
+        defaultsTo: 'dart2js',
+        help: 'The js compiler to use to build the app.');
```
munificent: Nit: "js" -> "JavaScript".

test/compilers/allows_import_in_dart_code_test.dart
```diff
@@ -9,7 +9,7 @@ import '../test_pub.dart';
 import '../serve/utils.dart';
```
munificent: Nit: Directory names are usually singular in pub, so can you call the directory "compiler" instead of "compilers"?

lib/src/barback/asset_environment.dart
```diff
@@ -135,9 +138,15 @@ class AssetEnvironment {
   /// Constants to passed to the built-in dart2js transformer.
   final Map<String, String> environmentConstants;
 
-  /// The [Transformer]s that should be appended by default to the root
-  /// package's transformer cascade. Will be empty if there are none.
-  final _builtInTransformers = <Transformer>[];
+  /// The [Transformer]s or [AggregateTransformer]s that should be appended by
+  /// default to the root package's transformer cascade. Will be empty if there
+  /// are none.
+  final _builtInTransformers = new Set();
```
natebosch: can we get union types ;)

I think we'd have to introduce a new supertype to mark them both.

lib/src/barback/ddc_transformer.dart
```diff
@@ -0,0 +1,500 @@
...
+    }
+    ddcArgs.addAll(filesToCompile.map((f) => p
+        .relative(f.path, from: tmpDir.path)
+        .replaceFirst('packages/', 'package:')));
+    var ddcPath = p.join(sdk.path, 'bin', 'dartdevc');
+    var result =
+        await Process.run(ddcPath, ddcArgs, workingDirectory: tmpDir.path);
+    if (result.exitCode != 0) {
+      var message = 'Failed to compile package:$basePackage with dartdevc '
+          'after ${watch.elapsed}:\n\n${result.stdout}';
```
jakemac53: dartdevc doesn't output to stderr, but I will add it anyways :)

lib/src/barback/ddc_transformer.dart
```diff
@@ -0,0 +1,500 @@
...
+}
+
+Set<AssetId> _findSummaryIds(package) {
+  // TODO(jakemac): Read build.yaml if available?
+  return new Set<AssetId>()
+    ..add(
+        new AssetId(package, p.url.join('lib', '$package.$_summaryExtension')));
+}
+
+Future<bool> _isEntryPoint(
```
jakemac53: done

lib/src/barback/ddc_transformer.dart
jakemac53: not supporting this at all anyways any more :)

lib/src/barback/ddc_transformer.dart
```diff
@@ -0,0 +1,500 @@
...
+    var file = _fileForId(id, tmpDir.path, packagesDir.path);
+    await _writeFile(file, readInput(id));
+    files.add(file);
+  }));
+  return files;
+}
+
+/// Crawls from [entryId] and finds all [Asset]s that are relative through
+/// relative uris (via all [UriBasedDirective]s).
+Future<Set<AssetId>> _findRelativeIds(AssetId entryId, TransformLogger logger,
```
jakemac53: I don't see this, it looks like it just has a `CompilerInputProvider` for dart2js which can read assets by uri, but it doesn't actually follow the import graph and only handles absolute uris.

lib/src/barback/ddc_transformer.dart
```diff
@@ -0,0 +1,500 @@
...
+      ? new File(p.joinAll(
+          [packagesDir, id.package]..addAll(p.url.split(id.path).skip(1))))
+      : new File(p.joinAll([rootDir]..addAll(p.url.split(id.path))));
+}
+
+Set<AssetId> _findSummaryIds(package) {
+  // TODO(jakemac): Read build.yaml if available?
+  return new Set<AssetId>()
+    ..add(
+        new AssetId(package, p.url.join('lib', '$package.$_summaryExtension')));
```
jakemac53: Changed this to just return an iterable, and returned a list.

lib/src/barback/ddc_transformer.dart
jakemac53: There isn't anything that supports relative imports under the same top level directory though.

lib/src/barback/asset_environment.dart
jakemac53: I removed these entirely actually and just inlined the logic inside of `getBuiltInTransformers`. I was abusing these to have ordering constraints anyways, and they really don't provide value as they are just used in one place.

lib/src/barback/asset_environment.dart
```diff
@@ -509,7 +554,6 @@ class AssetEnvironment {
     // other build directories in the root package by calling
     // [serveDirectory].
     await Future.wait(graph.packages.values.map((package) async {
-      if (graph.isPackageStatic(package.name)) return;
```
jakemac53: Oh whoops, I guess I never committed that change, will do so now

jakemac53:
Closing this pull request for now at least - I have an improved version coming which is separated out into smaller chunks and can handle things like package cycles (where possible, you still can't have cycles between packages in your import graph).

PR #1552, opened 2017-03-31
Handle http errors trying to find package

UnikZ:


abarth:
LGTM

whesse:
When the pub server has an internal failure, it returns a code 500.  This should either trigger a retry of the fetch by pub, or cause pub to fail with a specific message and error code for a "temporary server failure".

I'll file a new issue for this as well, if this pull request isn't the right place for it.

nex3:
Thanks for the patch!

PR #1553, opened 2017-04-03
Support UNLICENSE file

UnikZ:


nex3:
Thanks for the change! Note that for this to work like a LICENSE file in all ways, you'll also need to modify [the pub server](https://github.com/dart-lang/pub-dartlang-dart) to recognize this license.

PR #1561, opened 2017-04-07
Log HTTP response info at WARNING level for status codes >= 400

kevmoo:
Addresses https://github.com/flutter/flutter/issues/9237
Easier than https://github.com/flutter/flutter/pull/9238

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -194,7 +194,11 @@ class _PubHttpClient extends http.BaseClient {
     response.headers
         .forEach((name, value) => responseLog.writeln(_logField(name, value)));
 
-    log.fine(responseLog.toString().trim());
+    if (response.statusCode < 400) {
+      log.fine(responseLog.toString().trim());
+    } else {
+      log.warning(responseLog.toString().trim());
+    }
   }
 
   /// Returns a log-formatted string for the HTTP field or header with the given
--- a/test/get/hosted/get_test.dart
+++ b/test/get/hosted/get_test.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:pub/src/exit_codes.dart' as exit_codes;
+import 'package:test/test.dart';
 
 import '../../descriptor.dart' as d;
 import '../../test_pub.dart';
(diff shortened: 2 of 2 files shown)


kevmoo:
CC @Hixie

This feels a tiny bit weird, since the header output is just dumped to the terminal, but I think it'll be useful for us to figure out issues that may be happening with pub

BEFORE

```console
$ PUB_HOSTED_URL=https://google.com dart bin/pub.dart --verbosity=warning get
Could not find package analyzer at https://kemoohax-dot-dartlang-pu.appspot.com.
Depended on by:
- pub
```

AFTER

```console
$ PUB_HOSTED_URL=https://google.com dart bin/pub.dart --verbosity=warning get
HTTP response 404 Not Found for GET https://kemoohax-dot-dartlang-pu.appspot.com/api/packages/analyzer
took 0:00:00.156675
content-type: text/html; charset=UTF-8
alt-svc: quic=":443"; ma=2592000; v="37,36,35"
date: Fri, 07 Apr 2017 19:32:42 GMT
content-length: 1551
Could not find package analyzer at https://kemoohax-dot-dartlang-pu.appspot.com.
Depended on by:
- pub
```

kevmoo:
CC @yjbanov – I think you've been paying attention here, too

kevmoo:
> Warnings are enabled by default in pub and I don't think spewing HTTP headers are useful for most users most of the time. When they want them, it's easy to re-run pub with a higher verbosity level.

Understood. We're trying to get more info on sporadic failures – mostly on bots – where rerunning isn't practical and where we want to capture the sporadic information.

Running w/ full verbosity crashed Travis for our Flutter friends because the log got too big in the non-failure case.

Other ideas, @munificent ?

Hixie:
Unfortunately it's not as easy as one might think. We tried doing that to Flutter and the results were that our builds failed because there was too much output.

yjbanov:
How about `--verbosity=http`?

kevmoo:
> How about `--verbosity=http`?

You really want `--verbosity=http-warnings` – you only want HTTP info on >= 400...maybe just >= 500, honestly

yjbanov:
`--verbosity=http-warnings` LGTM

munificent:
`--verbosity=io` already exists and is the level we use to log details of more other IO (file and networking) operations. I'd stick it there.

Hixie:
It's not so much that it crashed Travis so much as Travis bails once you hit 10,000 lines of output.

kevmoo:
@nex3 thoughts on this?

I'm trying to help out the flutter folks without making them generate 100mb of log data to catch error cases...

Issue #1565, opened 2017-04-13
Create a ModuleConfigTransformer


jakemac53:
This should be an aggregate transformer that computes the ideal set of modules for a package and outputs a `.moduleConfig` file under each top level directory that lists all the modules in that directory.

This is a bit open ended but there are a few hard constraints:

- Modules can't cross package boundaries
- Modules can't cross top level dirs in the same package (lib/web/etc)
- Strongly connected components must be in the same module

jakemac53:
closed by https://github.com/dart-lang/pub/pull/1577

PR #1574, opened 2017-04-24
Add a module computer class which can compute modules for a package

jakemac53:
This is the first step towards resolving https://github.com/dart-lang/pub/issues/1565, it does everything but implement the actual transformer to create the `.moduleConfig` files.

The approach first takes the strongly connected components, and then merges modules based on the public entry points that import them. It does the merging by simply merging each "private" module into a new module whose name is a combination of all the entry point modules that transitively import that module. Entry point modules ("public" modules) are never merged.


lib/src/barback/dartdevc/module_config_transformer.dart
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
jakemac53: Whoops I meant to exclude this file and add it as a followup - I will do that tomorrow so please ignore for now.

lib/src/barback/dartdevc/util.dart
```diff
@@ -0,0 +1,46 @@
...
+          'Unable to resolve import. Only package: paths and relative '
+          'paths are supported, got `$url`.');
+    }
+  } else {
+    // Relative path.
+    var targetPath =
+        p.url.normalize(p.url.join(p.url.dirname(source.path), uri.path));
+    return new AssetId(source.package, targetPath);
+  }
+}
```
munificent: BarbackServer has `urlToId()` which is similar to this. There's also lib/src/barback.dart, which has `packagesUrlToId()`. You should be able to reuse at least some of the latter, it not all of it.

If you still need this function, move it into lib/src/barback.dart and give it a name that clarifies how it differs from the above.

lib/src/barback/dartdevc/util.dart
```diff
@@ -0,0 +1,46 @@
...
+        'under a directory.');
+  }
+  return parts.first;
+}
+
+/// Convert [url] found in [source] to an [AssetId].
+///
+/// Throws an [ArgumentError] if an [AssetId] cannot be created.
+///
+/// Returns [null] for `dart:` uris.
```
munificent: It would be good to note why it does this.

test/barback/dartdevc/module_computer_test.dart
```diff
@@ -0,0 +1,283 @@
...
+            srcs: ['a|lib/a.dart'],
+            directDependencies: ['a|lib/b.dart', 'a|lib/src/c.dart'])),
+        equalsModule(makeModule(
+            package: 'a',
+            name: 'lib__b',
+            srcs: ['a|lib/b.dart'],
+            directDependencies: ['a|lib/src/c.dart'])),
+        equalsModule(makeModule(
+            package: 'a',
+            name: 'lib__a\$lib__b',
```
munificent: What does this name mean? What does the module containing c and d have to do with a and b?

test/barback/dartdevc/module_computer_test.dart
```diff
@@ -0,0 +1,283 @@
...
+          ''',
+        'a|lib/src/c.dart': '''
+            import 'package:a/a.dart';
+          ''',
+      });
+
+      var expectedModules = [
+        equalsModule(makeModule(
+            package: 'a',
+            name: 'lib__a',
```
munificent: Is the name "a" here because that's the package name or one of the library names? How about using a different name for the package since make the distinction clearer? (Most other pub tests use "myapp".)

test/barback/dartdevc/util.dart
```diff
@@ -21,28 +37,73 @@ Set<AssetId> makeAssetIds({String package, String topLevelDir}) =>
...
+  assert(srcs == null || topLevelDir == null);
+  AssetId toAssetId(dynamic id) {
+    if (id is AssetId) return id;
+    assert(id is String);
+    return new AssetId.parse(id);
+  }
+
+  var id = makeModuleId(package: package, name: name);
+  srcs ??= makeAssetIds(package: id.package, topLevelDir: topLevelDir);
+  var assetIds = new Set<AssetId>()..addAll(srcs.map(toAssetId));
```
munificent:
```dart
srcs.map(toAssetId).toSet()
```

test/barback/dartdevc/module_computer_test.dart
jakemac53: It's based on the dart file name - renamed the package to "myapp".

test/barback/dartdevc/util.dart
jakemac53: done

lib/src/barback/dartdevc/util.dart
jakemac53: done

jakemac53:
I realized that I forgot to add tests for modules under `web`, adding those now.

lib/src/barback.dart
```diff
@@ -87,3 +89,43 @@ AssetId packagesUrlToId(Uri url) {
...
+/// `package:` imports and relative imports.
+///
+/// Returns [null] for `dart:` uris since they cannot be referenced properly
+///  by an `AssetId`.
+///
+/// Throws an [ArgumentError] if an [AssetId] can otherwise not be created.
+AssetId importUriToAssetId(AssetId source, String importUri) {
+  var parsedUri = Uri.parse(importUri);
+  if (parsedUri.isAbsolute) {
+    if (parsedUri.scheme == 'package') {
```
munificent: Maybe use a switch instead of chained ifs?

lib/src/io.dart
```diff
@@ -1132,3 +1132,21 @@ class PubProcessResult {
 
   bool get success => exitCode == exit_codes.SUCCESS;
 }
+
+/// Returns the top level directory in [path].
+///
+/// Throws an [ArgumentError] if [path] is just a filename with no directory.
+String topLevelDir(String pathString) {
```
munificent: Instead of `path`, I'd do `uri` to make it clear this works on URI-formatted paths, not local file system ones.

test/barback/dartdevc/module_computer_test.dart
munificent: By "entrypoint module", do you mean "entrypoint" in the Dart sense of a "script with main()"? Or you mean like a public module?

lib/src/barback.dart
jakemac53: done, also added a better comment highlighting the cases in which this method will throw

lib/src/io.dart
jakemac53: done

jakemac53:
re: your other question about "entrypoint" (I can't respond on the outdated diff):

I mean either a public module with files under `lib` (and at least one not under `lib/src`) or an actual entrypoint script with a `main` outside of `lib`. See description here https://github.com/dart-lang/pub/pull/1574/files#diff-bcdcea609fd3947070ec6894daf5dbcaR40.

Issue #1568, opened 2017-04-13
Update the DartDevcTransformers to respect module configs


jakemac53:
Instead of just globbing all of lib it should read in the `.moduleConfig` files from https://github.com/dart-lang/pub/issues/1565 and use the corresponding linked summaries from https://github.com/dart-lang/pub/issues/1565 to create modules.

PR #1580, opened 2017-04-27
Add the DartDevcModuleTransformer

jakemac53:
This transformer compiles modules to js based on `.moduleConfig` files using dartdevc and linked summaries from previous phases.

Closes https://github.com/dart-lang/pub/issues/1568.


lib/src/barback/dartdevc/dartdevc_module_transformer.dart
```diff
@@ -0,0 +1,120 @@
...
+
+/// Compiles [module] using the `dartdevc` binary from the SDK to a relative
+/// path under the package that looks like `$outputDir/${module.id.name}.js`.
+Future _createDartdevcModule(
+    Module module,
+    String outputDir,
+    TempEnvironment tempEnv,
+    Set<AssetId> linkedSummaryIds,
+    Transform transform) async {
+  var logger = transform.logger;
```
munificent: This local variable doesn't add a lot of value. Maybe just call `transform.logger` directly below?

lib/src/barback/dartdevc/dartdevc_module_transformer.dart
```diff
@@ -0,0 +1,120 @@
...
+/// be read in to create the modules.
+class DartDevcModuleTransformer extends Transformer {
+  @override
+  String get allowedExtensions => moduleConfigName;
+
+  DartDevcModuleTransformer();
+
+  @override
+  Future apply(Transform transform) async {
+    TempEnvironment tempEnv;
```
munificent: It looks like there's a lot of similarity in this code and [this stuff](https://github.com/dart-lang/pub/pull/1579/files#diff-76378714c8a4b4f95e6ef458e6269a20R36). Can it be refactored to reuse some of it? Maybe move more logic into Module itself?

lib/src/barback/dartdevc/dartdevc_module_transformer.dart
```diff
@@ -0,0 +1,120 @@
...
+    Transform transform) async {
+  var logger = transform.logger;
+  var jsOutputId = new AssetId(
+      module.id.package, p.url.join(outputDir, '${module.id.name}.js'));
+  var jsOutputFile = tempEnv.fileFor(jsOutputId);
+  var sdk_summary = p.url.join(sdkDir.path, 'lib/_internal/ddc_sdk.sum');
+  var request = new WorkRequest();
+  request.arguments.addAll([
+    '--dart-sdk-summary=$sdk_summary',
+    // TODO(jakemac53): Remove when no longer needed.
```
munificent: Can you file a tracking bug for this and mention it here?

lib/src/barback/dartdevc/dartdevc_module_transformer.dart
```diff
@@ -0,0 +1,120 @@
...
+    '--library-root=${p.dirname(jsOutputFile.path)}',
+    '--summary-extension=${linkedSummaryExtension.substring(1)}',
+    '-o',
+    jsOutputFile.path,
+  ]);
+  // Add all the linked summaries as summary inputs.
+  for (var id in linkedSummaryIds) {
+    request.arguments.addAll(['-s', tempEnv.fileFor(id).path]);
+  }
+  // Add url mappings for all the package: files to tell ddc where to find them.
```
munificent: Nits: "url" -> "URL". "ddc" -> "DDC".

lib/src/barback/dartdevc/dartdevc_module_transformer.dart
```diff
@@ -0,0 +1,120 @@
...
+  var response = await dartdevcDriver.doWork(request);
+
+  // TODO(jakemac53): Fix the ddc worker mode so it always sends back a bad
+  // status code if something failed. Today we just make sure there is an output
+  // js file to verify it was successful.
+  if (response.exitCode != EXIT_CODE_OK || !jsOutputFile.existsSync()) {
+    // We only log warnings for ddc modules because technically they don't all
+    // need to compile successfully, only the ones imported by an entrypoint do.
+    logger.warning('Error compiling dartdevc module: ${module.id}.\n'
+        '${response.output}');
```
munificent: This seems... dubious to me. I see the value in it, but I'm not sold.

By that same argument, we could warn if there are errors in methods, as long as you don't call them. But, in a statically typed language, I think users expect "zero errors" to mean "all the code correctly compiled".

Do you know for certain that we really need to be loose here? Is there a way we can let users control it?

lib/src/barback/dartdevc/workers.dart
```diff
@@ -13,4 +13,8 @@ final analyzerDriver = new BazelWorkerDriver(() => Process.start(
     p.join(sdkDir.path, 'bin', 'dartanalyzer'),
     ['--build-mode', '--persistent_worker']));
 
+/// Manages a shared set of persistent dartdevc workers.
+final dartdevcDriver = new BazelWorkerDriver(() => Process
+    .start(p.join(sdkDir.path, 'bin', 'dartdevc'), ['--persistent_worker']));
```
munificent: Is this process ever shut down?

jakemac53: The `BazelWorkerDriver` manages the processes, it has some options for how many idle workers to keep alive and also max number of active workers to spawn, currently I am just using the defaults (4 for both).

All the processes are shut down when the pub process shuts down.

lib/src/barback/dartdevc/dartdevc_module_transformer.dart
jakemac53: done

lib/src/barback/dartdevc/dartdevc_module_transformer.dart
jakemac53: done (change to DartDevc though for consistency)

lib/src/barback/dartdevc/dartdevc_module_transformer.dart
jakemac53: The main reason for this is that today we compile all packages you depend on (including dev dependencies). Many of these packages probably are not a part of your actual web app, and they don't need to compile properly (although ideally they still would).

If you are using dart2js you can effectively have some bad entrypoints/code you depend on as long as everything in your actual app works - and I wanted to mirror that behavior. Dart2js does this by being lazy, so you only get errors if you actually request a bad entrypoint. For ddc we can't be lazy (right now at least), so I chose to just log a warning instead.

What if we made it an error in release mode and a warning in debug mode?

lib/src/barback/dartdevc/workers.dart
jakemac53: It actually spins up a few of these (4 by default). They are shut down when pub shuts down (we want to keep them hot).

Eventually I think we might want to make the number configurable, the BazelWorkerDriver accepts both a max number of idle workers and a max number of total workers, both default to 4 (same as bazel itself).

PR #1581, opened 2017-04-27
Add the DartDevcResourceTransformer and unit tests

jakemac53:
This copies the required support files for dartdevc from the sdk into each entrypoint directory. These include the compiled dart sdk and require.js since we are using amd (at least for now).


lib/src/barback/dartdevc/dartdevc_resource_transformer.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'dart:async';
+import 'dart:io';
+
+import 'package:barback/barback.dart';
+import 'package:cli_util/cli_util.dart' as cli_util;
+import 'package:path/path.dart' as p;
+
+import '../../io.dart';
+
+/// Copies the `dart_sdk.js` and `require.js` amd files from the SDK into each
```
munificent: Nit: "amd" -> "AMD".

lib/src/barback/dartdevc/dartdevc_resource_transformer.dart
```diff
@@ -0,0 +1,53 @@
...
+  String classifyPrimary(AssetId id) {
+    if (p.extension(id.path) != '.dart') return null;
+    var dir = topLevelDir(id.path);
+    if (dir == 'lib' || dir == 'bin') return null;
+    return p.url.dirname(id.path);
+  }
+
+  @override
+  Future apply(AggregateTransform transform) async {
+    var sdk = cli_util.getSdkDir();
```
munificent: It makes me sad that this isn't a getter named `sdkDir`, but I guess that's out of scope for this patch.

lib/src/barback/dartdevc/dartdevc_resource_transformer.dart
jakemac53: done

lib/src/barback/dartdevc/dartdevc_resource_transformer.dart
jakemac53: acknowledged

PR #1585, opened 2017-05-04
Some windows fixes

jakemac53:
Also deleted temp_environment.dart which snuck back in during a bad merge :(.


lib/src/barback/dartdevc/workers.dart
```diff
@@ -8,13 +8,16 @@ import 'package:bazel_worker/driver.dart';
 import 'package:cli_util/cli_util.dart' as cli_util;
 import 'package:path/path.dart' as p;
 
+String get _batOnWindows => Platform.isWindows ? '.bat' : '';
```
munificent: Maybe something like `_scriptExtension`?

jakemac53: done

Issue #1586, opened 2017-05-04
Angular Transformer gets errors with --compiler=dartdevc


jakemac53:
When running the angular components example app (https://github.com/dart-lang/angular_components_example) I get some errors from the angular2 transformer.

I think this is probably related to `isPackageStatic` always returning false under dartdevc mode, not something related to the actual dartdevc transformers.

PR #1588, opened 2017-05-10
fix listFiles for precompiled package dependencies

jakemac53:
Fixes https://github.com/dart-lang/pub/issues/1586


jakemac53:
Added a test, but it seems to actually passing before and after this change :(. Investigating further...

jakemac53:
Ok, added a unit test which does in fact fail before and pass now. Leaving the other test in place as it does provide some value as well even if it was passing before.

PR #1589, opened 2017-05-10
only support listing the lib dir for cached packages

jakemac53:
`CachedPackage` previously would fall back on serving from the original hosted package cache if you request files outside of lib, but that gives the untransformed sources which I don't think is ever correct to do.

This makes it an error to request anything outside of `lib` from a `CachedPackage`. As far as I know this should be fine (and the existing tests do pass), since the only package you should request files outside of lib from is the root package, and that package is never cached since it is always a path dependency?

Let me know if you think there is a use case this would break, merging this isn't super urgent so I will wait for @nex3 to comment.

### DIFF
--- a/lib/src/cached_package.dart
+++ b/lib/src/cached_package.dart
@@ -54,16 +54,13 @@ class CachedPackage extends Package {
   /// is within a cached directory, but not otherwise.
   List<String> listFiles(
       {String beneath, recursive: true, bool useGitIgnore: false}) {
-    if (beneath == null) {
-      return super.listFiles(recursive: recursive, useGitIgnore: useGitIgnore);
+    if (beneath == null || !_pathInCache(beneath)) {
+      throw new UnsupportedError(
+          'Cached packages can only list their `lib` dir.');
     }
 
-    if (_pathInCache(beneath)) {
-      return listDir(p.join(_cacheDir, beneath),
-          includeDirs: false, recursive: recursive);
-    }
-    return super.listFiles(
-        beneath: beneath, recursive: recursive, useGitIgnore: useGitIgnore);
+    return listDir(p.join(_cacheDir, beneath),
+        includeDirs: false, recursive: recursive);
   }
 
   /// Returns whether [relativePath], a path relative to the package's root,
--- a/test/cached_package_test.dart
+++ b/test/cached_package_test.dart
@@ -13,37 +13,54 @@ import 'descriptor.dart' as d;
 import 'test_pub.dart';
 
 main() {
-  // Regression test for https://github.com/dart-lang/pub/issues/1586.
-  integration('Can list the cached lib dir and compute relative paths', () {
-    d.dir('cache', [
+  group('CachedPackage', () {
+    CachedPackage cachedPackage;
+
+    setUpCachedPackage() {
+      d.dir('cache', [
+        d.dir('app', [
+          d.dir('lib', [
+            d.file('cached.txt', 'hello'),
+            d.file('original.txt', 'world'),
+          ]),
+        ]),
+      ]).create();
       d.dir('app', [
         d.dir('lib', [
-          d.file('cached.txt', 'hello'),
-          d.file('original.txt', 'world'),
-        ]),
-      ]),
-    ]).create();
-    d.dir('app', [
-      d.dir('lib', [
-        d.file('original.txt'),
-      ])
-    ]).create();
+          d.file('original.txt'),
+        ])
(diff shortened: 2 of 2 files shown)


kevmoo:
@nex3 ?

nex3:
Something about this smells fishy to me. How does `listDir()` work before this change on transformed, uncached packages?

jakemac53:
> Something about this smells fishy to me. How does listDir() work on transformed, uncached packages before this change?

I think it just reads from the regular pub cache (not the `.pub/...` dir in the package like a `CachedPackage`).

nex3:
But that'll give non-transformed data, right? Are we relying on that anywhere?

jakemac53:
See https://github.com/dart-lang/pub/pull/1588 which is what prompted this pr. It wouldn't surprise me if this is also broken for regular non-cached but transformed packages as well.

nex3:
What's the plan for fixing that? It seems like we need a way to fall back to listing via Barback in order to make this totally resilient. And if we do that, it seems like a naive top-level `listDir()` is *always* going to break in a Barback context. Is there a way we can make that it more statically clear that that's not allowed?

jakemac53:
I guess I don't fully understand in which context `listDir()` is _meant_ to be used. Today it _is_ used by the `PubPackageProvider` to implement `getAllAssetIds` [here](https://github.com/dart-lang/pub/blob/master/lib/src/barback/pub_package_provider.dart#L160). This is called by `barback.getAllAssetIds` which is specifically what I was trying to use and was getting untransformed sources from.

nex3:
> Today it _is_ used by the `PubPackageProvider` to implement `getAllAssetIds` [here](https://github.com/dart-lang/pub/blob/master/lib/src/barback/pub_package_provider.dart#L160). This is called by `barback.getAllAssetIds` which is specifically what I was trying to use and was getting untransformed sources from. 

That's only for a `StaticPackageProvider`, which is guaranteed not to need any active transformation.

jakemac53:
> That's only for a StaticPackageProvider, which is guaranteed not to need any active transformation.

Right, but they might have some already applied transformation (in the case of a `CachedPackage`). In that case listing anything other than the lib dir really isn't valid, because you won't get transformed sources. That is why this cl starts making that an error (it doesn't actually happen today as far as I can tell).

nex3:
For a `CachedPackage`, it seems like listing any part of it is valid—only the `lib/` directory can be transformed (as far as downstream packages are concerned), and the rest of the assets are valid as-written.

But I'm also wondering if we call `listDir()` at all for a dependency in a barback context, since it seems like that would be invalid for packages that are transformed but not cached.

jakemac53:
[barback:src/barback.dart#105](https://github.com/dart-lang/barback/blob/f33c25d48eb1deeaed76ff06bc96d94dc26530c5/lib/src/barback.dart#L105) => [barback:lib/src/graph/package_graph.dart#L178](https://github.com/dart-lang/barback/blob/f33c25d48eb1deeaed76ff06bc96d94dc26530c5/lib/src/graph/package_graph.dart#L178) => [barback:lib/src/graph/static_asset_cascade.dart#L50](https://github.com/dart-lang/barback/blob/f33c25d48eb1deeaed76ff06bc96d94dc26530c5/lib/src/graph/static_asset_cascade.dart#L50) => [pub:lib/src/barback/pub_package_provider.dart](https://github.com/dart-lang/pub/blob/master/lib/src/barback/pub_package_provider.dart#L159) is a path that gets you to this code in a barback context.

nex3:
I mean, do we call `listDir()` for a dependency that runs transformers dynamically (that is, not through `StaticPackageProvider`/`PackageCache`)?

jakemac53:
> I mean, do we call listDir() for a dependency that runs transformers dynamically (that is, not through StaticPackageProvider/PackageCache)?

As far as I can tell, no. The `StaticAssetCascade` is the only path I could find that ends up using `getAllAssetIds` which takes you through the path I linked above..

nex3:
I don't think the code in this PR is the right way to add safety here. A `CachedPackage` is one situation in which listing above `lib/` *is* valid, because nothing else in the package can be transformed. If we can find a way to ensure that `listDir()` isn't called for dynamically-transformed packages, though, that might be useful.

jakemac53:
I am fine with closing this PR, but I don't agree that listing anything above `lib/` is valid for a `CachedPackage`. That will fall back on the default implementation which reads the untransformed sources. In practice this never happens, so it doesn't matter all that much.

PR #1590, opened 2017-05-10
Move shared code into shared libraries, clean up args.

jakemac53:
For the most part this is just moving code around, with some minor cleanup.

The apis no longer take a `Transform` for any of these methods so they are more general, and they all return either a `Map<AssetId, Future<Asset>>` or `Map<AssetId, Asset>` synchronously.

### DIFF
--- a/lib/src/barback/dartdevc/dartdevc.dart
+++ b/lib/src/barback/dartdevc/dartdevc.dart
@@ -37,7 +37,7 @@ Map<AssetId, Future<Asset>> bootstrapDartDevcEntrypoint(
     AssetId dartEntrypointId, BarbackMode mode, ModuleReader moduleReader) {
   var bootstrapId = dartEntrypointId.addExtension('.bootstrap.js');
   var jsEntrypointId = dartEntrypointId.addExtension('.js');
-  var jsMapEntrypointId = jsEntrypointId.addExtension('.js.map');
+  var jsMapEntrypointId = jsEntrypointId.addExtension('.map');
 
   var outputCompleters = <AssetId, Completer<Asset>>{
     bootstrapId: new Completer<Asset>(),
@@ -120,8 +120,10 @@ Map<AssetId, Future<Asset>> createDartdevcModule(
     logError(String message)) {
   var outputCompleters = <AssetId, Completer<Asset>>{
     module.id.jsId: new Completer<Asset>(),
-    module.id.jsSourceMapId: new Completer<Asset>(),
   };
+  if (mode == BarbackMode.DEBUG) {
+    outputCompleters[module.id.jsSourceMapId] = new Completer<Asset>();
+  }
 
   () async {
     var jsOutputFile = scratchSpace.fileFor(module.id.jsId);
@@ -181,8 +183,12 @@ Map<AssetId, Future<Asset>> createDartdevcModule(
     // status code if something failed. Today we just make sure there is an output
     // js file to verify it was successful.
     if (response.exitCode != EXIT_CODE_OK || !jsOutputFile.existsSync()) {
-      logError(
-          'Error compiling dartdevc module: ${module.id}.\n${response.output}');
+      var message =
+          'Error compiling dartdevc module: ${module.id}.\n${response.output}';
+      logError(message);
+      outputCompleters.values.forEach((completer) {
+        completer.completeError(message);
+      });
     } else {
       outputCompleters[module.id.jsId].complete(
           new Asset.fromBytes(module.id.jsId, jsOutputFile.readAsBytesSync()));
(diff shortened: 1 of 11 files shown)


lib/src/barback/dartdevc/dartdevc.dart
```diff
@@ -0,0 +1,231 @@
...
+/// with two additional JS files:
+///
+/// * A `$dartEntrypointId.js` file which is the main entrypoint for the app. It
+///   injects a script tag whose src is `require.js` and whose `data-main`
+///   attribute points at a `$dartEntrypointId.bootstrap.js` file.
+/// * A `$dartEntrypointId.bootstrap.js` file which invokes the top level `main`
+///   function from the entrypoint module, after performing some necessary SDK
+///   setup.
+///
+/// In debug mode an empty sourcemap will be output for the entrypoint js file
```
munificent: Nit: "js" -> "JS", here and below. :)

lib/src/barback/dartdevc/dartdevc.dart
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  () async {
+    var module = await moduleReader.moduleFor(dartEntrypointId);
+
+    // The path to the entrypoint js module as it should appear in the call to
+    // `require` in the bootstrap file.
+    var moduleDir = topLevelDir(dartEntrypointId.path);
+    var appModulePath = p.relative(p.join(moduleDir, module.id.name),
+        from: p.dirname(dartEntrypointId.path));
```
munificent: Are these local file paths or URIs? Should these be `p.url` here? What about below?

lib/src/barback/dartdevc/dartdevc.dart
```diff
@@ -0,0 +1,231 @@
...
+    // This is used to invoke `main()` from within the bootstrap script.
+    //
+    // TODO(jakemac53): Sane module name creation, this only works in the most
+    // basic of cases.
+    //
+    // See https://github.com/dart-lang/sdk/issues/27262 for the root issue which
+    // will allow us to not rely on the naming schemes that dartdevc uses
+    // internally, but instead specify our own.
+    var appModuleScope = p.url
+        .split(p.withoutExtension(
```
munificent: p.url?

lib/src/barback/dartdevc/dartdevc.dart
```diff
@@ -0,0 +1,231 @@
...
+/// immediately what assets will be output.
+Map<AssetId, Future<Asset>> createDartdevcModule(
+    Module module,
+    ScratchSpace scratchSpace,
+    Set<AssetId> linkedSummaryIds,
+    Map<String, String> environmentConstants,
+    BarbackMode mode,
+    logError(String message)) {
+  var outputCompleters = <AssetId, Completer<Asset>>{
+    module.id.jsId: new Completer<Asset>(),
```
munificent: Nit: I think you can leave off the `<Asset>` here. It should be inferred.

lib/src/barback/dartdevc/dartdevc.dart
```diff
@@ -0,0 +1,231 @@
...
+    if (response.exitCode != EXIT_CODE_OK || !jsOutputFile.existsSync()) {
+      var message =
+          'Error compiling dartdevc module: ${module.id}.\n${response.output}';
+      logError(message);
+      outputCompleters.values.forEach((completer) {
+        completer.completeError(message);
+      });
+    } else {
+      outputCompleters[module.id.jsId].complete(
+          new Asset.fromBytes(module.id.jsId, jsOutputFile.readAsBytesSync()));
```
munificent: Can you just return `Asset.fromPath()` here and delay reading the file off disc until needed?

lib/src/barback/dartdevc/dartdevc.dart
jakemac53: done

lib/src/barback/dartdevc/dartdevc.dart
jakemac53: Good catch - p.url is correct.

lib/src/barback/dartdevc/dartdevc.dart
jakemac53: yep

lib/src/barback/dartdevc/dartdevc.dart
jakemac53: Good point, done (also above)

lib/src/barback/dartdevc/dartdevc.dart
jakemac53: The problem is the file can get deleted at any time, so this is really the only safe thing to do (without a ton more logic at least).

PR #1595, opened 2017-05-12
start building transitive js deps eagerly

jakemac53:
Whenever you request a JS module, this will now start a build for all transitive js modules. If they have already been built they will be skipped using the normal caching strategy.


lib/src/barback/dartdevc/dartdevc_environment.dart
```diff
@@ -145,6 +146,13 @@ class DartDevcEnvironment {
       var jsId = id.extension == '.map' ? id.changeExtension('') : id;
       assets = createDartdevcModule(jsId, _moduleReader, _scratchSpace,
           _environmentConstants, _mode, logError);
+      // Pre-emptively start building all transitive js deps under the
+      // assumption they will be needed in the near future.
+      () async {
+        var module = await _moduleReader.moduleFor(jsId);
+        var deps = await _moduleReader.readTransitiveDeps(module);
+        deps.forEach((moduleId) => getAssetById(moduleId.jsId));
+      }();
```
munificent: What happens if an error occurs here?

jakemac53: Bad things - updated to run in an error handling zone.

lib/src/barback/dartdevc/dartdevc_environment.dart
```diff
@@ -144,6 +145,18 @@ class DartDevcEnvironment {
       var jsId = id.extension == '.map' ? id.changeExtension('') : id;
       assets = createDartdevcModule(jsId, _moduleReader, _scratchSpace,
           _environmentConstants, _mode, logError);
+      // Pre-emptively start building all transitive js deps under the
```
munificent: "JS". :)

PR #1596, opened 2017-05-12
increase timeout for require js modules for initial loads

jakemac53:


lib/src/barback/dartdevc/dartdevc.dart
```diff
@@ -100,6 +100,7 @@ Map<AssetId, Future<Asset>> bootstrapDartDevcEntrypoint(
 
     var bootstrapContent = '''
 require.config({
+    waitSeconds : 30,
```
munificent: Nit: No space before `:`.

PR #1597, opened 2017-05-12
add an integration test for pub run test with ddc

jakemac53:


test/barback/dartdevc/pub_run_test.dart
```diff
@@ -0,0 +1,72 @@
...
+      ])
+    ]).create();
+
+    pubGet();
+    pubServe(args: ['test', '--compiler', 'dartdevc']);
+    schedule(() {
+      var testPort = serverPorts['test'];
+      expect(testPort, isNotNull);
+      var process = pubRun(args: ['test:test', '--pub-serve=$testPort']);
+      process.shouldExit(0);
```
munificent: This will exit with exit code zero even if it does nothing, won't it? Maybe test that it produces some output around the number of tests that passed?

jakemac53:
I think this test does provide value, as its one of the easiest ways to confirm that an app in general actually works (a test is an app after all).

However, it looks like I get an error from within the bowels of scheduleTest when running through the test runner even though it works fine if invoked directly. I don't really want to fight that right now so I am going to close this pr for now.

PR #1600, opened 2017-05-15
better error handling, especially around uncaught errors

jakemac53:
- public entrypoints now use runZoned to make sure we don't leak exceptions
- we now complete with real errors instead of null for most things
- error logging is handled at a higher level now based on caught exceptions


jakemac53:
Closing for now after discussion with @munificent, going to see if we can avoid all the runZoned madness entirely

PR #1601, opened 2017-05-15
Results instead of futures

jakemac53:


lib/src/barback/dartdevc/dartdevc.dart
```diff
@@ -65,40 +65,46 @@ Map<AssetId, Future<Asset>> bootstrapDartDevcEntrypoint(
   }
 
   () async {
-    var module = await moduleReader.moduleFor(dartEntrypointId);
+    try {
+      var module = await moduleReader.moduleFor(dartEntrypointId);
+      if (module == null) {
+        outputCompleters.values.forEach((c) =>
```
natebosch: [nit] consider using for loops rather than forEach throughout

https://www.dartlang.org/guides/language/effective-dart/usage#avoid-using-iterableforeach-with-a-function-literal

lib/src/barback/dartdevc/dartdevc_environment.dart
```diff
@@ -253,24 +255,26 @@ AssetId _entrypointDartId(AssetId id) {
 class _AssetCache {
   /// [Asset]s are indexed first by package and then path, this allows us to
```
natebosch: have you measured the performance difference against
```
for(var assetId in _assets.keys) {
  if(assetId.package == packageName) _assets.remove(packageName);
}
```

It makes the rest of the class a lost simpler if we avoid the nested map...

lib/src/barback/dartdevc/dartdevc.dart
jakemac53: Refactored the shared logic for this out to a method, and used for/each.

lib/src/barback/dartdevc/dartdevc_environment.dart
jakemac53: I haven't, but I don't think this introduces undue complexity. The extra logic is confined to `_getResult` and `[]=`, and it simplifies `invalidatePackage` a bit.

The number of assets in this map is O(#modules), which isn't generally going to be that large, but if we introduce other module schemes (like one module per strongly connected component) it could blow up fairly quickly.

PR #1608, opened 2017-05-18
Allow modules that aren't reachable from entrypoints

jakemac53:
This came up in https://github.com/dart-lang/test/pull/596, and apparently can come up in some angular apps as well (they have some deprecated things you can import from src directly which aren't reachable from a public entrypoint).

These modules will be very fine grained (one per strongly connected component), so they aren't super efficient but they at least allow this type of thing to work.


jakemac53:
cc @zoechi

PR #1612, opened 2017-05-22
add travis config

jakemac53:


jakemac53:
Actually I am just going to close this because even though analyzer/dartfmt checks provide benefit not running tests probably gives a false sense of security that we don't want.

PR #1615, opened 2017-05-22
add an e2e_test for dartdevc

jakemac53:
To run this test you need to run `pub serve --compiler=dartdevc` and then `pub run test -P e2e`.

I couldn't come up with anything better right now in terms of validating the full end to end workflow for dartdevc. A simple web test like this validates that applications can actually run (with relative and package  imports), which has a lot of value.

The main issue with this test imo is it isn't very discoverable, we could add some sort of high level test.sh script or more ideally integrate with travis and make sure it runs there. However I was having issues getting travis running (lots of errors regarding tar it seemed like).

### DIFF
--- a/dart_test.yaml
+++ b/dart_test.yaml
@@ -1,0 +1,6 @@
+presets:
+  e2e:
+    paths: [test/e2e]
+    platforms:
+      - chrome
+    pub_serve: 8080
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -43,3 +43,7 @@ dev_dependencies:
 
 environment:
   sdk: ">=1.24.0-dev.2.0 <2.0.0"
+
+transformers:
+- test/pub_serve:
+    $include: test/e2e/**/*_test.dart
--- a/test/e2e/pub_run_test/hello.dart
+++ b/test/e2e/pub_run_test/hello.dart
@@ -1,0 +1,5 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+String get hello => 'Hello';
--- a/test/e2e/pub_run_test/pub_run_test_test.dart
+++ b/test/e2e/pub_run_test/pub_run_test_test.dart
@@ -1,0 +1,13 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+@TestOn('browser')
+import 'package:test/test.dart';
+
+import 'hello.dart';
+
+main() {
+  test('hello world', () {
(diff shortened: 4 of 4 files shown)


nex3:
I think this should be a test that invokes `pub serve` itself, followed by verifying that the files it serves look (more or less) like what we want—that would be more consistent with how the rest of pub's tests work. To do the validation you could either invoke `phantomjs` and ensure that it prints the right thing, or just inspect the contents and make some assertions about it. It probably won't be the case that we'll end up with compiled JS that does the wrong thing, so the latter should be acceptable.

jakemac53:
> I think this should be a test that invokes pub serve itself, followed by verifying that the files it serves look (more or less) like what we want—that would be more consistent with how the rest of pub's tests work.

The tests today do work this way, I essentially took all the dart2js tests (which do this) and migrated them to run with each compiler (and added some more specific unit tests as well).

However, when it comes to DDC there is more logic involved in pub than just invoking the compiler, because we have to bootstrap the app as well. I can validate that the bootstrap code looks like I think it should, but that doesn't inspire a lot of confidence imo. Having a real test that actually runs the js code is a lot more confidence inspiring.

Fwiw I tried running this as a scheduled_test integration test, but I couldn't get that to work in the test runner, mostly because I don't know what port pub serve is going to run on until after pub serve is running.

nex3:
If you're using the infrastructure in [`test/serve/utils.dart`](https://github.com/dart-lang/pub/blob/master/test/serve/utils.dart), you can call `getServeUrl()` to get the full URL of a `pub serve` instance.

jakemac53:
> If you're using the infrastructure in test/serve/utils.dart, you can call getServeUrl() to get the full URL of a pub serve instance.

The problem is I can't get the port until pub serve is actually running, and I can't then schedule a new task at that point (when running in the test runner at least, it does actually work if I run the test using the dart vm directly which is even more confusing).

nex3:
The idea of scheduled tests is that you don't do *anything* when the test is running—you just queue up tasks beforehand.

jakemac53:
> The idea of scheduled tests is that you don't do anything when the test is running—you just queue up tasks beforehand.

Ya, it seemed that way (which makes sense generally) which is why I went with this approach.

nex3:
I mean, you should be able to schedule anything without needing to do stuff at runtime (except within a `schedule()` call), this included.

jakemac53:
> I mean, you should be able to schedule anything without needing to do stuff at runtime (except within a schedule() call), this included.

The problem is I need to set up the arguments ahead of time (at least how all the helpers work today in pub tests, but I think ScheduledProcess in general requires that?).

I could force a certain pub serve port, but that is likely to be fragile.

nex3:
`ScheduledProcess` can take a mix of `String` and `Future<String>` arguments.

nex3:
If you want to wait on this for a bit, I'm working on migrating pub's tests away from `scheduled_test` which should make it simpler.

jakemac53:
Sure, I am fine with waiting on this. Migrating away from `scheduled_test` should greatly simplify things.

PR #1616, opened 2017-05-23
Clean up analyzer hints

jakemac53:


nex3:
I missed this on the first review, but this needs to bump the `source_span` dependency to the version that adds `new SourceFile.fromString()`.

jakemac53:
whoops, I will send a followup

Issue #1618, opened 2017-05-23
DartDevc compilation errors should be more discoverable


jakemac53:
Today if you get a compilation error it just fails to serve the compiled module which gives an error like this in the chrome console:

```
Uncaught Error: Script error for "my_module", needed by: my_other_module
```

In order to get the actual compilation error you have to go look at the console output.

Ideally we would plumb through any compilation errors to either the chrome console or a visible widget on the page.

PR #1619, opened 2017-05-23
Improve/unify error handling, plumb through compilation errors to the browser console

jakemac53:
Added some new Exception types which are thrown instead of AssetNotFoundExceptions, and added uniform handling for them which creates a new asset at `originalId.addExtension('.errors')`.

This new asset is requested by the app when it fails to load a module and it prints the contents to the console so you can get compilation errors in your browser.

Partially resolves https://github.com/dart-lang/pub/issues/1618, I think we could still do something better with these errors but this is a good first step.

### DIFF
--- a/lib/src/dartdevc/dartdevc.dart
+++ b/lib/src/dartdevc/dartdevc.dart
@@ -14,7 +14,7 @@ import 'package:path/path.dart' as p;
 
 import '../dart.dart';
 import '../io.dart';
-import 'module.dart';
+import 'errors.dart';
 import 'module_reader.dart';
 import 'scratch_space.dart';
 import 'summaries.dart';
@@ -80,10 +80,7 @@ Future<bool> isAppEntryPoint(
 /// Synchronously returns a `Map<AssetId, Future<Asset>>` so that you can know
 /// immediately what assets will be output.
 Map<AssetId, Future<Asset>> bootstrapDartDevcEntrypoint(
-    AssetId dartEntrypointId,
-    BarbackMode mode,
-    ModuleReader moduleReader,
-    Future<Asset> getAsset(AssetId id)) {
+    AssetId dartEntrypointId, BarbackMode mode, ModuleReader moduleReader) {
   var bootstrapId = dartEntrypointId.addExtension('.bootstrap.js');
   var jsEntrypointId = dartEntrypointId.addExtension('.js');
   var jsMapEntrypointId = jsEntrypointId.addExtension('.map');
@@ -100,7 +97,6 @@ Map<AssetId, Future<Asset>> bootstrapDartDevcEntrypoint(
 
   return _ensureComplete(outputCompleters, () async {
     var module = await moduleReader.moduleFor(dartEntrypointId);
-    if (module == null) return;
 
     // The path to the entrypoint JS module as it should appear in the call to
     // `require` in the bootstrap file.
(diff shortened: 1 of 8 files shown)


lib/src/dartdevc/dartdevc.dart
```diff
@@ -186,6 +182,24 @@ for (let moduleName of Object.getOwnPropertyNames(modulePaths)) {
     }
 
     bootstrapContent.write('''
+// Whenever we fail to load a js module, try and request the corresponding
```
munificent: Grammar nit: "try and" -> "try to".

Also... "js" -> "JS".

lib/src/dartdevc/errors.dart
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:barback/barback.dart';
+
+/// An [Exception] that is thrown when the analyzer fails to create a summary.
+class AnalyzerSummaryException implements Exception {
+  /// The module that couldn't be compiled.
+  final AssetId assetId;
```
munificent: We know it's an AssetId from the type. How about a more meaningful name. Maybe "failedModule" or just "module"?

lib/src/dartdevc/errors.dart
```diff
@@ -0,0 +1,43 @@
...
+
+  AnalyzerSummaryException(this.assetId, this.error);
+
+  String toString() => 'Error creating summary for module: $assetId\n\n'
+      '$error';
+}
+
+/// An [Exception] that is thrown when dartdevc compilation fails.
+class DartDevcCompilationException implements Exception {
+  /// The js module that couldn't be compiled.
```
munificent: "js".

lib/src/dartdevc/errors.dart
```diff
@@ -0,0 +1,43 @@
...
+  DartDevcCompilationException(this.assetId, this.error);
+
+  String toString() => 'Error compiling dartdevc module: $assetId\n\n'
+      '$error';
+}
+
+/// An [Exception] that is thrown when a module is not found for an [AssetId].
+class MissingModuleException implements Exception {
+  /// The asset that a module could not be found for.
+  final AssetId assetId;
```
munificent: There's a lot of duplication between these classes. How about a base class that contains the AssetId? Or maybe just a base class for the first two that has an AssetId and the dartdevc worker error?

lib/src/dartdevc/dartdevc.dart
jakemac53: done

lib/src/dartdevc/errors.dart
jakemac53: Created base class where this in named `failedAsset`, constructor for this class takes an arg called `summaryId`.

lib/src/dartdevc/errors.dart
jakemac53: Created base class where this in named failedAsset, constructor for this class takes an arg called jsId.

lib/src/dartdevc/errors.dart
jakemac53: created a `_WorkerException` base class

PR #1620, opened 2017-05-23
handle sourcemap requests for js resources gracefully

jakemac53:


jakemac53:
Updated the test as well with the new resources, and to confirm that sourcemaps give a 404 as expected.

PR #1624, opened 2017-05-25
add web compiler option to pubspec.yaml, change --compiler to --web-compiler

jakemac53:
With the top level key in the pubspec being called `js-compiler`, it no longer makes sense for the command line arg to be just `compiler` (same argument holds about future compilers conflicting). Since `js-compiler` is really long I talked with @munificent and we landed on just `--js`.

Let me know if there are objections to changing the flag from `--compiler` to `--js`. If we could come up with a single character shorthand for `--js-compiler` that might be preferable, but I don't know of one that really makes sense.

cc @kevmoo @munificent @jwren @devoncarew 

EDIT: See comment below about naming changes

### DIFF
--- a/lib/src/command/barback.dart
+++ b/lib/src/command/barback.dart
@@ -35,20 +35,20 @@ abstract class BarbackCommand extends PubCommand {
             "The --dart2js flag can't be used with the --web-compiler arg. "
             "Prefer using the --web-compiler arg as --[no]-dart2js is "
             "deprecated.");
+      } else {
+        log.warning("The --dart2js flag is deprecated, please use "
+            "--web-compiler=dart2js option instead.");
       }
-      if (argResults['dart2js']) {
+      if (argResults["dart2js"]) {
         return Compiler.dart2JS;
       } else {
         return Compiler.none;
       }
-    } else if (argResults.options.contains("web-compiler") &&
-        argResults.wasParsed("web-compiler")) {
+    } else if (argResults.options.contains("web-compiler")) {
       return Compiler.byName(argResults["web-compiler"]);
     } else {
-      var compilerName = entrypoint.root.pubspec.webCompiler[mode.name];
-      return compilerName != null
-          ? Compiler.byName(compilerName)
-          : Compiler.dart2JS;
+      var compiler = entrypoint.root.pubspec.webCompiler[mode.name];
+      return compiler ?? Compiler.dart2JS;
     }
   }
 
--- a/lib/src/command/barback.dart
+++ b/lib/src/command/barback.dart
@@ -27,23 +27,25 @@ abstract class BarbackCommand extends PubCommand {
 
   /// The current compiler mode.
   Compiler get compiler {
-    if (argResults.options.contains('dart2js') &&
-        argResults.wasParsed('dart2js')) {
-      if (argResults.options.contains("compiler") &&
-          argResults.wasParsed("compiler")) {
-        usageException(
-            "The --dart2js flag can't be used with the --compiler arg. "
-            "Prefer using the --compiler arg as --[no]-dart2js is deprecated.");
+    if (argResults.options.contains("dart2js") &&
+        argResults.wasParsed("dart2js")) {
+      if (argResults.options.contains("js") && argResults.wasParsed("js")) {
+        usageException("The --dart2js flag can't be used with the --js arg. "
+            "Prefer using the --js arg as --[no]-dart2js is deprecated.");
       }
       if (argResults['dart2js']) {
         return Compiler.dart2JS;
       } else {
         return Compiler.none;
       }
-    } else if (argResults.options.contains("compiler")) {
-      return Compiler.byName(argResults["compiler"]);
+    } else if (argResults.options.contains("js") &&
+        argResults.wasParsed("js")) {
+      return Compiler.byName(argResults["js"]);
     } else {
-      return Compiler.dart2JS;
+      var compilerName = entrypoint.root.pubspec.jsCompiler[mode.name];
+      return compilerName != null
+          ? Compiler.byName(compilerName)
+          : Compiler.dart2JS;
     }
   }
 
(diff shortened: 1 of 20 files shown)


jakemac53:
Ok, after some offline discussion I will update this to change the arg to `--web-compiler`, and I am going to change the pubspec config to have a top level `web` key, with `compiler` as a key under that, so it will look like this:

```yaml
web:
  compiler:
    debug: dartdevc
```

jakemac53:
Updated now to reflect the new configuration.

munificent:
👍  to the name change.

lib/src/command/barback.dart
```diff
@@ -27,23 +27,28 @@ abstract class BarbackCommand extends PubCommand {
...
+      if (argResults.options.contains("web-compiler") &&
+          argResults.wasParsed("web-compiler")) {
         usageException(
-            "The --dart2js flag can't be used with the --compiler arg. "
-            "Prefer using the --compiler arg as --[no]-dart2js is deprecated.");
+            "The --dart2js flag can't be used with the --web-compiler arg. "
+            "Prefer using the --web-compiler arg as --[no]-dart2js is "
+            "deprecated.");
       }
       if (argResults['dart2js']) {
```
nex3: Shouldn't this produce a warning as well?

lib/src/command/barback.dart
```diff
@@ -27,23 +27,28 @@ abstract class BarbackCommand extends PubCommand {
...
         return Compiler.dart2JS;
       } else {
         return Compiler.none;
       }
-    } else if (argResults.options.contains("compiler")) {
-      return Compiler.byName(argResults["compiler"]);
+    } else if (argResults.options.contains("web-compiler") &&
+        argResults.wasParsed("web-compiler")) {
```
nex3: Why not just `argResults["web-compiler"]`?

lib/src/pubspec.dart
```diff
@@ -354,6 +355,57 @@ class Pubspec {
...
+        _error('"compiler" keys must be strings.', key.span);
+      }
+
+      final keyPattern = new RegExp(r"^[a-zA-Z0-9_-]+$");
+      if (!keyPattern.hasMatch(key.value)) {
+        _error(
+            '"compiler" keys may only contain letters, '
+            'numbers, hyphens and underscores.',
+            key.span);
+      }
```
nex3: Why this restriction? As far as I remember, there aren't any such restrictions on Barback mode names.

lib/src/pubspec.dart
```diff
@@ -354,6 +355,57 @@ class Pubspec {
 
   Map<String, String> _executables;
 
+  /// The settings for which web compiler to use in which mode.
+  ///
+  /// It is a map of strings to string. Each key is the name of a mode, and
+  /// the value is the name of the web compiler to use in that mode.
+  ///
+  /// Valid compiler values are all of [Compiler.names].
+  Map<String, String> get webCompiler {
```
nex3: Why not `Map<String, Compiler>`?

test/pubspec_test.dart
```diff
@@ -592,5 +592,61 @@ executables:
         expect(pubspec.executables['command'], equals('command'));
       });
     });
+
+    group("web", () {
+      test("can be empty", () {
+        var pubspec = new Pubspec.parse('web: {}', sources);
+        expect(pubspec.webCompiler, isEmpty);
+      });
```
nex3: Also can be null.

lib/src/command/barback.dart
jakemac53: done

lib/src/command/barback.dart
jakemac53: removed, there was a default value but I removed it as well.

lib/src/pubspec.dart
jakemac53: Oh sure, I like that, updated.

lib/src/pubspec.dart
jakemac53: Oh ok, I guess I was just assuming this, removed.

test/pubspec_test.dart
jakemac53: I don't get a YamlMap in that case, but I added this to the "throws if not a map" test.

nex3: I believe in general we allow users to explicitly specify `null` for fields that are allowed to not exist, which means `web:` without anything after it is valid.

lib/src/command/barback.dart
```diff
@@ -35,20 +35,20 @@ abstract class BarbackCommand extends PubCommand {
...
       }
-      if (argResults['dart2js']) {
+      if (argResults["dart2js"]) {
         return Compiler.dart2JS;
       } else {
         return Compiler.none;
       }
-    } else if (argResults.options.contains("web-compiler") &&
-        argResults.wasParsed("web-compiler")) {
+    } else if (argResults.options.contains("web-compiler")) {
```
nex3: I'm still not quite sure why this isn't just `argResults["web-compiler"]` with a default value of `false`.

test/pubspec_test.dart
jakemac53: ok, done. I also added a test for a null compiler value

lib/src/command/barback.dart
jakemac53: It doesn't feel right to have a default value that isn't a valid compiler name to me. If you want I can do that though.

nex3: Oh I see, I was thinking this was a boolean value for some reason.

kevmoo:
Cherry pick request to sdk?

From mobile. Sorry for brevity and/or spelling.

jakemac53:
cl out to merge into sdk here https://codereview.chromium.org/2908783003/ then I will send out the cherry pick request

jakemac53:
https://github.com/dart-lang/sdk/issues/29741

PR #1628, opened 2017-06-01
Log HTTP requests/responses to the IO log

kevmoo:
...instead of info.

Allows capturing response information in the case of 5xx errors without
requiring verbose logging

Related to https://github.com/dart-lang/pub-dartlang-dart/issues/94

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -177,7 +177,7 @@ class _PubHttpClient extends http.BaseClient {
       }
     }
 
-    log.fine(requestLog.toString().trim());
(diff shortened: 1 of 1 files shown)


kevmoo:
@Hixie @cbracken – If this lands, you'll want to change `flutter_tools` to use `--verbosity=io` instead of `--verbosity=warning`

kevmoo:
@Hixie @cbracken @pq  – FYI, this is in Dart https://github.com/dart-lang/sdk/releases/tag/1.25.0-dev.1.0

PR #1629, opened 2017-06-01
Stop using scheduled_test.

nex3:
This switches pub's tests over to using the stream matchers built into
test, as well as the test_descriptor, test_process, and
shelf_test_handler packages. It uses async/await rather than
scheduled_test's task queue.

This removes a layer of indirection which should make it much easier to
work on pub's test infrastructure, especially for people unfamiliar with
scheduled_test.


nex3:
I've verified locally that all the tests still pass.

nex3:
The test files are pretty mechanical, but everything that doesn't end in `_test.dart` could use a review:

* test/compiler/utils.dart
* test/dependency_computer/utils.dart
* test/descriptor.dart
* test/descriptor/git.dart
* test/descriptor/packages.dart
* test/descriptor/tar.dart
* test/descriptor_server.dart
* test/global/binstubs/utils.dart
* test/lish/utils.dart
* test/oauth2/utils.dart
* test/package_server.dart
* test/serve/caches_unchanged_assets.dart
* test/serve/does_not_cache_changed_asset.dart
* test/serve/utils.dart
* test/test_pub.dart
* test/validator/utils.dart

PR #1636, opened 2017-06-12
only fetch .errors file for script load errors

jakemac53:
`e.originalError` doesn't always exist (for timeout errors for instance), now we will only try to fetch the `.errors` file (and log the errors to the console) if `e.originalError.srcElement` exists.


jakemac53:
Currently no - now that we have migrated off of scheduled_test I might be able to somewhat reasonably make one though. I will try and work on a followup PR.

jakemac53:
https://github.com/dart-lang/pub/issues/1640

PR #1637, opened 2017-06-12
fix cache invalidation for module configuration in dartdevc

jakemac53:
Previously the actual module configuration didn't get updated which means reorganizing your package could give weird errors :(. Unfortunately the bug report didn't come through until just before the stable 1.24 release but I will see what it takes to get this in a +1 release as people are definitely likely to run into it.

### DIFF
--- a/lib/src/dartdevc/module_reader.dart
+++ b/lib/src/dartdevc/module_reader.dart
@@ -121,7 +121,7 @@ class ModuleReader {
         _modulesByModuleId[module.id] = module;
         for (var id in module.assetIds) {
           if (_modulesByAssetId.containsKey(id)) {
-            throw new StateError('Assets can only exist in one module, but $id'
+            throw new StateError('Assets can only exist in one module, but $id '
                 'was found in both ${_modulesByAssetId[id].id} and '
                 '${module.id}');
           }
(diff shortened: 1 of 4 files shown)


lib/src/dartdevc/module_reader.dart
```diff
@@ -131,4 +131,19 @@ class ModuleReader {
       return modules;
     });
   }
+
+  /// Invalidates all [Module]s for [package].
+  void invalidatePackage(String package) {
+    // `map` must be of type <AssetId|ModuleId, dynamic>
+    removePackage(String package, Map map) {
```
nex3: I'd either avoid passing `package` here, since it's already in the closure scope, or make `_removePackage` a private method rather than a local function.

test/dartdevc/dartdevc_test.dart
```diff
@@ -1,15 +1,20 @@
...
 
 import '../descriptor.dart' as d;
 import '../serve/utils.dart';
 import '../test_pub.dart';
 
 main() {
-  test("can compile js files for modules under lib and web", () async {
+  test(
+      "can compile js files for modules under lib and web, and handle new "
+      "files and edits.", () async {
```
nex3: I strongly encourage you to make this three different tests. Each test should generally focus on a single chunk of behavior.

test/dartdevc/module_reader_test.dart
```diff
@@ -157,6 +169,26 @@ void main() {
               packageCLibModule.id,
             ]));
       });
+
+      test('invalidatePackage', () async {
+        var modules = await moduleReader.readModules(packageAModuleConfig);
+        expect(
+            modules
+                .firstWhere((m) => m.id == packageAModuleC.id)
```
nex3: Nit: avoid one-letter variable names.

lib/src/dartdevc/module_reader.dart
jakemac53: Ah yep good point, removed the arg here.

test/dartdevc/module_reader_test.dart
jakemac53: done, I usually only do this for lambdas or for loops but I don't feel strongly

test/dartdevc/dartdevc_test.dart
jakemac53: Created a group and moved shared setup code into a setUp to reduce the duplicate code and get the clean separation of tests.

jakemac53:
@nex3 updated

PR #1641, opened 2017-06-13
add an integration test for pub run test with ddc

jakemac53:
Add an integration test validating using pub serve in dartdevc mode with package:test.


nex3:
Currently, all the tests of `pub serve` plus the test runner live in the `test` package, and it kind of feels like this one should as well. WDYT?

jakemac53:
Sure, I can move it there. Should I make this test run with both compilers as well?

nex3:
We already have tests for dart2js. If the API has changed, then yes, but otherwise no.

PR #1647, opened 2017-06-23
Add basic feature support.

nex3:
This allows packages to declare features with dependencies, and to opt
in to those features in dependency declarations.

### DIFF
--- a/lib/src/pubspec.dart
+++ b/lib/src/pubspec.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:collection/collection.dart';
 import 'package:path/path.dart' as path;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:source_span/source_span.dart';
@@ -114,7 +115,8 @@ class Pubspec {
   /// The additional packages this package depends on.
   List<PackageRange> get dependencies {
     if (_dependencies != null) return _dependencies;
-    _dependencies = _parseDependencies('dependencies');
+    _dependencies =
+        _parseDependencies('dependencies', fields.nodes['dependencies']);
     _checkDependencyOverlap(_dependencies, _devDependencies);
     return _dependencies;
   }
@@ -124,7 +126,8 @@ class Pubspec {
   /// The packages this package depends on when it is the root package.
   List<PackageRange> get devDependencies {
     if (_devDependencies != null) return _devDependencies;
-    _devDependencies = _parseDependencies('dev_dependencies');
+    _devDependencies = _parseDependencies(
+        'dev_dependencies', fields.nodes['dev_dependencies']);
     _checkDependencyOverlap(_dependencies, _devDependencies);
     return _devDependencies;
   }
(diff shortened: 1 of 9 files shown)


test/pubspec_test.dart
```diff
@@ -654,5 +658,75 @@ web:
...
+            (pubspec) => pubspec.features);
+      });
+
+      test("parses valid dependency specifications", () {
+        var pubspec = new Pubspec.parse(
+            '''
+features:
+  foobar:
+    dependencies:
+      baz: 1.0.0
```
jakemac53: What happens if `baz` is also a regular dependency, or appears in multiple features (with different version constraints in each case)? My intuition would be for it to just take the intersection of the versions, but it would be good to test that case.

nex3: The plan is to only allow the dependency in `dependencies` to include a version range. I'll add verification for that in a later CL.

jakemac53: Ah ok - that is probably fine (and I don't think it would be a breaking change to switch it later to support versions everywhere if we decided that was important).

lib/src/pubspec.dart
```diff
@@ -630,6 +673,34 @@ class Pubspec {
...
+  Map<String, bool> _parseDependencyFeatures(YamlNode node) {
+    if (node?.value == null) return const {};
+    if (node is! YamlMap) _error('Features must be a map.', node.span);
+
+    return mapMap((node as YamlMap).nodes,
+        key: (nameNode, _) => _validateFeatureName(nameNode),
+        value: (_, valueNode) {
+          var value = valueNode.value;
+          if (value is bool && value) return value;
+          _error('Features must be set to true.', valueNode.span);
```
munificent: Can you leave a comment explaining that this will be expanded later to allow other values? Right now, it looks silly to require `true`.

test/version_solver_test.dart
```diff
@@ -1300,6 +1301,241 @@ void downgrade() {
...
+
+    await d.appDir({'foo': '1.0.0'}).create();
+    await expectResolves(result: {'foo': '1.0.0'});
+  });
+
+  test("enables a feature if it's required", () async {
+    await servePackages((builder) {
+      builder.serve('foo', '1.0.0', pubspec: {
+        "features": {
+          "feature": {
```
munificent: Nit: Can you use a slightly less confusing name for the feature? Maybe just "foof"?

test/version_solver_test.dart
```diff
@@ -1300,6 +1301,241 @@ void downgrade() {
...
+        'version': '^1.0.0',
+        'features': {'feature': true}
+      },
+      'baz': '1.0.0'
+    }).create();
+    await expectResolves(
+        result: {'foo': '1.0.0', 'bar': '1.0.0', 'baz': '1.0.0'}, tries: 2);
+  });
+
+  test("the root package's features are disabled by default", () async {
```
munificent: Is there a way to enable a feature in the root package?

test/version_solver_test.dart
nex3: Done.

test/version_solver_test.dart
nex3: Currently, only if a dependency enables it. We can talk about more explicit ways to enable them, but I think that's out of scope for this CL.

nex3:
@munificent I discovered a bug and added a commit that fixes it; PTAL

PR #1650, opened 2017-06-28
Add support for Git dependencies on subdirectories

### LINKED ISSUE #1305: Add `path` option to git dependencies
Today you can do:

``` yaml
dependencies:
  my_package:
    git:
      ref: my_branch
      url: https://github.com/username/repo_name.git
```

It is assumed the path lives at the root of the repo. I propose an additional `path` option, which allows you to specify a path to the root of the package inside the specified repo.

nex3:
Closes #1305

### DIFF
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -277,6 +277,9 @@ class BoundGitSource extends CachedSource {
       if (!entryExists(revisionCachePath)) {
         await _clone(_repoCachePath(ref), revisionCachePath);
         await _checkOut(revisionCachePath, id.description['resolved-ref']);
+        _writePackageList(revisionCachePath, [id.description['path']]);
+      } else {
+        _updatePackageList(revisionCachePath, id.description['path']);
       }
     });
 
@@ -304,16 +307,23 @@ class BoundGitSource extends CachedSource {
 
     var packages = listDir(systemCacheRoot)
         .where((entry) => dirExists(p.join(entry, ".git")))
-        .map((packageDir) {
-          try {
-            return new Package.load(null, packageDir, systemCache.sources);
-          } catch (error, stackTrace) {
-            log.error("Failed to load package", error, stackTrace);
-            var name = p.basename(packageDir).split('-').first;
-            failures.add(new PackageId(name, source, Version.none, '???'));
-            tryDeleteEntry(packageDir);
-            return null;
-          }
+        .expand((revisionCachePath) {
+          return _readPackageList(revisionCachePath).map((relative) {
+            // If we've already failed to load another package from this
+            // repository, ignore it.
+            if (!dirExists(revisionCachePath)) return null;
+
+            var packageDir = p.join(revisionCachePath, relative);
+            try {
+              return new Package.load(null, packageDir, systemCache.sources);
+            } catch (error, stackTrace) {
+              log.error("Failed to load package", error, stackTrace);
+              var name = p.basename(revisionCachePath).split('-').first;
+              failures.add(new PackageId(name, source, Version.none, '???'));
+              tryDeleteEntry(revisionCachePath);
+              return null;
+            }
+          });
         })
         .where((package) => package != null)
         .toList();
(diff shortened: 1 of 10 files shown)


test/get/git/path_test.dart
```diff
@@ -0,0 +1,128 @@
...
+      ])
+    ]).validate();
+
+    await d.appPackagesFile({
+      'sub1': pathInCache('git/foo-${await repo.revParse('HEAD')}/subdir1'),
+      'sub2': pathInCache('git/foo-${await repo.revParse('HEAD')}/subdir2')
+    }).validate();
+  });
+
+  test('depends on packages in the same subdirectory at different revisions',
```
jakemac53: haha, nice test :)

test/get/git/path_test.dart
```diff
@@ -0,0 +1,128 @@
...
+
+import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
+
+import 'package:pub/src/io.dart';
+
+import '../../descriptor.dart' as d;
+import '../../test_pub.dart';
+
+main() {
```
jakemac53: Might as well throw in a test for a subdirectory that is more than one level deep as well, just to make sure.

jakemac53: I am also curious what happens if you have a package nested inside another package. I think its totally fine to not support that, but if it isn't supported lets make sure it throws a nice error.

For example, can you depend on both of these packages:
```
my_repo/
  pkg1/
    pubspec.yaml
    pkg2/
      pubspec.yaml
```

jakemac53: Oh also it might be good to validate that we error if `path` reaches outside of the main package dir using `../`

lib/src/source/git.dart
```diff
@@ -350,6 +417,37 @@ class BoundGitSource extends CachedSource {
...
+    if (packages.contains(path)) return;
+
+    _writePackageList(revisionCachePath, packages..add(path));
+  }
+
+  /// Returns the list of packages in [revisionCachePath].
+  List<String> _readPackageList(String revisionCachePath) {
+    var path = _packageListPath(revisionCachePath);
+
+    // If there's no pubspec list file, this cache was created by an older
```
nex3: Done.

lib/src/source/git.dart
```diff
@@ -350,6 +417,37 @@ class BoundGitSource extends CachedSource {
...
+    // version of pub where pubspecs were only allowed at the root of the
+    // repository.
+    if (!fileExists(path)) return ['./pubspec.yaml'];
+    return readTextFile(path).split("\n");
+  }
+
+  /// Writes a package list indicating that we care about [packages] exist in
```
nex3: Done.

lib/src/source/git.dart
```diff
@@ -390,10 +488,22 @@ class BoundGitSource extends CachedSource {
...
+    var repoCacheName = '${_repoName(ref)}-${sha1(ref.description['url'])}';
     return p.join(systemCacheRoot, 'cache', repoCacheName);
   }
+
+  /// Returns a short, human-readable name for the repository URL in [packageName].
+  ///
+  /// This name is not guaranteed to be unique.
+  String _repoName(PackageName packageName) {
+    var name = p.url.basename(packageName.description['url']);
+    if (name.endsWith('.git')) name = name.substring(0, name.length - 4);
```
nex3: Done.

test/cache/repair/git_test.dart
```diff
@@ -12,108 +12,195 @@ import '../../descriptor.dart' as d;
...
+    });
+
+    test('reinstalls previously cached git packages', () async {
+      // Find the cached foo packages for each revision.
+      var gitCacheDir = path.join(d.sandbox, cachePath, "git");
+      var fooDirs = listDir(gitCacheDir)
+          .where((dir) => path.basename(dir).startsWith("foo-"))
+          .toList();
+
+      // Delete "foo.dart" from them.
```
nex3: Done.

test/get/git/path_test.dart
nex3: Done.

nex3: This is totally fine by default (and is already doable with, say a `path` dependency). I'm not sure it's worth adding a separate test, though, since it doesn't tickle any different code paths.

nex3: Good catch, I don't think we had any validation there. Done.

WillBeebe:
Thank you!!

PR #1654, opened 2017-07-05
reduce --build-delay test flakyness

jakemac53:


jakemac53:
Agreed, on both accounts :).

This should be better than what we have now, if it doesn't make the bots happy I will try and figure out a unit test that isn't flaky.

test/serve/build_delay_test.dart
```diff
@@ -71,6 +49,7 @@ main() {
   });
 
   test("continual fast edits won't cause multiple builds", () async {
+    // Set a larg-ish delay of 100ms to reduce flakyness on bots.
```
munificent: "large-ish".

jakemac53:
closing in favor of https://github.com/dart-lang/pub/pull/1655 which should not be flaky

PR #1656, opened 2017-07-05
Add support for opting out of features

nex3:


lib/src/package_name.dart
```diff
@@ -153,39 +153,62 @@ class PackageRange extends PackageName {
...
+    }
+
+    if (disabledFeatures.isNotEmpty) {
+      description += "without ${toSentence(disabledFeatures)}";
+    }
+    return description;
+  }
+
   String toString() {
     if (isRoot) return "$name $constraint (root)";
```
munificent: Can the root package enable/disable features? If so, probably don't want to early return here.

nex3: Done.

PR #1658, opened 2017-07-07
Add support for declaring SDK constraints in features

nex3:


lib/src/pubspec.dart
```diff
@@ -279,40 +283,48 @@ class Pubspec {
...
 
-    var yaml = fields['environment'];
+    var pair = _parseEnvironment(fields);
+    _dartSdkConstraint = pair.first;
+    _flutterSdkConstraint = pair.last;
+  }
+
+  /// Parses the "environment" field in [parent] and returns the Dart and
+  /// Flutter SDK constraints, respectively.
+  Pair<VersionConstraint, VersionConstraint> _parseEnvironment(YamlMap parent) {
```
nex3: I think it's a little clearer that it's going to be exactly two constraints. I'd also feel a lot weirder making the positions in a list semantically meaningful than the positions in a pair.

Issue #1593, opened 2017-05-12
Add package features


matanlurey:
*Edited by @nex3*

I'm updating this to track progress on the package features feature I outlined below. In particular, we need:

* [x] The ability to declare and depend on off-by-default features.
* [x] The ability to declare on-by-default features, and to disable them.
* [x] The ability to declare SDK constraints in features.
* [x] The ability to depend on a feature if it's available, but allow versions of the package that don't expose it.
* [ ] The ability for one feature to depend on another in the same package.
* [x] The ability to enable or disable features in `pub global activate`.
* [x] Handle multiple dependencies on the same package with different features.
* [x] Ensuring that a package that exposes or uses features has an appropriate SDK constraint so it's not visible to older `pub` versions.

----

Rather, it should be possible to create a `local_storage` library with 3 implementations:

* Using `PlatformMessages` for `dart:ui` / Flutter
* Using `localStorage` for `dart:html`
* Using a data directory for `dart:io`

Today this doesn't really work, because I can't conditionally depend on the Flutter SDK. Might I suggest that we allow a dependency that itself depends on Flutter, even if we don't have the Flutter SDK installed?

/cc @jakemac53 who is working on pub these days.

nex3:
This is definitely a hole in our current package story—one that'll take a lot of careful design to fix properly. My bet is that this will ultimately require some degree of language support to resolve, some way of determining at runtime what packages are available and in what form.

One idea for this that I've floated before is *package features* (which I totally stole from Rust): a package would be able to declare multiple optional "features", each of which has its own dependencies. Then dependencies on that package could declare which features they need, and only those dependencies would be included in the package graph. Something like:

```yaml
# local_storage/pubspec.yaml
features:
  flutter:
    platform_messages:
      version: ^1.2.0
      sdk: flutter
  io:
    path: ^1.4.0

# app/pubspec.yaml
dependencies:
  local_storage:
    version: ^1.0.0
    features: [flutter, io]
```

The trick is providing `local_storage` the ability to load flutter when flutter is running and to load IO otherwise, and more generally to be able to change its behavior based on which features were selected. Something like configuration-specific imports, combined with the ability for pub to define environment constants, could do this.

matanlurey:
@nex3:

> This is definitely a hole in our current package story—one that'll take a lot of careful design to fix properly. My bet is that this will ultimately require some degree of language support to resolve, some way of determining at runtime what packages are available and in what form.

[...]

PR #1662, opened 2017-07-11
Allow multiple dependencies from one package onto another

nex3:
Enabling multiple dependencies is necessary for some use-cases for
issue #1593 like having a package expose a feature that enables a
feature in a dependency, or having a package expose a "testing"
feature and also depend on test from its dev dependencies.

For consistency with features, this also removes the restriction that
a pubspec may not have a normal and dev dependency on the same
package.

### DIFF
--- a/test/version_solver_test.dart
+++ b/test/version_solver_test.dart
@@ -288,7 +288,9 @@ void devDependency() {
   group("with both a dev and regular dependency", () {
     test("succeeds when both are satisfied", () async {
       await servePackages((builder) {
+        builder.serve('foo', '1.0.0');
         builder.serve('foo', '2.0.0');
+        builder.serve('foo', '3.0.0');
       });
 
       await d.dir(appPath, [
@@ -2153,7 +2155,7 @@ void features() {
           "dependencies": {
             'bar': {
               'version': '1.0.0',
-              'features': {'stuff': true}
+              'features': {'stuff': false}
             },
           },
           "features": {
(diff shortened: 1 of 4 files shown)


test/version_solver_test.dart
```diff
@@ -284,6 +284,130 @@ void devDependency() {
     await d.appDir({'foo': '1.0.0'}).create();
     await expectResolves(result: {'foo': '1.0.0'});
   });
+
+  group("with both a dev and regular dependency", () {
+    test("succeeds when both are satisfied", () async {
+      await servePackages((builder) {
+        builder.serve('foo', '2.0.0');
```
munificent: Might be good to serve 1.0.0 and 3.0.0 to ensure that the chosen package is also satisfies both ranges.

test/version_solver_test.dart
```diff
@@ -1978,6 +2102,94 @@ void features() {
...
+        'foo': {
+          'version': '1.0.0',
+          'features': {'stuff': true}
+        }
+      }).create();
+      await expectResolves(
+          result: {'foo': '1.0.0', 'bar': '1.0.0', 'baz': '1.0.0'});
+    });
+
+    test("can't disable features", () async {
```
munificent: How is this test trying to disable the feature? It looks like both foo's regular dependency on bar and the feature dependency on bar both turn on stuff. Should one be false?

test/version_solver_test.dart
nex3: Done.

test/version_solver_test.dart
nex3: Yeah, this was a mistake. Done.

PR #1663, opened 2017-07-12
Add a default SDK constraint

### LINKED ISSUE #1657: Treat missing upper bound in Dart SDK constraint as `<2.0.0`
This will ensure packages have to explicitly support the v2 SDK

nex3:
Packages that don't declare an SDK constraint will now be considered
compatible with SDK versions matching <2.0.0-dev.infinity.

Closes #1657

### DIFF
--- a/lib/src/pubspec.dart
+++ b/lib/src/pubspec.dart
@@ -23,6 +23,12 @@ import 'utils.dart';
 final _packageName = new RegExp(
     "^${identifierRegExp.pattern}(\\.${identifierRegExp.pattern})*\$");
 
+/// The default SDK constraint for packages that don't declare one.
+///
+/// This allows 2.0.0 dev versions to make the migration proecss smoother.
+final _defaultSdkConstraint =
+    new VersionConstraint.parse("<2.0.0-dev.infinity");
+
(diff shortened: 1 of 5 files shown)


lib/src/pubspec.dart
```diff
@@ -23,6 +23,12 @@ import 'utils.dart';
 final _packageName = new RegExp(
     "^${identifierRegExp.pattern}(\\.${identifierRegExp.pattern})*\$");
 
+/// The default SDK constraint for packages that don't declare one.
+///
+/// This allows 2.0.0 dev versions to make the migration proecss smoother.
+final _defaultSdkConstraint =
+    new VersionConstraint.parse("<2.0.0-dev.infinity");
```
munificent: Perhaps this is a little *too* forward looking, but is it worth making this automatically be "<X.0.0-dev.infinity" where X is the current major version of the SDK?

nex3: I'd like *all* packages to have SDK constraints by the time 3.0.0-dev rolls around, which should happen if we ensure that no constraint-less package is ever considered compatible with 2.0.0.

munificent: 👍

PR #1667, opened 2017-07-14
Add the ability for features to depend on one another

### LINKED ISSUE #1593: Add package features

nex3:
Closes #1593

### DIFF
--- a/test/pubspec_test.dart
+++ b/test/pubspec_test.dart
@@ -250,9 +250,7 @@ dependencies:
 
     test("throws if transformers isn't a list", () {
       expectPubspecException(
-          'transformers: "not list"',
-          (pubspec) => pubspec.transformers,
-          '"transformers" field must be a list');
+          'transformers: "not list"', (pubspec) => pubspec.transformers);
     });
 
     test("throws if a transformer isn't a string or map", () {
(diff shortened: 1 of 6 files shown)


test/pubspec_test.dart
```diff
@@ -733,6 +732,26 @@ features:
...
+        test("must be a list", () {
+          expectPubspecException('features: {foobar: {requires: baz}, baz: {}}', (pubspec) => pubspec.features);
+        });
+
+        test("must be a string list", () {
+          expectPubspecException('features: {foobar: {requires: [12]}}', (pubspec) => pubspec.features);
+        });
+
+        test("must refer to features that exist in the pubspec", () {
+          expectPubspecException('features: {foobar: {requires: [baz]}}', (pubspec) => pubspec.features);
```
munificent: Got some long lines here.

PR #1670, opened 2017-07-19
Deprecated cli util

kevmoo:


lib/src/dartdevc/workers.dart
```diff
@@ -12,12 +12,12 @@ String get _scriptExtension => Platform.isWindows ? '.bat' : '';
...
     ['--build-mode', '--persistent_worker']));
 
 /// Manages a shared set of persistent dartdevc workers.
 final dartdevcDriver = new BazelWorkerDriver(() => Process.start(
-    p.join(sdkDir.path, 'bin', 'dartdevc$_scriptExtension'),
+    p.join(sdkDir, 'bin', 'dartdevc$_scriptExtension'),
     ['--persistent_worker']));
 
-final sdkDir = cli_util.getSdkDir();
+final sdkDir = cli_util.getSdkPath();
```
devoncarew: This is API, so should specify the type :)

kevmoo: top-level inference FTW!

lib/src/dartdevc/dartdevc.dart
```diff
@@ -300,12 +299,12 @@ Map<AssetId, Future<Asset>> createDartdevcModule(
...
     await scratchSpace.ensureAssets(allAssetIds);
     var jsOutputFile = scratchSpace.fileFor(module.id.jsId);
-    var sdk_summary = p.url.join(sdkDir.path, 'lib/_internal/ddc_sdk.sum');
+    var sdk_summary = p.url.join(sdkDir, 'lib/_internal/ddc_sdk.sum');
     var request = new WorkRequest();
     request.arguments.addAll([
       '--dart-sdk-summary=$sdk_summary',
       '--modules=amd',
-      '--dart-sdk=${sdkDir.path}',
+      '--dart-sdk=${sdkDir}',
```
natebosch: [nit] surrounding code avoids unnecessary '{}' in string interpolation

kevmoo: I missed you, @natebosch

lib/src/dartdevc/workers.dart
natebosch: FWIW the variable name could apply to a `Directory` and the method we're calling could apply to a `Path` (if that existed) so the type isn't super obvious. I'm generally in the camp that leans on inference but this is a case where I may have included an explicit type annotation.

PR #1671, opened 2017-07-19
Set the content type during publish to application/gzip

kevmoo:
Fixes https://github.com/dart-lang/pub-dartlang-dart/issues/230

### DIFF
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:http/http.dart' as http;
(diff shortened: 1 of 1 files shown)


kevmoo:
There is no changelog :-)

lib/src/command/lish.dart
```diff
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:http/http.dart' as http;
+import 'package:http_parser/http_parser.dart' as http_parser;
```
nex3: This is not typically imported with a prefix.

Issue #1684, opened 2017-08-10
Make sure analysis_options.yaml can only impact DDC builds on the local package


natebosch:
We *might* want to look for analysis_options.yaml in other packages and try to honor those - but at the least the local analysis options should not leak into DDC behavior when compiling other modules.

I think we disallow strongly connected components between packages so it should always be possible to determine the analysis_options.yaml for a module.

This will mean recreating the logic that the analyzer uses since we'll have to pass --options explicitly rather than let it get picked up based on directory structure. cc @bwilkerson

natebosch:
Some alternatives:

- always pass `--options=/dev/null` - this is in opposition to @leafpetersen and @jmesserly described intended behavior in https://github.com/dart-lang/sdk/issues/29713
- juggle temp directories or however we're invoking DDC such that the modules we build for other packages don't appear to be under a directory with the local analysis options - @jakemac53 do you know if there is something we can do to that end?

natebosch:
We might also consider adding a config in the pubspec to let us override which analysis options is passed to DDC.

jakemac53:
Going to try out passing an empty analysis options file for all dependencies, but not the root package so that will still get whatever settings you have.

This will not support custom analysis options in dependencies but I think that is probably fine.

Issue #1687, opened 2017-08-18
summary generation test failing in latest dev sdk


jakemac53:
You can repro by running `pub run test test/dartdevc/summaries_test.dart`.

My guess is this is related to recent changes relating to analysis driver in analyzer_cli?

I will be out starting 8/22 but can take a look when I get back if this is still an issue.

jakemac53:
cc @scheglov

scheglov:
Thank you for fixing this!

PR #1688, opened 2017-08-18
dont use root analysis options for dependencies

jakemac53:
fixes https://github.com/dart-lang/pub/issues/1684

### DIFF
--- a/lib/src/dartdevc/common.dart
+++ b/lib/src/dartdevc/common.dart
@@ -10,10 +10,7 @@ final defaultAnalysisOptionsId =
     new AssetId('_internal_', 'lib/analysis_options.yaml');
 
 final defaultAnalysisOptions =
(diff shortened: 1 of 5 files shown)


lib/src/dartdevc/common.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:barback/barback.dart';
+
+import 'scratch_space.dart';
+
+final defaultAnalysisOptionsId =
+    new AssetId('_internal_', 'lib/analysis_options.yaml');
+
+final defaultAnalysisOptions =
```
natebosch: I think this should work with an empty file

jakemac53: done

jakemac53:
fyi I am expecting one travis failure because of https://github.com/dart-lang/pub/issues/1687

PR #1694, opened 2017-08-31
Hide `features` support behind a ENABLE_PUB_FEATURES environment variable

jakemac53:


jakemac53:
This does have the unfortunate side-effect of making the features related tests fail by default, you have to enable them via the environment variable.

I could add a specific mention of the environment variable but then it also becomes more discoverable which we may not actually want?

kevmoo:
A lot of tests already set up the environment variables before running the client.

Wonder if you should do that instead – otherwise we'll get breaks on the pkg/sdk bots

jakemac53:
@kevmoo these tests don't spawn a subprocess though so I can't set environment variables afaik?

kevmoo:
> these tests don't spawn a subprocess though so I can't set environment variables afaik?

I guess we'd have to get fancy w/ a top-level getter + some other way to override the value

Hrm...

Not a huge deal – but I do worry this will blow up on the pkg/sdk bots

jakemac53:
Added a field which you can override in tests etc, not my favorite but it should get the job done.

kevmoo:
Not perfect, but preferable I think.

nex3:
Can someone provide some context on this?

jakemac53:
We needed to merge pub into the sdk for some other changes but weren't prepared to release the "features" feature for general consumption as we didn't know the status, so we just disabled it for now. If it's ready to release we can go ahead and revert this and merge into the sdk for the next release.

nex3:
Yeah it's good to go! I'll prepare a revert.

jakemac53:
cool, and welcome back!

Issue #1695, opened 2017-09-01
Consider adding option for controlling dartdevc build resource usages


dam0vm3nt:
Building with dartdevc requires a great amount of resources. This can be a problem and make build fail in serveral contexes (like in workstation/ci server with less than 8gb of ram available which is not so uncommon).

Please consider adding some option to limit resources usages (like number of worker threads and they memory limit) during dartdevc build such that the developer can trade off between build speed and resource usages.

jakemac53:
I see ~4 ways of handling this:

- pubspec.yaml configuration, probably an additional option under the `web` key.
- command line arg configuration, `--max-workers=x` or similar.
- system environment variable `PUB_MAX_WORKERS` or similar
- Dart environment variable ( `-DPUB_MAX_WORKERS` command line arg)

It should be noted there are also two types of workers (analyzer and ddc), so any of these numbers would be multiplied by 2 in actuality. We could make this more obvious by letting you configure it per-worker (--max-ddc-workers, --max-analyzer-workers, etc).

cc @natebosch @kevmoo @munificent for thoughts

jakemac53:
I am going to run with the system environment variable for now as that works the nicest with Travis I think and it will have to exist regardless to override settings for bot builds.

We can add more options later if needed, but I would like to keep it simple for now.

jakemac53:
Going to call it `PUB_MAX_WORKERS_PER_TASK` so try and make it clear that its actually multiplier

jakemac53:
I am working on a merge into the sdk this week with some other changes, not sure when the next release will be though.

In the meantime you can add a git dev dependency on pub and run `pub run pub ...`.

The dev dependency would look like:

```
dev_dependencies:
  pub:
    git: https://github.com/dart-lang/pub.git
```

dam0vm3nt:
FWIW : using the latest pub and setting `PUB_MAX_WORKERS_PER_TASK` to `1` now my builds are working on our CI servers too. Thanks!

PR #1697, opened 2017-09-05
Add support for PUB_MAX_WORKERS_PER_TASK system environment variable

jakemac53:
Fixes https://github.com/dart-lang/pub/issues/1695.

Not sure on a testing strategy for this....

### DIFF
--- a/lib/src/dartdevc/workers.dart
+++ b/lib/src/dartdevc/workers.dart
@@ -8,16 +8,31 @@ import 'package:bazel_worker/driver.dart';
 import 'package:cli_util/cli_util.dart' as cli_util;
 import 'package:path/path.dart' as p;
 
+import '../log.dart';
+
 String get _scriptExtension => Platform.isWindows ? '.bat' : '';
 
+const _defaultMaxWorkers = 4;
+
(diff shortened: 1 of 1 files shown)


lib/src/dartdevc/workers.dart
```diff
@@ -8,16 +8,31 @@ import 'package:bazel_worker/driver.dart';
...
+
 String get _scriptExtension => Platform.isWindows ? '.bat' : '';
 
+const _defaultMaxWorkers = 4;
+
+const _maxWorkersEnvVar = 'PUB_MAX_WORKERS_PER_TASK';
+
+final int _maxWorkersPerTask = int.parse(
+    Platform.environment[_maxWorkersEnvVar] ?? '$_defaultMaxWorkers',
+    onError: (value) => warning(
```
natebosch: I don't see where onError returns the default

jakemac53: Lol good catch, it happened to work because the default for the bazel worker package is the same and it was falling back on that.

jakemac53: done

PR #1698, opened 2017-09-05
change default sdk constraint to <2.0.0

jakemac53:


jakemac53:
So I realized now after updating some more tests that this effectively adds a _required_ field to all pubspecs for 2.x versions (`environment: {sdk: <something-that-allows-my-version>}`).

Is that what we want? For root packages it seems unfortunate :(.

jakemac53:
Alright, we are going to change this so the default constraint does not apply to the root package.

That opens up the door to having a package that works locally but can't version solve once depended on by another project (even when using the same sdk as before), so we will additionally start enforcing an upper bound sdk constraint when publishing packages.

I will update the logic here to not apply the defaults on the root package and follow up with another pr to make the missing upper bound a failure when publishing.

lib/src/pubspec.dart
```diff
@@ -517,8 +529,11 @@ class Pubspec {
             devDependencies == null ? null : devDependencies.toList(),
         _dependencyOverrides =
             dependencyOverrides == null ? null : dependencyOverrides.toList(),
-        _dartSdkConstraint = dartSdkConstraint ?? _defaultSdkConstraint,
+        _dartSdkConstraint = dartSdkConstraint ?? includeDefaultSdkConstraint
+            ? _defaultUpperBoundSdkConstraint
+            : VersionConstraint.any,
         _flutterSdkConstraint = flutterSdkConstraint,
+        _includeDefaultSdkConstraint = false,
```
kevmoo: I see the ctor arg – then this setting the field ... not consistent?

jakemac53: I can see how it would seem that way :), but I am just shortcutting here and setting the `_dartSdkConstraint` directly. I could leave this as `true` but it wouldn't do anything.

jakemac53: Oh wait nvm you are right because you can pass in a special dartSdkConstraint so ya I should respect this setting as well blergs.

jakemac53:
@kevmoo this is ready now, all tests passing and addressed feedback

PR #1701, opened 2017-09-07
require an upper bound on the sdk to publish

jakemac53:
We should talk about what this means for flutter packages...

### DIFF
--- a/lib/src/validator/sdk_constraint.dart
+++ b/lib/src/validator/sdk_constraint.dart
@@ -34,7 +34,7 @@ class SdkConstraintValidator extends Validator {
       if (dartConstraint.max == null) {
         errors.add(
             'Published packages should have an upper bound constraint on the '
-            'Dart sdk (typically this should restrict to less than the next '
+            'Dart SDK (typically this should restrict to less than the next '
             'major version to guard against breaking changes).\n'
             'See https://www.dartlang.org/tools/pub/pubspec#sdk-constraints for '
             'instructions on setting an sdk version constraint.');
--- a/lib/src/validator/sdk_constraint.dart
+++ b/lib/src/validator/sdk_constraint.dart
@@ -20,15 +20,25 @@ class SdkConstraintValidator extends Validator {
 
   Future validate() async {
     var dartConstraint = entrypoint.root.pubspec.dartSdkConstraint;
-    if (dartConstraint is VersionRange &&
-        dartConstraint.toString().startsWith("^")) {
-      errors
-          .add("^ version constraints aren't allowed for SDK constraints since "
-              "older versions of pub don't support them.\n"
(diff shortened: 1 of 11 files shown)


lib/src/validator/sdk_constraint.dart
```diff
@@ -20,15 +20,25 @@ class SdkConstraintValidator extends Validator {
...
+            "Expand it manually instead:\n"
+            "\n"
+            "environment:\n"
+            "  sdk: \">=${dartConstraint.min} <${dartConstraint.max}\"");
+      }
+
+      if (dartConstraint.max == null) {
+        errors.add(
+            'Published packages should have an upper bound constraint on the '
+            'Dart sdk (typically this should restrict to less than the next '
```
munificent: "SDK".

lib/src/validator/sdk_constraint.dart
```diff
@@ -20,15 +20,25 @@ class SdkConstraintValidator extends Validator {
...
+        errors.add(
+            "^ version constraints aren't allowed for SDK constraints since "
+            "older versions of pub don't support them.\n"
+            "Expand it manually instead:\n"
+            "\n"
+            "environment:\n"
+            "  sdk: \">=${dartConstraint.min} <${dartConstraint.max}\"");
+      }
+
+      if (dartConstraint.max == null) {
```
munificent: So we require an upper bound if they have an SDK constraint, but we still allow completely omitting the SDK constraint. Should we require having one?

jakemac53: We shouldn't allow a completely omitted one either, I will add an additional test to verify that behavior.

jakemac53:
> I'm not opposed to this, though I'm not sure how much actual user value it will provide in normal day-to-day life.

It only really impacts users during major version transitions, but I think it will provide good value at those times. By forcing all packages to have sane sdk constraints it should help ensure that people can more safely migrate to the new sdk versions (and highlight which packages are not verified compatible when they do).

> If we are going to do this, we should very clearly announce it ahead of time and get feedback first, since it makes it harder for everyone to publish a package.

I don't think the barrier is very large, but its true that its an extra barrier. @kevmoo do you have any suggestions on how/when we should announce this generally? Is it worth a news post?

lib/src/validator/sdk_constraint.dart
jakemac53: test added

jakemac53:
Oh also with respect to my initial concern about flutter it looks like we already required a lower bound sdk constraint for those users, so requiring an upper bound as well isn't much more burden imo.

lib/src/validator/sdk_constraint.dart
jakemac53: done

Issue #1692, opened 2017-08-24
Support PUB_ALLOW_PRERELEASE_SDK environment variable


kevmoo:
Details TBD

kevmoo:
Consider adding support for `PUB_ENFORCE_SDK`
The default for `--enforce-sdk-constraint` becomes
`Platform.environment['PUB_ENFORCE_SDK'] ?? 'warn'`

This will make it easier to roll Travis and some other tools where it's easier to set one environment variable than to dig into every invocation of pub

jakemac53:
Are we just doing the environment variable for now or should I do the flag as well?

PR #1704, opened 2017-09-14
Add default upper bound Dart SDK constraint override from 2.0.0 -> 2.0.0-dev.infinity

jakemac53:
Also introduces the PUB_ALLOW_PRERELEASE_SDK environment variable which can override this behavior. Valid values are `true` (default), `false`, and `quiet` (true but no log message about the override).

Fixes https://github.com/dart-lang/pub/issues/1692

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -22,6 +22,7 @@ import 'log.dart' as log;
 import 'package.dart';
 import 'package_name.dart';
 import 'package_graph.dart';
+import 'pubspec.dart';
 import 'sdk.dart' as sdk;
 import 'solver/version_solver.dart';
 import 'source/cached.dart';
@@ -193,6 +194,23 @@ class Entrypoint {
       bool packagesDir: false}) async {
     var result = await resolveVersions(type, cache, root,
         lockFile: lockFile, useLatest: useLatest);
+
+    // Log once about all overridden packages.
+    if (warnAboutPreReleaseTwoDotZeroSdkOverrides && result.pubspecs != null) {
+      var overriddenPackages = result.pubspecs.values
+          .where((pubspec) => pubspec.dartSdkWasOverridden)
+          .map((pubspec) => pubspec.name)
+          .join(', ');
+      if (overriddenPackages.isNotEmpty) {
+        log.message(log.yellow(
+            'Overriding Dart SDK constraint from <2.0.0 to <2.0.0-dev.infinity'
+            ' for the following packages:\n\n$overriddenPackages\n\n'
+            'To disable this you can set the PUB_ALLOW_PRERELEASE_SDK system '
+            'environment variable to `false`, or you can silence this message '
+            'by setting it to `quiet`.'));
+      }
+    }
+
     if (!result.succeeded) throw result.error;
 
     result.showReport(type);
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -197,14 +197,16 @@ class Entrypoint {
 
     // Log once about all overridden packages.
     if (warnAboutPreReleaseTwoDotZeroSdkOverrides && result.pubspecs != null) {
-      var overriddenPackages = result.pubspecs.values
-          .where((pubspec) => pubspec.dartSdkWasOverridden)
-          .map((pubspec) => pubspec.name)
+      var overriddenPackages = (result.pubspecs.values
+              .where((pubspec) => pubspec.dartSdkWasOverridden)
+              .map((pubspec) => pubspec.name)
+              .toList()
+                ..sort())
           .join(', ');
       if (overriddenPackages.isNotEmpty) {
         log.message(log.yellow(
             'Overriding Dart SDK constraint from <2.0.0 to <2.0.0-dev.infinity'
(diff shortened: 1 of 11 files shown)


lib/src/pubspec.dart
```diff
@@ -29,13 +30,53 @@ final _packageName = new RegExp(
...
+/// PUB_ALLOW_PRERELEASE_SDK system environment variable.
+final bool allowPreReleaseTwoDotZeroSdk = () {
+  var userSetting =
+      Platform.environment["PUB_ALLOW_PRERELEASE_SDK"]?.toLowerCase() ?? 'true';
+  switch (userSetting) {
+    case "true":
+      return true;
+      break;
+    case "quiet":
+      warnAboutPreReleaseTwoDotZeroSdkOverrides = false;
```
kevmoo: PM nit...

This makes me sad. I hate mutable top-level things.

Could their be a private field w/ an enum type and then two public bool properties that just key off it?

test/version_solver_test.dart
```diff
@@ -1040,7 +1040,62 @@ void dartSdkConstraint() {
...
+        }
+      })
+    ]).create();
+
+    await expectResolves(
+      environment: {'_PUB_TEST_SDK_VERSION': '2.0.0-dev.99'},
+      // Log output should mention the PUB_ALLOW_RELEASE_SDK environment
+      // variable and mention the foo and bar packages specifically.
+      output: allOf(contains('PUB_ALLOW_PRERELEASE_SDK'),
+          anyOf(contains('foo, bar'), contains('bar, foo'))),
```
kevmoo: Or you could just sort it? Might make it easier to parse for the user, too...

lib/src/entrypoint.dart
```diff
@@ -193,6 +194,23 @@ class Entrypoint {
...
+
+    // Log once about all overridden packages.
+    if (warnAboutPreReleaseTwoDotZeroSdkOverrides && result.pubspecs != null) {
+      var overriddenPackages = result.pubspecs.values
+          .where((pubspec) => pubspec.dartSdkWasOverridden)
+          .map((pubspec) => pubspec.name)
+          .join(', ');
+      if (overriddenPackages.isNotEmpty) {
+        log.message(log.yellow(
+            'Overriding Dart SDK constraint from <2.0.0 to <2.0.0-dev.infinity'
```
munificent: > "SDK"

You did it! 🎉

lib/src/pubspec.dart
```diff
@@ -29,13 +30,53 @@ final _packageName = new RegExp(
 /// The default SDK upper bound constraint for packages that don't declare one.
 ///
 /// This provides a sane default for packages that don't have an upper bound.
-final _defaultUpperBoundSdkConstraint = new VersionConstraint.parse("<2.0.0");
+final VersionRange _defaultUpperBoundSdkConstraint =
+    new VersionConstraint.parse("<2.0.0");
+
+/// The upper bound contraint that matches the dev sdk.
```
munificent: "sdk", here and elsewhere.

lib/src/pubspec.dart
```diff
@@ -29,13 +30,53 @@ final _packageName = new RegExp(
...
+The environment variable PUB_ALLOW_PRERELEASE_SDK is set as `$userSetting`.
+The expected value is either `true`, `quiet` (true but no logging), or `false`.
+Using a default value of `true`.
+'''));
+      return true;
+  }
+}();
+
+/// Whether or not to warn about pre-release sdk overrides.
+bool warnAboutPreReleaseTwoDotZeroSdkOverrides = true;
```
munificent: This returns the wrong value if it's read before `allowPreReleaseTwoDotZeroSdk` is. Like @kevmoo suggests, you probably want a single lazy-initialized backing field and then two bool getters.

test/version_solver_test.dart
munificent: Or just:

```dart
allOf(contains('PUB_ALLOW_PRERELEASE_SDK'), contains('foo'), contains('bar'))
```

test/version_solver_test.dart
```diff
@@ -1040,7 +1040,62 @@ void dartSdkConstraint() {
...
+    );
+  });
+
+  test(
+      'pub doesn\'t log about pre-release sdk overrides if '
```
munificent: Nit: I think it's a little cleaner to use a double-quoted string instead of escaping the apostrophe.

lib/src/entrypoint.dart
jakemac53: :D

lib/src/pubspec.dart
jakemac53: done

lib/src/pubspec.dart
jakemac53: done (didn't use an enum though)

lib/src/pubspec.dart
jakemac53: done

test/version_solver_test.dart
jakemac53: went ahead and sorted it since this is actually a long list in many cases so its a bit easier to parse

test/version_solver_test.dart
jakemac53: done

PR #1705, opened 2017-09-15
expand the sdk prerelease override logic to work for any pre-release

jakemac53:


lib/src/entrypoint.dart
```diff
@@ -196,7 +196,7 @@ class Entrypoint {
         lockFile: lockFile, useLatest: useLatest);
 
     // Log once about all overridden packages.
-    if (warnAboutPreReleaseTwoDotZeroSdkOverrides && result.pubspecs != null) {
+    if (warnAboutPreReleaseoSdkOverrides && result.pubspecs != null) {
```
munificent: "Releaseo" -> "Release"

lib/src/pubspec.dart
```diff
@@ -376,24 +373,45 @@ class Pubspec {
...
+          includeMax: true);
     } else {
       _dartSdkConstraint = parsedDartSdkConstraint;
     }
 
     _flutterSdkConstraint = pair.last;
   }
 
+  /// Whether or not we should override [sdkConstraint] to be <= the current
+  /// users SDK version.
```
munificent: "current users" -> "user's current".

lib/src/pubspec.dart
```diff
@@ -376,24 +373,45 @@ class Pubspec {
...
+  ///
+  /// This is true if the following conditions are met:
+  ///
+  ///   - [allowPreReleaseSdk] is `true`
+  ///   - The current users SDK is a pre-release version.
```
munificent: Ditto.

test/pubspec_test.dart
```diff
@@ -443,17 +444,31 @@ dependencies:
     });
 
     group("environment", () {
+      /// Checking for the default sdk constraint based on the current sdk.
```
munificent: "sdk" -> "SDK" x 2.

test/pubspec_test.dart
```diff
@@ -443,17 +444,31 @@ dependencies:
...
+      void expectDefaultSdkConstraint(Pubspec pubspec) {
+        var sdkVersionString = sdk.version.toString();
+        if (sdkVersionString.startsWith('2.0.0') && sdk.version.isPreRelease) {
```
munificent: I'm not a fan of conditional test expectations.

Can you just override the SDK version number somehow to test it explicitly?

test/pubspec_test.dart
```diff
@@ -443,17 +444,31 @@ dependencies:
...
+
       test("allows an omitted environment", () {
         var pubspec = new Pubspec.parse('name: testing', sources);
-        expect(pubspec.dartSdkConstraint,
-            equals(new VersionConstraint.parse("<2.0.0-dev.infinity")));
+        expectDefaultSdkConstraint(pubspec);
         expect(pubspec.flutterSdkConstraint, isNull);
       });
 
       test("default sdk constraint can be ommited with empty environment", () {
```
munificent: "ommited" -> "omitted".

test/version_solver_test.dart
```diff
@@ -1134,9 +1134,89 @@ void dartSdkConstraint() {
 
     await expectResolves(
         environment: {'_PUB_TEST_SDK_VERSION': '2.0.0'},
-        error: 'Package foo requires SDK version <2.0.0-dev.infinity but the '
+        error: 'Package foo requires SDK version <2.0.0 but the '
             'current SDK is 2.0.0.');
   });
+
+  test("prerelease override requires major sdk versions to match", () async {
```
munificent: "prerelease" -> "pre-release" (here and below).

"sdk" -> "SDK", of course. :)

test/version_solver_test.dart
```diff
@@ -1134,9 +1134,89 @@ void dartSdkConstraint() {
...
+        'name': 'myapp',
+        'environment': {'sdk': '<1.2.3'}
+      })
+    ]).create();
+
+    await expectResolves(
+        environment: {'_PUB_TEST_SDK_VERSION': '1.2.3-dev.1.0'},
+        output: allOf(contains('PUB_ALLOW_PRERELEASE_SDK'),
+            contains('<=1.2.3-dev.1.0'), contains('myapp')));
+  });
```
munificent: It might be worth adding a test for how it works with a "+..." build version too. The SDK doesn't use those now, but just in case we do in the future.

lib/src/entrypoint.dart
jakemac53: doh, done

lib/src/pubspec.dart
jakemac53: done

lib/src/pubspec.dart
jakemac53: done

test/pubspec_test.dart
jakemac53: done

test/pubspec_test.dart
jakemac53: with these tests I can't since they aren't spinning up a new process so I can't set an environment variable... I could more explicitly just do an `anyOf` check with the two allowed versions though?

I do have more explicit checks around this behavior in the version solver test (where I can set a fake sdk version), this is really just a way to get this test passing and confirming there is some sensible default constraint.

test/pubspec_test.dart
jakemac53: done

test/version_solver_test.dart
jakemac53: done

test/version_solver_test.dart
jakemac53: done, added one to confirm it doesn't override if there is a dev release constraint explicitly as well since I forgot that

PR #1711, opened 2017-09-21
add back dart:io import to pubspec.dart

jakemac53:
Looks like the features env variable revert removed a dart:io import which was relied upon by a susequent change.


lib/src/pubspec.dart
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) 2012, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+import 'dart:io';
```
nex3: Nit: newline above.

PR #1727, opened 2017-10-20
Generic method cleanup

kevmoo:


kevmoo:
> You didn't have to make these different pull requests... different commits in the existing pull request would have been fine.

Good to know!

PR #1740, opened 2017-11-16
Fix analysis hints

nex3:


lib/src/asset/dart/serialize/aggregate_transform.dart
```diff
@@ -136,10 +136,8 @@ class ForeignAggregateTransform extends _ForeignBaseAggregateTransform
             deserializeStream(transform['primaryInputs'], deserializeAsset),
         super(transform);
 
-  Future<Asset> getInput(AssetId id) {
-    return call(_port, {'type': 'getInput', 'id': serializeId(id)})
-        .then(deserializeAsset);
-  }
+  Future<Asset> getInput(AssetId id) async => await deserializeAsset(
```
natebosch: [nit] the outer `await` shouldn't be necessary. Similar below

nex3: Done.

PR #1742, opened 2017-11-16
Move pub to the new dart2js API

hterkelsen:


lib/src/barback/dart2js_transformer.dart
```diff
@@ -225,36 +225,16 @@ class _BarbackCompilerProvider implements dart.CompilerProvider {
...
-  /// Whether or not verbose info messages should be logged.
-  var _verbose = false;
-
-  /// Whether an exception should be thrown on an error to stop compilation.
-  final _throwOnError = false;
-
-  /// This gets set after a fatal error is reported to quash any subsequent
-  /// errors.
-  var _isAborting = false;
+  _BarbackCompilerInput inputProvider;
```
nex3: Can you make these final?

lib/src/barback/dart2js_transformer.dart
```diff
@@ -277,10 +257,24 @@ class _BarbackCompilerProvider implements dart.CompilerProvider {
...
+class _BarbackCompilerInput extends compiler.CompilerInput {
+  final Transform _transform;
+  final AssetEnvironment _environment;
+  final Map<String, SourceFile> _sourceFiles;
 
-  /// A [CompilerInputProvider] for dart2js.
-  Future<String> provideInput(Uri resourceUri) {
+  _BarbackCompilerInput(this._transform, this._environment, this._sourceFiles);
+
+  @override
```
nex3: Nit: pub doesn't use `@override`.

lib/src/barback/dart2js_transformer.dart
```diff
@@ -293,8 +287,47 @@ class _BarbackCompilerProvider implements dart.CompilerProvider {
     });
   }
 
-  /// A [CompilerOutputProvider] for dart2js.
-  EventSink<String> provideOutput(String name, String extension) {
+  Future<String> _readResource(Uri url) {
+    return new Future.sync(() {
```
nex3: Can you convert this to `async`/`await`?

lib/src/barback/dart2js_transformer.dart
hterkelsen: Done

lib/src/barback/dart2js_transformer.dart
hterkelsen: I can't because they need to share the _sourceFiles map

lib/src/barback/dart2js_transformer.dart
hterkelsen: I thought I could get away with still using `compiler_unsupported` v1.22 but there have been changes to the new API since then. I will land a new version of `compiler_unsupported` then update this change.

The diagnostic handler is still a copy of `FormattingDiagnosticHandler`

hterkelsen: Done

hterkelsen:
I thought I could get away with still using `compiler_unsupported` v1.22 but there have been changes to the new API since then. I will land a new version of `compiler_unsupported` then update this change.

The diagnostic handler is still a copy of `FormattingDiagnosticHandler`

kevmoo:
No longer needed.

PR #1743, opened 2017-11-22
Represent Pubspec.dependencies as a map

nex3:
This cleans up some existing code, but more importantly it makes some
algorithms for the new version solver more efficient.

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -55,15 +55,18 @@ class Package {
   Map<String, PackageRange> get devDependencies => pubspec.devDependencies;
 
   /// The dependency overrides this package specifies in its pubspec.
-  Map<String, PackageRange> get dependencyOverrides => pubspec.dependencyOverrides;
+  Map<String, PackageRange> get dependencyOverrides =>
+      pubspec.dependencyOverrides;
 
   /// All immediate dependencies this package specifies.
   ///
   /// This includes regular, dev dependencies, and overrides.
   Map<String, PackageRange> get immediateDependencies {
-    return {}..addAll(dependencies)..addAll(devDependencies)
-    // Make sure to add these last so they replace normal dependencies.
-    ..addAll(dependencyOverrides);
+    // Make sure to add overrides last so they replace normal dependencies.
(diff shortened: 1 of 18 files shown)


kevmoo:
There's a plan for a new solver? I'm so excited!

nex3:
There's not just a plan, there's a working prototype!

lib/src/package.dart
```diff
@@ -49,31 +49,21 @@ class Package {
...
-    addToMap(dep) {
-      deps[dep.name] = dep;
-    }
-
-    dependencies.forEach(addToMap);
-    devDependencies.forEach(addToMap);
-
+  Map<String, PackageRange> get immediateDependencies {
+    return {}..addAll(dependencies)..addAll(devDependencies)
     // Make sure to add these last so they replace normal dependencies.
```
munificent: Putting the comment in the middle of the cascade is kind of weird. How about moving it up and changing "these" -> "overrides"?

lib/src/pubspec.dart
```diff
@@ -597,11 +597,11 @@ class Pubspec {
...
       SourceRegistry sources})
       : _version = version,
-        _dependencies = dependencies == null ? null : dependencies.toList(),
+        _dependencies = dependencies == null ? null : new Map.fromIterable(dependencies, key: (range) => range.name),
         _devDependencies =
-            devDependencies == null ? null : devDependencies.toList(),
+            devDependencies == null ? null : new Map.fromIterable(devDependencies, key: (range) => range.name),
         _dependencyOverrides =
-            dependencyOverrides == null ? null : dependencyOverrides.toList(),
+            dependencyOverrides == null ? null : new Map.fromIterable(dependencyOverrides, key: (range) => range.name),
```
munificent: Dartfmt?

lib/src/package.dart
nex3: Done.

lib/src/pubspec.dart
nex3: Done.

PR #1769, opened 2018-01-02
Throw an error if publish_to isn't an absolute URL

### LINKED ISSUE #1766: Require `publish_to` be in the form `http(s)://`
See https://www.dartlang.org/tools/pub/pubspec#supported-fields
See https://github.com/dart-lang/site-www/issues/494

Just 'localhost:8080' doesn't work. Neither does the example `pub.dartlang.org`

Would be good to fail early (during pubspec parse) than during upload w/ a weird error

nex3:
Closes #1766

### DIFF
--- a/lib/src/pubspec.dart
+++ b/lib/src/pubspec.dart
@@ -443,8 +443,12 @@ class Pubspec {
 
       // It must be "none" or a valid URL.
       if (publishTo != "none") {
-        _wrapFormatException(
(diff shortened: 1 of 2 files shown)


kevmoo:
dartfmt is sad https://travis-ci.org/dart-lang/pub/jobs/324351060

nex3:
@kevmoo Fixed.

PR #1772, opened 2018-01-03
Don't adjust SDK constraints whose min versions are prerelease

### LINKED ISSUE #1770: Invalid argument(s): Minimum version ("2.0.0-dev.15.0") must be less than maximum ("2.0.0-dev.14.0").
Running `pub --trace upgrade` causes this exception

[log.txt](https://github.com/dart-lang/pub/files/1600816/log.txt)

`pub get` (with existing `pubspec.lock`) completes without an error.

nex3:
Previously, if the user was using 1.2.3-dev.1 and a package had an SDK
constraint ">=1.2.3-dev.2", we would try to create an invalid
">=1.2.3-dev.2 <1.2.3-dev.1" constraint.

Closes #1770

### DIFF
--- a/lib/src/pubspec.dart
+++ b/lib/src/pubspec.dart
@@ -391,6 +391,9 @@ class Pubspec {
     if (!allowPreReleaseSdk) return false;
     if (!sdk.version.isPreRelease) return false;
     if (sdkConstraint.includeMax) return false;
+    if (sdkConstraint.min != null && sdkConstraint.min.isPreRelease) {
+      return false;
+    }
     if (sdkConstraint.max == null) return false;
     if (sdkConstraint.max.isPreRelease) return false;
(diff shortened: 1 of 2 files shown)


lib/src/pubspec.dart
```diff
@@ -391,6 +391,7 @@ class Pubspec {
     if (!allowPreReleaseSdk) return false;
     if (!sdk.version.isPreRelease) return false;
     if (sdkConstraint.includeMax) return false;
+    if (sdkConstraint.min?.isPreRelease ?? false) return false;
```
munificent: Maybe it's just me, but I think this would be easier to read as the more explicit:

```dart
if (sdkConstraint.min != null && sdkConstraint.min.isPreRelease) {
  return false;
}
```

nex3: Done.

PR #1779, opened 2018-01-11
Various refactors in preparation for adding error reporting

nex3:
Sending these out on their own to avoid dumping another giant pull
request.

See #912


lib/src/solver/incompatibility_cause.dart
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
natebosch: year += 1

lib/src/solver/failure.dart
```diff
@@ -2,201 +2,22 @@
...
-  final Version version;
-  final String feature;
-
-  String get _message =>
-      "$package $version doesn't have a feature named $feature";
 
-  MissingFeatureException(String package, this.version, this.feature,
-      Iterable<Dependency> dependencies)
-      : super(package, dependencies);
+  String toString() => "Tough luck, Chuck!";
```
munificent: Add a TODO to put something real here?

lib/src/solver/incompatibility.dart
```diff
@@ -44,39 +52,84 @@ class Incompatibility {
...
+      assert(terms.length == 1);
+      assert(terms.first.isPositive);
+
+      if (terms.first.constraint.isAny) {
+        return "no versions of ${_terseRef(terms.first)} "
+            "are compatible with the current SDK";
+      } else {
+        return "${terms.first.package.toTerseString()} is incompatible with "
+            "the current SDK";
+      }
```
munificent: Might be useful to show the SDK version number here, no?

lib/src/solver/incompatibility.dart
```diff
@@ -44,39 +52,84 @@ class Incompatibility {
...
+          return byRef.values;
+        }).toList(),
+        cause);
   }
 
-  Incompatibility._(this.terms);
+  Incompatibility._(this.terms, this.cause);
 
   String toString() {
+    if (cause == IncompatibilityCause.dependency) {
```
munificent: What do you think of moving most of this logic into IncompatibilityCause? Something like:

```dart
class Incompability {
  String toString() => cause.describe(terms);
}

abstract class IncompatibilityCause {
  String describe(List<Term> terms);
}

class SdkCause implements IncompatibilityCause {
  String describe(List<Term> terms) {
    // ...
  }
}
```

?

lib/src/solver/failure.dart
nex3: Done.

lib/src/solver/incompatibility.dart
nex3: It always weirds me out a bit to have a method like `Incompatibility.describe()` that's only ever supposed to be invoked by one specific caller... it feels like kind of the wrong separation of concerns.

lib/src/solver/incompatibility.dart
nex3: Added a TODO.

lib/src/solver/incompatibility_cause.dart
nex3: Done.

PR #1785, opened 2018-01-19
Add error reporting for the new version solver

nex3:


kevmoo:
A thought (for my contrived case)

```
Because repo_manager depends on angular 5.0.0-alpha+4 which depends on quiver >=0.22.0 <0.28.0, quiver >=0.22.0 <0.28.0 is required.
So, because repo_manager depends on quiver ^0.21.0, version solving failed.
```

My attempt at tweaks...

```
repo_manager depends on angular 5.0.0-alpha+4 which depends on quiver >=0.22.0 <0.28.0
  so quiver >=0.22.0 <0.28.0 is required.
but, repo_manager also depends on quiver ^0.21.0, with is incompatible with quiver >=0.22.0 <0.28.0
so version solving failed.
```

This is 1000x than what we have now. Feel free to tuck this away as a p2 for later.

nex3:
@kevmoo What would your proposed output do if the terminal was too narrow to fit the entire first line?

> ```
> repo_manager also depends on quiver ^0.21.0, with is incompatible with quiver >=0.22.0 <0.28.0
> ```

My instinct is that including this level of explanation at every step would make the output in even moderately complex cases a *lot* more verbose. There comes a point at which explicitness is actually unhelpful because there's so much raw text that peoples eyes gloss over. Is this worth it?

lib/src/solver/failure.dart
```diff
@@ -19,6 +23,294 @@ class SolveFailure implements ApplicationException {
...
   }
 
-  // TODO(nweiz): Produce a useful error message.
-  String toString() => "Tough luck, Chuck!";
+  /// Describes how [incompatibility] was derived, and thus why version solving
+  /// failed.
+  ///
+  /// See https://github.com/dart-lang/pub/tree/master/doc/solver.md#error-reporting
+  /// for details on how this algorithm works.
+  String toString() {
```
munificent: This is quite the method! Given that it has state and a number of sub-functions with their own doc comments, I think it's worth moving it out to a separate class that this calls.

lib/src/solver/failure.dart
```diff
@@ -19,6 +23,294 @@ class SolveFailure implements ApplicationException {
...
+        lines.add(new Pair(message, null));
+      }
+    }
+
+    /// Returns whether we can collapse the derivation of [incompatibility].
+    ///
+    /// If [incompatibility] is only used to derive one other incompatibility,
+    /// it may make sense to skip that derivation and just derive the second
+    /// incompatibility directly from three causes. This is usually clear enough
+    /// to the user, and makes the proof much terser.
```
munificent: Can you give an example? (If that's not easy, feel free to ignore this.)

lib/src/solver/failure.dart
nex3: Done.

lib/src/solver/failure.dart
nex3: Done.

PR #1796, opened 2018-02-02
Use .dart_tool/pub as the cache directory

nex3:


kevmoo:
If we're really going to run with this

1) Awesome! So glad!
2) We should create a doc (or meta bug) to track the places that should be updated
  - stagehand
  - open source templates
  - github ignores for languages
  - site-www
3) email dart-core-eng an FYI
4) email announce@ an FYI

CC @kwalrath

kevmoo:
Moved my comment to https://github.com/dart-lang/sdk/issues/32030

PR #1797, opened 2018-02-03
Added more verbose error description for upload size validator

### LINKED ISSUE #1791: SizeValidator should give a better error message when the project is not in VCS and the size exceeds the limit
# Current behavior
When running `pub publish`, and when the project is not in Git, you might get the following error (especially with a Flutter plugin package, which can get quite large):

> Your package is 180 MB. Hosted packages must be smaller than 100 MB.

The reason for this is that if the project is not in Git, the `.gitignore` file is ignored, and every file in the project gets included in the package. 

This might not be immediately obvious to newcomers, although it's [documented here](https://www.dartlang.org/tools/pub/publishing#what-files-are-published). Searching the error message on Google doesn't help solving it.

# Expected behavior
If [SizeValidator#validate](https://github.com/dart-lang/pub/blob/2895b1d52c7893854a81ba5feda0ae612998e278/lib/src/validator/size.dart#L20) detects that the package size is over the limit, it also checks if the project is in source control (or if the `.gitignore` file is ignored).

If the project is not in source control, something like this appears:

> Your package is 180 MB. Hosted packages must be smaller than 100 MB.
> Your .gitignore has no effect since your project does not appear to be in version control.

Or maybe something even more explicit.

The reason why I didn't have my package in Git was that I'm embarrassed about the initial code quality, and also am such a git noob that I didn't know how to use the interactive rebase. Also, I should've just RTFM, but I was being lazy. I'm pretty sure a lot of people could be like me here. :)

gagoman:
Closes #1791

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -109,7 +109,7 @@ class Package {
   }
 
   /// Returns whether or not this package is in a Git repo.
-  bool get _inGitRepo {
+  bool get inGitRepo {
     if (_inGitRepoCache != null) return _inGitRepoCache;
 
     if (dir == null || !git.isInstalled) {
@@ -242,7 +242,7 @@ class Package {
     // path package, since re-parsing a path is very expensive relative to
     // string operations.
     Iterable<String> files;
-    if (useGitIgnore && _inGitRepo) {
+    if (useGitIgnore && inGitRepo) {
       // List all files that aren't gitignored, including those not checked in
       // to Git. Use [beneath] as the working dir rather than passing it as a
       // parameter so that we list a submodule using its own git logic.
--- a/lib/src/validator/size.dart
+++ b/lib/src/validator/size.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:math' as math;
 
 import '../entrypoint.dart';
+import '../io.dart' as io;
 import '../validator.dart';
 
 /// The maximum size of the package to upload (100 MB).
@@ -21,8 +22,21 @@ class SizeValidator extends Validator {
     return packageSize.then((size) {
       if (size <= _MAX_SIZE) return;
       var sizeInMb = (size / math.pow(2, 20)).toStringAsPrecision(4);
-      errors.add("Your package is $sizeInMb MB. Hosted packages must be "
-          "smaller than 100 MB.");
+      // Current implementation of Package.listFiles skips hidden files
+      var ignoreExists = io.fileExists(entrypoint.root.path('.gitignore'));
+
+      var error = new StringBuffer("Your package is $sizeInMb MB. Hosted "
+          "packages must be smaller than 100 MB.");
+
+      if (ignoreExists && !entrypoint.root.inGitRepo) {
+        error.write(" Your .gitignore has no effect since your project "
+            "does not appear to be in version control.");
+      } else if (!ignoreExists && entrypoint.root.inGitRepo) {
+        error.write(" Consider adding a .gitignore to avoid including "
+            "temporary files.");
+      }
+
+      errors.add(error.toString());
     });
   }
 }
--- a/lib/src/validator/size.dart
+++ b/lib/src/validator/size.dart
@@ -6,7 +6,7 @@ import 'dart:async';
 import 'dart:math' as math;
 
 import '../entrypoint.dart';
-import '../io.dart' as io;
+import '../io.dart';
 import '../validator.dart';
 
 /// The maximum size of the package to upload (100 MB).
@@ -23,7 +23,7 @@ class SizeValidator extends Validator {
       if (size <= _MAX_SIZE) return;
       var sizeInMb = (size / math.pow(2, 20)).toStringAsPrecision(4);
       // Current implementation of Package.listFiles skips hidden files
-      var ignoreExists = io.fileExists(entrypoint.root.path('.gitignore'));
+      var ignoreExists = fileExists(entrypoint.root.path('.gitignore'));
 
       var error = new StringBuffer("Your package is $sizeInMb MB. Hosted "
           "packages must be smaller than 100 MB.");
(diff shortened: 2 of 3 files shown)


lib/src/package.dart
```diff
@@ -108,6 +108,9 @@ class Package {
     }));
   }
 
+  /// Returns whether or not this package is under source control system.
+  bool get isUnderSourceControl => [_inGitRepo].any((e) => e);
```
gagoman: As it is a public property, I thought it will make more sense to name it more general. E.g. Mercurial support will be added.

test/validator/size_test.dart
```diff
@@ -18,14 +18,56 @@ ValidatorCreator size(int size) {
...
+          contains("Hosted packages must be smaller than 100 MB."),
+          contains(
+              "Your .gitignore has no effect since your project does not appear to be in version control.")));
+      expectValidationError(size(100 * math.pow(2, 20) + 1),
+          errorsMatcher: errorsMatcher);
+    });
+
+    test('package is under source control and no .gitignore exists', () async {
+      await d.validPackage.create();
+      await d.git(appPath).create();
```
gagoman: There are tests that call `ensureGit` and tests without it. Which way is more preferred?

lib/src/validator/size.dart
```diff
@@ -21,8 +24,21 @@ class SizeValidator extends Validator {
...
+
+      var error = new StringBuffer(
+          "Your package is $sizeInMb MB. Hosted packages must be smaller than 100 MB.");
+
+      if (ignoreExists && !entrypoint.root.isUnderSourceControl) {
+        error.write(
+            " Your .gitignore has no effect since your project does not appear to be in version control.");
+      } else if (!ignoreExists && entrypoint.root.isUnderSourceControl) {
+        error.write(
+            " All files were included in the package since your project does not have .gitignore.");
```
gagoman: English is not my native language so I suppose this can be written better.

lib/src/package.dart
nex3: I think it's fine to just make `_inGitRepo` public here. We can always change it later if we do indeed add support for other systems.

lib/src/validator/size.dart
```diff
@@ -21,8 +24,21 @@ class SizeValidator extends Validator {
     return packageSize.then((size) {
       if (size <= _MAX_SIZE) return;
       var sizeInMb = (size / math.pow(2, 20)).toStringAsPrecision(4);
-      errors.add("Your package is $sizeInMb MB. Hosted packages must be "
-          "smaller than 100 MB.");
+      // Current implementation of Package.listFiles skips hidden files
+      var ignoreExists = fileExists(join(entrypoint.root.dir, '.gitignore'));
```
nex3: You can write the path as `entrypoint.root.path('.gitignore')`.

lib/src/validator/size.dart
```diff
@@ -8,6 +8,9 @@ import 'dart:math' as math;
 import '../entrypoint.dart';
 import '../validator.dart';
 
+import 'package:path/path.dart';
```
nex3: Style note: the `path` package is generally imported with the prefix `p`, since it exposes a bunch of very terse names that would otherwise be likely to conflict.

lib/src/validator/size.dart
```diff
@@ -8,6 +8,9 @@ import 'dart:math' as math;
...
+import 'package:pub/src/io.dart';
```
nex3: It's best to import files within the same `lib/` directory using relative imports. Otherwise you can run into canonicalization issues where the Dart VM ends up loading the same file twice.

lib/src/validator/size.dart
nex3: I'd write "Consider adding a .gitignore to avoid including temporary files."

test/validator/utils.dart
```diff
@@ -10,8 +10,9 @@ void expectNoValidationError(ValidatorCreator fn) {
   expect(validatePackage(fn), completion(pairOf(isEmpty, isEmpty)));
 }
 
-void expectValidationError(ValidatorCreator fn) {
-  expect(validatePackage(fn), completion(pairOf(isNot(isEmpty), anything)));
+void expectValidationError(ValidatorCreator fn,
+    {Matcher errorsMatcher = isNotEmpty}) {
+  expect(validatePackage(fn), completion(pairOf(errorsMatcher, anything)));
```
nex3: Rather than doing this, I'd add an `expectSizeValidationError()` function to `size_test.dart`. See `dependency_test.dart`'s `expectDependencyValidationWarning()`.

lib/src/validator/size.dart
```diff
@@ -21,8 +24,21 @@ class SizeValidator extends Validator {
     return packageSize.then((size) {
       if (size <= _MAX_SIZE) return;
       var sizeInMb = (size / math.pow(2, 20)).toStringAsPrecision(4);
-      errors.add("Your package is $sizeInMb MB. Hosted packages must be "
-          "smaller than 100 MB.");
+      // Current implementation of Package.listFiles skips hidden files
+      var ignoreExists = fileExists(join(entrypoint.root.dir, '.gitignore'));
+
+      var error = new StringBuffer(
+          "Your package is $sizeInMb MB. Hosted packages must be smaller than 100 MB.");
```
nex3: A lot of these strings are longer than the 80-character line length. Unfortunately, the Dart formatter will only change whitespace, so you'll have to manually split the strings up to get them to fit.

test/validator/utils.dart
gagoman: Is it too general? Adding `expectSizeValidationError` (similar to `expectDependencyValidationWarning`) will result in almost same code.

lib/src/validator/size.dart
gagoman: Will fix.

lib/src/validator/size.dart
gagoman: Will do.

lib/src/validator/size.dart
gagoman: Sure.

lib/src/validator/size.dart
gagoman: `path` will not be used in next revision. Thank you for a tip.

nex3:
Thanks!

PR #1799, opened 2018-02-06
Triage tests that use the new solver

nex3:
This fixes a few bugs and marks a bunch of tests as skipped that
exercise functionality that's currently not supported by the solver.
As that functionality gets added, we'll re-enable those tests.

See #912


test/dependency_override_test.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+@Skip()
```
munificent: Leave a message explaining why it's skipped? You don't have to at all of them, but it would be good to have some kind of TODO to make sure these are addressed before it gets merged into master.

PR #1810, opened 2018-02-21
Improve the output of the new version solver

nex3:


lib/src/solver/incompatibility.dart
```diff
@@ -223,25 +222,24 @@ class Incompatibility {
         .map((term) => _terse(term, details))
         .join(' or ');
 
+    var buffer =
+        new StringBuffer("${_terse(thisPositive, details, allowEvery: true)} ");
```
munificent: The interpolation isn't buying you much. Maybe just:

```dart
new StringBuffer(_terse(thisPositive, details, allowEvery: true) + " ");
```

nex3: Done.

PR #1815, opened 2018-02-28
Gracefully handle unknown sources

nex3:
Also add some related improvements.


lib/src/solver/version_solver.dart
```diff
@@ -305,6 +306,16 @@ class VersionSolver {
         .toList();
     if (unsatisfied.isEmpty) return null;
 
+    // If we require a packages from an unknown source, add an incompatibility
```
munificent: "package"

PR #1818, opened 2018-03-07
Improve error output for mismatched SDK constraints

nex3:


lib/src/log.dart
```diff
@@ -336,39 +336,46 @@ void processResult(String executable, PubProcessResult result) {
 }
 
 /// Logs an exception.
-void exception(exception, [StackTrace trace]) {
-  if (exception is SilentException) return;
+void exception(toLog, [StackTrace trace]) {
```
munificent: Maybe annotate `toLog` with `Object`? I think we generally want to move away from implicit dynamic.

nex3: I'm still holding out hope that by the time we actually start enforcing no implicit dynamic, explicitly-un-annotated variables like this will default to `Object`...

PR #1822, opened 2018-03-09
Allow dart2js to compile with the CFE.

sigmundch:
Dart2js needs an extra argument to locate the dart2js_platform.dill file. The provider for the compiler needs to read this file as a binary file.

This change is necessary before we can make dart2js use the CFE by default.

Two questions:
* do you need `compiler_unsupported` to get updated? Its current version doesn't have the `--platform-binaries` flag. I'm investigating what's the process to update that package, but it appears we are not running dart2js as part of pub tests (when I run `pub run test`, all tests pass.)
* do we need to indicate a minimal sdk version where this pub will work? (I assume that it might not be needed, because pub is always shipped with the sdk.

### DIFF
--- a/lib/src/barback/dart2js_transformer.dart
+++ b/lib/src/barback/dart2js_transformer.dart
@@ -161,7 +161,8 @@ class Dart2JSTransformer extends Transformer implements LazyTransformer {
         suppressPackageWarnings:
             _configBool('suppressPackageWarnings', defaultsTo: true),
         terse: _configBool('terse'),
-        includeSourceMapUrls: _generateSourceMaps);
+        includeSourceMapUrls: _generateSourceMaps,
+        platformBinaries: provider.libraryRoot.resolve('lib/_internal/').path);
   }
 
   /// Parses and returns the "commandLineOptions" configuration option.
@@ -395,7 +396,13 @@ class _BarbackCompilerProvider implements dart.CompilerProvider {
     return new Future.sync(() {
       // Find the corresponding asset in barback.
       var id = _sourceUrlToId(url);
-      if (id != null) return _transform.readInputAsString(id);
+      if (id != null) {
+        if (id.extension == '.dill') {
+          return _transform.readInput(id);
+        } else {
+          return _transform.readInputAsString(id);
+        }
+      }
 
       // Don't allow arbitrary file paths that point to things not in packages.
       // Doing so won't work in Dartium.
--- a/lib/src/barback/dart2js_transformer.dart
+++ b/lib/src/barback/dart2js_transformer.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:analyzer/analyzer.dart';
+import 'package:async/async.dart';
 import 'package:barback/barback.dart';
 import 'package:collection/collection.dart';
 import 'package:path/path.dart' as p;
@@ -281,7 +282,7 @@ class _BarbackCompilerProvider implements dart.CompilerProvider {
   }
 
   /// A [CompilerInputProvider] for dart2js.
-  Future provideInput(Uri resourceUri) {
+  Future/* <String | List<int>> */ provideInput(Uri resourceUri) {
     // We only expect to get absolute "file:" URLs from dart2js.
     assert(resourceUri.isAbsolute);
     assert(resourceUri.scheme == "file");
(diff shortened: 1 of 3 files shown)


lib/src/barback/dart2js_transformer.dart
```diff
@@ -395,7 +396,13 @@ class _BarbackCompilerProvider implements dart.CompilerProvider {
     return new Future.sync(() {
       // Find the corresponding asset in barback.
       var id = _sourceUrlToId(url);
-      if (id != null) return _transform.readInputAsString(id);
+      if (id != null) {
+        if (id.extension == '.dill') {
+          return _transform.readInput(id);
```
nex3: `_readResource()` is declared to return a `Future<String>`, but `readInput()` returns a `Stream<List<int>>`.

lib/src/barback/pub_package_provider.dart
```diff
@@ -147,8 +148,10 @@ class PubPackageProvider implements StaticPackageProvider {
           files = files.map((file) => file.replaceAll(trailingUnderscore, ""));
         }
 
+        bool isIncluded(extension) =>
+            extension == '.dart' || extension == '.dill';
```
nex3: I would just inline this function.

natebosch:
Should we focus on getting this working with `build_runner` instead? We hope to remove dart2js from pub sometime soon.

cc @kevmoo @jakemac53 @matanlurey

nex3:
I for one would be happy to see this infrastructure deleted as soon as possible.

sigmundch:
+1

However, the reason I'm doing this is, is because this blocks us from enabling the FE in dart2js by default (which we were hoping to land this week in bleeding edge). Until everyone is not using pub-build, I'm afraid we cannot avoid small changes like these to unblock the rollout process.

We have a flag that enables the old frontend, and I was considering using that here instead, but I'd be afraid of breaking users without us realizing it.

@nex3 - thanks for your comments, please do not review yet the new commit, it is not complete, I'll let you know when it is.

kevmoo:
We're not ready to pull the world off pub – I think this is prudent

lib/src/barback/dart2js_transformer.dart
sigmundch: Done. Dart2js expects both String or List<int>, however I don't understand how returning a Stream didn't cause a crash for me before :-/. Anyhow, I just re-tested this on the SDK repo and it appears to work as expected.

lib/src/barback/pub_package_provider.dart
sigmundch: Done

matanlurey:
@kevmoo:

> We're not ready to pull the world off pub – I think this is prudent

Can we discuss this internally? This is going to be a non-trivial amount of time spent, and we don't know how successful we can be/what benefit this gives customers over build_runner.

lib/src/barback/dart2js_transformer.dart
```diff
@@ -280,7 +281,7 @@ class _BarbackCompilerProvider implements dart.CompilerProvider {
   }
 
   /// A [CompilerInputProvider] for dart2js.
-  Future<String> provideInput(Uri resourceUri) {
+  Future provideInput(Uri resourceUri) {
```
nex3: I'd write `Future/*<String | List<int>>*/`

lib/src/barback/dart2js_transformer.dart
```diff
@@ -391,11 +392,20 @@ class _BarbackCompilerProvider implements dart.CompilerProvider {
...
+    return new Future.sync(() async {
       // Find the corresponding asset in barback.
       var id = _sourceUrlToId(url);
-      if (id != null) return _transform.readInputAsString(id);
+      if (id != null) {
+        if (id.extension == '.dill') {
+          var stream = _transform.readInput(id);
+          var bytes = <int>[];
+          await stream.forEach(bytes.addAll);
+          return bytes;
```
nex3: Use [`collectBytes()`](https://www.dartdocs.org/documentation/async/1.13.3/async/collectBytes.html) to do this more efficiently.

lib/src/barback/dart2js_transformer.dart
sigmundch: Done (I had to add the extra space to avoid an analyzer warning about generic comment syntax, though)

lib/src/barback/dart2js_transformer.dart
sigmundch: Done, thanks!

PR #1829, opened 2018-03-15
Reenable the common front-end in dart2js by creating a .packages file

sigmundch:
Note: the `.packages` file is not available as an asset, so it wont be emitted in the output build folder. If you want to make it a asset, could you please give me some hints as to where/how I should do so?

### DIFF
--- a/lib/src/dart.dart
+++ b/lib/src/dart.dart
@@ -51,8 +51,8 @@ abstract class CompilerProvider {
 /// Uses [provider] to communcate between dart2js and the caller. Returns a
 /// future that completes when compilation is done.
 ///
-/// By default, the package root is assumed to be adjacent to [entrypoint], but
-/// if [packageRoot] is passed that will be used instead.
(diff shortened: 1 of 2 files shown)


lib/src/dart.dart
```diff
@@ -102,21 +101,22 @@ Future compile(String entrypoint, CompilerProvider provider,
...
+    packageConfig = p.join(p.dirname(entrypoint), '.packages');
   } else {
-    packageRoot = p.normalize(p.absolute(packageRoot));
+    packageConfig = p.normalize(p.absolute(packageConfig));
   }
 
   await compiler.compile(
       p.toUri(entrypoint),
       provider.libraryRoot,
-      p.toUri(appendSlash(packageRoot)),
```
nex3: Are we still using `appendSlash()`? If not, can you delete it?

sigmundch: Done.

sigmundch:
Sounds good. Thank you!

PR #1836, opened 2018-03-20
Refactor some solver stuff

nex3:


lib/src/package_name.dart
```diff
@@ -243,11 +246,13 @@ class PackageRange extends PackageName {
 
     var buffer = new StringBuffer(name);
     if (detail.showVersion ??
-        !constraint.isAny || (source is! PathSource && source is! GitSource)) {
+        !isRoot &&
+            (!constraint.isAny ||
+                (source is! PathSource && source is! GitSource))) {
```
munificent: Operators nested four deep is pretty rough. Maybe break some of this out into local variables so it's clearer what they represent?

Or perhaps add a getter on Source that means what `(source is! PathSource && source is! GitSource)` means? Is it "has multiple versions"?

nex3: We do have `Source.hasMultipleVersions`, but I don't think it's exactly the right fit here—we usually expect users to include version constraints for SDK dependencies, even though they technically only have one version available at a time.

I've factored this out into a getter to reduce nesting.

Issue #1853, opened 2018-03-30
Need help understanding `pub get` failure


aam:
Please help understand error  of (recently updated) `pub get` when run in  flutter/package/flutter-tools project:
```ERR : Because no versions of analyzer from path match <0.31.2-alpha.1 or >0.31.2-alpha.1 and analyzer
    |   0.31.2-alpha.1 from path depends on front_end 0.1.0-alpha.11, every version of analyzer from path
    |   requires front_end 0.1.0-alpha.11.
    | So, because flutter_tools depends on both front_end from path and analyzer from path, version
    |   solving failed.
```

`pub get` started to fail like that after upgrade from 875d35005a7d33f367d70a3e31e9d3bad5d1ebd8 to bc2b5ff80e50b4b1c317230e8963c1fada3cfe73.

Here is full output:
```
$FH/flutter/packages/flutter_tools $ $FH/flutter/bin/cache/dart-sdk/bin/pub -v get

FINE: Pub 2.0.0-edge.af1436931b93e755d38223c487d33a0a1f5eadf5
SLVR: fact: pub itself depends on barback >=0.15.0 <0.15.3
SLVR: fact: when barback is in use pub itself depends on source_span ^1.0.0
SLVR: fact: when barback is in use pub itself depends on stack_trace >=0.9.1 <2.0.0
SLVR: fact: when barback is in use pub itself depends on async >=1.8.0 <3.0.0
SLVR: fact: flutter_tools is 0.0.0
SLVR: derived: flutter_tools
SLVR: fact: flutter_tools depends on shelf_static 0.2.7
SLVR: fact: flutter_tools depends on file 3.0.0
SLVR: fact: flutter_tools depends on mustache 1.0.0
SLVR: fact: flutter_tools depends on web_socket_channel 1.0.7
SLVR: fact: flutter_tools depends on meta 1.1.2
SLVR: fact: flutter_tools depends on petitparser 1.7.6
SLVR: fact: flutter_tools depends on matcher 0.12.1+4
SLVR: fact: flutter_tools depends on boolean_selector 1.0.3
SLVR: fact: flutter_tools depends on source_map_stack_trace 1.1.4
SLVR: fact: flutter_tools depends on mime 0.9.6
SLVR: fact: flutter_tools depends on watcher 0.9.7+7
SLVR: fact: flutter_tools depends on quiver 0.28.0
SLVR: fact: flutter_tools depends on package_config 1.0.3
[...]

PR #1855, opened 2018-03-30
Improve version solver output

nex3:
Partially addresses #1853

### DIFF
--- a/test/version_solver_test.dart
+++ b/test/version_solver_test.dart
@@ -202,6 +202,27 @@ void withLockFile() {
       'newdep': '2.0.0'
     }, tries: 2);
   });
+
+  // Issue 1853
+  test(
+      "produces a nice message for a locked dependency that's the only "
+      "version of its package", () async {
+    await servePackages((builder) {
+      builder.serve('foo', '1.0.0', deps: {'bar': '>=2.0.0'});
+      builder.serve('bar', '1.0.0');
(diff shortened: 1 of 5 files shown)


test/version_solver_test.dart
```diff
@@ -638,6 +659,27 @@ void badSource() {
...
+      builder.serve('baz', '1.0.0');
+    });
+    await d.dir('baz', [d.libPubspec('baz', '1.0.0')]).create();
+
+    await d.appDir({
+      'foo': 'any',
+      'baz': {'path': '../baz'}
+    }).create();
+    await expectResolves(error: equalsIgnoringWhitespace('''
+      Because every version of foo depends on bar any which depends on baz any,
```
munificent: The "any" doesn't seem to add much value here and makes it a little harder to read. What do you think of just showing the package name if the constraint includes all versions?

> Because every version of foo depends on bar which depends on baz, every version of foo requires baz from hosted.
> So, because myapp depends on both baz from path and foo, version solving failed.

?

nex3: I considered that, but I think it makes sense to reproduce the constraints from the packages' pubspecs here even if they're `any`. Just saying `bar depends on baz` risks making it unclear that there's no other version of `bar` that could work.

Also, since we discourage `any` constraints in real packages, this is unlikely to come up very often in practice.

PR #1858, opened 2018-04-04
Clarify the meaning of negative terms

matklad:
Hi! 

    not foo ^1.0.0

could mean two things: either “some foo, but not 1.*.*” or “foo 1.*.* is not selected”. I don’t know which meaning is correct, but I bet on the second one. Let’s clarify this in the docs?


matklad:
Submitted the cla just in case!

PR #1865, opened 2018-04-10
Generalize support for optional SDKs

nex3:


lib/src/entrypoint.dart
```diff
@@ -24,33 +23,30 @@ import 'package.dart';
...
+  // ```
+  //
+  // and the new-style constraint:
+  //
+  // ```yaml
+  // sdks:
+  //   dart: ">=1.2.3 <2.0.0"
+  // ```
+  var sdkNames = sdks.keys.map((name) => "  " + name).join('|');
+  return new RegExp(r'^(' + sdkNames + r'|sdk): "?([^"]*)"?$', multiLine: true);
```
natebosch: Perhaps a bit out of scope for the review - but why do we need this regex to be multiLine?

lib/src/lock_file.dart
```diff
@@ -262,12 +249,12 @@ class LockFile {
...
-      sdks['flutter'] = flutterSdkConstraint.toString();
-    }
-
-    var data = {'sdks': sdks, 'packages': packageMap};
+    var data = {
+      'sdks': mapMap(sdkConstraints,
+          value: (_, constraint) => constraint.toString()),
+      'packages': packageMap
+    };
+    print(data);
```
natebosch: Is this intended? It doesn't look like there used to be a `print` and this is the first in the file

PR #1866, opened 2018-04-10
Run dart2_fix to get lower case constants

natebosch:
- Rename log.json to log.jsonLog
- Rename json when used as a var


natebosch:
Closing in favor of https://github.com/dart-lang/pub/pull/1868 since it uses `jsonEncode` and doesn't need to rename the variables named `json`.

Issue #1756, opened 2017-12-06
git dependencies on windows causes error with missing name comment


ride4sun:
I run flutter packages upgrade on windows with this pubspec.yaml file.
This works well on Mac OSX and Travis.
Attached log output on the end.
```
name: lookatmybaby_flutter
description: Look at my baby Flutter client app.
version: 0.1.0

publish_to: none # prevent accidental publishing

dev_dependencies:
  bwu_grinder_tasks: ^0.1.1-dev
  grinder: ^0.8.0
  http: ^0.11.3
  stack_trace: '>= 1.8.2 < 1.10.0'
  uuid: ^0.5.3

  flutter_test:
    sdk: flutter

  flutter_driver:
    sdk: flutter

dependencies:
#  Pluto:
#    git: git://github.com/FaisalAbid/pluto
  flutter:
    sdk: flutter
#  lookatmybaby_shared:
#    path: '../lookatmybaby_shared'
  built_collection: ^2.1.0
  built_redux: ^7.0.0
#  built_redux_thunk: ^0.4.0
  built_value: ^4.0.0
  collection: ^1.14.0
#  contact_picker: ^0.0.1+2
  device_info: ^0.0.4
  google_sign_in: #^1.0.1
    git:
      url: git://github.com/flutter/plugins.git
      path: 'packages/google_sign_in'
  googleapis: ^0.47.0
  googleapis_auth: ^0.2.3+6
  firebase_analytics: #^0.0.4
    git:
      url: git://github.com/flutter/plugins.git
      path: 'packages/firebase_analytics'
  firebase_auth: #^0.1.1
    git:
      url: git://github.com/flutter/plugins.git
      path: 'packages/firebase_auth'

  firebase_database: #^0.0.12
    git:
      url: git://github.com/flutter/plugins.git
      path: 'packages/firebase_database'

  firebase_storage: #0.0.6
    git:
      url: git://github.com/flutter/plugins.git
      path: 'packages/firebase_storage'

#  firebase_storage:
#    git:
#      url: git://github.com/franzsilva/plugins.git
#      path: packages/firebase_storage
  flutter_image: ^0.0.2
  image_picker: #^0.1.1
    git:
      url: git://github.com/flutter/plugins.git
      path: 'packages/image_picker'

  intl: ^0.15.0
  logging: ^0.11.3
  observable: ^0.20.4+1
  path_provider: ^0.2.1+1
  shared_preferences: #^0.2.5
    git:
[...]

PR #1871, opened 2018-04-14
Drop build and serve

### LINKED ISSUE #1857: Remove build & serve
Master issue: https://github.com/dart-lang/sdk/issues/32593

Blocked by [Dart2 Beta 3](https://github.com/dart-lang/sdk/milestone/53) – this should land ASAP after (but not before)

natebosch:
Closes #1857 

- Drop `lib/src/{barback,dartdevc,asset}` and relevant tests
- Add `lib/src/asset/id.dart` since non-building code uses this type
  from barback.
- Clean up analyzer errors, droping any code paths that relate to
  dealing with Transformers, for example the `run` command's ability to
  run with transformed code, or any parsing and plumbing for transformer
  config.
- Clean up any tests that were checking behavior of transformers or the
  build or serve command.
- Remove any dependencies for which I could no longer find an import.


natebosch:
Note that a `pub run` on a package which _does_ expected to be transformed will now likely fail in mysterious ways since it will be run directly without transformers. I tried to see if we could cheaply keep around just enough of the code paths to support a warning for this case but it didn't seem like it would be worth doing.

natebosch:
The `build` or `serve` commands now `exit(1)` after printing in red to stderr:

```
Dart 2 has a new build system. Learn how to migrate from pub build and
pub serve: https://webdev.dartlang.org/dart-2
```

kevmoo:
one failing test – and one unformatted file. Nice...

natebosch:
@nex3 - can you help me understand how this test works?
https://github.com/dart-lang/pub/blob/master/test/global/run/uses_old_lockfile_test.dart

I see that it's setting up a fake package for `foo`. How does the import to `bar` work here? I don't see where `bar.main()` is getting set up with some behavior. The failure `Could not find bin/script.dart in package foo.` makes sense to me, how did this work before?

nex3:
Will do a full review in a bit, but:

> Note that a `pub run` on a package which _does_ expected to be transformed will now likely fail in mysterious ways since it will be run directly without transformers. I tried to see if we could cheaply keep around just enough of the code paths to support a warning for this case but it didn't seem like it would be worth doing.

I think this is fine. I think the vast majority of users of transformers in practice used them for building web apps.

> @nex3 - can you help me understand how this test works?
> https://github.com/dart-lang/pub/blob/master/test/global/run/uses_old_lockfile_test.dart
> 
> I see that it's setting up a fake package for `foo`. How does the import to `bar` work here? I don't see where `bar.main()` is getting set up with some behavior. The failure `Could not find bin/script.dart in package foo.` makes sense to me, how did this work before?

This tries to simulate what `pub global activate` would have created with older pub versions—the `foo` package isn't fake, it's just added to the cache manually via `pub cache add` on line 26.

The `bar` import worked because `pub` would load up an asset environment every time it ran an executable. The right behavior now is probably to detect an executable without a snapshot and try to compile that snapshot.

natebosch:
> The `bar` import worked because `pub` would load up an asset environment every time it ran an executable.

If I'm understanding the test correctly it seems to expect that the asset environment would have successfully provided that file, allow the `pub run` to execute the script, and somehow make `bar.main()` return `"bar 1.0.0"`

Or am I misunderstanding this?

natebosch:
The immediate problem is that without the Barback asset environment we only know how to find a package [if it has a `.packages` file](https://github.com/dart-lang/pub/pull/1871/files#diff-6987800bc939a3fd5bb3d711da38b037R115).

I _think_ the "1.6-style lockfile" for global packages might only work through the asset environment.

Before I spend time on a fix - is this something we feel we need to continue to support?

nex3:
The immediate problem is that without the Barback asset environment we only know how to find a package [if it has a `.packages` file](https://github.com/dart-lang/pub/pull/1871/files#diff-6987800bc939a3fd5bb3d711da38b037R115).

> I _think_ the "1.6-style lockfile" for global packages might only work through the asset environment.

Right. Before we started creating snapshots (which was well after 1.6), all `pub global run` invocations used the asset environment, so they weren't set up to work in any other way.

> Before I spend time on a fix - is this something we feel we need to continue to support?

If it's a huge amount of effort you could drop it, but I *think* all you should need to do is something like

```dart
if (!fileExists(snapshotPath)) await _precompileExecutables(package);
```

There's a lot of psychological value for users in guaranteeing that the package manager will never choke on data it itself created, even in edge cases like this.

lib/src/asset/id.dart
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Like the AssetId class from barback but without the barback dependency.
+class AssetId {
```
nex3: Is it possible to get rid of this entirely? Looking through the remaining uses, it seems like almost all of them exist in contexts where the package is determined by context, so you could just use the executable name.

lib/src/cached_package.dart
```diff
@@ -3,23 +3,14 @@
...
-import 'pubspec.dart';
 
 /// A [Package] whose `lib` directory has been precompiled and cached.
 ///
 /// When users of this class request path information about files that are
-/// cached, this returns the cached information. It also wraps the package's
-/// pubspec to report no transformers, since the transformations have all been
-/// applied already.
+/// cached, this returns the cached information.
 class CachedPackage extends Package {
```
nex3: This shouldn't be necessary anymore—there's nothing for pub to precompile in the lib directory.

lib/src/command/barback.dart
```diff
@@ -2,231 +2,23 @@
...
-        pluralize(singularVerb, directoryNames.length, plural: pluralVerb);
-
-    var result = "$directories $names $verb";
-    if (suffix != null) result += " $suffix";
-    result += ".";
-
-    return result;
+    log.error(log.red("Dart 2 has a new build system. Learn how to migrate "
+        "from ${log.bold('pub build')} and\n"
+        "${log.bold('pub serve')}: https://webdev.dartlang.org/dart-2\n"));
```
nex3: Nit: use `fail()` instead of `log.error()` and `exit()`.

lib/src/command/global_run.dart
```diff
@@ -20,14 +19,9 @@ class GlobalRunCommand extends PubCommand {
...
   bool get allowTrailingOptions => false;
 
-  /// The mode for barback transformers.
-  BarbackMode get mode => new BarbackMode(argResults["mode"]);
-
   GlobalRunCommand() {
     argParser.addFlag("checked",
         abbr: "c", help: "Enable runtime type checks and assertions.");
-    argParser.addOption("mode",
-        defaultsTo: "release", help: 'Mode to run transformers in.');
```
nex3: If we're going to give `build` and `serve` helpful error messages, we might as well do so for `--mode` as well (also below).

lib/src/dart.dart
```diff
@@ -8,116 +8,13 @@ import 'dart:io';
 import 'dart:isolate';
```
nex3: It looks like `isPart()` in this file is no longer used.

lib/src/executable.dart
```diff
@@ -81,8 +71,8 @@ Future<int> runExecutable(Entrypoint entrypoint, String package,
   // "bin".
   if (p.split(executable).length == 1) executable = p.join("bin", executable);
 
-  var executableUrl = await _executableUrl(entrypoint, package, executable,
-      isGlobal: isGlobal, mode: mode);
+  var executableUrl =
```
nex3: Since we aren't going to be running executables over HTTP anymore, I'd make this a path rather than a URL.

This should also let you clean up some of the logic below that's conditional based on whether or not we're running an executable from the filesystem, since we always are now.

lib/src/global_packages.dart
```diff
@@ -232,13 +229,7 @@ class GlobalPackages {
...
       // possible. This is faster and produces better error messages.
       var package = entrypoint.packageGraph.packages[packageName];
-      if (entrypoint.packageGraph
-          .transitiveDependencies(packageName)
-          .every((package) => package.pubspec.transformers.isEmpty)) {
-        return _precompileExecutablesWithoutBarback(package, binDir);
-      } else {
-        return _precompileExecutablesWithBarback(entrypoint, package, binDir);
-      }
+      return _precompileExecutablesWithoutBarback(package, binDir);
```
nex3: Rename to just `_precompileExecutables()`.

lib/src/entrypoint.dart
```diff
@@ -396,24 +277,12 @@ class Entrypoint {
...
-        await _precompileExecutablesWithBarback(executables, packagesToLoad);
-      }
+      await _precompileExecutables(executables);
     });
   }
 
   //// Precompiles [executables] to snapshots from the filesystem.
-  Future _precompileExecutablesWithoutBarback(
-      Map<String, List<AssetId>> executables) {
+  Future _precompileExecutables(Map<String, List<AssetId>> executables) {
```
nex3: For example, the package names in the `executables` map are totally redundant; this could just be `Map<String, List<String>>`. The only reason we represented executables as `AssetId`s before was for convenience when passing them to Barback.

natebosch:
> If it's a huge amount of effort you could drop it, but I *think* all you should need to do is something like
>
>```dart
>if (!fileExists(snapshotPath)) await _precompileExecutables(package);
>```

`_precompileExecutables` would end up going through [`snapshot`](https://github.com/dart-lang/pub/blob/7f0650299fe0cec722aae09b7e090d45c287e8e6/lib/src/dart.dart#L211-L213) which requires a `.packages` or `packages/` as well (I don't know what "a package spec is inferred from the executable's location" means, but I suspect it won't happen since we're calling `dart` which doesn't understand `package_name.lock` files AFAIK).

I think in order to support this we'd need to rewrite some of the AssetEnvironment stuff without barback.

lib/src/pubspec.dart
```diff
@@ -245,82 +243,6 @@ class Pubspec {
 
   Map<String, Feature> _features;
 
-  /// The configurations of the transformers to use for this package.
-  List<Set<TransformerConfig>> get transformers {
```
nex3: It would be good to add validators that produce warnings when trying to publish packages that include a `transformers` field or a `web.compiler` field.

lib/src/solver/version_solver.dart
```diff
@@ -80,29 +79,7 @@ class VersionSolver {
...
-
-    pubConstraints.forEach((name, constraint) {
-      if (_root.pubspec.dependencyOverrides.containsKey(name)) return;
-
-      _addIncompatibility(new Incompatibility([
-        new Term(barbackRef.withConstraint(VersionConstraint.any), true),
-        new Term(
-            _systemCache.sources.hosted.refFor(name).withConstraint(constraint),
-            false)
-      ], IncompatibilityCause.pubDependency));
```
nex3: Please also remove `IncompatibilityCause.pubDependency` and its other referents.

test/global/run/uses_old_lockfile_test.dart
```diff
@@ -1,55 +0,0 @@
-// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
nex3: We should still behave nicely with old-style packages—ideally recompile them, if not that then create a `.packages` file, or at the very least produce a useful error message telling users how to fix the problem.

test/package_server.dart
```diff
@@ -38,14 +38,6 @@ Future servePackages(void callback(PackageServerBuilder builder)) async {
 /// This will always replace a previous server.
 Future serveNoPackages() => servePackages((_) {});
 
-/// A shortcut for [servePackages] that serves the version of barback used by
-/// pub.
-Future serveBarback() {
-  return servePackages((builder) {
-    builder.serveRealPackage('barback');
```
nex3: I think you can delete `serveRealPackage()` as well.

nex3:
> `_precompileExecutables` would end up going through [`snapshot`](https://github.com/dart-lang/pub/blob/7f0650299fe0cec722aae09b7e090d45c287e8e6/lib/src/dart.dart#L211-L213) which requires a `.packages` or `packages/` as well (I don't know what "a package spec is inferred from the executable's location" means, but I suspect it won't happen since we're calling `dart` which doesn't understand `package_name.lock` files AFAIK).

I suppose that's true, but you can generate a new `.packages` file really easily given the lockfile.

natebosch:
Ok, I've got a change that passes the test. https://github.com/dart-lang/pub/pull/1871/commits/bcc70cdad61dd1e10a6a3d72387a6e2f811ad2f2

It requires passing the system cache through, we could maybe do this check earlier though to avoid it introducing a new import in `executable.dart`

lib/src/asset/id.dart
natebosch: I tried that at first and hit at least one case where we were storing a collection of AssetIds and lost context of their associated packages.

I can take another look, but I'd prefer to do that in a new PR.

lib/src/cached_package.dart
natebosch: Done

lib/src/command/barback.dart
natebosch: Done

lib/src/dart.dart
natebosch: Removed along with `runInIsolate`.

lib/src/executable.dart
```diff
@@ -109,10 +111,14 @@ Future<int> runExecutable(Entrypoint entrypoint, String package,
...
 Future<Uri> _executableUrl(Entrypoint entrypoint, String package, String path,
-    {bool isGlobal: false}) async {
+    {bool isGlobal: false, SystemCache cache}) async {
   assert(p.isRelative(path));
 
-  if (!fileExists(entrypoint.packagesFile)) return null;
+  if (!fileExists(entrypoint.packagesFile)) {
+    if (!isGlobal) return null;
+    await writeTextFile(
+        entrypoint.packagesFile, entrypoint.lockFile.packagesFile(cache));
```
nex3: Add a comment indicating why a global entrypoint without a `.packages` file would exist.

natebosch: Done

lib/src/executable.dart
natebosch: Done

lib/src/asset/id.dart
natebosch: I overestimated the scope. I replaced `id` with a `packageName` and `pathInPackage` argument in the `snapshot` method, but everything else was, as you pointed out, already in a context where we had the package name.

nex3:
Give me a poke when you want me to take another look.

natebosch:
I think I have addressed all feedback except for:

- Validation that there is no transformers or web compiler config on publish
- Warning when using the `--mode` argument to run

If you're OK with doing those in a follow up this is ready for another review.

jakemac53:
Fwiw - I am going to forgo doing an actual review here since @nex3 is more familiar with the codebase anyways and its quite large.

I will note that when it comes to supporting the old lockfiles for `pub run` the chance of those globally activated packages being able to run with old package versions on the Dart 2 sdk is essentially zero anyways, so breaking compatibility with them is perfectly fine with me.

lib/src/dart.dart
```diff
@@ -49,73 +43,20 @@ class _DirectiveCollector extends GeneralizingAstVisitor {
...
 /// from the executable's location.
 ///
-/// If [id] is passed, it's used to describe the executable in logs and error
-/// messages.
+/// If [packageName] and [pathInPackage] are passed, they are used to describe the
+/// executable in logs and error messages.
 Future snapshot(Uri executableUrl, String snapshotPath,
-    {Uri packagesFile, AssetId id}) async {
-  var name = log.bold(id == null
+    {Uri packagesFile, String packageName, String pathInPackage}) async {
```
nex3: Consider replacing the `packageName` and `pathInPackage` parameters with an `name` parameter and make the caller responsible for deciding what that name could look like. Then the first line could just be `name ??= executableUrl.toString()`.

lib/src/executable.dart
```diff
@@ -85,43 +85,39 @@ Future<int> runExecutable(Entrypoint entrypoint, String package,
...
-    // helpful for the subprocess to be able to spawn Dart with
-    // Platform.executableArguments and have that work regardless of the working
-    // directory.
-    packageConfig = p.toUri(p.absolute(entrypoint.packagesFile));
-  }
+  // We use an absolute path here not because the VM insists but because it's
+  // helpful for the subprocess to be able to spawn Dart with
+  // Platform.executableArguments and have that work regardless of the working
+  // directory.
+  Uri packageConfig = p.toUri(p.absolute(entrypoint.packagesFile));
```
nex3: Nit: don't type-annotate local variables.

lib/src/validator/deprecated_fields.dart
```diff
@@ -0,0 +1,27 @@
...
+import '../entrypoint.dart';
+import '../validator.dart';
+
+/// A validator that validates that a pubspec is not including deprecated fields
+/// which are no longer read.
+class DeprecatedFieldsValidator extends Validator {
+  DeprecatedFieldsValidator(Entrypoint entrypoint) : super(entrypoint);
+
+  Future validate() {
+    return new Future.sync(() {
```
nex3: Nit: I'd just make the method `async`.

lib/src/validator/deprecated_fields.dart
```diff
@@ -0,0 +1,27 @@
...
+      if (entrypoint.root.pubspec.fields.containsKey('transformers')) {
+        warnings.add('Your pubpsec.yaml includes a "transformers" section which'
+            ' is no longer used and may be removed');
```
nex3: Nit: warnings should end with a period.

test/validator/deprecated_fields_test.dart
```diff
@@ -0,0 +1,40 @@
...
+import '../descriptor.dart' as d;
+import '../test_pub.dart';
+import 'utils.dart';
+
+Validator deprecatedFields(Entrypoint entrypoint) =>
+    new DeprecatedFieldsValidator(entrypoint);
+
+main() {
+  setUp(d.validPackage.create);
+
```
nex3: It would be good to add a test that this doesn't warn for a pubspec that doesn't have either deprecated field.

lib/src/dart.dart
natebosch: Done

lib/src/executable.dart
natebosch: Done

lib/src/validator/deprecated_fields.dart
natebosch: Done

lib/src/validator/deprecated_fields.dart
natebosch: Done

test/validator/deprecated_fields_test.dart
natebosch: Done

nex3:
giphy 3

denniskaselow:
I don't want to ruin the party, but running pub get/upgrade from master now hangs on windows after commit 5ca86eb6e7be45ade4e44cbb7402d296cdf8b32f.

It does
`IO  : Spawning "cmd /c dart --packages=.packages --snapshot=.\.dart_tool/pub\bin\test\bin\test.dart.snapshot file:///C:/Users/dennis/AppData/Roaming/Pub/Cache/hosted/pub.dartlang.org/test-0.12.34/bin%5Ctest.dart" in F:\Programming\workspaces\dart\dartemis_builder\.`

instead of

`IO  : Spawning "cmd /c dart --packages=.packages --snapshot=.\.dart_tool/pub\bin\test\bin\test.dart.snapshot file:///C:/Users/dennis/AppData/Roaming/Pub/Cache/hosted/pub.dartlang.org/test-0.12.34/bin/test.dart" in F:\Programming\workspaces\dart\dartemis_builder\.`

Stumbled over it, when I tested  the fix for #1756.

natebosch:
This is not synced in the SDK so it is not published. What configuration are you using such that you are running this version of pub?

denniskaselow:
I cloned the repository to test that other bugfix.

kevmoo:
@denniskaselow – could we open a separate issue for this? in the pub repo...

Please be clear on the commit SHA you're hitting the issue at...

PR #1879, opened 2018-04-24
Manipulate executable paths only as paths

### LINKED ISSUE #1878: Precompiling executables uses wrong paths on windows
See https://github.com/dart-lang/pub/pull/1871#issuecomment-383813274

natebosch:
Fixes #1878

### DIFF
--- a/lib/src/command/deps.dart
+++ b/lib/src/command/deps.dart
@@ -275,7 +275,7 @@ class DepsCommand extends PubCommand {
   /// Lists all Dart files in the `bin` directory of the [package].
   ///
   /// Returns file names without extensions.
-  List<String> _getExecutablesFor(Package package) => package.executableIds
+  List<String> _getExecutablesFor(Package package) => package.executablePaths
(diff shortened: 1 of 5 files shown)


kevmoo:
...might consider adding appveyor to the mix for pub – help w/ this?

natebosch:
There seem to be a lot of failing tests on windows outside of this.

I believe this would have been caught, but I don't think we'd get much value out of appveyor without a bunch of work to cleanup the existing tests.

kevmoo:
1) We could always flag the failing tests and exclude them on app veyor
2) It'd be great to understand which tests fail
3) It'd be great to at least make sure the non-failing tests don't start
randomly failing :-)

nex3:
Pub's tests passed on Windows for a long time. The fact that they aren't passing now is as likely to indicate Windows bugs as issues with the tests. It seems like time spent triaging them and ensuring that they all pass would be well-spent.

PR #1881, opened 2018-04-25
Remove a number of unused members

kevmoo:
Made members only used within a library private
Moved a number of items only used it tests closer to their usage

### DIFF
--- a/lib/src/log.dart
+++ b/lib/src/log.dart
@@ -42,14 +42,12 @@ const _MAX_TRANSCRIPT = 10000;
 
 /// The list of recorded log messages. Will only be recorded if
 /// [recordTranscript()] is called.
-Transcript<Entry> _transcript;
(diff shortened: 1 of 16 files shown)


test/io_test.dart
```diff
@@ -344,25 +344,14 @@ void testExistencePredicate(String name, bool predicate(String path),
       });
     }
   });
-
-  group('topLevelDir', () {
```
kevmoo: test for util that was only used in this test...

lib/src/log.dart
```diff
@@ -620,57 +564,3 @@ class _JsonLogger {
     print(jsonEncode(message));
   }
 }
-
-/// Collapses a series of collapsible messages into a single line of output if
-/// they happen within a short window of time.
-class _Collapser {
```
kevmoo: This class was never instantiated...

PR #1888, opened 2018-05-01
Use Platform.version for the SDK version

kevmoo:
Fixes https://github.com/dart-lang/pub/issues/


kevmoo:
CC @jakemac53 @natebosch @grouma – we discussed this in the BAT meeting

kevmoo:
@nex3 – less reading of special URLs and more using SDK APIs seems a good move, right?

PR #1889, opened 2018-05-01
Retry on pub server status 500

cbracken:
Previously, pub retried up to 5 times when it received a response with
HTTP status 502, 503, 504. This adds 500 to the list of retryable
response statuses.

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -227,7 +227,8 @@ final httpClient = new ThrottleClient(
     16,
     new _ThrowingClient(new RetryClient(_pubClient,
         retries: 5,
-        when: (response) => const [502, 503, 504].contains(response.statusCode),
+        when: (response) =>
+            const [500, 502, 503, 504].contains(response.statusCode),
         whenError: (error, stackTrace) {
           if (error is! IOException) return false;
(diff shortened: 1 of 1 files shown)


kevmoo:
CC @nex3 – 100% agree we should do this

tvolkert:
Why not just check if `response.statusCode % 100 == 5`?

cbracken:
Should probably be: `response.statusCode ~/ 100 == 5`

I'm fine with that but I noticed the surrounding code all seems to call out individual codes.

tvolkert:
> response.statusCode ~/ 100 == 5

Derp - yeah 😄

natebosch:
I'm not sure if it necessarily makes sense to retry on 511, but it might not hurt and could be safer to always retry on 5xx than to hardcode specific codes.

kevmoo:
@cbracken – let's get this landed and get a sync to the SDK repo going

cbracken:
Thanks!

kevmoo:
@cbracken @tvolkert – this is out in Dart 2.0.0-dev.54.0 – as you're pondering when you sync the SDK

PR #1890, opened 2018-05-02
Support the latest pub_semver

nex3:
This rewrites failures' incompatibility graphs so that they refer to
real version numbers rather than the pessimistic ranges that
pub_semver can now generate.

Note that this is expected to have one failing test until
dart-lang/pub_semver#30 lands and pub_semver 1.4.1 is released.

### DIFF
--- a/lib/src/solver/reformat_ranges.dart
+++ b/lib/src/solver/reformat_ranges.dart
@@ -1,0 +1,109 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:pub_semver/pub_semver.dart';
+
+import '../package_name.dart';
+import '../utils.dart';
+import 'incompatibility.dart';
+import 'incompatibility_cause.dart';
+import 'package_lister.dart';
+import 'term.dart';
+
+/// Replaces version ranges in [incompatibility] and its causes with more
+/// human-readable (but less technically-accurate) ranges.
+///
+/// We use a lot of ranges in the solver that explicitly allow pre-release
+/// versions, such as `>=1.0.0-0 <2.0.0` or `>=1.0.0 <2.0.0-∞`. These ensure
+/// that adjacent ranges can be merged together, which makes the solver's job
+/// much easier. However, they're not super human-friendly, and in practice most
+/// package versions don't actually have pre-releases available.
+///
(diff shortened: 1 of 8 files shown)


lib/src/solver/reformat_ranges.dart
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  var terms = incompatibility.terms.map((term) {
+    var versions = packageListers[term.package.toRef()]?.cachedVersions ?? [];
+
+    if (term.package.constraint is! VersionRange) return term;
+    var range = term.package.constraint as VersionRange;
+
+    var min = range.min;
+    if (min != null && range.includeMin && min.isFirstPreRelease) {
```
natebosch: We could avoid some of the reassignments if we pull out some methods.

`var min = _normalizedRangeMin(range, versions);`

lib/src/solver/reformat_ranges.dart
```diff
@@ -0,0 +1,109 @@
...
+    var minIsPreReleaseOfMax =
+        min != null && min.isPreRelease && equalsIgnoringPreRelease(min, max);
+    if (max != null &&
+        !includeMax &&
+        !max.isPreRelease &&
+        !minIsPreReleaseOfMax) {
+      var index = _lowerBound(versions, max);
+      var previous = index == 0 ? null : versions[index - 1].version;
+
+      print("original range: $range, previous version: $previous");
```
natebosch: is this run inside some zone that will capture print calls?

lib/src/solver/reformat_ranges.dart
```diff
@@ -0,0 +1,109 @@
...
+            min: min,
+            max: max,
+            includeMin: range.includeMin,
+            includeMax: includeMax,
+            alwaysIncludeMaxPreRelease: true)),
+        term.isPositive);
+  });
+
+  return new Incompatibility(terms.toList(), cause);
+}
```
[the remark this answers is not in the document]
munificent: +1

lib/src/solver/reformat_ranges.dart
```diff
@@ -0,0 +1,109 @@
...
+
+/// Returns the first index in [ids] (which is sorted by version) whose version
+/// is greater than or equal to [version].
+///
+/// Returns `ids.length` if all the versions in `ids` are less than [version].
+///
+/// We can't use the `collection` package's `lowerBound()` function here because
+/// [version] isn't the same as [ids]' element type.
+PackageVersion _lowerBound(List<PackageId> ids, Version version) {
```
munificent: The return type should be `int`, shouldn't it?

PR #1892, opened 2018-05-02
Add back build/serve argparser options

### LINKED ISSUE #1891: `pub serve --port` doesn't give the desired error
If you pass any arguments to pub serve, you get the following instead of the error pointing you to the new build system:

```
Could not find an option named "port".

Usage: pub serve [arguments]
-h, --help    Print this usage information.

```

natebosch:
Closes #1891

It's not possible to customize the `argParser` in a `Command` such that
it ignores these invalid options, instead customize the UsageException
to have the same message as running the command.

The difference now is that with arguments it will exit with code 64 and
without arguments it will exit with code 1.

### DIFF
--- a/lib/src/command/barback.dart
+++ b/lib/src/command/barback.dart
@@ -2,21 +2,17 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'package:args/command_runner.dart';
-
 import '../command.dart';
 import '../log.dart' as log;
 import '../utils.dart';
 
 /// Shared base class for [BuildCommand] and [ServeCommand].
 abstract class BarbackCommand extends PubCommand {
-  @override
-  usageException(_) => throw new UsageException(_deprecationError, '');
-
-  String get _deprecationError =>
-      log.red("Dart 2 has a new build system. Learn how to migrate "
-          "from ${log.bold('pub build')} and\n"
-          "${log.bold('pub serve')}: https://webdev.dartlang.org/dart-2\n");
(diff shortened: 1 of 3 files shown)


lib/src/command/barback.dart
```diff
@@ -2,21 +2,31 @@
...
 
-import '../command.dart';
+import 'package:args/command_runner.dart';
+
 import '../log.dart' as log;
 import '../utils.dart';
 
 /// Shared base class for [BuildCommand] and [ServeCommand].
-abstract class BarbackCommand extends PubCommand {
+abstract class BarbackCommand extends Command {
```
nex3: I'm worried that making this not extend `PubCommand` could cause problems later on if we need to rely on all commands implementing that interface. Why can't you just add hidden options that match the old options?

natebosch: Mostly because there were a lot of options so this was _just_ allowing anything and the alternative is a lot more code.

It also means that if you misspell and option you'll get a misleading usage exception making you think if you correct it it will work only to find a new failure when you fix the options.

I can add them back though.

natebosch: In case it helps I _could_ make this extend `PubCommand` again...

nex3: Can you override `PubCommand.argParser` to return an `ArgParser.allowAnything()`?

natebosch: no, `ArgParser.allowAnything()` doesn't let you set options on it and since it's a `Command` it automatically gets a `help`.

`Unsupported operation: ArgParser.allowAnything().addFlag() isn't supported.`

PR #1893, opened 2018-05-03
Don't try to run snapshots in Dart 2 mode

nex3:
Closes #32188


kevmoo:
@nex3 – thoughts on the travis breaks?

natebosch:
Looks like the Dart VM parser (_without_ --preview-dart-2) has a bug?

Can hack around it with:

```dart
final isDart2 = _checkType<String>() == String;
_checkType<T>() => T;
```

Issue #1793, opened 2018-01-30
"Invalid upload: no pubspec.yaml file" on macosx sierra


dam0vm3nt:
Seams that `pub` is using the wrong options when calling macosx `bsdtar` command, so `pub publish` with fail.

I've worked around it be `brew` installing gnu tar and adding a wrapper script that will replace `@/dev/stdin`  with `-T/dev/stdin` .

nex3:
Unfortunately, pub only works with the operating system's built-in tar command currently. See #1602 for details.

dam0vm3nt:
mmh, there's something wrong then, I've installed `gnu-tar` because it was not working with the native command.

natebosch:
@dam0vm3nt - can you reopen the issue with a repro using `bsdtar`?

What options was `pub` using that are unsupported? What version of `bsdtar` do you have?

I don't see any changes in our usage of `tar` in a while, and my mac should be up to date and has `bsdtar 2.8.3 - libarchive 2.8.3` which has been the version for a while AFAIK and should work with the options `pub` is passing.

dam0vm3nt:
I'll investigate it a bit more. Meanwhile I can confirm that `bsdtar` is `bsdtar 2.8.3 - libarchive 2.8.3`. 

When launching the same command that `pub` executes (captured from the logs launching with `--verbose`) in the terminal :

    git ls-files --cached --others --exclude-standard | tar --format=ustar --create --gzip --directory $PWD @/dev/stdin > /tmp/file.tar.gz

the resulting file was always 10240 bytes long (while the correct size would be greater) and it was actually empty.

Also from the man page : 

![image](https://user-images.githubusercontent.com/2409466/35592826-80d2322c-060e-11e8-887c-853ee3da7146.png)

Seams like the `@` char is wrong here.

dam0vm3nt:
@natebosch I've got no power to reopen the issue, the only thing I can do is open another one citing this one. Do you want me to do it ?

natebosch:
AFAIK the `@` is legal and being used correctly here.

Can you paste the results of the first part in that pipeline? `git ls-files --cached --others --exclude-standard`

dam0vm3nt:
Yep!

Running `git ls-files --cached --others --exclude-standard` will give : 

![image](https://user-images.githubusercontent.com/2409466/35673910-32e7f5fc-0743-11e8-85b0-290696a9dcdc.png)

...

[...]

PR #1895, opened 2018-05-09
fix space issue on macOS pub publish

dnfield:
Fixes https://github.com/dart-lang/pub/issues/1793

One thing I'm not 100% on is the best way to test this.  I'm hesitant to upload a package to pub to test this that can then never be deleted.  However, I've tested it using the raw commands and it all looks good in terms of creating a valid tar.gz

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -1020,7 +1020,12 @@ ByteStream createTarGz(List contents, {String baseDir}) {
 
         // We need a newline at the end, otherwise the last file would get
         // ignored.
-        stdin = mtreeHeader + contents.join("\n") + "\n";
+        stdin = mtreeHeader +
+            contents
+                .map((String content) =>
+                    content.replaceAll(new RegExp(r' '), r'\040'))
+                .join("\n") +
+            "\n";
       }
 
       // Setting the working directory should be unnecessary since we pass an


dgrove:
@nex3 is OOO for the next 1.5 weeks, and will take a look when she gets back (always possible that someone else will before then, but not guaranteed)

dnfield:
Thanks for the update.  FWIW, I didn't want that file included anyway and just removed it from my package - I'm not being blocked by this issue.

dnfield:
@nex3  for some reason I was under the impression that the list might contain `String` or `FileSystemEntity` objects.  But checking all the call sites it's only ever a String.

I'm not really sure why the unit test is failing, it's not failing for me locally ☹️

nex3:
> @nex3 for some reason I was under the impression that the list might contain String or FileSystemEntity objects. But checking all the call sites it's only ever a String.

That's what the documentation said so it's reasonable for you to think that! It's probably just a relic of how that code was used when it was originally written.

> I'm not really sure why the unit test is failing, it's not failing for me locally ☹️ 

That's just https://github.com/dart-lang/pub/issues/1898. I'll merge this in once we get a clean SDK build.

kevmoo:
Travis is green now on this...

Issue #1887, opened 2018-05-01
SDK version used by pub does not align with Platform.version on Flutter


kevmoo:
We'll get pub reporting/using `Dart pub 2.0.0-edge.0d5cf900b021bf5c9fa593ffa12b15bcd1cc5fe0`

Where Platform.version is something like `2.0.0-dev.48.0.flutter-fe606f890b`

nex3:
I think this is a Flutter/SDK issue, not a Pub issue. We just look at the `version` file in the root of the Dart SDK directory.

kevmoo:
Turns out this is not fixed. Hoping #1897 does the trick

natebosch:
Is this what is causing https://github.com/flutter/flutter/issues/17656 ?

PR #1897, opened 2018-05-18
Use Platform.version even when running from the repo

kevmoo:
Follow-up to https://github.com/dart-lang/pub/commit/3d7ae0cdf0
Hopefully really fixes https://github.com/dart-lang/pub/issues/1887


natebosch:
Do we not run `pub get` from within the Dart SDK repo anymore?

kevmoo:
I verified this works – I think this will fix the issue where we're not picking up the flutter-flavor of the Dart sdk.

natebosch:
You've tried pulling this in to the Dart SDK and verified the build works?

kevmoo:
Yup. Well, I modified the version of pub in the SDK – and confirmed that
the version is still correct...

kevmoo:
The travis failures were there before – should be investigated

Issue #1899, opened 2018-05-18
Support all operations with `--preview-dart-2`


natebosch:
We've so far validated the `pub run` code path. We need to verify Dart 2 runtime semantics for all other pub commands.

So far @kevmoo has hit a problem with `pub upgrade` - can you add details?

kevmoo:
I think this is due to `pkg:async` being a dot release behind in the SDK. Trying now...

kevmoo:
I think it comes down to this: https://github.com/dart-lang/http_throttle/pull/7

kevmoo:
This helps a bit with `pub upgrade` – but there are still issues https://dart-review.googlesource.com/c/sdk/+/55988

dgrove:
What needs to happen to do that?

kevmoo:
Run the tests with DART_VM_OPTIONS=--preview-dart-2

See failures. Fix failures.

I published pkg:test last night that works with the flag. Some tests do
pass...

nex3:
> What needs to happen to do that?

There's some work that needs to be done making the tests run pub processes in Dart 2 mode, and of course https://github.com/dart-lang/pub/issues/1898 needs to be fixed, but the biggest blocker is performance. Pub has 272 test files, and those tests contain 537 invocations of pub as a subprocess. If each test file takes five seconds to parse and each pub invocation takes ten, that's almost two hours of pure overhead for each full test iteration, on top of the time it takes to actually *run* the tests. We need to bring that down substantially before it's feasible to iterate on these tests or run them on CI to ensure that we don't regress Dart 2 semantics.

dgrove:
At tip of bleeding edge app-jit snapshots are mostly working (there seems
to be a race condition that I suspect shows up if you're not building from
a clean SDK build - see https://github.com/dart-lang/sdk/issues/33167 ).

There should enough functioning now to make progress here.

nex3:
Creating snapshots isn't an issue--since the slowdown is all in the parsing phase, script snapshots give us all the speedup we'd need anyway. The issue is that we'd need to recompile snapshots for each test file as we iterate on the code, which takes even longer than just running the tests from source.

natebosch:
We've been running with Dart 2 for a bit now and nothing new has come up.

Issue #1898, opened 2018-05-18
pub tests failing on -dev.55+


kevmoo:
See https://travis-ci.org/dart-lang/pub/builds/380860559

-dev.54: fine
-dev.55: broken

CC @natebosch

natebosch:
Broken by https://github.com/dart-lang/sdk/issues/33166

nex3:
We need that in a release to get the tests green.

kevmoo:
Fixed in Dart -dev.57 - https://travis-ci.org/dart-lang/pub/builds/380859567

PR #1902, opened 2018-05-22
Fix Dart 2 runtime errors

nex3:
I believe this should address #1899 well enough for now to make pub usable in Dart 2 mode. I don't think we should close the issue until we have CI set up that runs everything in Dart 2 mode, though.


nex3:
Holding off on landing this until #1898 is fixed and we can get the tests green.

kevmoo:
Travis is green now on this...

kevmoo:
Let's get this landed ASAP and pulled into the SDK. Blocking many uses of --preview-dart-2

Issue #1904, opened 2018-05-22
Allow running Dart scripts upon package installation


thosakwe:
Somewhat related to the long-closed #1674, but I believe this is a better solution.

What if, upon running `pub get`, a package could execute one or more scripts written in Dart? This would be useful for doing things like building native extensions, or other configuration necessary for the execution of a given package.

There could be a field called `after_install` with either a path to a single Dart script, or a list of paths. These paths would be resolved relative to the current working directory.

### Example
```yaml
name: foo
after_install:
  - tool/download_bar.dart
  - tool/do_something_else_.dart
```

The `pub` tool could spin up a new isolate instead of a new VM.

### Nested usage (in dependencies)
`pub` should also be able to run `after_install` scripts from packages that are dependencies. For example, if you depend on `package:foo` that bundles a `tool/bar.dart`, you'll get the following:

```bash
$ pub get
Resolving dependencies...

Got dependencies!

package:foo is requesting to run the script `tool/bar.dart`.
Allow? (y/n/View contents)

Contents of `tool/bar.dart`:

import 'dart:io';

void main() {
  print(new File('foo.bar').readAsStringSync());
}

package:foo is requesting to run the script `tool/bar.dart`.
Allow? (y/N)
```

### Caveats
* Executing arbitrary code is always something of a concern. To combat this, the "View contents" option would be the default, which would allow users to at least know what they're running before it runs.
* Each script would assume it were being run in the root directory of its corresponding package. For example, `package:foo`'s scripts would run in `~/.pub-cache/hosted/pub.dartlang.org/foo/x.y.z`.
* There would need to be some sort of of cache file (`~/.pub-cache/after_install_cache.yaml`) present containing timestamps of when each script was run, so that the user would not be repeatedly nagged to run scripts from every package they depend on. This would be in the pub cache, so that even if you use the same package in multiple projects, you'd only have to run things like native extension builders once globally.

## Conclusion
Overall, I think that the ability to run a callback script can go a long way to improving the Dart ecosystem. Being able to build things like native extensions on-the-fly would be great for Dart IMO.

P.S. I'm wholly willing to send a PR, and would actually love to do so ASAP. Thanks in advance for consideration.

zoechi:
Similar to what is mentioned in https://github.com/dart-lang/pub/issues/39#issuecomment-170665664

thosakwe:
[...]

PR #1908, opened 2018-05-23
After install scripts (#1904)

thosakwe:
This is a working implementation of the changes proposed in #1904.

I've added tests for parsing the necessary value (an `after_install` field in the pubspec), but I'm not so sure of how to test the functionality, other than physically running it myself.

To clarify, the behavior verifiably works when run by a human, but I'm not so sure how to run it via computer.

One other thing I'd like to propose is an additional check upon running `pub lish` that throws an error if a script specified the pubspec does not actually exist on disk (to prevent breaks on end-user machines).

I'm willing to fix anything that needs to be fixed to land this PR. I really feel like something so simple as support for post-install scripts could go a long way to growing the Dart ecosystem.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -4,13 +4,15 @@
 
 import 'dart:async';
 import 'dart:io';
+import 'dart:isolate';
 
+import 'package:async/async.dart';
 import 'package:collection/collection.dart';
 import 'package:package_config/packages_file.dart' as packages_file;
 import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 
-import 'after_install.dart' as after_install;
+import 'after_install.dart';
 import 'dart.dart' as dart;
 import 'exceptions.dart';
 import 'http.dart' as http;
@@ -648,6 +650,7 @@ class Entrypoint {
 
   /// Execute any outstanding Dart scripts in `after_install`.
   Future runAfterInstallScripts(SolveResult result) async {
+    // Count how many scripts we could potentially run.
     var allScripts = <String, List<String>>{};
 
     for (var package in result.changedPackages) {
@@ -655,19 +658,71 @@ class Entrypoint {
       if (scripts.isNotEmpty) allScripts[package] = scripts;
     }
 
-    if (root.pubspec.afterInstall.isNotEmpty)
-      allScripts[root.name] = root.pubspec.afterInstall;
-
     // If no scripts need to be run, don't bother updating the cache.
     if (allScripts.isEmpty) return;
 
     // Figure out what the last time every script was run.
     // If a script has not changed since the last time it was run,
     // Don't run it.
+    var systemScriptCache = await AfterInstallCache.load(cache.rootDir);
+    var localScriptCache = await AfterInstallCache.load(cachePath);
+    var mergedCache = systemScriptCache.merge(localScriptCache);
 
-    for (var package in allScripts.keys) {
+    // Run scripts in dependencies.
+    for (var package in result.changedPackages) {
       var scripts = allScripts[package];
-      print(scripts);
+      if (scripts.isEmpty) continue;
+
+      var pkg = package == root.name
+          ? root
+          : cache.load(result.packages
+              .firstWhere((a) => a.name == package, orElse: () => null));
+
+      for (var script in scripts) {
+        var absolutePath = p.normalize(p.absolute(p.join(pkg.dir, script)));
+
+        // Don't run the script against there are new changes.
+        if (!await mergedCache.isOutdated(absolutePath)) continue;
+
+        // Never trust any third-party scripts. Show a prompt before running scripts.
+        if (true || pkg != root) {
+          var message = 'package:$package wants to run the script "$absolutePath".';
+
+          // 
+        }
+
+        var scriptFile = new File(absolutePath);
+
+        if (await scriptFile.exists()) {
+          // We need to change into the package's directory to run its scripts.
+          var currentDir = Directory.current;
+          Directory.current = pkg.dir;
+
+          // Spawn an isolate for the script.
+          var onError = new ReceivePort();
+          var onExit = new ReceivePort();
+          var onComplete = new Completer();
+
+          onExit.listen(
+              (x) => onComplete.isCompleted ? null : onComplete.complete());
+          onError.listen((x) =>
+              onComplete.isCompleted ? null : onComplete.completeError(x));
+
+          try {
+            Isolate.spawnUri(p.toUri(absolutePath), [], null,
+                onExit: onExit.sendPort, onError: onError.sendPort);
+            await onComplete.future;
+          } catch (e) {
+            // TODO: Handle errors in scripts!
+            print(e);
+          } finally {
+            // Switch back!
+            Directory.current = currentDir;
+            onError.close();
+            onExit.close();
+          }
+        }
+      }
     }
   }
 }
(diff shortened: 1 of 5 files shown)


nex3:
I won't have time to review this any time soon.

stepancar:
@kevmoo any progress on this PR?

kevmoo:
Still waiting on @nex3 to get back in the Office.

Again, this is a tough one. It's not about the code quality or understanding the use case – it's about security concerns.

thosakwe:
Yeah, I can certainly understand that. If something like a post install hook could be used to do some sneaky business on users' computers, that certainly wouldn't be good.

I included an option for users to preview the source of all scripts that a package is requesting to run, so that could at least be a start. Also, the fact that `after_install` can only execute Dart scripts could potentially close at least a few holes.

nex3:
I don't have the time budgeted to review this, nor am I likely to any time soon.

lib/src/entrypoint.dart
```diff
@@ -642,4 +646,127 @@ class Entrypoint {
...
+
+      for (var script in pkg.pubspec.afterInstall) {
+        var absolutePath = p.normalize(p.absolute(p.join(pkg.dir, script)));
+        var scriptFile = new File(absolutePath);
+
+        // Don't run the script against there are new changes.
+        if (!await mergedCache.isOutdated(absolutePath)) continue;
+
+        if (!await scriptFile.exists()) {
+          if (pkg == root)
```
jakemac53: nit, use block syntax if it can't fit on one line

lib/src/entrypoint.dart
```diff
@@ -642,4 +646,127 @@ class Entrypoint {
...
+            log.warning(log.yellow(
+                '"after_install" script "$absolutePath" does not exist.'));
+        } else {
+          // Never trust any third-party scripts. Show a prompt before running scripts.
+          //
+          // The user can opt to print the contents of the script.
+          YesNoPrintResponse response = YesNoPrintResponse.yes;
```
jakemac53: nit: omit the type on the left (just use `var`)

lib/src/entrypoint.dart
```diff
@@ -642,4 +646,127 @@ class Entrypoint {
...
+
+          if (pkg != root) {
+            var message =
+                '\npackage:$package wants to run the script "$absolutePath".';
+
+            do {
+              response = await confirmOrPrint(message);
```
jakemac53: We would need to figure out how to handle this on CI systems like travis - or within editors like VsCode/Intellij

lib/src/entrypoint.dart
```diff
@@ -642,4 +646,127 @@ class Entrypoint {
...
+                log.message(log.gray(await scriptFile.readAsString()));
+              }
+            } while (response == YesNoPrintResponse.print);
+          }
+
+          if (response != YesNoPrintResponse.yes) continue;
+
+          // We need to change into the package's directory to run its scripts.
+          var currentDir = Directory.current;
+          Directory.current = pkg.dir;
```
jakemac53: We shouldn't need to do this (and its probably not what many scripts would want, they may want to look at the current package and would expect that to be the current directory).

We definitely _dont_ want to run inside the package cache like this would do, as touching anything in there is forbidden.

lib/src/entrypoint.dart
```diff
@@ -642,4 +646,127 @@ class Entrypoint {
...
+              (x) => onComplete.isCompleted ? null : onComplete.complete());
+          onError.listen((x) => onComplete.isCompleted
+              ? null
+              : onComplete.completeError(x[0], x[1]));
+
+          try {
+            Isolate.spawnUri(p.toUri(absolutePath), [], null,
+                onExit: onExit.sendPort,
+                onError: onError.sendPort,
+                automaticPackageResolution: true);
```
jakemac53: We should use the `packageConfig` argument and pass in the config from the current isolate, instead of `automaticPackageResolution` argument, similar to `pub run` behavior. This likely gets around the issue that was causing you to change directories to the other package?

thosakwe: That's probably the solution I was looking for, haha. Is there a way to specify a working directory for an isolate, though? In my head, the scripts would execute in the folder for the specific version that package, so that things like native extensions could be built to the correct paths.

Otherwise, though, people publishing scripts could still just `Isolate.resolvePackageUri`, probably

lib/src/entrypoint.dart
thosakwe: Sounds good, more than likely could be removed

lib/src/entrypoint.dart
thosakwe: That's a valid concern, and something I didn't consider at all. At least for Travis, in theory a command-line arg or env variable could be used to auto-deny (or maybe auto-allow) every script?

jakemac53: Yes - I think the default behavior would need to be to auto-deny, with a flag or environment variable to auto-allow.

lib/src/entrypoint.dart
jakemac53: > In my head, the scripts would execute in the folder for the specific version that package, so that things like native extensions could be built to the correct paths.

We really don't want to give access to the actual package cache. That should only contain the exact sources as they exist on pub because it is shared between all projects on your machine. The output of the `after_install` script could be different depending on the version solve of any given project. Therefore each individual project should have to re-run any `after_install` scripts and should generally write the results to the `.dart_tool/<package-name>` directory (by convention).

thosakwe: Okay, that's fair, and makes a lot of sense.

One more question: what can end-users do to depend on sources built to `.dart_tool`? By passing a custom package config? I'm not 100% sure how `build` does it.

jakemac53: That is actually a really good question haha - basically the build system knows how to surface those files via the `serve` command and within the build and treat it as an overlay file system. That isn't a general solution for all of .dart_tool though, its very specific to the `.dart_tool/build/generated` directory and the exact files that it knows it wrote there.

We would have to come up with some other solution here for pub, possibly writing custom paths to the `.packages` file that point to `.dart_tool` and copying the original package + generated sources there.

lib/src/entrypoint.dart
natebosch: should we consider being able to whitelist scripts from certain packages?

jakemac53: > should we consider being able to whitelist scripts from certain packages?

I like that idea

thosakwe:
I've put more thought into this, and in the long run, I'm actually not so sure that `after_install` scripts will really be worth the potential security risks they introduce to Pub.

The main use case I had in mind for this functionality was distribution of native extensions; however, I've been able to get an early version of said functionality use existing infrastructure in `build_native`:
https://github.com/thosakwe/build_native

So, the merits of `after_install` scripts might actually be... Nothing?

mit-mit:
Closing as it's my understanting we have decided not to move ahead with this.

thosakwe:
If something like this is ever added to Pub, it occurred to me that instead of packages you depend having permission to run scripts, it would make more sense if *your package* could run hooks on `get` and `upgrade`, etc.

For example, you could create a `.dart_tool/pub/hooks/post-get`:

```bash
#!/usr/bin/env bash

# Build dependencies
pub run build_runner build --delete-conflicting-outputs

# Run some tool that compiles FFI libraries
pub run some_tool compile
```

That being said, I'm not sure if this would be useful to the majority of people using Pub, and it's already easy enough to just create a script that invokes `pub get`, rather than the other way around.

PR #1912, opened 2018-05-30
Use public analyzer APIs.

scheglov:


matanlurey:
LGTM at face value, but @jakemac53 or someone on the BaT team is a better reviewer.

lib/src/validator/strict_dependencies.dart
```diff
@@ -18,6 +18,9 @@ import 'package:stack_trace/stack_trace.dart';
 
 /// Validates that Dart source files only import declared dependencies.
 class StrictDependenciesValidator extends Validator {
+  final AnalysisSessionManager analysisSessionManager =
```
natebosch: should this be a global? what are the costs to having a separate instance for different use cases?

lib/src/dart.dart
```diff
@@ -90,3 +80,78 @@ Future snapshot(Uri executableUrl, String snapshotPath,
...
+    for (var context in _contexts) {
+      if (context.contextRoot.isAnalyzed(path)) {
+        return context.currentSession;
+      }
+    }
+
+    throw new StateError('Unable to find the context to $path');
+  }
+
+  /**
```
natebosch: Use `///`

lib/src/dart.dart
```diff
@@ -90,3 +80,78 @@ Future snapshot(Uri executableUrl, String snapshotPath,
...
+  AnalysisSession _getSession(String path) {
+    _throwIfNotAbsolutePath(path);
+
+    for (var context in _contexts) {
+      if (context.contextRoot.isAnalyzed(path)) {
+        return context.currentSession;
+      }
+    }
+
+    var dirPath = p.dirname(path);
```
natebosch: would it be sensible to search upwards for an apparent package root? Will the current setup involve creating multiple contexts if I first analyze `package/lib/some/deep/path/file.dart` and `package/lib/some/deep/different/file2.dart`? What will the performance costs be?

lib/src/validator/strict_dependencies.dart
scheglov: The cost is not very high.
Pub uses only parsed units, so there is not much to share.

lib/src/dart.dart
scheglov: Done.

lib/src/dart.dart
scheglov: That's good question, thank you.

We don't discover package roots by looking "up", only "down".
But Pub knows `Package`, and we can pre-create analysis contexts for them, so any request for a file will be handled by an existing analysis context.

PR #1928, opened 2018-06-20
Fix pub now that the VM no longer supports "--preview-dart-2".

munificent:
Pub fails to create snapshots on install because it's passing a flag
the VM no longer recognizes. This fixes that.

Also fixed a couple of runtime cast errors in a test.


.travis.yml
```diff
@@ -17,7 +17,7 @@ dart_task:
 # snapshot if it's available.
 before_script:
   - dart --no-checked --snapshot=bin/pub.dart.snapshot --snapshot-kind=app-jit bin/pub.dart --help
```
natebosch: this might need a `--no-preview-dart-2` for consistency.

munificent: Done.

PR #1933, opened 2018-06-21
Move pub fully onto Dart 2 and off Dart 1.

munificent:
- Have the tests run pub in Dart 2 mode. This unfortunately makes them
  much slower when running from source, but creating a pub snapshot
  helps.
- Only generate Dart 2 snapshots when precompiling.
- Remove tests around Dart 1 snapshots.
- Remove test utility code for running pub in Dart 1 mode.
- Change pub [global] run [un]checked mode tests to use assert().

The tests around outdated snapshots are failing because the old code
for detecting a bad snapshot version error does not seem to work in the
Dart 2 VM. I have an email thread going for that, but I wanted to send
this out in the meantime.


lib/src/executable.dart
```diff
@@ -63,12 +63,8 @@ Future<int> runExecutable(Entrypoint entrypoint, String package,
     log.verbosity = log.Verbosity.WARNING;
   }
 
-  // Snapshots are compiled in Dart 1 mode, so in Dart 2 mode we always run
-  // executables from source.
+  // Uncached packages are run from source.
   if (snapshotPath != null) {
-    // Look for the Dart 2-specific snapshot when running in Dart 2 mode.
-    if (isDart2) snapshotPath += '.dart2';
```
natebosch: I think we need to avoid finding the snapshot if we're not in Dart 2 mode otherwise we're going to attempt to run a Dart 2 snapshot from a Dart 1 vm in some cases.

If we're OK with a hard cutover we could skip this, but it might be nice to have a fallback of running from source.

lib/src/io.dart
```diff
@@ -484,7 +484,7 @@ final bool runningFromTest = Platform.environment.containsKey('_PUB_TESTING');
 /// Whether pub is running from within the Dart SDK, as opposed to from the Dart
 /// source repository.
 final bool _runningFromSdk =
-    !runningFromTest && Platform.script.path.endsWith('.snapshot');
+    !runningFromTest && Platform.script.path.endsWith('.snapshot.dart2');
```
natebosch: I think this might be an unrelated "snapshot" postfix. This one is coming from how we snapshot pub in the SDK build right?

lib/src/executable.dart
munificent: My intent was to do a hard cutover and no longer support pub being run in Dart 1 mode. (I think we'll be making similar hard cut-offs to other packages. As soon as any code starts using `whereType()` or optional `new`, you are no longer in Dart 1 land.)

lib/src/io.dart
munificent: Oops, yes, you're right. Fixed.

lib/src/executable.dart
natebosch: synced offline, we'll recommend avoiding the latest SDK if Dart 1 support is necessary with `pub run`.

PR #1945, opened 2018-07-09
TEST ONLY: set up tests to run with build_runner

jakemac53:


jakemac53:
ok, so this works now but until I can resolve https://github.com/dart-lang/sdk/issues/33952 this doesn't actually make sense

PR #1953, opened 2018-07-27
Add deps for testing with build_runner

natebosch:
Testing from kernel files saves time over testing from source. This does
not switch Travis to test this way yet, but adds the deps to make it
easier to use locally.

Add a non-src lib file so get better behavior out of the coarse module
strategy. Without this file every library ends up becoming it's own
module.

### DIFF
--- a/lib/pub.dart
+++ b/lib/pub.dart
@@ -1,0 +1,1 @@
+export 'package:pub/src/command_runner.dart';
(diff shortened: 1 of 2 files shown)


natebosch:
> Open an issue to track the Travis work.

https://github.com/dart-lang/pub/issues/1954

PR #1955, opened 2018-07-27
Don't delete existing packages directories

### LINKED ISSUE #1951: "pub get" is deleting my source code.
When using `flutter packages get`, which just calls `pub get` under the covers, I see some dangerous behavior: If there happens to be a directory called `packages` in the same directory as the pubspec.yaml, the entire directory gets silently removed, even if the contents have nothing to do with pub's legacy `packages` directory.

Pub should no longer remove this directory, or should at least verify that it only contains what the legacy `packages` directory contained before doing so (e.g. check that there are only symlinks being removed).

See https://github.com/flutter/flutter/issues/19859 for a more detailed description, but basically, pub is nuking my source code.

natebosch:
Fixes #1951

Continues to allow _creating_ packages directories for now, but if they
exist and aren't being created they won't be touched.

### DIFF
--- a/lib/pub.dart
+++ b/lib/pub.dart
@@ -1,0 +1,1 @@
+export 'package:pub/src/command_runner.dart';
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -238,8 +238,6 @@ class Entrypoint {
     // Install the packages and maybe link them into the entrypoint.
     if (packagesDir) {
       cleanDir(packagesPath);
-    } else {
-      deleteEntry(packagesPath);
     }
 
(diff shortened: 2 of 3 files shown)


lib/pub.dart
```diff
@@ -0,0 +1 @@
+export 'package:pub/src/command_runner.dart';
```
jakemac53: nit: you could just import this instead, so that there is still nothing exposed by the package

natebosch: hmm, this was covered in https://github.com/dart-lang/pub/pull/1953, not sure why it's in this diff.

will change that to an import in a new PR

Issue #1952, opened 2018-07-27
Remove support for packages directory altogether


natebosch:
Drop the `--packages-dir` flag, clean up the dead code and tests.

Nothing else in the ecosystem supports this, pub shouldn't either.

PR #1958, opened 2018-08-03
Drop support for --packages-dir flag

natebosch:
Towards #1952

Drops support for the flag and updates tests that were depending on it
or referring to it, but does not remove the supporting code. That will
be done in a followup.

- Parse but ignore the flag and warn when it is used.
- Drop tests specifically for the packages dir behavior.
- Remove unnecessary references to the flag.
- Drop utility test code and constants for the packages dir.
- Wherever possible replace checks for content within the packages dir
  with checks for content in the package spec, and sometimes cache.
- When working with git shas which are unpredictable and have an ugly
  format in the package spec check instead for lines that contain the
  sha, or which have a package spec which either specificially did or
  did not chagne after an upgrade.


kevmoo:
@natebosch – need to `dartfmt`

test/pub_get_and_upgrade_test.dart
```diff
@@ -34,57 +34,21 @@ main() {
...
-            [d.file('myapp_name.dart', 'main() => "myapp_name";')])
-      ]).validate();
+      await pubCommand(command);
 
       await d.dir("myapp", [
         d.packagesFile({"myapp_name": "."})
       ]).validate();
     });
 
-    test(
```
jakemac53: Should these be checking the package spec? Or does it just not really matter?

test/upgrade/git/upgrade_to_nonexistent_pubspec_test.dart
```diff
@@ -1,44 +0,0 @@
...
-// for details. All rights reserved. Use of this source code is governed by a
-// BSD-style license that can be found in the LICENSE file.
-
-import 'package:test/test.dart';
-
-import '../../descriptor.dart' as d;
-import '../../test_pub.dart';
-
-main() {
-  test("upgrades Git packages to a nonexistent pubspec", () async {
```
jakemac53: this test doesn't seem --packages-dir specific?

test/pub_get_and_upgrade_test.dart
natebosch: The current behavior is to include the package in the package spec regardless of whether is has a lib directory. I'm assuming that behavior is OK...

test/upgrade/git/upgrade_to_nonexistent_pubspec_test.dart
natebosch: hmm, yeah I am not sure why I thought it was... Restored it

lib/src/command/downgrade.dart
```diff
@@ -27,17 +27,17 @@ class DowngradeCommand extends PubCommand {
...
-    argParser.addFlag('packages-dir',
-        negatable: true,
-        help: "Generate a packages/ directory when installing packages.");
+    argParser.addFlag('packages-dir', negatable: true, hide: true);
   }
 
   Future run() async {
+    if (argResults.wasParsed('packages-dir')) {
+      log.warning(
+          log.yellow('The --packages-dir flag is an unsupported no-op'));
```
munificent: "no-op" is jargon. How about:

> The --packages-dir flag is no longer used and does nothing.

Here and below.

test/downgrade/doesnt_change_git_dependencies_test.dart
```diff
@@ -18,23 +18,15 @@ main() {
...
 
-    // TODO(rnystrom): Remove "--packages-dir" and validate using the
-    // ".packages" file instead of looking in the "packages" directory.
-    await pubGet(args: ["--packages-dir"]);
+    await pubGet();
 
-    await d.dir(packagesPath, [
-      d.dir('foo', [d.file('foo.dart', 'main() => "foo";')])
-    ]).validate();
+    final originalFooSpec = packageSpecLine('foo');
```
munificent: Nit: We don't use "final" for local variables in pub even when we could. I don't think it adds much benefit in general, and I think it's good to be consistent.

lib/src/command/downgrade.dart
natebosch: Good call. Done

test/downgrade/doesnt_change_git_dependencies_test.dart
natebosch: Done.

PR #1961, opened 2018-08-06
Add some stricter analysis

natebosch:
Enabled a bunch of lints which are (hopefully) not controversial and fix
existing violations. Most of the fixes are pretty mechanical.

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -787,7 +787,7 @@ _doProcess(Function fn, String executable, List<String> args,
   // Spawning a process on Windows will not look for the executable in the
   // system path. So, if executable looks like it needs that (i.e. it doesn't
   // have any path separators in it), then spawn it through a shell.
-  if (Platform.isWindows && executable.contains('\\')) {
+  if (Platform.isWindows && !executable.contains('\\')) {
     args = ["/c", executable]..addAll(args);
     executable = "cmd";
   }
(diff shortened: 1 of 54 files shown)


lib/src/io.dart
```diff
@@ -783,12 +782,12 @@ class _PubProcess {
...
     {String workingDir,
     Map<String, String> environment,
-    bool runInShell: false}) {
+    bool runInShell = false}) {
   // TODO(rnystrom): Should dart:io just handle this?
   // Spawning a process on Windows will not look for the executable in the
   // system path. So, if executable looks like it needs that (i.e. it doesn't
   // have any path separators in it), then spawn it through a shell.
-  if (Platform.isWindows && executable.indexOf('\\') == -1) {
+  if (Platform.isWindows && executable.contains('\\')) {
```
jakemac53: this should be `!executable.contains`

lib/src/io.dart
```diff
@@ -963,10 +962,10 @@ ByteStream createTarGz(List<String> contents, {String baseDir}) {
   return new ByteStream(StreamCompleter.fromFuture(new Future.sync(() async {
     var buffer = new StringBuffer();
     buffer.write('Creating .tar.gz stream containing:\n');
-    contents.forEach((file) => buffer.write('$file\n'));
+    buffer.writeAll(contents, '\n');
```
jakemac53: technically this is not emitting the last newline that would have been emitted before

lib/src/io.dart
natebosch: Done

lib/src/io.dart
natebosch: Switched to `contents.forEach(buffer.writeln);`

Issue #1922, opened 2018-06-15
Remove `-c, --[no-]checked` from CLI usage text


chalin:
From https://github.com/dart-lang/pub/issues/1807#issuecomment-397573190:

The option `--[no-]checked` should be removed/replaced.

```console
> pub run -h
Run an executable from a package.

Usage: pub run <executable> [args...]
-h, --help            Print this usage information.
-c, --[no-]checked    Enable runtime type checks and assertions.

Run "pub help" to see global options.
```

cc @kevmoo @kwalrath

kevmoo:
Maybe just add `--[no-]enable-asserts` and hit `--checked`?

kevmoo:
Sorry. This is for pub client, not pkg:test – sorry for the fire drill @grouma

natebosch:
> In Dart2 strong mode, the checked parameter only controls asserts, but not type checks.

https://api.dartlang.org/dev/2.0.0-dev.63.0/dart-isolate/Isolate/spawnUri.html

`--enable-asserts` would have the same implementation as `--checked` today. I'd be fine adding an alias for this and hiding the original.

natebosch:
@kevmoo - what should the single character flag be?

kevmoo:
`e`? `a`?

chalin:
> `--enable-asserts` would have the same implementation as `--checked`

In the context of Dart 2, since the old and the new flags have the same effect, how about this: renaming the new flag to `--check-asserts` and reusing `-c` as a shorthand.

Of course `--no-checked` and `--checked` would be dropped.

chalin:
Any update on this?

kevmoo:
@jakemac53 @natebosch ?

natebosch:
Fixed with 9f00679ef47bc79cadc18e143720ade6c06c0100

PR #1964, opened 2018-08-07
Rename --checked to --check-asserts

natebosch:
Keep the old flag as an alias, run with checked mode if either is set.
Behavior will be weird for cases like `--checked --no-check-asserts` but
it's not worth the extra complexity to track that down.

Update tests using the flag.

Use consistent single quotes in the files that were being edited anyway.


kevmoo:
Shouldn't we be going with `--enable-asserts` ? To be consistent with...everything else?

natebosch:
See https://github.com/dart-lang/pub/issues/1922#issuecomment-398352483

natebosch:
discussed offline. Went with `--enable-asserts` and update the help string to match `dart --help`

lib/src/command/global_run.dart
```diff
@@ -13,28 +13,28 @@ import '../utils.dart';
...
       "NOTE: We are currently optimizing this command's startup time.";
-  String get invocation => "pub global run <package>:<executable> [args...]";
+  String get invocation => 'pub global run <package>:<executable> [args...]';
   bool get allowTrailingOptions => false;
 
   GlobalRunCommand() {
-    argParser.addFlag("checked",
-        abbr: "c", help: "Enable runtime type checks and assertions.");
-    argParser.addOption("mode", help: "Deprecated option", hide: true);
+    argParser.addFlag('enable-asserts', help: 'Enable assert statements.');
```
chalin: The following seems clearer / more accurate as help text: **Enable runtime checks of assert statements**. WDYT? /cc @kwalrath @kevmoo

lib/src/command/run.dart
```diff
@@ -14,20 +14,20 @@ import '../utils.dart';
...
+  String get name => 'run';
+  String get description => 'Run an executable from a package.';
+  String get invocation => 'pub run <executable> [args...]';
   bool get allowTrailingOptions => false;
 
   RunCommand() {
-    argParser.addFlag("checked",
-        abbr: "c", help: "Enable runtime type checks and assertions.");
-    argParser.addOption("mode", help: "Deprecated option", hide: true);
+    argParser.addFlag('enable-asserts', help: 'Enable assert statements.');
```
chalin: Ditto

lib/src/executable.dart
```diff
@@ -22,7 +22,7 @@ import 'utils.dart';
 ///
 /// Arguments from [args] will be passed to the spawned Dart application.
 ///
-/// If [checked] is true, the program is run in checked mode.
+/// If [checked] is true, the program is run with assertions enabled.
```
chalin: Here and elsewhere: how about "with assertion checking enabled"?

lib/src/command/global_run.dart
natebosch: I took this directly from `dart --help`. If we want to change this I think we should do it in both places.

PR #1972, opened 2018-08-16
Add check for a scheme on PUB_HOSTED_URL

natebosch:
If the environment variable does not have a scheme it will produce a
confusing error like `Bad state: Cannot use origin without a scheme`.

- Add a `ConfigException` to correspond to the config exit code.
- Change some non-reference usages of `[]` to backticks.
- Stop setting a default of the `server` option since it depends on user
  config and could be incorrect. We never read the default value anyway
  and now access of that field can throw and this would throw before
  proper exception handling is set up.
- Attempt to parse the scheme when we read `PUB_HOSTED_URL` and throw a
  more actionable exception when it is missing.
- Update tests hardcoding the usage string.

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -35,11 +35,19 @@ class HostedSource extends Source {
           : BoundHostedSource(this, systemCache);
 
   /// Gets the default URL for the package server for hosted dependencies.
-  String get defaultUrl {
-    var url = io.Platform.environment["PUB_HOSTED_URL"];
-    if (url != null) return url;
-
-    return "https://pub.dartlang.org";
+  String get defaultUrl =>
+      _defaultUrl ??= _pubHostedUrlConfig() ?? 'https://pub.dartlang.org';
+  String _defaultUrl;
+  String _pubHostedUrlConfig() {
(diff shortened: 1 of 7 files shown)


lib/src/source/hosted.dart
```diff
@@ -35,11 +35,19 @@ class HostedSource extends Source {
...
-
-    return "https://pub.dartlang.org";
+  String get defaultUrl =>
+      _defaultUrl ??= _pubHostedUrlConfig() ?? 'https://pub.dartlang.org';
+  String _defaultUrl;
+  String _pubHostedUrlConfig() {
+    var url = io.Platform.environment['PUB_HOSTED_URL'];
+    if (url == null) return null;
+    var uri = Uri.parse(url);
+    if (uri.scheme?.isEmpty ?? true) {
```
jakemac53: nit: to me `!= false` reads a bit nicer here, but up to you

jakemac53:
> Do you have suggestions on how we can know for sure?

I don't - my assumption is currently that it wouldn't break anything that wasn't already broken anyways. It will change the particular exit code, and the reporting of the failure though. I think those are just fine.

PR #1975, opened 2018-08-21
Update default SDK upper bound

zanderso:
The dev SDK is at 2.1.0-... so pub misbehaves when packages default to a 2.0.0 upper bound.

### DIFF
--- a/lib/src/pubspec.dart
+++ b/lib/src/pubspec.dart
@@ -32,7 +32,7 @@ final _packageName =
 ///
 /// This provides a sane default for packages that don't have an upper bound.
 final VersionRange _defaultUpperBoundSdkConstraint =
-    VersionConstraint.parse("<2.0.0");
+    VersionConstraint.parse("<3.0.0");
 
 /// Whether or not to allow the pre-release SDK for packages that have an
 /// upper bound Dart SDK constraint of <2.0.0.


jakemac53:
The reason for the <2.0.0 default is that all old existing packages on pub probably do _not_ support dart 2, and many of then do _not_ have an upper bound constraint on the sdk version. Ideally we would have forced an upper bound sdk constraint for all packages from the beginning, but we didn't and so we are somewhat stuck here.

matanlurey:
Can we enforce a lower and upper bound, i.e. `>=2.0.0 <3.0.0`?

jakemac53:
I believe that the intention however was to only default the constraints for your dependencies, but not for the current package. This would allow applications to omit sdk constraints. If it is not working that way today then I think it is a bug.

jakemac53:
> Can we enforce a lower and upper bound, i.e. >=2.0.0 <3.0.0?

Agreed that we should enforce both lower and upper on publish if we aren't already (I thought we are but not 100% sure)

natebosch:
What do we mean by "misbehaves"? What is the current behavior and what is the new behavior?

zanderso:
When an SDK constraint is not declared, we get:

```
Because $PACKAGE depends on [e.g.] lorem any which requires SDK version <2.0.0, version solving failed.
```

A default upper bound of 2.0.0 makes it impossible to depend on packages that are Dart 2 compliant, but do not declare an SDK upper bound.

jakemac53:
> Depending on the package we might also be able to contact the author and ask to take over the existing package name and publish a new version.

We have had good success with this before, @kevmoo is good at reaching out to these people.

> In the short term I believe we can hack around it with dependency_overrides pinning a specific version.

This has to happen at the application level though... so if its a transitive dependency of all fuschia apps that doesn't really work.

zanderso:
This default number will have to be bumped or removed at some point. What are the criteria for doing that?

natebosch:
> This default number will have to be bumped or removed at some point. What are the criteria for doing that?

Why would it ever need to be bumped or removed? We expect that most packages published without upper bound were intended for Dart 1 and should instead prefer to enforce or encourage all new published packages to include an upper constraint (We may already do that?)

jakemac53:
Fwiw, we do have validation now that requires an upper bound constraint, https://github.com/dart-lang/pub/blob/43a91d1245e1213ce6437b4a261033f50dbc5aff/lib/src/validator/sdk_constraint.dart#L37. 

We don't need to change the default in the future because we require this going forward.

kevmoo:
@jakemac53 @zanderso – can this be closed?

PR #1993, opened 2018-10-19
Check for dead links

neelkamath:
One dead link was found via this

### DIFF
--- a/test/dead_link_test.dart
+++ b/test/dead_link_test.dart
@@ -1,0 +1,22 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:link_checker/link_checker.dart';
+import 'package:test/test.dart';
+
+void main() => test('project for dead links', () async {
+      var badLinks = <BadLink>[];
+      await for (BadLink badLink
+          in getBadLinksInDirectory(blacklistedLinksRegexes: [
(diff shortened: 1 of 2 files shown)


natebosch:
@mit-mit  - can you triage?

mit-mit:
Sorry, I missed this one. This seems like a reasonable change to me. cc @jonasfj

natebosch:
Note that this will require a new package pulled into the SDK repo.

test/dead_link_test.dart
```diff
@@ -0,0 +1,22 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:link_checker/link_checker.dart';
+import 'package:test/test.dart';
+
+void main() => test('project for dead links', () async {
+      var badLinks = <BadLink>[];
+      await for (BadLink badLink
+          in getBadLinksInDirectory(blacklistedLinksRegexes: [
```
jonasfj: `getBadLinksInDirectory` assumes that `Directory.current` is the project root. I don't know if this always holds? It might...

jonasfj:
I would prefer that this is something we run occasionally instead...

Maybe we could find (or someone could build) a github service that scans a repository for broken links and files issues :)

ayan-b:
> Maybe we could find (or someone could build) a github service that scans a repository for broken links and files issues :)

A GitHub action cron job can be written which runs weekly/quarterly/monthly. In that way, it doesn't intervene with travis and get the job done.

PR #1995, opened 2018-10-26
2 commits - lints and cleaning up dependencies

kevmoo:


kevmoo:
The Travis failure is the same as w/ current master

lib/src/http.dart
```diff
@@ -42,7 +42,7 @@ class _PubHttpClient extends http.BaseClient {
   http.Client _inner;
 
   _PubHttpClient([http.Client inner])
-      : this._inner = inner == null ? http.Client() : inner;
+      : _inner = inner == null ? http.Client() : inner;
```
natebosch: for next time: `_inner = inner ?? http.client();`

kevmoo:
I didn't enable that lint :-)

Issue #2003, opened 2018-11-24
Feature request: warn if version to be published is absent from changelog


RedBrogdon:
I have a devil of a time remembering to update my changelog prior to publishing a new version of a package, and I know the `publish` command is capable of kicking out warnings prior to publication if it finds anything that looks fishy. I'd love it if pub automatically scanned the changelog.md file in my package for the version number I'm trying to publish, and warned me if it wasn't present.

jakemac53:
That seems pretty reasonable to me, fwiw @RedBrogdon this would be fairly easy to add yourself if you wanted to make a contribution :).

jakemac53:
Ok sure, also not a requirement just mentioning that is an option (and probably a faster one than waiting for somebody else to come fix this)

jakemac53:
See https://github.com/dart-lang/pub/blob/master/lib/src/validator/name.dart for an example validator

jakemac53:
cc @jonasfj @sigurdm how you do both feel about this? Willing to accept a contribution for it?

sigurdm:
I believe we already have this: https://github.com/dart-lang/pub/blob/master/lib/src/validator/changelog.dart#L46

Please reopen if something else was meant.

RedBrogdon:
I think this is my fault for leaving the bug open. I was still new to working with the Dart team back then and probably didn't feel confident claiming I'd actually fixed something. 😄

jakemac53:
Huh, I just accidentally published a package yesterday which had no changelog at all which this check appears to allow https://github.com/dart-lang/pub/issues/2003#L22.

I would consider making that a warning as well as all packages should have a changelog.

sigurdm:
I agree we should warn about that. Opened https://github.com/dart-lang/pub/issues/2337

PR #2016, opened 2018-12-18
Adds a changelog validator that checks if the current version is mentioned.

RedBrogdon:
Fixes Issue #2003 by adding a new Validator class that looks for a CHANGELOG file and, if one is located, verifies that the current package version can be found somewhere in it.

This is my first contribution to pub, so literally anything in here could be wrong. I based the code mostly off of the README validator and the tests pass, so I've at least got that going for me. 😄

### DIFF
--- a/lib/src/validator.dart
+++ b/lib/src/validator.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:meta/meta.dart';
+import 'package:pub/src/validator/changelog.dart';
 import 'package:pub_semver/pub_semver.dart';
 
 import 'entrypoint.dart';
@@ -120,6 +121,7 @@ abstract class Validator {
       ExecutableValidator(entrypoint),
       CompiledDartdocValidator(entrypoint),
       ReadmeValidator(entrypoint),
+      ChangelogValidator(entrypoint),
       SdkConstraintValidator(entrypoint),
       StrictDependenciesValidator(entrypoint),
     ];
--- a/lib/src/validator/changelog.dart
+++ b/lib/src/validator/changelog.dart
@@ -1,0 +1,49 @@
+// Copyright (c) 2012, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+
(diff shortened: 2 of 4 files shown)


kevmoo:
@mit-mit Do you guys want to run w/ this?

RedBrogdon:
<gentle ping>

lib/src/validator/changelog.dart
```diff
@@ -0,0 +1,46 @@
...
+      } on FormatException catch (_) {
+        warnings.add("$changelog contains invalid UTF-8.\n"
+            "This will cause it to be displayed incorrectly on "
+            "pub.dartlang.org.");
+      }
+
+      final version = entrypoint.root.pubspec.version.toString();
+
+      if (!contents.contains(version)) {
+        warnings.add("Your package includes a changelog file that doesn\'t "
```
sigurdm: Drive by comment: this message should probably mention `$changelog`

lib/src/validator/changelog.dart
```diff
@@ -0,0 +1,46 @@
...
+    return Future.sync(() {
+      final changelog = entrypoint.root.changelogPath;
+
+      if (changelog == null) {
+        // No changelog was found, which is fine. Return with no warnings.
+        return;
+      }
+
+      var bytes = readBinaryFile(changelog);
+      var contents = '';
```
sigurdm: Drive by comment: you might prefer:
```
String contents;
```

lib/src/validator.dart
```diff
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:meta/meta.dart';
+import 'package:pub/src/validator/changelog.dart';
```
sigurdm: Drive by comment: this import does not follow the pattern of the others. (should be a relative import)

lib/src/validator/changelog.dart
RedBrogdon: Good point. Revised to better match the format used by the UTF-8 check.

RedBrogdon:
Thanks for the review! PTAL.

RedBrogdon:
Any news here? I'm sure the team has larger responsibilities, but I'd really like to get this closed out one way or another.

mit-mit:
cc @jonasfj @sigurdm can we take this?

RedBrogdon:
Thanks!

PR #2021, opened 2019-01-03
Update/add www.dartlang.org URLs

### LINKED ISSUE #2001: Update doc URLs
The help strings for each command have doc links that work but are slightly outdated. For example:

> See http://dartlang.org/tools/pub/cmd/pub-cache.html for detailed documentation.

We should update them to use https and no terminal ".html":

> See https://dartlang.org/tools/pub/cmd/pub-cache for detailed documentation.

It looks like this should be an easy change, and I'd be happy to create a PR.

kwalrath:
Fixes #2001


kwalrath:
It looks like all the build jobs are passing, except for the one that's fixed by #2020.

Issue #2039, opened 2019-01-30
Support global activate from a different pub server


evanweible-wf:
`pub global activate` already supports hosted, path, and git sources, but the `hosted` option defaults to the main pub server with no way to activate from a separate server. We have a private pub server for our internal packages and it would be nice to be able to globally activate some of them using the standard version constraints.

It looks like this idea was at least considered previously in this issue: https://github.com/dart-lang/pub/issues/1044

> pub global activate foo ">=1.0 <2.0" --server http://...

Is there any reason not to support this? And if not, would it be something that you would accept a PR for? Looking at the implementation of the global activate command, it looks like it might require a separate cache for each pub server, but otherwise seems to be straightforward.

evanweible-wf:
I've put up a proposed solution here: https://github.com/dart-lang/pub/pull/2041

kevmoo:
Fixed in https://github.com/dart-lang/pub/commit/980fc947db32bca6ae7d07fbb4ad0fcdba0050ec

Tracking issue to include this in an SDK - https://github.com/dart-lang/sdk/issues/36026

PR #2041, opened 2019-01-31
Add support for global activating package from a custom pub URL

evanweible-wf:
Proposed solution for #2039, looking for feedback on whether or not this would be an acceptable addition.

The solution ended up being very straightforward since pub already knows how to handle dependencies from custom hosted pub servers and activating a package globally is essentially just performing a `pub get` in a dummy package with a single dependency. All that was needed was a command-line flag to specify the custom pub URL that is then passed through to the `.refFor()` call used to build the dependency ref.

I went with `--source-url` for the CLI option mainly in an attempt to note its relation to `--source` while keeping it brief, but there may be a better name.

Additionally, since this flag is only applicable when activating from a hosted source, it might be worth failing if it is used with a non-hosted source.

### DIFF
--- a/lib/src/command/global_activate.dart
+++ b/lib/src/command/global_activate.dart
@@ -112,9 +112,7 @@ class GlobalActivateCommand extends PubCommand {
 
         validateNoExtraArgs();
         return globals.activateHosted(package, constraint, executables,
-            features: features,
-            overwriteBinStubs: overwrite,
-            url: hostedUrl);
+            features: features, overwriteBinStubs: overwrite, url: hostedUrl);
 
       case "path":
         if (features.isNotEmpty) {
--- a/lib/src/command/global_activate.dart
+++ b/lib/src/command/global_activate.dart
@@ -38,6 +38,11 @@ class GlobalActivateCommand extends PubCommand {
     argParser.addFlag("overwrite",
         negatable: false,
         help: "Overwrite executables from other packages with the same name.");
+
+    argParser.addOption("source-url",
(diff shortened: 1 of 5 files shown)


lib/src/command/global_activate.dart
```diff
@@ -38,6 +38,11 @@ class GlobalActivateCommand extends PubCommand {
     argParser.addFlag("overwrite",
         negatable: false,
         help: "Overwrite executables from other packages with the same name.");
+
+    argParser.addOption("source-url",
```
jonasfj: `--server-url` or `--hosted-url` might be better, the key for this in the pubspec is `dependencies.<packageName>.hosted.url`

Or maybe `--hosted-source-url`

test/global/activate/custom_source_url_test.dart
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
natebosch: [nit] 2019

lib/src/command/global_activate.dart
natebosch: +1 on `--hosted-url`

test/global/activate/custom_source_url_test.dart
evanweible-wf: Fixed

evanweible-wf:
@natebosch @jonasfj Updated the command line option to `--hosted-url`, let me know if there's any other changes you'd like to see.

PR #2067, opened 2019-02-26
Remove validation for library file paths

### LINKED ISSUE #2024: Invalid lint for part files
I am trying to publish built value generated code and I'm getting the following suggestion:

* The name of "lib/data/log_to_paths_request.g.dart", "log_to_paths_request.g", may only contain letters, numbers, and underscores.

This seems like an invalid lint since periods should be allowed in file names. I would imagine Angular code has similar issues as Angular applications often contain "file.template.dart".

natebosch:
Fixes #2024

There is nothing that requires a library name to be a valid Dart
identifier, and it's common practices to at least have `part of` files
which include a `.` in the file name.

### DIFF
--- a/lib/src/validator/name.dart
+++ b/lib/src/validator/name.dart
@@ -10,21 +10,16 @@ import '../entrypoint.dart';
 import '../utils.dart';
 import '../validator.dart';
 
-/// A validator that validates the name of the package and its libraries.
+/// A validator that the name of a package is legal and matches the library name
+/// in the case of a single library.
 class NameValidator extends Validator {
   NameValidator(Entrypoint entrypoint) : super(entrypoint);
 
   Future validate() {
     return Future.sync(() {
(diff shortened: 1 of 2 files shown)


lib/src/validator/name.dart
```diff
@@ -10,21 +10,16 @@ import '../entrypoint.dart';
...
-      _checkName(entrypoint.root.name, 'Package name "${entrypoint.root.name}"',
-          isPackage: true);
+      _checkName(entrypoint.root.name);
 
       var libraries = _libraries;
-      for (var library in libraries) {
-        var libName = path.basenameWithoutExtension(library);
-        _checkName(libName, 'The name of "$library", "$libName",',
-            isPackage: false);
-      }
```
jonasfj: I think this check is valuable, but as you point out we're not running the analyzer here and we can't spot whether the file is using `part of`, so this check is too simple.

Hence, I agree we should omit it...

PR #2093, opened 2019-03-05
Fix #28 -- escaping arguments as mentioned in TODO

### LINKED ISSUE #28: Implement proper shell escaping for pub command error reproduction
<a href="https://github.com/whesse"></img></a> **Issue by [whesse](https://github.com/whesse)**
_Originally opened as dart-lang/sdk#23505_

---

Error reporting from pub includes printing a command line to rerun the failing command.  The arguments need to be properly escaped, depending on the shell (OS) that is running.  We currently put double quotes around any argument containing spaces.

jonasfj:
I just want to close: https://github.com/dart-lang/pub/issues/28

### DIFF
--- a/lib/src/command_runner.dart
+++ b/lib/src/command_runner.dart
@@ -180,8 +180,13 @@ class PubCommandRunner extends CommandRunner {
       if (topLevelResults['trace']) {
         log.dumpTranscript();
       } else if (!isUserFacingException(error)) {
-        // TODO(23505): Implement proper shell escaping, not a partial hack.
-        protectArgument(String x) => x.contains(' ') ? '"$x"' : x;
+        // Escape the argument for users to copy-paste in bash.
+        // Wrap with single quotation, and use '\'' to insert single quote, as
+        // long as we have no spaces this doesn't create a new argument.
+        protectArgument(String x) =>
+            RegExp(r'^[a-zA-Z0-9-_]+$').stringMatch(x) == null
+                ? "'${x.replaceAll("'", r"'\''")}'"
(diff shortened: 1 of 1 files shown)


lib/src/command_runner.dart
```diff
@@ -180,8 +180,11 @@ class PubCommandRunner extends CommandRunner {
       if (topLevelResults['trace']) {
         log.dumpTranscript();
       } else if (!isUserFacingException(error)) {
-        // TODO(23505): Implement proper shell escaping, not a partial hack.
-        protectArgument(String x) => x.contains(' ') ? '"$x"' : x;
+        // Escape the argument for users to copy-paste in bash.
+        protectArgument(String x) =>
+            RegExp(r'^[a-zA-Z0-9-_]+$').stringMatch(x) == null
+                ? "'${x.replaceAll("'", "'\\''")}'"
```
sigurdm: You cannot escape a single quote inside a single quote:
https://www.gnu.org/software/bash/manual/html_node/Single-Quotes.html#Single-Quotes
You probably need to use double quotes and escape all of ‘$’, ‘`’, ‘\’

jonasfj: Not really... Read carefully I escape single quote, by ending single quotation then having an escaped single quote and starting single quotation again :)

If there is no space between end and start of single quotation it doesn't become a new argument...

sigurdm: Explain this in a comment ...

sigurdm: You could use `r"'\''"`

Issue #1813, opened 2018-02-27
Warning on `pub publish` a library that includes an `/example` with separate pubspec


mehaase:
I have written a library for some web components, and I have also written a little demo web app to show off those components. Here is a simplified layout of my library:

```
|-- CHANGELOG.md
|-- LICENSE
|-- MAINTENANCE.md
|-- README.md
|-- example
|   |-- README.md
|   |-- lib
|   |   |-- about.dart
|   |   |-- about.html
|   |   |-- body.dart
|   |   | (32 more...)
|   |-- pubspec.yaml
|   '-- web
|       |-- index.html
|       '-- main.dart
|-- lib
|   |-- ng_modular_admin.dart
|   |-- src
|   |   |-- components
|   |   |   |-- app.css
|   |   |   |-- app.dart
|   |   |   |-- app.html
|   |   |   | (78 more...)
|   |   |-- framework.dart
|   |   '-- services
|   |       |-- side_nav.dart
|   |       '-- toast.dart
|   |-- util.dart
|   '-- validators.dart
|-- pubspec.yaml
'-- tool
    '-- build-styles.dart
```

The library is called `ng_modular_admin` and it contains a `/lib` directory with a bunch of Angular components. The demo app is called `ng_modular_admin_demo` and it lives in `/example` and it has its own `/example/pubspec.yaml`, `/example/lib`, and `/example/web`. The demo app depends on the library. In the demo app, the `/example/web/main.dart` imports an application component like this: `package:ng_modular_admin_demo/body.dart`. 

This works fine when I'm testing the library and the demo, but when I run `pub publish`, it issues a warning about importing from the `ng_modular_admin_demo` package without depending on it. I think I understand why it is warning on this, because `ng_modular_admin` does not depend on `ng_modular_admin_demo`, and I guess `pub publish` ignores the nested `pubspec.yaml`, but I'm not clear on whether or how I should address this warning.

Is this user error? A violation of packaging guidelines? Or should `pub` handle a nested `/example/pubspec.yaml` more gracefully? If this is user error, I would be happy to submit a PR to clarify the documentation.

The source code for the library and demo is [on github](https://github.com/HyperionGray/ng_modular_admin). Full `pub publish` output is [in a gist](https://gist.github.com/mehaase/66fc5a9e6264a43b234c1b10917c83a9).

nex3:
[...]

PR #2107, opened 2019-03-26
Don't validate imports in `example/` if it contains its own pubspec.yaml

sigurdm:
Proposed fix of: https://github.com/dart-lang/pub/issues/1813

### DIFF
--- a/lib/src/validator/strict_dependencies.dart
+++ b/lib/src/validator/strict_dependencies.dart
@@ -74,7 +74,7 @@ class StrictDependenciesValidator extends Validator {
       ..add(entrypoint.root.name);
     var devDependencies = MapKeySet(entrypoint.root.devDependencies);
     _validateLibBin(dependencies, devDependencies);
-    _validateBenchmarkExampleTestTool(dependencies, devDependencies);
+    _validateBenchmarkTestTool(dependencies, devDependencies);
   }
 
   /// Validates that no Dart files in `lib/` or `bin/` have dependencies that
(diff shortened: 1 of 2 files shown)


lib/src/validator/strict_dependencies.dart
```diff
@@ -96,10 +96,15 @@ class StrictDependenciesValidator extends Validator {
...
   /// Validates that no Dart files in `benchmark/`, `example/, `test/` or
   /// `tool/` have dependencies that aren't in [deps] or [devDeps].
+  ///
+  /// `example/` is only validated if it does not contain its own pubspec.
   void _validateBenchmarkExampleTestTool(
       Set<String> deps, Set<String> devDeps) {
-    for (var usage
-        in _usagesBeneath(['benchmark', 'example', 'test', 'tool'])) {
+    var directories = ['benchmark', 'test', 'tool'];
+    if (!fileExists(entrypoint.root.path('example', 'pubspec.yaml'))) {
```
jonasfj: Hmm, it's it better to either care about the contents of the `example/` folder, or not care at all.

sigurdm: Makes sense.
Done

lib/src/validator/strict_dependencies.dart
```diff
@@ -98,8 +98,8 @@ class StrictDependenciesValidator extends Validator {
   /// `tool/` have dependencies that aren't in [deps] or [devDeps].
   void _validateBenchmarkExampleTestTool(
```
jonasfj: We should rename this

sigurdm: done

PR #2115, opened 2019-04-04
Outline for a minimal repository specification

jonasfj:
This is an attempt to create a specification from our current behavior.
The server doesn't care about the `Accept` header, and does not send
the correct `Content-Type`. But the client seems to get this right,
and it doesn't seems to care about the response header.

This makes is clear what we can't change, and how to deal with
versioning if we decide to change something. I want to further
document what API end-points `pub.dartlang.org` offers and how
stable they are. This is just a first step in that direction, since
we're probably not going to able to change end-points required by
old pub clients.

Expanding this to cover publishing would be nice, but at the momement
the authorization scheme is hardcoded to use the Goolge OAuth flow
for all hosted pub servers, which is a security concern should users
attept to publish packages to an untrusted server. It's also a security
concern for people writing custom pub servers, so that will probably
have to be redesigned a bit, before we can formalize it.

CC @isoos, @natebosch, @pq

### DIFF
--- a/doc/repository-spec-v2.md
+++ b/doc/repository-spec-v2.md
@@ -1,0 +1,102 @@
+Hosted Pub Repository Specification Version 2
+=============================================
+
+This document specifies the REST API that a hosted pub package repository must
+implement.
+
+A hosted pub package repository is a server from which packages can be
+downloaded, the default public pub server is `'https://pub.dartlang.org'`.
+This can be overwritten in the `dependencies` of a `pubspec.yaml` or specified
+with the environment variable `PUB_HOSTED_URL`. In the rest of this document we
+shall refer to the base URL as `PUB_HOSTED_URL`.
+
+## Forward Compatibility
+To ensure forward compatibility API requests should carry an
+`Accept: application/vnd.pub.v2+json` header. As future versions of the API
+may have different responses.
+
+## Fetching Packages
+
+The following API end-points MUST be supported by a hosted pub package
+repository.
+
+**Optional headers** used for popularity metrics.
+
+* `X-Pub-OS: windows | macos | linux | ...`
+* `X-Pub-Command: get | upgrade | ...`
+* `X-Pub-Session-ID: <UUID>`
+* `X-Pub-Reason: direct | dev`
+* `X-Pub-Environment: ...`
+* `User-Agent: ...`
+
+### List all versions of a package
+
+**GET** `<PUB_HOSTED_URL>/api/packages/<PACKAGE>`
+
+**Headers:**
+* `Accept: application/vnd.pub.v2+json`
+
+**Response**
+* `Content-Type: application/vnd.pub.v2+json`
+
+```js
+{
+  "name": "<PACKAGE>",
+  "latest": {
+    "version": "<VERSION>",
+    "archive_url": "<PUB_HOSTED_URL>/packages/<PACKAGE>/versions/<VERSION>.tar.gz",
+    "pubspec": {
+      /* pubspec contents as JSON object */
+    }
+  },
+  "versions": [
+    {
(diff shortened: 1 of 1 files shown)


doc/repository-spec-v2.md
```diff
@@ -0,0 +1,101 @@
...
+
+**GET** `<PUB_HOSTED_URL>/packages/<PACKAGE>/versions/<VERSION>.tar.gz`
+
+**Headers**
+* `Accept: application/octet-stream` (optional)
+
+**Response** (typically through a ´30x` redirect)
+* `Content-Type: application/octet-stream`
+
+**Important:** The server MAY redirect the client to a differnet URL, clients
```
sigurdm: different

doc/repository-spec-v2.md
```diff
@@ -0,0 +1,101 @@
...
+MUST support redirects.
+
+## Publishing Packages
+
+The API for authenticating and publishing packages is not formalized yet.
```
sigurdm: Perhaps add TODO(jonasfj)

sigurdm:
cc @nex3

jonasfj:
Note. I'm tempted to make the case that we should not use a custom content-type to signal API versioning, because:
 * It can't cover the tarball which is a binary response.
 * https://www.chromium.org/developers/design-documents/blocking-cross-site-documents
 * Everybody expects `application/json`
 * No other server implementation is likely to get this right.

That said, I do see the point that including version numbers in the URL is less than ideal :)

isoos:
> including version numbers in the URL is less than ideal

Why is it less than ideal? We can add things to version X until we have a breaking change, and then open a new endpoint under version X+1?

The complexity of compatibility won't go away with additional headers, it just makes testing and monitoring less obvious (compared to `/api/v2/` URLs anyway). I'd rather put version in the URL than anywhere else.

jonasfj:
For context:
https://stackoverflow.com/questions/389169/best-practices-for-api-versioning
https://stackoverflow.com/questions/10742594/versioning-rest-api

I personally think putting it in the URL is nice and simple, but I see the arguments that using `Accept` header is more REST-like. Only potential downside I see is security in the browser, but we have `nosniff` on everything.

doc/repository-spec-v2.md
jonasfj: I referenced the issue instead :)

doc/repository-spec-v2.md
```diff
@@ -0,0 +1,102 @@
+Hosted Pub Repository Specification Version 2
+=============================================
+
+This document specifies the REST API that a hosted pub package repository must
+implement.
+
+A hosted pub package repository is an server from which packages can be
```
pq: an => a

doc/repository-spec-v2.md
```diff
@@ -0,0 +1,102 @@
...
+downloaded, the default public pub server is `'https://pub.dartlang.org'`.
+This can be overwritten in the `dependencies` of a `pubspec.yaml` or specified
+with the environment variable `PUB_HOSTED_URL`. In the rest of this document we
+shall refer to base URL as `PUB_HOSTED_URL`.
```
pq: "to base" => "to the base" ?

kevmoo:
QQ – should this be defined in https://github.com/dart-lang/pub_server ?

I like the idea that we have a canonical request handler here and that this repo is just our implementation of it...

jonasfj:
I'm not convinced we should be using `pub_server`, it seems like a thing that adds unnecessary abstractions to an otherwise simple API.

It might have value as an minimalistic example of how to implement a server. Putting this document there just makes it harder to deprecated the repository, should we ever get that far ;)

PR #2134, opened 2019-05-20
A note on why we're not using pub.dev/api

jonasfj:
Just so that we don't forget why we didn't change this right now.

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -35,8 +35,21 @@ class HostedSource extends Source {
           : BoundHostedSource(this, systemCache);
 
   /// Gets the default URL for the package server for hosted dependencies.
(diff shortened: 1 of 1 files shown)


lib/src/source/hosted.dart
```diff
@@ -35,8 +35,20 @@ class HostedSource extends Source {
           : BoundHostedSource(this, systemCache);
 
   /// Gets the default URL for the package server for hosted dependencies.
-  String get defaultUrl =>
-      _defaultUrl ??= _pubHostedUrlConfig() ?? 'https://pub.dartlang.org';
+  String get defaultUrl {
+    // Changing this to pub.dev raises the following concerns:
+    //
+    //  1. It would below through users caches.
```
jakemac53: blow*

PR #2137, opened 2019-05-22
Use canonical oauth2 scope for email

jonasfj:
Now that we know this is the canonical scope, then let's use that.


jonasfj:
Note. we should roll to the dart-sdk when this is landed..

PR #2141, opened 2019-05-24
Fix most dartlang.org URLs

kevmoo:
Also remove codereview.settings


kevmoo:
CC @kwalrath

lib/src/validator/changelog.dart
```diff
@@ -31,7 +31,7 @@ class ChangelogValidator extends Validator {
       } on FormatException catch (_) {
         warnings.add("$changelog contains invalid UTF-8.\n"
             "This will cause it to be displayed incorrectly on "
-            "pub.dartlang.org.");
+            "pub.dev.");
```
natebosch: I wonder if we should be concerned about readability with this format - without the `https://` or a well-known TLD like `.org` I'm not sure if this will stand out as a URL...

README.md
```diff
@@ -6,7 +6,7 @@ Thanks for being interested in contributing to pub! Contributing to a new
 project can be hard: there's a lot of new code and practices to learn. This
 document is intended to get you up and running as quickly as possible. If you're
 looking for documentation on using pub, try
-[dartlang.org](https://www.dartlang.org/tools/pub).
+[pub.dev](https://dart.dev/tools/pub/cmd).
```
kwalrath: [pub.dev] -> the [pub tool documentation]

kwalrath: Actually, that's redundant. But you get the idea... the link text shouldn't be pub.dev.

lib/src/validator/changelog.dart
kwalrath: We could instead say "the Pub site (pub.dev)" or "the Pub site (https://pub.dev)".

kevmoo: ack

README.md
kevmoo: ack

README.md
```diff
@@ -4,9 +4,8 @@ Pub is the package manager for Dart.
 
 Thanks for being interested in contributing to pub! Contributing to a new
 project can be hard: there's a lot of new code and practices to learn. This
-document is intended to get you up and running as quickly as possible. If you're
-looking for documentation on using pub, try
-[pub.dev](https://dart.dev/tools/pub/cmd).
+document is intended to get you up and running as quickly as possible. More
+documentation is [here](https://dart.dev/tools/pub/cmd).
```
kwalrath: Not a big deal in this case, but just so you know, "here" is considered bad for accessibility reasons.

https://webaim.org/techniques/hypertext/link_text#text

Better would be: "For more information, see the [pub tool documentation](...)."

kevmoo: TIL! Fixed.

kevmoo:
@natebosch @kwalrath – LGTY?

Issue #2032, opened 2019-01-25
error publishing linter ("Bad state: Unable to find the context")


pq:
Hi!  I'm trying to publish the latest linter package ([0.1.79](https://github.com/dart-lang/linter/tree/0.1.79)) but it's failing.

```
[~/src/repos/linter] (master) $ pub --version

Pub 2.2.0-edge.983b77dda72acd73e1616bb63bf98acd8b5fee82
```

[pub_trace.txt](https://github.com/dart-lang/pub/files/2797554/pub_trace.txt)

Thanks for any help!

pq:
Publishing works w/ 2.1.0.

FWIW @bwilkerson suggested this might be the result of changes to `ContextManager` @stereotype441 made recently in HEAD...

stereotype441:
@bwilkerson can you clarify what changes you are thinking of, and what you are seeing in the log that suggests the failure might be related to those changes?  AFAICT the last change I made to ContextManager (other than comment changes) was in https://github.com/dart-lang/sdk/commit/993b4caf0e0b7f094ef5bb2ce91152e8c0e1468e, in May.

jonasfj:
I think we have a similar issue in: https://github.com/dart-lang/pub/issues/2087

pq:
@stereotype441, @bwilkerson : I may have mischaracterized the issue.  Sorry for that.  It does persist though!

pq:
FWIW this is still an issue and I've been resorting to an old version of `pub` for publishing `package:linter`.  Not a big deal at the moment but probably not desirable in the long run!

passsy:
I run into the same error with dart 2.2 and dart 2.3 when trying to publish [ktdart](https://github.com/passsy/kt.dart). I always have to switch to dart 2.1 to publish.

```
$ pub --version
Pub 2.3.0
$ pub publish -n        
Bad state: Unable to find the context to /Users/pascalwelsch/Projects/kt.dart/test/collection/list_test.dart
package:pub/src/dart.dart 146:5                            AnalysisContextManager._getExistingSession
package:pub/src/dart.dart 112:23                           AnalysisContextManager.parse
package:pub/src/dart.dart 127:16                           AnalysisContextManager.parseImportsAndExports
package:pub/src/validator/strict_dependencies.dart 40:36   StrictDependenciesValidator._findPackages
dart:core                                                  _SyncIterator.moveNext
[...]

PR #2143, opened 2019-05-27
Fix 2032, ignore analysis_options.yaml in pub

jonasfj:
The `analysis_options.yaml` is not part of the package.
It's for the editor, it should not affect which binaries are loaded
or how dependencies are analyzer when publishing.


jonasfj:
Yes, it's weird that we have different warnings in `pub` client and server.

passsy:
Thanks for fixing #2032

Issue #1867, opened 2018-04-10
Add a flag to ignore dependency overrides in pubspec


alorenzen:
In angular, we use a "mono_repo" approach, where we keep multiple related packages in the same repo, and test/publish them at once.

In order to support local development and testing on travis, we use `dependency_overrides` in our pubspec to point to the local packages and not the published versions. The only time we don't want this behavior is when we are actually publishing new packages.

Right now, we have to manually comment out our `dependency_overrides` sections in every `pubspec.yaml` before we can run `pub upgrade` and then `pub publish`. This manual step is a burden, and something blocking us from fully automating our publish process.

Can we add a `--ignore-overrides` flag to pub (specifically the `get`, `upgrade`, and `publish` commands) to remove the need for this manual removal step?

nex3:
I'm not sure what your local dependencies have to do with publishing...

jakemac53:
It allows you to test the package against the actual published versions of all dependencies as opposed to the local overrides.

alorenzen:
So `angular_forms` depends on `angular`, which depends on `angular_compiler`, which depends on `angular_ast`.

We have a dependency_override in `angular_forms` for `angular`, `angular_compiler`, and `angular_ast`.  https://github.com/dart-lang/angular/blob/master/angular_forms/pubspec.yaml#L22

*After* we publish `angular`, but *before* we publish `angular_forms`, we need to remove the dependency_overrides and verify that everything is working with the live versions. (I.e. we didn't forget to bump the version for `angular`).

nex3:
All right, sounds good to me. I've also been meaning to add a `--no-dev` flag for a while, which is similar.

jonasfj:
I don't think a `--no-dev` flag will work since you won't get dev-dependencies needed for tests.

But this might be worth taking into the discussion in #2036.

Note. It's also tempting to use `mono_repo` to rewrite pubspecs to comment in/out the `dependency_overrides`.

lemz90:
Docker Compose uses an override file to solve this - [docker-compose.override.yaml](https://docs.docker.com/compose/extends/)

If the override file exists, the content inside is used, otherwise it just uses the docker-compose.yaml like normal.

The override file can then be ignored in version control. I might work on a PR for this. Thoughts?

lemz90:
@nex3 I have some working code for my proposed solution in [my comment](https://github.com/dart-lang/pub/issues/1867#issuecomment-499182381) above.

[...]

PR #2149, opened 2019-06-11
add functionality and tests for pubspec.override.yaml

lemz90:
Hey my team is starting using Dart for our mobile and web applications, using a shared Dart business library we are publishing to a private pub repository. Currently, we use the 'dependency_overrides' pubspec.yaml key when we are working locally on the shared library.

As detailed in issue #1867, developers want a way to ignore the overrides. I proposed a solution mimicking the way docker-compose solves this issue. Here is my PR with a simple implementation and test for this solution. Let me know what you think. Thanks.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -128,6 +128,9 @@ class Entrypoint {
   /// The path to the entrypoint package's pubspec.
   String get pubspecPath => root.path('pubspec.yaml');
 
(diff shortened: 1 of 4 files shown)


lemz90:
Signed CLA

lemz90:
Ugh still made the commit with an unlogged in user or something - closing until I figure that out and fix these tests.

PR #2150, opened 2019-06-11
add functionality and tests for pubspec.override.yaml

lemz90:
Hey my team is starting using Dart for our mobile and web applications, using a shared Dart business library we are publishing to a private pub repository. Currently, we use the `dependency_overrides` pubspec.yaml key when we are working locally on the shared library.

As detailed in issue #1867, developers want a way to ignore the overrides. I proposed a solution mimicking the way docker-compose solves this issue. Here is my PR with a simple implementation and test for this solution. Let me know what you think. Thanks.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -128,6 +128,9 @@ class Entrypoint {
   /// The path to the entrypoint package's pubspec.
   String get pubspecPath => root.path('pubspec.yaml');
 
+  /// The path to the entrypoint package's pubspec.
+  String get pubspecOverridePath => root.path('pubspec.override.yaml');
+
   /// The path to the entrypoint package's lockfile.
   String get lockFilePath => root.path('pubspec.lock');
 
@@ -395,9 +398,14 @@ class Entrypoint {
 
     var pubspecModified = File(pubspecPath).lastModifiedSync();
     var lockFileModified = File(lockFilePath).lastModifiedSync();
+    var pubspecOverrideModified = File(pubspecOverridePath).existsSync()
+        ? File(pubspecOverridePath).lastModifiedSync()
+        : pubspecModified;
 
     var touchedLockFile = false;
-    if (lockFileModified.isBefore(pubspecModified) || hasPathDependencies) {
+    if (lockFileModified.isBefore(pubspecModified) ||
+        lockFileModified.isBefore(pubspecOverrideModified) ||
+        hasPathDependencies) {
       _assertLockFileUpToDate();
       if (_arePackagesAvailable()) {
         touchedLockFile = true;
--- a/lib/src/pubspec.dart
+++ b/lib/src/pubspec.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:collection/collection.dart';
@@ -452,7 +453,10 @@ class Pubspec {
   factory Pubspec.load(String packageDir, SourceRegistry sources,
       {String expectedName, bool includeDefaultSdkConstraint}) {
     var pubspecPath = path.join(packageDir, 'pubspec.yaml');
+    var pubspecOverridePath = path.join(packageDir, 'pubspec.override.yaml');
     var pubspecUri = path.toUri(pubspecPath);
+    var pubspecOverrideUri = path.toUri(pubspecOverridePath);
+
     if (!fileExists(pubspecPath)) {
       throw FileException(
           // Make the package dir absolute because for the entrypoint it'll just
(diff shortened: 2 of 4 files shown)


jonasfj:
How does this solve the original problem of developers wanting a flag to disable overrides?

I do think that using a separate file has some attractive properties (I'm still not sure if there's any good reason to override anything besides `dependencies`). That said, I'm not sure if it's wise to introduce yet another way to do overrides.

Perhaps it's better we fix how `dependency_overrides` work, and add a flag to disable dependency overrides.

**Note**: I think `dependency_overrides` is only supposed to work locally, when installed from `pub.dev` or using a package as a dependency from somewhere else, then `dependency_overrides` should be ignored. Afaik this is the case today, except for a bug in `pub global activate`.

We could also discuss removing the check that prevents publishing a package with `dependency_overrides`, since this field should be ignored when a package is used as a dependency.

jonasfj:
@lemz90,
If you're interested in improving the `dependency_overrides` functionality, I would suggest we aim for the following:

 * Adding a `--ignore-overrides` flag to `pub get`, `pub upgrade`, etc.
 * Ignore `dependency_overrides` in `pubspec.yaml` when a package is used as a dependency.
 * Ignore `dependency_overrides` in `pubspec.yaml` when a package is installed with `pub global activate`.
 * Remove check that prevents publishing packages that contain `dependency_overrides`, since this field should be ignored when a published package is consumed. We could also strip `dependency_overrides` from the pubspec when publishing, but this requires us to modify YAML, which is non-trivial (with comments and stuff).
 * We properly document this behavior in https://dart.dev/tools/pub/pubspec#dependencies 

That way, I think we get a decent setup... where you can test both with and without `dependency_overrides` in CI, without having to modify files.. And we clean-up the current state, where it's not entirely clear how `dependency_overrides` is supposed to work.

@sigurdm, @isoos, does this seem sane to you?

lemz90:
@jonasfj,
Hmm, you are right, I guess it doesn't solve the original issue of adding a flag; however, I think the overrides being in a second file has a few advantages.

Most importantly, you can _ignore it in version control_ and not have to edit `pubspec.yaml` anytime someone checks in their `dependency_overrides`.

I'm not sure I understand the case of wanting to test with `dependency_overrides` in CI. From my perspective, it's purely a local development feature, which seems to correspond with the other proposals you mentioned. I am probably missing a use case here.

That said, I think having a `--ignore-overrides` flag is a good idea. I might start a PR for this today :)

jonasfj:
Ability to `.gitignore` the overrides is a good argument..

However, if you're working in a mono_repo your overrides are typically path dependencies and they can use a relative path, so checking them in isn't a bad thing. Since local development would almost always want to use the other folders in the mono_repo rather than the published versions.

Mostly my concern is:
A) We already have `dependency_overrides`, so adding another method to accomplish the same is not great.
B) A `pubspec.override.yaml` can override more than just dependencies.
C) `pub publish` does not respect `.gitignore`, so recommending users `.gitignore` the `pubspec.override.yaml` file gives a false impression that it won't be published.

-------
If there is a lot of demand for this feature, I think we could add `pubspec.override.yaml`. But unless there is a lot of push for this from developers I suspect that (A) is a good reason not to have a second way to override. It's so easy to have a lot of files in the root of a package, there already is things like `analysis_options.yaml`, `.packages`, `pubspec.lock`, `.dart_tool/` and even more if you use `build_runner`.

The argument isn't that we shouldn't add new files, it's just that before we do that we should be sure it's what we want to do.

-----
I suggest we focus on making the existing mechanism (`dependency_overrides`) better, rather than adding a new one, see https://github.com/dart-lang/pub/pull/2150#issuecomment-501225959. If you're interested in doing some of this work, we're happy to accept PRs :)

sigurdm:
> Ignore dependency_overrides in pubspec.yaml when a package is installed with pub global activate.

I'm partial to this one. I think we should still be able to have overrides when `pub global activate`ing from path (and probably from git).

> Ignore dependency_overrides in pubspec.yaml when a package is used as a dependency.
According to my local tryouts this is already the case, but it should certainly be documented.

lemz90:
Ah ok, this saddens me because it would really help our workflow in my team.

Just for the sake of it, here is a rebuttal to your points:
> A) We already have `dependency_overrides`, so adding another method to accomplish the same is not great.

I definitely understand not wanting to add a second method when one currently exists

> B) A `pubspec.override.yaml` can override more than just dependencies.

This is a simple implementation detail

> C) `pub publish` does not respect `.gitignore`, so recommending users `.gitignore` the `pubspec.override.yaml` file gives a false impression that it won't be published.

According to [the documentation](https://dart.dev/tools/pub/publishing#what-files-are-published), files in .gitignore are not published. Note: I have not tested this.

Anyway, I will start work on `--ignore-overrides`.

jonasfj:
> According to the documentation, files in .gitignore are not published. Note: I have not tested this.

Oh, good point, I might be wrong... It's file not under git control, but also not in `.gitignore` that are also published, if I recall correctly...

StorminGorman:
I would like to +1 having something like a `pub get --override=override.yaml`

Overrides should never be checked into version control.  They are most always environment and developer specific because they involve _relative paths_ ... I think we can all agree that enforcing uniform relative paths across development environments is a bad thing.

The use case for this is a team working with two dart projects simultaneously where one depends on the other, but for whatever reason they should be released separately.  (Two projects sharing the same library for instance.). 

So the developer has to check out both code bases and does a "dev_override" so that he/she work on both code bases simultaneously.  **Checking in the override in this case is terrible**. I've seen the suggestion that we should have a flag like `pub get --ignore-overrides` and honestly I think this is not the way to go.  The overrides would have to be _checked in_ for it to make sense for CI systems to use this flag.  I hate to always have my pubspec.yaml file status always be edited when all I'm doing is overriding a dependency.  I support reopening this issue and exploring the idea of having a separate override file.. Thanks

jonasfj:
@StorminGorman, you raise some interesting concerns..

I want to note that for mono repositories, where multiple packages exist in the same repository, it does make some sense to check-in the `dependency_override` and be able to turn them on/off. Also in CI..

That said, we should be careful as another use case for `dependency_overrides` is when you as an application developer feel you know better than the constraint solver, and wish to override one of your transitive dependencies, or maybe force a specific version of some dependency. I imagine this could be useful for those cases where the constraint resolver fails, but you decide to overrule it.

Maybe these two use-cases are inherently different.

StorminGorman:
@jonasfj Yes I do understand that for mono repositories that it can make _some_ sense to check in a dependency override.. but I think an override file would work just as well in those cases and you have the option to check in the override file of course.  

But what about the poly repository approach which is a common case at least for me.  If I'm working on two open source projects neither of which I own but one depends on the other, it makes sense for me to engage an override after checking them out separately on my machine.  That way I can work on them in tandem. 

I would assume that dart is not taking a position that all developers must have the projects _and the dependency projects they are also working on_ within the same repository?  Or maybe I should pose the question .. is `pub` only meant to be compatible with mono repositories ?  or should we find a solution that works equally well for both mono and poly repository structures?

jonasfj:
> Or maybe I should pose the question .. is pub only meant to be compatible with mono repositories ?

No, we also have many repositories...

> or should we find a solution that works equally well for both mono and poly repository structures?

We don't have to solve all problems at once... I'm also very interested in **not** adding features (mistakes) we can't remove in the future.

But let's reopen this issue and debate a few options..

jonasfj:
Actually, let's move this discussion to https://github.com/dart-lang/pub/issues/2161

Sorry, for all the noise :)

-----
We should probably aim to find a good solution without painting ourselves into a corner. It's easy to add features and flags, it's much harder to remove them!

bsutton:
@lemz90 I'm trying to use this feature but it doesn't appear to be working.
Has it been released?
Is there any documentation on using it?

lemz90:
> @lemz90 I'm trying to use this feature but it doesn't appear to be working.
> Has it been released?
> Is there any documentation on using it?

As far as I know, the only way to override is using the `depedency_overrides:` pubspec.yaml key. You can see more discussion on #2161, but it seems like it's been dead for quite awhile and the maintainers were not really interested in my PR.

bsutton:
Thanks for the response.

I got the override working and I created an issue to improve the
documentation.

PR #2159, opened 2019-06-25
Process invalid git repo (Before fetch and after clone).

kangwang1988:
This will resolve the error below:
```dart
Resolving dependencies...
Git error. Command: git fetch
fatal: not a git repository (or any of the parent directories): .git
```

Including 
https://github.com/flutter/flutter/issues?q=is%3Aissue+%22not+a+git+repository%22+is%3Aclosed
https://github.com/flutter/flutter/issues/32556
https://github.com/flutter/flutter/issues/29853
https://github.com/flutter/flutter/issues/29310
https://github.com/flutter/flutter/issues/21495
etc

### DIFF
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -380,6 +380,8 @@ class BoundGitSource extends CachedSource {
     var path = _repoCachePath(ref);
     if (_updatedRepos.contains(path)) return;
 
+    _cleanInvalidGitRepoCache(path);
+
     if (!entryExists(path)) await _createRepoCache(ref);
 
     // Try to list the revision. If it doesn't exist, git will fail and we'll
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -397,7 +397,7 @@ class BoundGitSource extends CachedSource {
     var path = _repoCachePath(ref);
     if (_updatedRepos.contains(path)) return;
 
-    _checkIfValidGitRepo(Directory(path), delIfInvalid: true);
+    _cleanInvalidGitRepoCache(path);
 
     if (!entryExists(path)) {
       await _createRepoCache(ref);
@@ -415,8 +415,8 @@ class BoundGitSource extends CachedSource {
 
     try {
       await _clone(ref.description['url'], path, mirror: true);
-    } catch (error, _) {
-      _checkIfValidGitRepo(Directory(path), delIfInvalid: true);
+    } catch (_) {
+      _cleanInvalidGitRepoCache(path);
       rethrow;
     }
     _updatedRepos.add(path);
@@ -433,20 +433,18 @@ class BoundGitSource extends CachedSource {
     _updatedRepos.add(path);
   }
 
-  /// Check if [directory] is a valid git repo. If invalid and
-  /// [delIfInvalid] is set to be true, it will be deleted.
-  ///
-  bool _checkIfValidGitRepo(Directory directory, {bool delIfInvalid = false}) {
-    if (directory.existsSync()) {
-      var result = runProcessSync(git.command, ['log', '-1'],
+  ///  Clean-up [dirPath] if it's an invalid git repository
+  void _cleanInvalidGitRepoCache(String dirPath) {
+    final Directory directory = Directory(dirPath);
+    if (dirExists(dirPath)) {
+      var processResult = runProcessSync(
+          git.command, ['rev-parse', '--is-inside-git-dir'],
           workingDir: directory.path);
-      if (result.exitCode == 0) {
-        return true;
-      } else if (delIfInvalid == true) {
-        directory.deleteSync(recursive: true);
+      var result = processResult.stdout?.join('\n');
+      if (result != 'true') {
+        deleteEntry(dirPath);
       }
     }
-    return false;
   }
 
   /// Updates the package list file in [revisionCachePath] to include [path], if
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -397,6 +397,8 @@ class BoundGitSource extends CachedSource {
     var path = _repoCachePath(ref);
     if (_updatedRepos.contains(path)) return;
 
+    _checkIfValidGitRepo(Directory(path), delIfInvalid: true);
+
     if (!entryExists(path)) {
       await _createRepoCache(ref);
     } else {
@@ -411,7 +413,12 @@ class BoundGitSource extends CachedSource {
     var path = _repoCachePath(ref);
     assert(!_updatedRepos.contains(path));
 
-    await _clone(ref.description['url'], path, mirror: true);
+    try {
+      await _clone(ref.description['url'], path, mirror: true);
+    } catch (error, _) {
+      _checkIfValidGitRepo(Directory(path), delIfInvalid: true);
+      rethrow;
+    }
     _updatedRepos.add(path);
   }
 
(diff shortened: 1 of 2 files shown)


kangwang1988:
@kevmoo 
Can you review this?

lib/src/source/git.dart
```diff
@@ -426,6 +433,22 @@ class BoundGitSource extends CachedSource {
     _updatedRepos.add(path);
   }
 
+  /// Check if [directory] is a valid git repo. If invalid and
+  /// [delIfInvalid] is set to be true, it will be deleted.
+  ///
+  bool _checkIfValidGitRepo(Directory directory, {bool delIfInvalid = false}) {
+    if (directory.existsSync()) {
+      var result = runProcessSync(git.command, ['log', '-1'],
```
jonasfj: If we're just checking if this is a valid git repository why not use: `git rev-parse --is-inside-work-tree`

Is `git log` better?

I think `git rev-parse --is-inside-work-tree` was introduced in:
https://github.com/git/git/commit/892c41b98ae2e6baf3aa13901cb10db9ac67d2f3
Which shipped with git 1.5.3 in 2007, so it's probably fair to depend upon this today.

lib/src/source/git.dart
```diff
@@ -426,6 +433,22 @@ class BoundGitSource extends CachedSource {
     _updatedRepos.add(path);
   }
 
+  /// Check if [directory] is a valid git repo. If invalid and
+  /// [delIfInvalid] is set to be true, it will be deleted.
+  ///
+  bool _checkIfValidGitRepo(Directory directory, {bool delIfInvalid = false}) {
```
jonasfj: The return value is never used... And you only ever call it with `delIfInvalid: true`

How about renaming this to `_cleanInvalidGitRepository`?

jonasfj: Since all invocations pass a `String` wrapped as `Directory`, how about just passing a `String`?

And maybe using `dirExists` instead of `Directory.existsSync()`, I assume the `lib/src/io.dart` has `dirExists` defined because it's easier to use :)

lib/src/source/git.dart
```diff
@@ -426,6 +433,22 @@ class BoundGitSource extends CachedSource {
...
+    if (directory.existsSync()) {
+      var result = runProcessSync(git.command, ['log', '-1'],
+          workingDir: directory.path);
+      if (result.exitCode == 0) {
+        return true;
+      } else if (delIfInvalid == true) {
+        directory.deleteSync(recursive: true);
```
jonasfj: Let's use `deleteEntry` from `lib/src/io.dart`, I assume it exists in there for a reason..

lib/src/source/git.dart
```diff
@@ -411,7 +413,12 @@ class BoundGitSource extends CachedSource {
     var path = _repoCachePath(ref);
     assert(!_updatedRepos.contains(path));
 
-    await _clone(ref.description['url'], path, mirror: true);
+    try {
+      await _clone(ref.description['url'], path, mirror: true);
+    } catch (error, _) {
```
jonasfj: `} catch (_) {` should do

jonasfj: Maybe give this a one-line comment along the lines of: `// clean-up invalid git repositories after failed git clone`

jonasfj:
Also would it be possible to write a small test case for this? See examples in `test/get/git/...`

lib/src/source/git.dart
kangwang1988: '--is-inside-git-dir' would be the right one to check a git repo cache.
It's introduced in https://github.com/git/git/commit/c9bf7be238ac3c767f9ac6546d5bb7b11765253c 
on Jun 17, 2007

lib/src/source/git.dart
```diff
@@ -426,6 +433,20 @@ class BoundGitSource extends CachedSource {
     _updatedRepos.add(path);
   }
 
+  ///  Clean-up [dirPath] if it's an invalid git repository
+  void _cleanInvalidGitRepoCache(String dirPath) {
+    final Directory directory = Directory(dirPath);
```
jonasfj: What do you need this for anymore?

I'm just wondering if it makes if we pass `directory.path` in `runProcessSync`?

lib/src/source/git.dart
```diff
@@ -426,6 +433,20 @@ class BoundGitSource extends CachedSource {
...
+    if (dirExists(dirPath)) {
+      var processResult = runProcessSync(
+          git.command, ['rev-parse', '--is-inside-git-dir'],
+          workingDir: directory.path);
+      var result = processResult.stdout?.join('\n');
```
jonasfj: I would suggest reading the exit code instead, it feels more robust... If it's zero then it's a git repository.

kangwang1988:
@jonasfj 
Test case is added and verified.

jonasfj:
I'll need to study the test case a bit closer... But are we testing both the negative and positive case?

Ideally, I would like to make sure that we're not always purging the cache, while it would be safe it would also be wasteful, so we should have a test that ensures we don't.. In particularly because we don't want to accidentally have this happen next time someone hacks on this :)

kangwang1988:
@jonasfj 
What do mean by "purging the cache"?
Are you talking about the logic in clean_invalid_git_repo_cache_test.dart
```dart
    fooCacheDir.deleteSync(recursive: true);
    fooCacheDir.createSync();
```

jonasfj:
> What do mean by "purging the cache"?

I was thinking it would be worth checking if this would cause the git repository to always be deleted.
It doesn't because, this is already tested in: `doesnt_fetch_if_nothing_changes_test.dart`.

kangwang1988:
> > What do mean by "purging the cache"?
> 
> I was thinking it would be worth checking if this would cause the git repository to always be deleted.
> It doesn't because, this is already tested in: `doesnt_fetch_if_nothing_changes_test.dart` :)

Yes.
So please check if I need to do more changes, thanks.

test/get/git/clean_invalid_git_repo_cache_test.dart
```diff
@@ -0,0 +1,120 @@
...
+    ]).validate();
+
+    final String cacheDir =
+        path.join(d.sandbox, path.joinAll([cachePath, 'git', 'cache']));
+    final Directory fooCacheDir =
+        Directory(cacheDir).listSync().firstWhere((entity) {
+      if (entity is Directory &&
+          entity.path.split(Platform.pathSeparator).last.startsWith('foo'))
+        return true;
+    });
```
jonasfj: We can we move this into a helper function and write it as:
```dart
/// Invalidates a git clone in the pub-cache, by recreating it as empty-directory.
Future<void> _invalidateGitCache(String repo) async {
  final cacheDir = path.joinAll([d.sandbox, cachePath, 'git', 'cache']);
  final repoCache = Directory(cacheDir).list().firstWhere((e) {
    return e is Directory && path.basename(e.path).startsWith('foo');
  });
  await repoCache.delete(recursive: true);
  await repoCache.create();
}
```

Note:
The linter complains about `final String cacheDir = ...` for local variables, just wants us to use `final cacheDir = ...`.

jonasfj: hmm, maybe travis does run linters, we do have `omit_local_variable_types` in `analysis_options.yaml`.

I should probably review the travis setup for this repo...

kangwang1988: @jonasfj 
Issues resolved.

jonasfj:
Huge thanks for the work, I've filed https://dart-review.googlesource.com/c/sdk/+/108100
This should be part of the next dart-sdk.

PR #2166, opened 2019-07-08
Minor typo fixed in solver.md

buranmert:
## Summary
Wrong number in graph fixed

#### Description
Although there is no 6 in the graph, the doc reads:
> ...
> 5. root requires baz ^3.0.0.
> _6. So, because root depends on baz ^1.0.0,_
> 7. root isn't valid and version solving has failed.
> ...


buranmert:
I signed it!

Issue #2164, opened 2019-07-04
Test that pre-releases do not take precedence over stable releases


jonasfj:
Maybe I missed it, but I couldn't find any tests that if we have two versions of package foo `1.0.0` and `1.1.0-dev`, then `1.0.0` will be installed unless the other version is explicitly pinned.

We should have tests for this.

jakemac53:
There is actually no guarantee for this (I know because it has caused problems).

Pub only prefers stable release versions but it won't backtrack to try and find a solution without pre-release versions.

jakemac53:
To clarify, lets say you have the following pubspecs, where `a` and `b` are published on pub and `app` is the application package:

Latest `a` pubspec:
```yaml
name: a
version: 2.0.0
```

Pre-release `b` pubspec:
```yaml
name: b
version: 2.0.0-dev
dependencies:
  a: ^2.0.0
```


Stable `b` pubspec:
```yaml
name: b
version: 1.0.0
dependencies:
  a: ^1.0.0
```

`app` pubspec:
```yaml
name: app
version: 1.0.0
dependencies:
  a: '>=1.0.0 <3.0.0'
```

In this scenario, I believe that pub will always select version `2.0.0` of `a`, and `2.0.0-dev` of `b`, even though it could get the stable `1.0.0` of each of them.

This is because it first went to find a version of `a`, which chooses the stable `2.0.0` release. It then tries to select a version of `b` that is compatible, but only the pre-release version is, so it selects that. It does not backtrack to find a version solve with only stable releases. I believe that is purely for performance reasons.

I think this is also the case if there was a direct dependency on `b` from the `app` package.

jonasfj:
Hmm, that's super interesting..

I'm thinking I might be worth while exploring as I attempt to write up some documentation on how pre-releases can be used.

PR #2175, opened 2019-07-23
Test more pre-release behavior.

jonasfj:
Fixing https://github.com/dart-lang/pub/issues/2164

 * Cover common use-cases with caret notation.
 * Test that dependency on pre-release from stable release causes usage
   of the pre-release.
 * Test that direct dependency on stable release will not be overwritten
   by pre-release dependency from other stable dependency.
 * Test limitations of back-tracking for dependency on pre-releases.

-----
I'm not sure if the test case `back-tracks pre-release choice with direct-dependency` is actually stable. If this just an arbitrary result or if we have code that ensures this.

I suspect the behavior in the test case `back-tracking pre-release fails with indirect-dependency` is "not necessarily desired", it might be acceptable. Regardless we should certainly have test coverage, and a comment so we can remember why we made this compromise.

### DIFF
--- a/test/version_solver_test.dart
+++ b/test/version_solver_test.dart
@@ -1661,6 +1661,84 @@ void prerelease() {
     await d.appDir({'a': '<=2.0.0-dev'}).create();
     await expectResolves(result: {'a': '1.1.0'});
   });
+
+  test('use pre-release when desired', () async {
+    await servePackages((builder) {
+      builder.serve('a', '1.0.0');
+      builder.serve('a', '1.1.0-dev');
+    });
+
+    await d.appDir({'a': '^1.1.0-dev'}).create();
+    await expectResolves(result: {'a': '1.1.0-dev'});
+  });
+
+  test('can upgrade from pre-release', () async {
+    await servePackages((builder) {
+      builder.serve('a', '1.0.0');
+      builder.serve('a', '1.1.0-dev');
+      builder.serve('a', '1.1.0');
+    });
+
+    await d.appDir({'a': '^1.1.0-dev'}).create();
+    await expectResolves(result: {'a': '1.1.0'});
+  });
+
+  test('will use pre-release if depended on in stable release', () async {
+    // This behavior is desired because a stable package has dependency on a
+    // pre-release.
+    await servePackages((builder) {
+      builder.serve('a', '1.0.0', deps: {'b': '^1.0.0'});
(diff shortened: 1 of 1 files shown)


natebosch:
cc @munificent

test/version_solver_test.dart
```diff
@@ -1661,6 +1661,84 @@ void prerelease() {
...
+    });
+
+    await d.appDir({'a': '^1.0.0'}).create();
+    await expectResolves(result: {
+      'a': '1.1.0',
+      'b': '1.1.0-dev',
+    });
+  });
+
+  test('back-tracks pre-release choice with direct-dependency', () async {
```
munificent: Nit: "back-tracks" -> "backtracks".
"direct-dependency" -> "direct dependency".

Here and below.

test/version_solver_test.dart
```diff
@@ -1661,6 +1661,84 @@ void prerelease() {
...
+    }).create();
+    await expectResolves(result: {
+      'a': '1.0.0',
+      'b': '1.0.0',
+    });
+  });
+
+  test('back-tracking pre-release fails with indirect-dependency', () async {
+    // NOTE: This behavior is not necessarily desired. It is a performance
+    //       compromise, as back-tracking is expensive.
```
munificent: Is this actually a deliberate perf choice, or is it just that we didn't realize the solver would behave this way?

jonasfj: I heard @jakemac53 suggest it might be a performance choice.

It could also be unintended behavior.

I haven't looked at how complicated it would be to change the behavior. Regardless, it's probably fine to retrain the behavior for now, document it with a comment in the test case and then we can always review in the future. One would hope that being accidentally bumped to a pre-release rarely happens, and rarely has negative side-effects.

munificent: Yeah, I'm OK with the behavior as it is. I just don't want to leave a comment that claims a rationale unless we're sure that *is* the reason. Otherwise, it will just mislead future maintainers.

jakemac53: Ya, that this was a performance choice is just an assumption. It would certainly be more expensive to support backtracking and ranking previous possible results or something, but I don't know any actual measurements of the cost.

jonasfj:
@sethladd, @sigmundch, you two seems to have uploader permissions on pub.dev/packages/crypto, would either if you like to merge and publish this PR?

PR #2198, opened 2019-09-02
return the hashCode of the description

JakeChampion:
From reading through the code I think `_parseDescription.hashCode` was a possible typo and it was actually meant to be `description.hashCode`.

Should there be a corresponding test for this method?

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -95,7 +95,7 @@ class HostedSource extends Source {
   bool descriptionsEqual(description1, description2) =>
       _parseDescription(description1) == _parseDescription(description2);
 
-  int hashDescription(description) => description.hashCode;
+  int hashDescription(description) => _parseDescription(description).hashCode;
 
   /// Ensures that [description] is a valid hosted package description.
   ///
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -95,7 +95,7 @@ class HostedSource extends Source {
   bool descriptionsEqual(description1, description2) =>
       _parseDescription(description1) == _parseDescription(description2);
 
-  int hashDescription(description) => _parseDescription.hashCode;
+  int hashDescription(description) => description.hashCode;
 
   /// Ensures that [description] is a valid hosted package description.
   ///


jonasfj:
I'm not quite sure this is right... it doesn't seem that it makes tests pass...

From looking at `_parseDescription` it seems this is more a normalization method... which is probably also why `descriptionsEqual` uses it.

See `lib/src/source.dart`:
```dart
  /// Returns a hash code for [description].
  ///
  /// Descriptions that compare equal using [descriptionsEqual] should return
  /// the same hash code.
  int hashDescription(description);
```
It's clear that `hashDescription` relates to the behavior of `descriptionsEqual`, rather than the actual description.

-----

My guess is that multiple descriptions that are different can be the same... e.g. `{"name", "xxx"}` and `Pair<String,String>("xxx", "<defaultUrl>")`.

jonasfj:
If you can see something that this fixes please reopen.. From what I can see this does break tests.

JakeChampion:
I believe the code `int hashDescription(description) => _parseDescription.hashCode;` does not use `description` and instead returns the same `int` for every single invocation of `hashDescription`. If that's the case, then I really do think this is a bug. I'm not sure why the tests break, are there any tests which specifically test ` hashDescription` in isolation? I couldn't find any when I looked.

Having a look at every single other override of `hashDescription` we can see that they used the `hashCode` of `description` or of a derivation of `description`:
https://github.com/dart-lang/pub/blob/5962908a66e814dd78f6856a4e5f792c6db65fbc/lib/src/source/path.dart#L52-L53
https://github.com/dart-lang/pub/blob/fe288706b9d67f676b8b7b45923a099130a92181/test/lock_file_test.dart#L34
https://github.com/dart-lang/pub/blob/5962908a66e814dd78f6856a4e5f792c6db65fbc/lib/src/source/sdk.dart#L54
https://github.com/dart-lang/pub/blob/5962908a66e814dd78f6856a4e5f792c6db65fbc/lib/src/source/unknown.dart#L36
https://github.com/dart-lang/pub/blob/b6d9152a7303805dedc49f93b11dde728f0e9f3d/lib/src/source/hosted.dart#L98
https://github.com/dart-lang/pub/blob/4bfdc9b626accf10372334d422741c8d1cf8900b/test/pubspec_test.dart#L32
https://github.com/dart-lang/pub/blob/df0f72daaa724e29ed6075e0fb5549a6d6dc5daf/lib/src/source/git.dart#L174-L181

jonasfj:
I think you're onto something here :)

It should probably be:
```dart
int hashDescription(description) => _parseDescription(description).hashCode;
```

Given that `HostedSource.descriptionsEqual(Pair<String,String>('<name>', defaultUrl), {'name': '<name>'})` should return `true`, we probably have to do `_parseDescription(description).hashCode`, rather than just `description.hashCode`.

-------

Technically, `_parseDescription.hashCode` is a correct implementation because the only requirement is that two instances with the same value returns the hash code :see_no_evil: 

But now I wonder of `Pair.hashCode` is implemented, will it do the right thing.

lib/src/source/hosted.dart
```diff
@@ -95,7 +95,7 @@ class HostedSource extends Source {
   bool descriptionsEqual(description1, description2) =>
       _parseDescription(description1) == _parseDescription(description2);
 
-  int hashDescription(description) => _parseDescription.hashCode;
+  int hashDescription(description) => description.hashCode;
```
jonasfj: This probably has to be `_parseDescription(description).hashCode`, then I think we're on right track.

I wonder if this will yield any performance improvements :)

JakeChampion:
@jonasfj updated the implementation 👌

jonasfj:
Sweet, let's see what the tests does :)

jonasfj:
So I'm not sure this has much impact, since it's only called from `PackageName`:
https://github.com/dart-lang/pub/blob/92b52682e8fc6eed9ef2e77ed890647f75570165/lib/src/package_name.dart#L75-L80


The fact that the `name` of the package is in the `hashCode` probably reduces the number of possible collisions dramatically. In fact, I'm starting to think this whole `hashCode` thing might be overkill, since few projects have thousands of packages. But it's nice to have to ability to scale :)

jonasfj:
Test seems to pass, so I've merged this -- huge credits for spotting this :)

PR #2199, opened 2019-09-03
Don't mention 'transformed dependencies'.

jonasfj:
I think this documentation relates to when Dart 1.x when pub did more than
fetching dependencies and pre-compiling executables.

-----
Doing a bit of archaeology I find that it was introduced in c1405b945c6d818c8cfe78334e8d4b11fd913103 along with a similar phrase in documentation comment for `EntryPoint.acquireDependencies`, but this comment was removed by @natebosch in 5ca86eb6e7be45ade4e44cbb7402d296cdf8b32f which removed other Dart 1.x related features.

So I'm inclined to think this had to do with the code-transformer features in Dart 1.x.

### DIFF
--- a/lib/src/command/get.dart
+++ b/lib/src/command/get.dart
@@ -28,7 +28,7 @@ class GetCommand extends PubCommand {
 
     argParser.addFlag('precompile',
         defaultsTo: true,
-        help: "Precompile executables and transformed dependencies.");
+        help: "Precompile executables in immediate dependencies.");
 
     argParser.addFlag('packages-dir', negatable: true, hide: true);
   }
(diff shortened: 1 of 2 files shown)


lib/src/command/get.dart
```diff
@@ -27,8 +27,7 @@ class GetCommand extends PubCommand {
         help: "Report what dependencies would change but don't change any.");
 
     argParser.addFlag('precompile',
-        defaultsTo: true,
-        help: "Precompile executables and transformed dependencies.");
+        defaultsTo: true, help: "Precompile executables.");
```
natebosch: I think this would be fine - it might also be worthwhile to be a little more precise?

"Precompile executables in immediate dependencies."

natebosch:
> So I'm inclined to think this had to do with the code-transformer features in Dart 1.x.

Yes, I think you are correct.

PR #2203, opened 2019-09-06
Rephrase warnings for missing deps

### LINKED ISSUE #2204: The error message on pub publish for missing dependencies is confusing
For example, if you omit a pub dependency on package:meta you get the following error which made it sound like the import is unused.
```
* line 12, column 1 of lib/src/server.dart: This package doesn't depend on meta.
     ╷
  12 │ import 'package:meta/meta.dart';
     │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     ╵
```
A clearer error message could be:
```
* line 12, column 1 of lib/src/server.dart: This package uses package:meta but does not list it as a dependency in its pubspec.
     ╷
  12 │ import 'package:meta/meta.dart';
     │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     ╵
```
As a bonus, it would be nice if there was a lint that warned about this issue so that running the analyzer on a package could catch these issues before it is time to publish a package.

natebosch:
Fixes #2204

The phrase "doesn't depend on" can be interpreted as "doesn't use" and
considered an unnecessary import. Rephrase to be more clear that the
message is discussing the dependencies in pubspec.yaml.

### DIFF
--- a/lib/src/validator/strict_dependencies.dart
+++ b/lib/src/validator/strict_dependencies.dart
@@ -148,13 +148,15 @@ class _Usage {
 
   /// Returns an error message saying the package is not listed in dependencies.
   String dependencyMissingMessage() =>
-      _toMessage("This package doesn't depend on $package.");
+      _toMessage('This package does nt have $package in the `dependencies` '
+          'section of `pubspec.yaml`.');
 
   /// Returns an error message saying the package should be in `dependencies`.
   String dependencyMisplaceMessage() {
     var shortFile = p.split(p.relative(_file)).first;
     return _toMessage(
-        '$package is a dev dependency. Packages used in $shortFile/ must be '
-        'declared as normal dependencies.');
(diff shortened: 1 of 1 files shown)


lib/src/validator/strict_dependencies.dart
```diff
@@ -148,13 +148,15 @@ class _Usage {
...
+      _toMessage('This package does not have $package in the `dependencies` '
+          'section of `pubspec.yaml`.');
 
   /// Returns an error message saying the package should be in `dependencies`.
   String dependencyMisplaceMessage() {
     var shortFile = p.split(p.relative(_file)).first;
     return _toMessage(
-        '$package is a dev dependency. Packages used in $shortFile/ must be '
-        'declared as normal dependencies.');
+        '$package is in the `dev_dependencies` section of `pubsepc.yaml`. '
```
jonasfj: s/pubsepc.yaml/pubspec.yaml/

PR #2216, opened 2019-09-27
Remove unused Map `availableVersions`

JakeChampion:
I couldn't see this variable being used anywhere in the method, which makes me think it is safe to be removed :-)


JakeChampion:
One of the failures looks to be perhaps a bug in the formatter?
https://github.com/dart-lang/dart_style/issues/859

jakemac53:
Yes that is a known bug in the formatter which has been fixed but I think not yet released in a new sdk.

cc @munificent

PR #2228, opened 2019-10-20
Dart package option

jonasfj:
@sigurdm, I took your PR a bit further and fixed the case where we weren't checking if path dependencies had updated their lower bound SDK constraint (language version).

I also did some utility for ease of parsing/writing of package config and aligned it with a newer version of the specification..

We need to actually do a test case, I haven't gotten that far... I'm not sure to what extend we should make the test case magic. But we can look at this next week :)

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -9,6 +9,7 @@ import 'dart:io';
 import 'package:collection/collection.dart';
 import 'package:package_config/packages_file.dart' as packages_file;
 import 'package:path/path.dart' as p;
+import 'package:pub/src/package_config.dart';
 import 'package:pub_semver/pub_semver.dart';
 
 import 'dart.dart' as dart;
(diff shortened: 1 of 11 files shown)


lib/src/entrypoint.dart
```diff
@@ -500,50 +530,135 @@ class Entrypoint {
...
+  /// Determines whether or not the `.dart_tool/package_config.json` file is
+  /// out of date with respect to the lockfile.
+  ///
+  /// This will be `false` if the [lockfile] contains dependencies that are not
+  /// in the `.dart_tool/package_config.json` or that don't match what's in
+  /// there.
+  ///
+  /// Throws [DataException], if `.dart_tool/package_config.json` is not
+  /// up-to-date for some other reason.
+  bool _isPackageConfigUpToDate() {
```
sigurdm: We should probably have return type `void` and just throw where we now return `false`

jonasfj:
Closed in favor of #2238.

PR #2233, opened 2019-10-22
Warn about old flutter plugin registration format

jonasfj:
@amirh, does this look sane to you?

I'll write up test if you're happy with this, but is this the tests we want to do and is the wording of the messages reasonable to you?

### DIFF
--- a/lib/src/validator/flutter_plugin.dart
+++ b/lib/src/validator/flutter_plugin.dart
@@ -49,8 +49,8 @@ class FlutterPluginValidator extends Validator {
     // SDKs older than 1.10.0, then this is going to be a problem.
     if (usesNewPluginFormat &&
         pubspec.sdkConstraints['flutter'] != null &&
-        pubspec.sdkConstraints['flutter'].allows(Version.parse('1.9.0'))) {
-      errors.add('pubspec.yaml allows Flutter SDK version 1.9.0, which does '
+        pubspec.sdkConstraints['flutter'].allows(Version.parse('1.9.999'))) {
+      errors.add('pubspec.yaml allows Flutter SDK version 1.9.x, which does '
           'not support the flutter.plugin.platforms key.\n'
           'Please consider increasing the Flutter SDK requirement to '
           '^1.10.0 (environment.sdk.flutter)');
--- a/lib/src/validator/flutter_plugin.dart
+++ b/lib/src/validator/flutter_plugin.dart
@@ -1,0 +1,65 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:pub_semver/pub_semver.dart';
+
+import '../entrypoint.dart';
+import '../validator.dart';
+
+/// Validates that Flutter plugins doesn't use both new and old plugin format.
+///
+/// Warns if using the old plugin registration format.
+class FlutterPluginValidator extends Validator {
+  FlutterPluginValidator(Entrypoint entrypoint) : super(entrypoint);
+
+  Future validate() async {
+    final pubspec = entrypoint.root.pubspec;
+    // Ignore all packages that do not have the `flutter.plugin` property.
+    if (pubspec.fields['flutter'] is! Map ||
+        pubspec.fields['flutter']['plugin'] is! Map) {
+      return;
+    }
+    final plugin = pubspec.fields['flutter']['plugin'] as Map;
+
+    // Determine if this uses the old format by checking if `flutter.plugin`
+    // contains any of the following keys.
+    final usesOldPluginRegistrationFormat = {
+      'androidPackage',
+      'iosPrefix',
+      'pluginClass',
+    }.any(plugin.containsKey);
+
+    // Determine if this uses the new format by check if the:
+    // `flutter.plugin.platforms` keys is defined.
+    final usesNewPluginRegistrationFormat = plugin['platforms'] != null;
+
+    // Report an error, if both the new and the old format is used.
+    if (usesOldPluginRegistrationFormat && usesNewPluginRegistrationFormat) {
+      errors.add('In pubspec.yaml the flutter.plugin.platforms plugin '
+          'registration format, cannot be used in combination with the old '
+          'flutter.plugin.{androidPackage,iosPrefix,pluginClass} '
+          'plugin registration format.');
+    }
+
+    // If the new plugin format is used, and the flutter SDK dependency allows
+    // SDKs older than 1.12, then this is going to be a problem.
+    if (usesNewPluginRegistrationFormat &&
+        !pubspec.sdkConstraints['flutter'].allows(Version.parse('1.11'))) {
+      warnings.add('pubspec.yaml allows Flutter SDK version 1.11, which does '
+          'not support the flutter.plugin.platforms plugin registration '
+          'format. Please consider increasing the Flutter SDK requirement to '
+          '^1.12 (environment.sdk.flutter)');
+    }
+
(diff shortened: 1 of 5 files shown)


lib/src/validator/flutter_plugin.dart
```diff
@@ -0,0 +1,65 @@
...
+    // Report an error, if both the new and the old format is used.
+    if (usesOldPluginRegistrationFormat && usesNewPluginRegistrationFormat) {
+      errors.add('In pubspec.yaml the flutter.plugin.platforms plugin '
+          'registration format, cannot be used in combination with the old '
+          'flutter.plugin.{androidPackage,iosPrefix,pluginClass} '
+          'plugin registration format.');
+    }
+
+    // If the new plugin format is used, and the flutter SDK dependency allows
+    // SDKs older than 1.12, then this is going to be a problem.
```
amirh: The minimal sdk should be 1.10 for the new schema I believe

lib/src/validator/flutter_plugin.dart
```diff
@@ -0,0 +1,65 @@
...
+      warnings.add('pubspec.yaml allows Flutter SDK version 1.11, which does '
+          'not support the flutter.plugin.platforms plugin registration '
+          'format. Please consider increasing the Flutter SDK requirement to '
+          '^1.12 (environment.sdk.flutter)');
+    }
+
+    // Warn if using the old plugin registration format.
+    if (usesOldPluginRegistrationFormat) {
+      warnings.add('The flutter.plugin.{androidPackage,iosPrefix,pluginClass} '
+          'plugin registration format is deprecated. Consider using the '
```
amirh: nit: the term "plugin registration" may be a bit confusing here (I'd think it is talking about the generated plugin registrant), maybe saying: the {androidPackage,iosPrefix,pluginClass} keys are deprecated. Consider using the flutter.plugin.platforms key

lib/src/validator/flutter_plugin.dart
```diff
@@ -0,0 +1,65 @@
...
+          'flutter.plugin.platforms plugin registration format introduced in '
```
amirh: We should probably add a go-link which have some explanation. TODO(amirh)

mit-mit: Should probably be mentioned here? https://flutter.dev/docs/development/packages-and-plugins/developing-packages

jonasfj:
@amirh, would you mind taking another look, I just:
 * Report error if using the `flutter.plugin.platforms` key along with the `flutter.plugin.{androidPackage,iosPrefix,pluginClass}` keys.
 * Report error if using the `flutter.plugin.platforms` key with Flutter SDK constraint that allows `1.9.x`. 
 * Added a few test cases.

Please confirm that:
 A) Flutter 1.9.x will be the last series to not support the new format, and,
 B) Flutter 1.10.0 will support the new platform format.
Is this correct? (It would be unfortunate for users if we hardcoded in the wrong constraints).

lib/src/validator/flutter_plugin.dart
jonasfj: @amirh, if you would like us to put a link into these warnings, please determine what that should be :)

lib/src/validator/flutter_plugin.dart
```diff
@@ -0,0 +1,67 @@
...
+
+import 'dart:async';
+
+import 'package:pub_semver/pub_semver.dart';
+
+import '../entrypoint.dart';
+import '../validator.dart';
+
+/// Validates that Flutter plugins doesn't use both new and old plugin format.
+///
```
sigurdm: Add a TODO to point to a document describing the formats when it becomes available.

lib/src/validator/flutter_plugin.dart
```diff
@@ -0,0 +1,67 @@
...
+/// Warns if using the old plugin registration format.
+class FlutterPluginValidator extends Validator {
```
sigurdm: Perhaps: `FlutterPluginFormatValidator`

lib/src/validator/flutter_plugin.dart
```diff
@@ -0,0 +1,67 @@
...
+  FlutterPluginValidator(Entrypoint entrypoint) : super(entrypoint);
+
+  Future validate() async {
+    final pubspec = entrypoint.root.pubspec;
+    // Ignore all packages that do not have the `flutter.plugin` property, or
+    // which does not have an SDK constraint on Flutter.
```
sigurdm: does -> do

lib/src/validator/flutter_plugin.dart
```diff
@@ -0,0 +1,67 @@
...
+    if (pubspec.fields['flutter'] is! Map ||
+        pubspec.fields['flutter']['plugin'] is! Map ||
+        pubspec.sdkConstraints['flutter'] == null) {
+      return;
+    }
+    final plugin = pubspec.fields['flutter']['plugin'] as Map;
+
+    // Determine if this uses the old format by checking if `flutter.plugin`
+    // contains any of the following keys.
+    final usesOldPluginFormat = {
```
sigurdm: Consider making the set const.

lib/src/validator/flutter_plugin.dart
```diff
@@ -0,0 +1,67 @@
...
+      errors.add('In pubspec.yaml the flutter.plugin.platforms key cannot be '
+          'used in combination with the old '
+          'flutter.plugin.{androidPackage,iosPrefix,pluginClass} keys.');
+    }
+
+    // If the new plugin format is used, and the flutter SDK dependency allows
+    // SDKs older than 1.10.0, then this is going to be a problem.
+    if (usesNewPluginFormat &&
+        pubspec.sdkConstraints['flutter'] != null &&
+        pubspec.sdkConstraints['flutter'].allows(Version.parse('1.9.999'))) {
```
sigurdm: I think you should test for empty intersection with a version constraint <1.10.0-0

lib/src/validator/flutter_plugin.dart
```diff
@@ -0,0 +1,67 @@
...
+    // Determine if this uses the new format by check if the:
+    // `flutter.plugin.platforms` keys is defined.
+    final usesNewPluginFormat = plugin['platforms'] != null;
+
+    // Report an error, if both the new and the old format is used.
+    if (usesOldPluginFormat && usesNewPluginFormat) {
+      errors.add('In pubspec.yaml the flutter.plugin.platforms key cannot be '
+          'used in combination with the old '
+          'flutter.plugin.{androidPackage,iosPrefix,pluginClass} keys.');
+    }
```
sigurdm: Consider early return

lib/src/validator/flutter_plugin.dart
```diff
@@ -0,0 +1,67 @@
...
+    if (usesNewPluginFormat &&
+        pubspec.sdkConstraints['flutter'] != null &&
+        pubspec.sdkConstraints['flutter'].allows(Version.parse('1.9.999'))) {
+      errors.add('pubspec.yaml allows Flutter SDK version 1.9.x, which does '
+          'not support the flutter.plugin.platforms key.\n'
+          'Please consider increasing the Flutter SDK requirement to '
+          '^1.10.0 (environment.sdk.flutter)');
+    }
+
+    // Warn if using the old plugin format.
```
sigurdm: Comment seems redundant

test/validator/flutter_plugin_format_test.dart
```diff
@@ -0,0 +1,187 @@
...
+            },
+          },
+        },
+      };
+      await d.dir(appPath, [d.pubspec(pkg)]).create();
+      expectNoValidationError(flutterPluginFormat);
+    });
+  });
+
+  group('should consider a package invalid if it', () {
```
sigurdm: Nit: maybe split this group between warning and error cases.

PR #2237, opened 2019-10-24
Bumped YAML 

jonasfj:


tool/test.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -e
```
sigurdm: This has its own PR

sigurdm:
CC @nex3. The Yaml package started using insertion order for the maps, resulting in different resolution error messages. As far as I can see the changes are all benign. ?

jonasfj:
@nex3, I'm going to land this, and fix travis :)

But if the analysis of the ordering I did in the commit message for this PR isn't accurate, we'd be very interested to hear any concerns. As I understand it there's risk that the solution will be different, but this already the case as the package graph on pub.dev isn't stable.

nex3:
FYI the Yaml spec [explicitly says](https://yaml.org/spec/1.2/spec.html#id2763452):

> [...] while imposing a order on mapping keys is necessary for flattening YAML representations to a sequential access medium, this serialization detail must not be used to convey application level information.

If the YAML mapping entry order is affecting pub in any way, it's violating that principle; it should really sort those keys alphabetically or something to provide its own stable ordering.

Issue #2242, opened 2019-10-31
flutter packages get repeatedly tries to fetch packages if a packages in pubspec does not exist


escamoteur:
Hi,

just stumbled upon this here when trying to add a new package to the project and made a typo. Instead of just getting the message that the package doesn't exist I got this here

```
Running "flutter pub get" in compass_test...                           -Batchvorgang abbrechen (J/N)? j
PS C:\Entwicklung\projects\Aaronia\trunk\Sources\RTSAMobileDev\tests\compass_test> flutter packages get -v
[  +19 ms] executing: [C:\Entwicklung\flutter\] git -c log.showSignature=false log -n 1 --pretty=format:%H
[  +40 ms] Exit code 0 from: git -c log.showSignature=false log -n 1 --pretty=format:%H
[        ] 8e0799a657051bf540d98b8aecf9cfa1325f3f71
[        ] executing: [C:\Entwicklung\flutter\] git describe --match v*.*.* --first-parent --long --tags
[  +48 ms] Exit code 0 from: git describe --match v*.*.* --first-parent --long --tags
[        ] v1.10.14-351-g8e0799a65
[   +5 ms] executing: [C:\Entwicklung\flutter\] git rev-parse --abbrev-ref --symbolic @{u}
[  +26 ms] Exit code 0 from: git rev-parse --abbrev-ref --symbolic @{u}
[        ] origin/master
[        ] executing: [C:\Entwicklung\flutter\] git ls-remote --get-url origin
[  +26 ms] Exit code 0 from: git ls-remote --get-url origin
[   +3 ms] https://github.com/flutter/flutter
[  +50 ms] executing: [C:\Entwicklung\flutter\] git rev-parse --abbrev-ref HEAD
[  +24 ms] Exit code 0 from: git rev-parse --abbrev-ref HEAD
[        ] master
[  +31 ms] Artifact Instance of 'AndroidMavenArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'AndroidGenSnapshotArtifacts' is not required, skipping update.  
[        ] Artifact Instance of 'AndroidInternalBuildArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'IOSEngineArtifacts' is not required, skipping update.
[        ] Artifact Instance of 'FlutterWebSdk' is not required, skipping update.
[   +3 ms] Artifact Instance of 'WindowsEngineArtifacts' is not required, skipping update.       
[...]

PR #2254, opened 2019-11-11
Retry on all SocketExceptions

jonasfj:
Do not retry on OS specific error codes for `SocketException`.
These OS error codes are not documented in the code, and they
may change behavior depending on OS and architecture.

Furthermore, many seemingly persistent errors such as DNS lookup
failures should be retried.

Finally, if we retry a `SocketException` unnecessarily this is of
little impact. It means that if the user doesn't have a network
connection, or tries to use a `PUB_HOSTED_URL` pointing to a
non-existent pub server, then it will take a bit longer before
the user is informed of the problem.

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -63,32 +63,7 @@ class _PubHttpClient extends http.BaseClient {
     request.headers[HttpHeaders.userAgentHeader] = 'Dart pub ${sdk.version}';
     _logRequest(request);
 
-    http.StreamedResponse streamedResponse;
-    try {
-      streamedResponse = await _inner.send(request);
-    } on SocketException catch (error, stackTraceOrNull) {
-      // Work around issue 23008.
-      var stackTrace = stackTraceOrNull ?? Chain.current();
-
-      if (error.osError == null) rethrow;
-
-      if (error.osError.errorCode == 8 ||
-          error.osError.errorCode == -2 ||
-          error.osError.errorCode == -5 ||
-          error.osError.errorCode == 11001 ||
-          error.osError.errorCode == 11004) {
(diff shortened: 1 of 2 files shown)


lib/src/http.dart
```diff
@@ -63,32 +63,7 @@ class _PubHttpClient extends http.BaseClient {
...
-      // Work around issue 23008.
-      var stackTrace = stackTraceOrNull ?? Chain.current();
-
-      if (error.osError == null) rethrow;
-
-      if (error.osError.errorCode == 8 ||
-          error.osError.errorCode == -2 ||
-          error.osError.errorCode == -5 ||
-          error.osError.errorCode == 11001 ||
-          error.osError.errorCode == 11004) {
```
jonasfj: @nex3, you wouldn't happen to know where the error codes come from? and why we don't retry them..

sigurdm:
Might help on https://github.com/dart-lang/pub/issues/2242

jonasfj:
Hmm, test broken because the error message isn't clean. I think we should:
 * Retry the errors, and,
 * Print a nice error message when this error persists.

Notably, if we can't resolve the URL we should retry. I know that DNS resolvers aren't always reliable, not even in clouds like EC2.

lib/src/http.dart
nex3: I don't remember for sure, but I think they came up from manual testing on what happened when we tried to request obviously bogus or non-existent URLs on various OSes.

jonasfj: Thanks, that makes sense... I suspect we can make slight reliability improvements by retrying..

But th tests do reveal that straight up ignoring them produce worse error messages :)

PR #2270, opened 2019-12-02
Fix lints

jonasfj:
Fixed lints, also fixed a possible bug..

Note: Maybe we shouldn't merge this before cherry-picking to flutter.

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -298,15 +298,19 @@ class BoundHostedSource extends CachedSource {
     var cacheDir = p.join(systemCacheRoot, _urlToDirectory(source.defaultUrl));
     if (!dirExists(cacheDir)) return [];
 
-    return listDir(cacheDir).map((entry) {
-      try {
-        return Package.load(null, entry, systemCache.sources);
-      } catch (error, stackTrace) {
-        log.fine("Failed to load package from $entry:\n"
-            "$error\n"
-            "${Chain.forTrace(stackTrace)}");
-      }
-    }).toList();
+    return listDir(cacheDir)
+        .map((entry) {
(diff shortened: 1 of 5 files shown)


lib/src/source/hosted.dart
```diff
@@ -298,15 +298,19 @@ class BoundHostedSource extends CachedSource {
...
+          try {
+            return Package.load(null, entry, systemCache.sources);
+          } catch (error, stackTrace) {
+            log.fine("Failed to load package from $entry:\n"
+                "$error\n"
+                "${Chain.forTrace(stackTrace)}");
+            return null;
+          }
+        })
+        .where((e) => e != null)
```
jonasfj: I think this fixed a real bug, in cases where loading package from the file-system fails.

This is only called in `pub cache list`, so I don't think it's very important, but still nice :)

test/io_test.dart
```diff
@@ -347,7 +347,7 @@ void testExistencePredicate(String name, bool predicate(String path),
 }
 
 /// Like [withTempDir], but canonicalizes the path before passing it to [fn].
-Future _withCanonicalTempDir(Future fn(String path)) =>
+Future _withCanonicalTempDir(FutureOr fn(String path)) =>
```
sigurdm: Now we're here. Could we fill in the generic type?

lib/src/source/hosted.dart
sigurdm: Yes filtering out those nulls seems to be the right thing

natebosch: It would be better to split behavior changes and hint fixes in separate commits.

jonasfj:
Closing in favor of #2290.

PR #2273, opened 2019-12-05
Resolve API URL's relative to the base URL, consistently

thosakwe:
Hello! 

This pull request makes the way that API URL's resolved, consistent across all commands. This was originally part of https://github.com/dart-lang/pub/pull/2167, but has now been updated, and with unit tests added.

If this gets merged, then next I'll get working on the pull request to implement the rest of the features discussed in #2167 and #1381 (storing credentials for 3rd-party servers in `$PUB_CACHE/tokens.json`).

Thanks!!

## Changes to Existing Functionality
* Resolve the URLs for `api/versions/new`, `api/packages/:package/uploaders`, and `api/packages/:package/uploaders/:uploader` relative to the `$PUB_HOSTED_URL`. In my testing, I noticed that calls to `pub get`, `pub upgrade`, `pub global activate`, (basically any command that involved a download) worked fine when the `$PUB_HOSTED_URL` included a path, because their implementations append the desired URI to `$PUB_HOSTED_URL` (ex. `https://foo.com/my/path` -> `https://foo.com/my/path/api/packages/string_scanner`). However, since `pub uploaders` and `pub lish` were using `Uri.resolve` with absolute URIs, whatever path was in `$PUB_HOSTED_URL` would be obliterated.

### DIFF
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:http/http.dart' as http;
+import 'package:path/path.dart' as p;
 
 import '../ascii_tree.dart' as tree;
 import '../command.dart';
@@ -67,7 +68,8 @@ class LishCommand extends PubCommand {
         return log.progress('Uploading', () async {
           // TODO(nweiz): Cloud Storage can provide an XML-formatted error. We
           // should report that error and exit.
-          var newUri = server.resolve("api/packages/versions/new");
+          var newUri = server.replace(
+              path: p.join(server.path, 'api/packages/versions/new'));
           var response = await client.get(newUri, headers: pubApiHeaders);
           var parameters = parseJsonResponse(response);
 
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -67,7 +67,7 @@ class LishCommand extends PubCommand {
         return log.progress('Uploading', () async {
           // TODO(nweiz): Cloud Storage can provide an XML-formatted error. We
           // should report that error and exit.
-          var newUri = server.resolve("/api/packages/versions/new");
+          var newUri = server.resolve("api/packages/versions/new");
           var response = await client.get(newUri, headers: pubApiHeaders);
           var parameters = parseJsonResponse(response);
 
--- a/test/lish/uses_relative_resolution.dart
+++ b/test/lish/uses_relative_resolution.dart
@@ -4,6 +4,7 @@
 
 import 'dart:convert';
 
+import 'package:path/path.dart' as p;
 import 'package:shelf/shelf.dart' as shelf;
 import 'package:shelf_test_handler/shelf_test_handler.dart';
 import 'package:test/test.dart';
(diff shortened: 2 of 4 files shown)


test/lish/uses_relative_resolution.dart
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
sigurdm: I think the file name has to end with `_test.dart` for the framework to pick it up.

lib/src/command/lish.dart
```diff
@@ -67,7 +68,8 @@ class LishCommand extends PubCommand {
         return log.progress('Uploading', () async {
           // TODO(nweiz): Cloud Storage can provide an XML-formatted error. We
           // should report that error and exit.
-          var newUri = server.resolve("/api/packages/versions/new");
+          var newUri = server.replace(
```
sigurdm: Would this work?:
```
var newUri = server.resolve("api/packages/versions/new");
```

test/lish/uses_relative_resolution.dart
thosakwe: Ah, yes. I had been running `pub run test <filename>` directly, and didn’t notice. I’ll fix that.

lib/src/command/lish.dart
thosakwe: If your base URL is `example.com/sub/dir` then you end up with `example.com/sub/api/packages/versions/new`, I guess because `api` is technically a “file” path and not a “directory” path component. I *didn’t* test it with `example.com/sub/dir/`, though, but also, I think that if users were required to include the trailing slash in `$PUB_HOSTED_URL`, many people would be confused.

test/lish/uses_relative_resolution.dart
thosakwe: Also, it looks like there’s a failing test or two in the last commit that I didn’t see before, so I’ll have to check that out.

lib/src/command/lish.dart
sigurdm: >I think that if users were required to include the trailing slash in $PUB_HOSTED_URL, many people would be confused.

I agree

We have another strategy at: https://github.com/dart-lang/pub/blob/master/lib/src/source/hosted.dart#L316 , just doing sting concatenation. It somehow 'feels' wrong, but seems to work.

I would prefer using just Uri methods, and not package:path.

@jonasfj do you have any ideas?

jonasfj: @sigurdm, I think not using `package:path` is nice.. but maybe we should make it all consistent and just land it?

sigurdm:
It seems a number of tests are broken @thosakwe could you investigate?

mit-mit:
@thosakwe are you still working on this?

MisterJimson:
Seems like this may be related to #2551

thosakwe:
Sorry, I completely forgot this was open. It'll have to wait until this weekend, but I should be able to fix the broken tests.

PR #2274, opened 2019-12-06
Use separate credentials for each server

sigurdm:


sigurdm:
This is mainly meant as a suggestion - we could also consolidate everything in one file, and upgrade the current file format when encountered.

mit-mit:
Is this still being considered?

sigurdm:
No I think we need a better solution.

Issue #2147, opened 2019-06-11
fetching packages is slow, versions are being listed sequentially


jonasfj:
When `pub get` runs it fetches the list of versions for each package it needs.
When fetching the list of versions available for a given package, the constraint solving is paused, this all such requests are made sequentially.

We should be able to do this concurrently, at-least for all the dependencies within a given pubspec.

This would increase performance significantly.

jonasfj:
We fixed this in https://github.com/dart-lang/pub/pull/2280

PR #2280, opened 2019-12-16
Parallel fetching of available versions

sigurdm:
Implements rate-limited parallel fetching of version information when solving.
Also does speculative pre-fetching of version information of dependencies of the newest versions of each package.

In informal benchmarks of `pub get` for a package whose only direct dependency is package:test the resolution time goes from 7.8 seconds to 1.5 seconds.

Also adds package:pedantic to our runtime dependencies


sigurdm:
This messes up the zones. That needs to be fixed to land this.

sigurdm:
the zones should work correctly now

jonasfj:
Fixes #2147 \o/

doc/repository-spec-v2.md
```diff
@@ -64,6 +64,9 @@ repository.
 
 ### Inspect a specific version of a package
```
jonasfj: Let's add `(deprecated)` to the section title

doc/repository-spec-v2.md
```diff
@@ -64,6 +64,9 @@ repository.
...
 
+**Deprecated** Use `<PUB_HOSTED_URL>/api/packages/<PACKAGE>` instead and find
```
jonasfj: I think it's enough to say:
```md
**Deprecated** as of Dart 2.8, use "List all versions of a package" instead.
```
In particular, we'll want to know what Dart version we stopped using this in. Such that when we decide to remove the end-point we know how many Dart versions will be affected.

I assume it'll be years before we remove the end-point, and we might have forgotten in the mean time.

lib/src/solver/version_solver.dart
```diff
@@ -99,6 +99,7 @@ class VersionSolver {
 
       return await _result();
     } finally {
+      _systemCache.hosted.stopPrefetching();
```
jonasfj: Maybe give a one-line comment like:
```dart
// Stop pre-fetching package/version listings from hosted repository, as resolution is done.
```

lib/src/utils.dart
```diff
@@ -292,20 +292,27 @@ T maxAll<T extends Comparable>(Iterable<T> iter,
       .reduce((max, element) => compare(element, max) > 0 ? element : max);
 }
 
-/// Like [minBy], but with an asynchronous [orderBy] callback.
+/// Returns the element of [values] for which [orderBy] returns the smallest
+/// value.
+///
+/// Returns the first such value in case of ties.
```
jonasfj: Maybe we should write an explicit test case for this utility, to ensure that we cover this?

lib/src/solver/version_solver.dart
sigurdm: Done

lib/src/utils.dart
sigurdm: I wrote a test of stable ordering.

doc/repository-spec-v2.md
sigurdm: Done

doc/repository-spec-v2.md
sigurdm: Done

lib/src/solver/version_solver.dart
```diff
@@ -99,6 +99,9 @@ class VersionSolver {
 
       return await _result();
     } finally {
+      // Stop pre-fetching package/version listings from hosted repository, as
+      // resolution is done.
+      _systemCache.hosted.stopPrefetching();
```
jonasfj: Hmm, are we building a foot-gun here?

I'm wondering if there's other places where we call `BoundHostedSource` and could accidentally be triggering a pre-fetch operation?

What if as a first step in the `try {` block above we initiate pre-fetching if everything that is unsatisfied, since this must be known at this point, right?

Then pre-fetching only happens when `_systemCache.hosted.startPrefetching(...)` has been called, and it always steops when `_systemCache.hosted.stopPrefetching()` is called. Well, I mean most importantly it is necer necessary to call `_systemCache.hosted.stopPrefetching()` if you haven't called `_systemCache.hosted.startPrefetching(...)`.

Thoughts? Or maybe pre-fetching is a zone thing? Just thinking here..

sigurdm: > I'm wondering if there's other places where we call BoundHostedSource and could accidentally be triggering a pre-fetch operation?

Good point. I kind of like the zone idea.
We enter a pefetching-zone, and the fetching operation of BoundHostedSource only does prefetching when inside that zone.

lib/src/retriever.dart
```diff
@@ -0,0 +1,159 @@
...
+    _started = true;
+    while (_queue.isNotEmpty) {
+      final resource = await _pool.request();
+      // This checks if [stop] has been called while waiting for a resource.
+      if (!_started) {
+        resource.release();
+        break;
+      }
+      // Take the highest priority task from the queue.
+      final taskWithZone = _queue.removeFirst();
```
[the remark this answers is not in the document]
sigurdm: I don't think that can happen, as stop would set `_started` to `false`, therefore the above condition would break the loop.

Do you agree?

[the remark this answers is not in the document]
jonasfj: Yes :)

lib/src/source/hosted.dart
```diff
@@ -144,35 +145,68 @@ class BoundHostedSource extends CachedSource {
...
       var id = source.idFor(ref.name, pubspec.version,
           url: _serverFor(ref.description));
-      memoizePubspec(id, pubspec);
+      return MapEntry(id, pubspec);
+    }));
+
+    // Prefetch the dependencies of the latest version, we are likely to need
+    // them later.
+    void prefetch() {
+      final latest = doc['latest'];
```
jonasfj: Should we consider using `result.values` which probably can be sorted based on version numbers, as `pub_semver` knows how to compare these..

This avoids us depending on a property from the server that we didn't depend on before. It also ensure that we use the same ordering as the solve uses...

This is a small thing, and maybe I'm just crazy, just thinking that we should consider if we can avoid trusting the server for anything new :)

lib/src/solver/version_solver.dart
```diff
@@ -91,13 +91,15 @@ class VersionSolver {
...
-      }
+      return await _systemCache.hosted.withPrefetching(() async {
+        var next = _root.name;
+        while (next != null) {
+          _propagate(next);
+          next = await _choosePackageVersion();
+        }
 
-      return await _result();
+        return _result();
```
jonasfj: `await result` for better stack trace maybe? or does that not matter?

sigurdm: Hmm - not sure - I guess you are right - I'll re-add the await to be on the safe side.

lib/src/source/hosted.dart
```diff
@@ -144,35 +145,68 @@ class BoundHostedSource extends CachedSource {
...
           expectedName: ref.name, location: url);
       var id = source.idFor(ref.name, pubspec.version,
           url: _serverFor(ref.description));
-      memoizePubspec(id, pubspec);
+      return MapEntry(id, pubspec);
+    }));
+
+    // Prefetch the dependencies of the latest version, we are likely to need
+    // them later.
+    void prefetch() {
```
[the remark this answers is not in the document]
sigurdm: It's not a method, but an inline function.

I  think I mainly did it like this to name this part of the code.

lib/src/source/hosted.dart
sigurdm: I think you are right that might be the more safe thing to do.
Done

lib/src/source/hosted.dart
```diff
@@ -144,35 +145,68 @@ class BoundHostedSource extends CachedSource {
...
+      final latest = doc['latest'];
+      if (latest is! Map) return;
+      final latestVersionString = latest['version'];
+      if (latestVersionString is! String) return;
+
+      final latestVersionId = PackageId(ref.name, source,
+          Version.parse(latestVersionString as String), ref.description);
+
+      final dependencies = result[latestVersionId]?.dependencies?.values ?? [];
+      if (dependencies.isNotEmpty) {
```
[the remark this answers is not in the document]
sigurdm: no - removed

lib/src/source/hosted.dart
```diff
@@ -144,35 +145,68 @@ class BoundHostedSource extends CachedSource {
...
+        withDependencyType(DependencyType.none, () async {
+          for (final packageRange in dependencies) {
+            retriever.prefetch(packageRange.toRef());
```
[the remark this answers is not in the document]
sigurdm: Good point. Now only prefetching hosted dependencies

lib/src/solver/version_solver.dart
natebosch: The `await` after a `return` currently impacts the behavior of a surrounding `try/catch`. I'm not entirely sure if `try/finally` is impacted. Tracking issue is https://github.com/dart-lang/sdk/issues/39204

lib/src/retriever.dart
```diff
@@ -0,0 +1,159 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import "dart:collection";
+
+import "package:pool/pool.dart";
+import "package:async/async.dart";
+
+/// Handles rate-limited scheduling of tasks.
```
natebosch: The name feels unrelated to the description. Should we rename from `Retriever` to `RateLimitedScheduler`? Are we mixing 2 concepts (retrieving and rate limiting) that should be split?

lib/src/retriever.dart
```diff
@@ -0,0 +1,159 @@
...
+///     (Uri uri, _) => return CancelableOperation.fromFuture(http.read(uri)));
+/// // Start fetching `pub.dev` in the background.
+/// retriever.prefetch(Uri.parse('https://pub.dev/'));
+/// // ... do time-consuming task.
+///
+/// // Now we actually need `pub.dev`.
+/// final pubDevBody = await retriever.fetch(Uri.parse('https://pub.dev/'));
+/// ```
+class Retriever<K, V> {
+  final CancelableOperation<V> Function(K, Retriever) _run;
```
natebosch: Should this `Retriever` be `Retriever<K,V>`? If not, can we make it explicit that it is a `Retriever<dynamic, dynamic>`?

lib/src/retriever.dart
```diff
@@ -0,0 +1,159 @@
...
+  final Queue<_TaskWithZone<K>> _queue = Queue<_TaskWithZone<K>>();
+
+  /// Rate limits the downloads.
+  final Pool _pool;
+
+  /// The currently active operations.
+  final Map<K, CancelableOperation<V>> _active = <K, CancelableOperation<V>>{};
+
+  /// True when the processing loop is running.
+  bool _started = false;
```
natebosch: How about `_isRunning`?

lib/src/retriever.dart
```diff
@@ -0,0 +1,159 @@
...
+                CancelableOperation.fromFuture(run(key, retriever)),
+            maxConcurrentOperations: maxConcurrentOperations);
+
+  /// Starts running operations from the queue. Taking the first items first.
+  void _process() async {
+    assert(!_started);
+    _started = true;
+    while (_queue.isNotEmpty) {
+      final resource = await _pool.request();
+      // This checks if [stop] has been called while waiting for a resource.
```
natebosch: The comments that explain what the code is about to do are redundant. Can we remove them? If the code is too hard to understand without explaining each step, can we make it easier to understand? Should we pull out some chunks of code into well named methods?

lib/src/source/hosted.dart
```diff
@@ -144,35 +145,68 @@ class BoundHostedSource extends CachedSource {
...
-  BoundHostedSource(this.source, this.systemCache);
+  BoundHostedSource(this.source, this.systemCache) {
+    _retriever =
+        Retriever.nonCancelable(_getVersions, maxConcurrentOperations: 10);
+  }
 
-  /// Downloads a list of all versions of a package that are available from the
-  /// site.
-  Future<List<PackageId>> doGetVersions(PackageRef ref) async {
+  Future<Map<PackageId, Pubspec>> _getVersions(
```
natebosch: How about `_fetchVersions`? https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

test/retriever_test.dart
```diff
@@ -0,0 +1,180 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
```
natebosch: [nit] Add a blank line separating `dart:` from `package:` imports.

https://dart.dev/guides/language/effective-dart/style#do-place-dart-imports-before-other-imports

test/utils_test.dart
```diff
@@ -122,4 +124,43 @@ b: {}"""));
       }
     });
   });
+
+  group('minByAsync', () {
+    test('is stable', () async {
+      {
+        final completers = <String, Completer>{};
+        getCompleter(k) => completers.putIfAbsent(k, () => Completer());
```
natebosch: How about `completer(k) =>`?

lib/src/retriever.dart
sigurdm: I updated the name - it is better.

The retrieving behavior is already abstracted, not sure how it would be split further.

lib/src/retriever.dart
sigurdm: It should indeed be <K,V> thanks!

lib/src/retriever.dart
sigurdm: better - thanks

lib/src/retriever.dart
sigurdm: I think most of the comments can be omitted.
Done

lib/src/source/hosted.dart
sigurdm: Better thanks

test/retriever_test.dart
sigurdm: Done

test/utils_test.dart
sigurdm: done

lib/src/source/hosted.dart
```diff
@@ -420,17 +439,12 @@ class BoundHostedSource extends CachedSource {
...
-    var parsed = source._parseDescription(id.description);
-    var server = parsed.last;
-    var package = Uri.encodeComponent(parsed.first);
-    var version = Uri.encodeComponent(id.version.toString());
-    return Uri.parse(pattern(server, package, version));
+  /// Enables speculative prefetching of dependencies of packages queried with
+  /// [getVersions].
+  Future<T> withPrefetching<T>(Future<T> Function() callback) async {
+    return await _scheduler.withPrescheduling((preschedule) async {
+      return await runZoned(callback, zoneValues: {#_pre: preschedule});
```
jonasfj: Mistyped symbole

lib/src/rate_limited_scheduler.dart
```diff
@@ -0,0 +1,151 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import "dart:collection";
+
+import "package:pool/pool.dart";
+
+/// Handles rate-limited scheduling of tasks.
+///
+/// Tasks are named with a key of type [J] (should be useful as a Hash-key) and
```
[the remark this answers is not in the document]
sigurdm: Done

jonasfj:
oh, I guess we'll also have to fix the travis tests.. but we'll land this soon :)

lib/src/rate_limited_scheduler.dart
```diff
@@ -0,0 +1,137 @@
...
+  Future<void> _processNextTask() async {
+    if (_queue.isEmpty) {
+      return;
+    }
+    final task = _queue.removeFirst();
+    final completer = _cache[task.jobId];
+    if (_started.contains(task.jobId)) {
+      return;
+    }
+    _started.add(task.jobId);
```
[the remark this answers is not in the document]
sigurdm: Done

lib/src/rate_limited_scheduler.dart
```diff
@@ -0,0 +1,137 @@
...
+  ///
+  /// If [jobId] has already run, the cached result will be returned.
+  /// If [jobId] is not yet running, it will go to the front of the work queue
+  /// to be scheduled next when there are free resources.
+  Future<V> schedule(J jobId) {
+    final completer = _cache.putIfAbsent(jobId, () => Completer());
+    if (!_started.contains(jobId)) {
+      final task = _Task(jobId, Zone.current);
+      _queue.addFirst(task);
+      scheduleMicrotask(() => _pool.withResource(_processNextTask));
```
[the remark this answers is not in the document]
jonasfj: Good point..

If we do `_pool.withResource(_processNextTask);` then let's leave it with a comment that we intentionally do not await the future..

lib/src/rate_limited_scheduler.dart
```diff
@@ -0,0 +1,137 @@
...
+      return;
+    }
+    final task = _queue.removeFirst();
+    final completer = _cache[task.jobId];
+    if (_started.contains(task.jobId)) {
+      return;
+    }
+    _started.add(task.jobId);
+
+    Future<V> runJob() async {
```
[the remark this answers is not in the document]
sigurdm: It is to handle the case that _runJob throws sync exceptions
Added an explanatory comment

lib/src/rate_limited_scheduler.dart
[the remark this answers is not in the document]
sigurdm: I used `unawaited()`

sigurdm:
Implements rate-limited parallel fetching of version information when solving.
Also does speculative pre-fetching of version information of dependencies of the newest versions of each package.

In informal benchmarks of pub get for a package whose only direct dependency is package:test the resolution time goes from 7.8 seconds to 1.5 seconds.

Also adds package:pedantic to the direct runtime dependencies

Issue #2214, opened 2019-09-20
Consider making a missing dependency an error and not a warning


jakemac53:
See https://github.com/dart-lang/sdk/issues/38482 and https://github.com/dart-lang/sdk/issues/38483 for recent problems regarding this.

Currently pub allows you to publish a package that is known to have fatal errors (missing dependencies).

I would like to propose making those errors, so that publishing would not be allowed at all in that case.

jonasfj:
Yeah, this seems like a solid warning.

I suspect there could be cases where such imports are intentionally ignored.
Say if `package:my_pkg/use_with_other_pkg.dart` imports `other_pkg`, but is only intended to be used if `other_pkg` is already a dependency. For such cases, we won't have proper versioning constraints since we don't support `peerDependencies`/`optionalDependencies`, so forbidding this behavior seems reasonable.

jakemac53:
Ya I really don't think there are any valid use cases here.

mit-mit:
Marking high pri as this affects the overall stability of the package system

jakemac53:
Fwiw I ran a test using `pub_crawl` and `surveyor` to check how many strict deps errors exist in a random sampling of around 1000 existing pub packages, and here is the output:

```
Found 18 packages with strict deps errors (out of 1008 total analyzed packages).
The package amplitude_flutter_example is missing the following deps {amplitude_flutter}
The package appmetrica_sdk_example is missing the following deps {appmetrica_sdk}
The package artemis is missing the following deps {glob, dart_style}
The package auto_animated_example is missing the following deps {auto_animated}
The package essential_components is missing the following deps {http}
The package example is missing the following deps {angular_forms}
The package flutter_local_notifications_example is missing the following deps {flutter_local_notifications}
The package fluxmap_example is missing the following deps {device, flutter_map, geopoint, latlong, map_controller, pedantic}
The package flutter_map_demo is missing the following deps {latlong, geopoint, pedantic}
The package example is missing the following deps {bloc}
The package example is missing the following deps {flutter_map, latlong, pedantic}
The package r_flutter is missing the following deps {glob}
The package sass is missing the following deps {js}
The package sass_migrator is missing the following deps {js}
The package signalr_core is missing the following deps {meta}
The package sqflite_example is missing the following deps {matcher, pedantic, synchronized, meta}
The package tencent_kit_example is missing the following deps {tencent_kit}
The package xmpp_stone is missing the following deps {meta}
```

[...]

PR #2289, opened 2020-01-02
promote strict deps failures to errors from warnings

jakemac53:
Part of https://github.com/dart-lang/pub/issues/2214

This should really have a server side validation component as well, but this will help some.

### DIFF
--- a/lib/src/validator/strict_dependencies.dart
+++ b/lib/src/validator/strict_dependencies.dart
@@ -60,7 +60,7 @@ class StrictDependenciesValidator extends Validator {
             (url.scheme == 'package' &&
                 (url.pathSegments.length < 2 ||
                     url.pathSegments.any((s) => s.isEmpty)))) {
-          warnings.add(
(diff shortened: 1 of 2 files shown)


jonasfj:
Agreed, however, I don't think we can check this at upload time on the server.. we can run some pana scoring and the definitely should reduce score.

Maybe we eventually do a form of staged publishing, and then it makes sense to do on the server. But it's not clear that we would block publishing waiting for such checks.

The problem with running it on the server is that we want analyzer usage to be sandboxed for performance, reliability and security.

jakemac53:
I am fine with client-side only validation if we don't have a mechanism in place for server side validation.

Ideally all the validation would actually be on the server side as it would no longer be dependent on the version of the SDK a given person is using, but that sounds like it would be a larger project.

jonasfj:
Yeah, cleaning up all the checks we have will take time... But conceptually most of the error producing check should be on the server.

PR #2290, opened 2020-01-02
Fix existing hints and lints

natebosch:
Fix 2 cases of `prefer_final_fields`. Fix some unnecessary_lambda
surfaced by making one field final.

Add a missing return on an argument to `List.map`.

Change the argument type on some utility methods to allow a function
returning `FutureOr` to avoid hints about missing returns on methods
expected to return a `Future`.


natebosch:
Possibly redundant against https://github.com/dart-lang/pub/pull/2270 ? 

When do we expect movement there?

jonasfj:
I completely forgot to land https://github.com/dart-lang/pub/pull/2270 :)

PR #2297, opened 2020-01-07
Use LOCALAPPDATA for system cache on windows

### LINKED ISSUE #1273: pub uses %APPDATA% and not %LOCALAPPDATA%
_Originally opened as dart-lang/sdk#22954_

_This issue was originally filed by guillermo.listas...&#064;gmail.com_

---

**What steps will reproduce the problem?**
1. Check pub's cache dir on Windows OR
2. Check pub global's exes dir on Windows

**What is the expected output? What do you see instead?**

Stuff is stored in %USERPROFILE%\AppData\Roaming...

i. e.

%APPDATA%...

Because pub does not store configuration data, but cached data and scripts, it seems strange that pub stores stuff in the roaming profile, as Dart many not be installed on other computers the user accesses. It seems more adequate to store the cache and pub global exes in the %LOCALAPPDATA% folder.

Note that if config data was stored instead, it would be more justifiable to use %APPDATA%.

**What version of the product are you using?**

1.8.5

**On what operating system?**

Windows

Jesse-Bakker:
Currently, the default location for pub's system cache is in %APPDATA%, which means it is in the roaming profile that gets synced across devices. This is less than ideal because it can get pretty large.

This PR moves the default location of the cache to %LOCALAPPDATA%, but will use %APPDATA% if the cache already exists there to avoid breaking existing installs linking to that location

Fixes #1273

### DIFF
--- a/lib/src/system_cache.dart
+++ b/lib/src/system_cache.dart
@@ -38,8 +38,9 @@ class SystemCache {
       // If a cache dir already exists in %APPDATA%, use it, else default to %LOCALAPPDATA%
       var appData = Platform.environment['APPDATA'];
       var appDataCacheDir = p.join(appData, 'Pub', 'Cache');
-      if (Directory(appDataCacheDir).existsSync())
+      if (Directory(appDataCacheDir).existsSync()) {
         return appDataCacheDir;
+      }
       var localAppData = Platform.environment['LOCALAPPDATA'];
       return p.join(localAppData, 'Pub', 'Cache');
     } else {
--- a/lib/src/system_cache.dart
+++ b/lib/src/system_cache.dart
@@ -38,7 +38,7 @@ class SystemCache {
       // If a cache dir already exists in %APPDATA%, use it, else default to %LOCALAPPDATA%
       var appData = Platform.environment['APPDATA'];
       var appDataCacheDir = p.join(appData, 'Pub', 'Cache');
-      if (Directory(appDataCacheDir).existsSync()) {
+      if (dirExists(appDataCacheDir)) {
         return appDataCacheDir;
       }
       var localAppData = Platform.environment['LOCALAPPDATA'];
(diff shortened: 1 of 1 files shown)


Jesse-Bakker:
@googlebot I signed it!

lib/src/system_cache.dart
```diff
@@ -35,8 +35,14 @@ class SystemCache {
     if (Platform.environment.containsKey('PUB_CACHE')) {
       return Platform.environment['PUB_CACHE'];
     } else if (Platform.isWindows) {
+      // If a cache dir already exists in %APPDATA%, use it, else default to %LOCALAPPDATA%
       var appData = Platform.environment['APPDATA'];
-      return p.join(appData, 'Pub', 'Cache');
+      var appDataCacheDir = p.join(appData, 'Pub', 'Cache');
+      if (Directory(appDataCacheDir).existsSync()) {
```
jonasfj: nit, but for consistency with the rest of the code base can we consider using [`dirExists(dir)`](https://github.com/dart-lang/pub/blob/ceaa86f2c7eb35f044b4c202268a17361de68d13/lib/src/io.dart#L303-L307).

lib/src/system_cache.dart
```diff
@@ -35,8 +35,14 @@ class SystemCache {
     if (Platform.environment.containsKey('PUB_CACHE')) {
       return Platform.environment['PUB_CACHE'];
     } else if (Platform.isWindows) {
+      // If a cache dir already exists in %APPDATA%, use it, else default to %LOCALAPPDATA%
```
jonasfj: nit, Can we briefly explain that we prefer `%APPDATA%`  if it already exists because older pub versions created the pub-cache here.
And maybe state that using `%LOCALAPPDATA%` is better because it's not synchronized.

PR #2306, opened 2020-01-17
Prefer system tar, instead of tar in path

sigurdm:
This should fix: https://github.com/dart-lang/pub/issues/2286

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -842,6 +842,8 @@ Future<HttpServer> bindServer(String host, int port) async {
   return server;
 }
 
+String _tarPath = _findTarPath();
+
 /// Find a tar. Prefering system installed tar.
 ///
 /// On linux tar should always be /bin/tar [See FHS 2.3][1]
@@ -877,8 +879,7 @@ Future extractTarGz(Stream<List<int>> stream, String destination) async {
     destination
   ];
 
-  final tarPath = _findTarPath();
-  var process = await _startProcess(tarPath, args);
+  var process = await _startProcess(_tarPath, args);
 
   // Ignore errors on process.std{out,err}. They'll be passed to
   // process.exitCode, and we don't want them being top-levelled by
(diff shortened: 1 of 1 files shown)


lib/src/io.dart
```diff
@@ -887,7 +906,7 @@ Future extractTarGz(Stream<List<int>> stream, String destination) async {
 /// silenced by this flag.
 final bool _noUnknownKeyword = _computeNoUnknownKeyword();
 bool _computeNoUnknownKeyword() {
-  if (!Platform.isLinux) return false;
+  if (Platform.isLinux) return false;
```
jonasfj: I don't think this is intended

lib/src/io.dart
```diff
@@ -842,10 +842,28 @@ Future<HttpServer> bindServer(String host, int port) async {
...
+      'Could not find a system `tar` installed in /bin/tar or /usr/bin/tar, '
+      'attempting to use tar from PATH');
+  return 'tar';
+}
+
 /// Extracts a `.tar.gz` file from [stream] to [destination].
 Future extractTarGz(Stream<List<int>> stream, String destination) async {
   log.fine('Extracting .tar.gz stream to $destination.');
-  if (Platform.isWindows) {
+  if (!Platform.isWindows) {
```
jonasfj: I don't think this change is intentional..

lib/src/io.dart
```diff
@@ -842,10 +842,28 @@ Future<HttpServer> bindServer(String host, int port) async {
...
 
+/// Find a tar. Prefering system installed tar.
+///
+/// On linux tar should always be /bin/tar [See FHS 2.3][1]
+/// On MacOS it seems to always be /usr/bin/tar.
+///
+/// [1]: https://refspecs.linuxfoundation.org/FHS_2.3/fhs-2.3.pdf
+String _findTarPath() {
+  for (final file in ['/bin/tar', '/usr/bin/tar']) {
+    if (File(file).existsSync()) {
```
jonasfj: nit: use `fileExists(file)` defined in this file and use throughout pub.

lib/src/io.dart
sigurdm: indeed :)

lib/src/io.dart
sigurdm: done

sigurdmeldgaard:
The value should be cached...

sigurdm:
> The value should be cached...

Done

PR #2325, opened 2020-02-12
Remove April 1st. easter egg.

jonasfj:
I'm not actually sure I want to remove this.. it's just a bit tempting.

Originally shipped in:
 * https://codereview.chromium.org/1852513004
 * https://codereview.chromium.org/344413004

Mostly, because it'll be harder to keep in sync, once [+134600](https://dart-review.googlesource.com/c/sdk/+/134600) lands.


jwren:
Ping... is this still going to land.  As soon as it does, I will make the equivalent change in `pkg/dartdev/`

natebosch:
We should not be removing functionality, even trivial functionality, because of a bad pattern we started using elsewhere.

mit-mit:
>  We should not be removing functionality, even trivial functionality, because of a bad pattern we started using elsewhere.

@natebosch, please elaborate; what actual use does this have?

jakemac53:
Easter eggs provide value in terms of user happiness and love for a product. It's an intangible for sure but I do think there is real value in things like this. Consider for instance the google doodles. I would argue this falls under the same category as that.

The bigger question that @natebosch was bringing up I think is why this is being removed? I don't think this feature in itself has caused any sort of maintenance burden up to this point. The only reason it is being removed now is because we are copy-pasting large amounts of arg parsing boilerplate. That pattern _is_ going to be a maintenance burden regardless of how many commands/arguments exist in pub. This one in particular hasn't changed in ages so it isn't making the problem any worse imo.

mit-mit:
I agree that some easter eggs can provide emotional value. I am not arguing against easter in general. I was merely making a comment about this *specific one*. I personally find it very confusing to have an easter egg that is triggered with no user action. Partly this limits the value of it (it will be triggered very rarely), and partly I think this is a very poor UX (something as crucial as command help should be easy to parse, and never change).

I do agree on the general point that we should avoid to duplicating code. @jwren I thought we had a plan for sharing this similar to what we did for `dartfmt`?

jakemac53:
I don't think the help output changing (to make two flags visible) is a problem, but regardless we could remove that logic while still leaving the easter egg in place (as always hidden flags).

Note that there is already some logic in there to not do this when running from tests so golden tests won't be broken by it.

jwren:
Yes, there is a path forward to share code/ remove the duplicate implementations.  For both `pub` and `dartfmt` I hope that we get to a place where there is a single implementation:

`flutter [format | pub]` can delegate to `dart [format | pub]`

`pub` and `dartfmt` can be removed from the dart sdk, and aliases mapping `pub` to `dart pub`, etc.


As for the flags, I am okay with fun... When these flags were added, `pub` had a different owners.  My expectation is that since they are triggered on April 1st, they should change every year (Doodles and April Fools Day jokes change every year).  We should remove it at this point, they were added almost 4 years ago.  With so much work that has gone into the pub.dev website, perhaps we should change it up to have sparkle points on packages this year on the web app...?

jonasfj:
It doesn't really cost much to keep it, yet. But I was a bit surprised when I found it :)

Given that the joke has played out, I figured it might be time to remove it.

It also seems a bit out of place that this only works for one subcommand in `dartdev`. I think it would be nice to harmonize the different subcommands when possible. Hence, why I filed #2329, I suspect we should also consider using [package:cli_util](https://pub.dev/packages/cli_util) for print to terminal.

PR #2332, opened 2020-02-19
Enable and fix a number of additional lints

kevmoo:


analysis_options.yaml
```diff
@@ -9,33 +9,54 @@ analyzer:
...
     - directives_ordering
     - empty_statements
+    - file_names
     - implementation_imports
     - iterable_contains_unrelated_type
     - list_remove_unrelated_type
     - literal_only_boolean_expressions
+    - missing_whitespace_between_adjacent_strings
     - no_adjacent_strings_in_list
+    - no_runtimeType_toString
```
natebosch: Why do we care about this one?

lib/src/solver/version_solver.dart
```diff
@@ -185,6 +185,7 @@ class VersionSolver {
     // [incompatibility] is satisfied and we have a conflict.
     if (unsatisfied == null) return #conflict;
 
+    // ignore: missing_whitespace_between_adjacent_strings
```
natebosch: Is there a bug in the lint? Have you filed an issue?

lib/src/utils.dart
```diff
@@ -241,7 +241,8 @@ bool isLoopback(String host) {
 
   try {
     return InternetAddress(host).isLoopback;
-  } on ArgumentError catch (_) {
+  } on ArgumentError catch (_) // ignore: avoid_catching_errors
```
natebosch: 😞 

https://github.com/dart-lang/sdk/issues/40692

lib/src/solver/version_solver.dart
kevmoo: mostly for JS compile. 🤷‍♂

analysis_options.yaml
kevmoo: mostly for JS compile. 🤷‍♂

lib/src/solver/version_solver.dart
kevmoo: No bug. In this case, the lint is not valid – because we don't need whitespace here

natebosch: Right, so we should file a bug on the lint.

kevmoo: No. The lint is doing the right thing. It makes sure that

```dart
"This is my long sentence error message"
"that's missing a space"
```

is correct. But you get false positives.

kevmoo: fixed

kevmoo:
@natebosch – LGTY?

PR #2334, opened 2020-02-20
Change logic for asserting package_config.json is up-to-date.

jonasfj:
* Documented the reasoning behind these assertions.
 * Changed how we assert to forbid packages that are not in the
   lockfile.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -430,6 +430,12 @@ class Entrypoint {
 
     var touchedLockFile = false;
     if (lockFileModified.isBefore(pubspecModified) || hasPathDependencies) {
+      // If `pubspec.lock` is newer than `pubspec.yaml` or we have path
+      // dependencies, then we check that `pubspec.lock` is a correct solution
+      // for the requirements in `pubspec.yaml`. This aims to:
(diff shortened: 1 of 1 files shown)


lib/src/entrypoint.dart
```diff
@@ -646,9 +672,12 @@ class Entrypoint {
     } on FormatException {
       badPackageConfig();
     }
-    if (cfg.configVersion != 2 ||
-        cfg.generator != 'pub' ||
-        cfg.generatorVersion != sdk.version) {
+
+    // Version 2 is the initial version number for `package_config.json`,
+    // because `.packages` was version 1 (even if it was a different file).
+    // If the version is differnet from 2, then it must be a newer incompatible
```
sigurdm: type: different

lib/src/entrypoint.dart
```diff
@@ -665,18 +694,29 @@ class Entrypoint {
...
-        continue; // Ignore entries that aren't in the lockFile
+        assert(
+          false,
+          'unnecessary package_config.json entries should be forbidden by '
+          '_isPackagePathsMappingUpToDateWithLockfile',
+        );
+        continue;
       }
 
+      // If a package is cached, then it's universially immutable and we need
```
sigurdm: typo: universally

Issue #2337, opened 2020-02-24
pub should warn (error?) when publishing without a changelog.


sigurdm:
Came up here: https://github.com/dart-lang/pub/issues/2003

jonasfj:
Can we warn if it's not called `CHANGELOG.md`? And then update docs so that this is our recommended file name?

jakemac53:
We do already recommend that file name https://dart.dev/tools/pub/package-layout#changelog

jakemac53:
We do already recommend that file name https://dart.dev/tools/pub/package-layout#changelog

jonasfj:
I think our recommendation could be more clear. In fact, we should probably just make it the only filename we allow, filed https://github.com/dart-lang/site-www/issues/2289 to discuss this. As the first step is updating documentation for the packaging format.

jonasfj:
@sigurdm, didn't we fix this?

sigurdm:
Fixed in https://github.com/dart-lang/pub/pull/2339

PR #2339, opened 2020-02-24
Give a warning if no CHANGELOG is found.

sigurdm:
Fix of: https://github.com/dart-lang/pub/issues/2337

### DIFF
--- a/lib/src/validator/changelog.dart
+++ b/lib/src/validator/changelog.dart
@@ -25,7 +25,7 @@ class ChangelogValidator extends Validator {
       }
 
       if (!changelog.endsWith('CHANGELOG.md')) {
-        warnings.add('Please consider renaming $changelog to `CHANGELOG.md`.'
+        warnings.add('Please consider renaming $changelog to `CHANGELOG.md`. '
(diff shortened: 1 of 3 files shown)


sigurdm:
I added a warning if the CHANGELOG is not named CHANGELOG.md
PTAL.

lib/src/validator/changelog.dart
```diff
@@ -19,10 +19,16 @@ class ChangelogValidator extends Validator {
       final changelog = entrypoint.root.changelogPath;
 
       if (changelog == null) {
-        // No changelog was found, which is fine. Return with no warnings.
+        warnings.add('Please add a`CHANGELOG.md` to your package. '
+            'See https://dart.dev/tools/pub/publishing#important-files.');
         return;
       }
 
+      if (!changelog.endsWith('CHANGELOG.md')) {
```
jonasfj: @sigurdm, why do we tests using `endsWith`? Can't we use extract the filename from the path and test for equality?

sigurdm: Sure - that would be better. I'll make a fix

PR #2356, opened 2020-03-10
Warn when non-prerelease depends on pre-release SDK or package

sigurdm:
Fixes https://github.com/dart-lang/pub-dev/issues/3375

### DIFF
--- a/lib/src/validator/dependency.dart
+++ b/lib/src/validator/dependency.dart
@@ -76,16 +76,18 @@ class DependencyValidator extends Validator {
               "Older versions of pub don't support Git path dependencies.");
         }
       } else {
-        _warnAboutPrerelease(dependency);
         if (constraint.isAny) {
           _warnAboutNoConstraint(dependency);
-        } else if (constraint is Version) {
-          _warnAboutSingleVersionConstraint(dependency);
         } else if (constraint is VersionRange) {
-          if (constraint.min == null) {
-            _warnAboutNoConstraintLowerBound(dependency);
-          } else if (constraint.max == null) {
-            _warnAboutNoConstraintUpperBound(dependency);
+          if (constraint is Version) {
+            _warnAboutSingleVersionConstraint(dependency);
+          } else {
+            _warnAboutPrerelease(dependency.name, constraint);
+            if (constraint.min == null) {
+              _warnAboutNoConstraintLowerBound(dependency);
+            } else if (constraint.max == null) {
+              _warnAboutNoConstraintUpperBound(dependency);
+            }
           }
           _hasCaretDep = _hasCaretDep || constraint.toString().startsWith('^');
         }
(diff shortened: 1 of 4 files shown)


kevmoo:
We're just warning? Not blocking?

Also: we should have this implemented on the server, not just the client!

CC @mit-mit

sigurdm:
If it should indeed block, it should be checked on the server -- agree.

jonasfj:
@kevmoo, this follows the discussion in https://github.com/dart-lang/pub-dev/issues/3375, let's continue the discussion of what is desirable there.

lib/src/validator/sdk_constraint.dart
```diff
@@ -42,6 +47,21 @@ class SdkConstraintValidator extends Validator {
...
+      if (constrainMin != null &&
+          constrainMin.isPreRelease &&
+          !packageVersion.isPreRelease) {
+        warnings.add(
+            'The Dart SDK constraint in `pubspec.yaml` is on pre-release '
+            'version $constrainMin, but you are publishing a '
+            'non-pre-release version $packageVersion.\n'
+            'Consider publishing a pre-release instead.\n'
+            'See https://dart.dev/tools/pub/publishing#publishing-prereleases '
+            'For more information on pre-releases.');
```
jonasfj: Can't we use stronger language...

```
Packages with an SDK constraint on a pre-release of the Dart SDK should themselves be published as a pre-release version. If this package needs Dart version $constraintMin, consider publishing the package as a pre-release instead.
See: https://dart.dev/tools/pub/publishing#publishing-prereleases
For more information on pre-releases.
```

sigurdm: Done

sigurdm:
We also want to warn about dependencies on prereleases of packages

jonasfj:
> We also want to warn about dependencies on prereleases of packages

Yes, otherwise we can have transitive dependency on a pre-release SDK.

sigurdm:
Now it also warns about dependencies on prereleases of packages

lib/src/validator/dependency.dart
```diff
@@ -75,18 +75,20 @@ class DependencyValidator extends Validator {
...
-        } else if (constraint.max == null) {
-          _warnAboutNoConstraintUpperBound(dependency);
+      } else {
+        _warnAboutPrerelease(dependency);
+        if (constraint.isAny) {
+          _warnAboutNoConstraint(dependency);
+        } else if (constraint is Version) {
+          _warnAboutSingleVersionConstraint(dependency);
+        } else if (constraint is VersionRange) {
+          if (constraint.min == null) {
```
jonasfj: Can't we move ` _warnAboutPrerelease(dependency);` down here? And call it as `_warnAboutPrerelease(dependency.name, constraint);` that way we don't need to do `is VersionRange` twice.

sigurdm: I would like to also catch the `Version` case. (`Version` implements `VersionRange`) That's why I took it outside.

But I think I can restructure the logic to avoid the double check.

Issue #2365, opened 2020-03-16
Use package archive URL from list versions response to download packages


jonasfj:
In: https://github.com/dart-lang/pub/blob/ffb1a3c06df62113d95963b376c5ae7ed39a08a3/lib/src/source/hosted.dart#L359

We have a hardcoded end-point, but the list-versions API returns a URL for the archive. We should always use this URL.

PR #2366, opened 2020-03-16
Download packages from the archive link instead of hardcoded uri

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/2365

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -368,8 +368,6 @@ class BoundHostedSource extends CachedSource {
     final url = versions[id]?.archiveUrl ??
         (throw PackageNotFoundException(
             'Could not find archive of package $id'));
-
-    ;
     log.io('Get package from $url.');
     log.message('Downloading ${log.bold(id.name)} ${id.version}...');
 
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -151,6 +151,13 @@ class HostedSource extends Source {
   }
 }
 
+class _VersionInfo {
+  final Pubspec pubspec;
+  final Uri archiveUrl;
+
+  _VersionInfo(this.pubspec, this.archiveUrl);
+}
+
 /// The [BoundSource] for [HostedSource].
 class BoundHostedSource extends CachedSource {
   @override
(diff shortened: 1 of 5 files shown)


lib/src/source/hosted.dart
```diff
@@ -186,7 +193,10 @@ class BoundHostedSource extends CachedSource {
           expectedName: ref.name, location: url);
       var id = source.idFor(ref.name, pubspec.version,
           url: _serverFor(ref.description));
-      return MapEntry(id, pubspec);
+      final archiveUrlValue = map['archive_url'];
+      final archiveUrl =
+          archiveUrlValue is String ? Uri.tryParse(archiveUrlValue) : null;
```
jonasfj: What do we do if the value is `null`?

lib/src/source/hosted.dart
```diff
@@ -354,11 +363,13 @@ class BoundHostedSource extends CachedSource {
 
   /// Downloads package [package] at [version] from [server], and unpacks it
   /// into [destPath].
-  Future _download(
-      String server, String package, Version version, String destPath) async {
-    var url = Uri.parse('$server/packages/$package/versions/$version.tar.gz');
+  Future _download(PackageId id, String destPath) async {
+    final versions = await _scheduler.schedule(id.toRef());
+    final url = versions[id]?.archiveUrl ??
+        (throw PackageNotFoundException(
```
jonasfj: Being unable to find the version and being unable to find the archiveUrl is two different issues.

If we want find the package we should fallback to `PackageNotFoundException`, if we can't find the `archiveUrl` I propose that we could fallback to the old URL pattern?

Honestly, I would prefer to avoid this, but there might be 3rd party pub servers out there which doesn't have the `archiveUrl`.

lib/src/source/hosted.dart
```diff
@@ -151,21 +151,28 @@ class HostedSource extends Source {
   }
 }
 
+class _VersionInfo {
```
jonasfj: Can we give this a one-liner comment, something like:
`/// Information about a package version as served from pub server`

lib/src/source/hosted.dart
sigurdm: Well, then we cannot download the file.
We could make a fallback to the hardcoded value.

lib/src/source/hosted.dart
sigurdm: Ok
Done.

Now it will fall back to the old URL pattern if the version exists in the version info from the api/package/versions but has no archive url. If it does not exist in the version listing we throw up front.

You like?

lib/src/source/hosted.dart
```diff
@@ -354,11 +363,13 @@ class BoundHostedSource extends CachedSource {
 
   /// Downloads package [package] at [version] from [server], and unpacks it
   /// into [destPath].
-  Future _download(
-      String server, String package, Version version, String destPath) async {
-    var url = Uri.parse('$server/packages/$package/versions/$version.tar.gz');
+  Future _download(PackageId id, String destPath) async {
+    final versions = await _scheduler.schedule(id.toRef());
```
jonasfj: Won't this introduce an extra round-trip if installing from a `pubspec.lock` (I'm not sure, just guessing).

This might also be acceptable.

sigurdm: Yes I think it would - not sure how much a problem this is, all the packages are AFAIK downloaded in parallel, so it will not take the time of an extra round-trip per package.

On the positive side, it gives us freedom to move things around by changing the archive_url.

If we get around to cache the response from api/package/versions on disk, the issue will diminish even further.

PR #2369, opened 2020-03-17
Drop redundant stream utils

natebosch:
- `streamFirst` does not behave differently from `stream.first`. The
  comment about being able to use it with single-subscription streams is
  a lie.
- `streamLines` is redundant against `LineSplitter`.


kevmoo:
You can tell your code is old when...

jonasfj:
Please check that CI passes first :)

PR #2375, opened 2020-03-18
Stop using package:package_config

sigurdm:
We only used it for parsing and writing .packages.
And that API is deprecated.

The newer api for writing .packages together with .dart_tools/package_config.json is not catered to our purposes.


lib/src/package_config.dart
```diff
@@ -274,3 +276,24 @@ String extractLanguageVersion(VersionConstraint c) {
...
   return '${minVersion.major}.${minVersion.minor}';
 }
+
+/// Parses the (now obsolete) .packages file.
+Map<String, Uri> parsePackagesFile(List<int> bytes, Uri base) {
+  final result = <String, Uri>{};
+  for (var line in utf8.decode(bytes).split('\n')) {
+    line = line.trim();
+    final commentLocation = line.indexOf('#');
+    if (commentLocation != -1) line = line.substring(0, commentLocation);
```
jonasfj: I think `#` is only allowed on the first character of a line, see:
https://github.com/dart-lang/package_config/blob/master/lib/packages_file.dart#L33-L97

Interpreting it as comment elsewhere is probably a mistake.

lib/src/package_config.dart
```diff
@@ -274,3 +276,24 @@ String extractLanguageVersion(VersionConstraint c) {
...
+    if (line.isEmpty) continue;
+    final colonLocation = line.indexOf(':');
+    if (colonLocation == -1) {
+      throw FormatException('Malformed line in .packages');
+    }
+    final packageName = line.substring(0, colonLocation);
+    if (result.containsKey(packageName)) {
+      throw FormatException('.packages contains $packageName twice');
+    }
+    result[packageName] = base.resolve(line.substring(colonLocation + 1));
```
jonasfj: Should we catch `FomatException` from `base.resolve` and throw it as a `FormatException` with a nicer message?

lib/src/package_config.dart
```diff
@@ -274,3 +276,24 @@ String extractLanguageVersion(VersionConstraint c) {
   }
   return '${minVersion.major}.${minVersion.minor}';
 }
+
+/// Parses the (now obsolete) .packages file.
+Map<String, Uri> parsePackagesFile(List<int> bytes, Uri base) {
+  final result = <String, Uri>{};
+  for (var line in utf8.decode(bytes).split('\n')) {
+    line = line.trim();
```
jonasfj: I think this is too much... From what I can see the original code only ignores CR and LF:
https://github.com/dart-lang/package_config/blob/master/lib/packages_file.dart#L33-L97

Treating both of them as line breaks.. and it allows empty lines.

PR #2377, opened 2020-03-20
Download hosted packages in parallel when repairing cache

sigurdm:
This makes `pub cache repair` a lot faster.

The max number of parallel downloads is still throttled to 16 by the ThrottleClient in `src/http.dart`.

I did not make the git cache repairs parallel as they are somewhat rarer (I expect).

### DIFF
--- a/lib/src/command/cache_repair.dart
+++ b/lib/src/command/cache_repair.dart
@@ -26,17 +26,21 @@ class CacheRepairCommand extends PubCommand {
 
   @override
   Future run() async {
-    var successes = [];
-    var failures = [];
-
     // Repair every cached source.
-    for (var source in cache.sources.all.map(cache.source)) {
-      if (source is CachedSource) {
-        var results = await source.repairCachedPackages();
-        successes.addAll(results.first);
-        failures.addAll(results.last);
-      }
-    }
+    final repairResults = (await Future.wait(
+            cache.sources.all.map(cache.source).map((source) async {
+      return source is CachedSource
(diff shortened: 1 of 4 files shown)


lib/src/source/cached.dart
```diff
@@ -67,8 +67,14 @@ abstract class CachedSource extends BoundSource {
...
-  /// repaired.
-  Future<Pair<List<PackageId>, List<PackageId>>> repairCachedPackages();
+  /// Returns a list of results indicating for each if that package was
+  /// successfully repaired.
+  Future<Iterable<RepairResult>> repairCachedPackages();
+}
+
+/// The result of repairing a single cache entry.
+class RepairResult {
+  final bool success;
```
jonasfj: Can we give this a comment, saying that this is whether or not the repair was successful.

Also maybe explain in comment what it means when repair for a package failed, is it now missing from cache?

lib/src/command/cache_repair.dart
```diff
@@ -26,17 +26,21 @@ class CacheRepairCommand extends PubCommand {
...
+          : <RepairResult>[];
+    })))
+        .expand((x) => x);
+
+    final successes = [
+      for (final result in repairResults) if (result.success) result.package
+    ];
+    final failures = [
+      for (final result in repairResults) if (!result.success) result.package
+    ];
```
jonasfj: Nice :D

lib/src/source/cached.dart
sigurdm: Done

PR #2379, opened 2020-03-20
Fix `pub outdated` for circular dependency on root.

jonasfj:
When root package is required by one of its dependencies, resolution
fails as the stripped `PubSpec` didn't retain the verion number.
This addresses this issue, and adds a test case covering circular
dependency on the root package.


jonasfj:
credits @mit-mit for finding this :)

PR #2381, opened 2020-03-23
Warn about deprecated LICENSE and README.md file names

sigurdm:


test/validator/readme_test.dart
```diff
@@ -47,21 +45,32 @@ void main() {
...
       await d.dir(appPath, [
         d.file('README', [192])
       ]).create();
       expectValidationWarning(readme);
     });
+
+    test('has only a non-primary readme', () async {
+      await d.validPackage.create();
+      deleteEntry(p.join(d.sandbox, 'myapp/README.md'));
+      await d.dir(appPath, [d.file('README.whatever')]).create();
```
jonasfj: or test `README.mdown` and `README.markdown` those are alternative suffixes we're working to deprecate.

sigurdm: Done

sigurdm:
> I wonder if we should warn about README.md and LICENSE being empty files, that's probably a separate issue.

I think warning about empty files leads us down a long never-ending path of warnings.

I like the warning about the presence of the files. That will remind authors that these files are needed. But in the end we cannot prevent people from putting nonsense in them.

Issue #2186, opened 2019-08-07
Avoid using gzip from the system


jonasfj:
This causes problems on OSX when pigz is installed, and the dart-sdk does ship with a gzip implementation already.

See: https://github.com/dart-lang/pub/issues/2126

natebosch:
More discussion in https://github.com/dart-lang/pub/issues/1602

jonasfj:
Yes, but even if we just used gzip provided with dart that might be an improvement on OS X where pigz is installed.

That said I generally agree we should do a tar-stream implementation in Dart.

PR #2382, opened 2020-03-23
Use gzip decoder from dart:io.

sigurdm:
And stop using `tar` in streaming mode. It seems to give us problems.
Instead the decompressed tar file is stored to disk (that was always the
case on windows).

Also streamline the tar.gz extraction logic between
windows/non-windows a bit.

Has been tested with the tool/extract_all_pub_dev.dart tool on linux, and partial runs with windows and mac os.

Fixes: #2186 , https://github.com/dart-lang/pub/issues/2308

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -847,7 +847,7 @@ Future extractTarGz(Stream<List<int>> stream, String destination) async {
     final tarFile = path.join(tempDir, 'archive.tar');
     await _createFileFromStream(decompressed, tarFile);
     return (Platform.isWindows)
-        ? runProcess(_pathTo7zip, ['x', '$tarFile', '-o$destination'])
+        ? runProcess(_pathTo7zip, ['x', '$tarFile'], workingDir: destination)
         : runProcess(_tarPath, [
             if (_noUnknownKeyword) '--warning=no-unknown-keyword',
             '--extract',
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -844,7 +844,7 @@ Future extractTarGz(Stream<List<int>> stream, String destination) async {
   // We used to stream directly to `tar`,  but that was fragile in certain
   // settings.
   final processResult = await withTempDir((tempDir) async {
-    final tarFile = path.join(tempDir, '${path.basename(destination)}.tar');
+    final tarFile = path.join(tempDir, 'archive.tar');
     await _createFileFromStream(decompressed, tarFile);
     return (Platform.isWindows)
         ? runProcess(_pathTo7zip, ['x', '"$tarFile"', '-o"$destination"'])
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -226,8 +226,8 @@ String createTempDir(String base, String prefix) {
 /// 'pub_' with characters appended to it to make a unique name.
 ///
 /// Returns the path of the created directory.
-String _createSystemTempDir() {
-  var tempDir = Directory.systemTemp.createTempSync('pub_');
+Future<String> _createSystemTempDir() async {
+  var tempDir = await Directory.systemTemp.createTemp('pub_');
   log.io('Created temp directory ${tempDir.path}');
   return tempDir.resolveSymbolicLinksSync();
 }
@@ -797,7 +797,7 @@ void touch(String path) => File(path).setLastModifiedSync(DateTime.now());
 /// Returns a future that completes to the value that the future returned from
 /// [fn] completes to.
 Future<T> withTempDir<T>(FutureOr<T> Function(String path) fn) async {
-  var tempDir = _createSystemTempDir();
+  var tempDir = await _createSystemTempDir();
   try {
     return await fn(tempDir);
   } finally {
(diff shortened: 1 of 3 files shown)


lib/src/io.dart
```diff
@@ -867,38 +840,38 @@ String _findTarPath() {
...
-  if (Platform.isWindows) {
-    return await _extractTarGzWindows(stream, destination);
-  }
+  final decompressed = stream.transform(GZipCodec().decoder);
+  // We used to stream directly to `tar`,  but that was fragile in certain
+  // settings.
+  PubProcessResult processResult;
+  await withTempDir((tempDir) async {
+    final tarFile =
+        path.join(tempDir, path.basenameWithoutExtension(destination));
```
jonasfj: Why not just always call this intermediate file for `path.join(tempDir, 'archive.tar')` ?

sigurdm:
Hmm - not sure why travis is not triggered

test/io_test.dart
```diff
@@ -344,6 +348,59 @@ void testExistencePredicate(String name, bool Function(String path) predicate,
       });
     }
   });
+  group('extractTarGz', () {
+    test('decompresses simple archive', () async {
+      await withTempDir((tempDir) async {
+        await extractTarGz(
```
jonasfj: Can we make a test case that we default to `skip: true` where we iterate through all packages and all versions on `pub.dev` and check if the tar archive that is downloaded can be extracted?

Then we just manually enable or run the test when we want to run it. Maybe we can create a test tag called `extremely_expensive` and [tag the test case](https://pub.dev/packages/test#tagging-tests) with this and configuring the tag to be skipped by default.

lib/src/io.dart
```diff
@@ -867,38 +840,38 @@ String _findTarPath() {
...
 Future extractTarGz(Stream<List<int>> stream, String destination) async {
   log.fine('Extracting .tar.gz stream to $destination.');
-  if (Platform.isWindows) {
-    return await _extractTarGzWindows(stream, destination);
-  }
+  final decompressed = stream.transform(GZipCodec().decoder);
+  // We used to stream directly to `tar`,  but that was fragile in certain
+  // settings.
+  PubProcessResult processResult;
+  await withTempDir((tempDir) async {
```
jonasfj: Since you plan to do many of these concurrently, maybe we should consider making `_createSystemTempDir` async

test/io_test.dart
sigurdm: Sounds good.
I'll look into that.

lib/src/io.dart
sigurdm: Done

natebosch:
Should mark #2186 as fixed with this.

jonasfj:
Not sure why travis won't update status, but windows tests fails with:
```
[e] FINE: Extracting .tar.gz stream to C:\Users\travis\AppData\Local\Temp\dart_test_ab95ba0e-6d11-11ea-94ed-00155d45060c\cache\_temp\dirabfd6df3-6d11-11ea-94ed-00155d45060c.
[e] IO  : Created temp directory C:\Users\travis\AppData\Local\Temp\pub_abfd6df4-6d11-11ea-94ed-00155d45060c
[e] IO  : Creating C:\Users\travis\AppData\Local\Temp\pub_abfd6df4-6d11-11ea-94ed-00155d45060c\dirabfd6df3-6d11-11ea-94ed-00155d45060c.tar from stream.
[e] FINE: Created C:\Users\travis\AppData\Local\Temp\pub_abfd6df4-6d11-11ea-94ed-00155d45060c\dirabfd6df3-6d11-11ea-94ed-00155d45060c.tar from stream.
[e] IO  : Spawning "C:\Users\travis\dart-sdk\lib\_internal\pub\asset\7zip\7za.exe x "C:\Users\travis\AppData\Local\Temp\pub_abfd6df4-6d11-11ea-94ed-00155d45060c\dirabfd6df3-6d11-11ea-94ed-00155d45060c.tar" -o"C:\Users\travis\AppData\Local\Temp\dart_test_ab95ba0e-6d11-11ea-94ed-00155d45060c\cache\_temp\dirabfd6df3-6d11-11ea-94ed-00155d45060c"" in C:\Users\travis\AppData\Local\Temp\dart_test_ab95ba0e-6d11-11ea-94ed-00155d45060c\myapp\.
[e] IO  : Finished C:\Users\travis\dart-sdk\lib\_internal\pub\asset\7zip\7za.exe. Exit code 2.
[e]     | stdout:
[e]     | | 
[e]     | | 7-Zip (A) 9.20  Copyright (c) 1999-2010 Igor Pavlov  2010-11-18
[e]     | | 
[e]     | | 
[e]     | | Error:
[e]     | | cannot find archive
[e]     | Nothing output on stderr.
```

sigurdm:
Yes, I saw, I have to load my windows VM up It think

test/io_test.dart
sigurdm: I ended up putting it in tools/ instead. It is hard to run in a single session, so I made it save a status file, so you can pick up from where you left off.

tool/extract_all_pub_dev.dart
```diff
@@ -0,0 +1,109 @@
...
+
+import 'package:http/http.dart' as http;
+import 'package:pool/pool.dart';
+import 'package:pub/src/http.dart';
+import 'package:pub/src/io.dart';
+
+const statusFilename = 'extract_all_pub_status.json';
+
+Stream<String> allPackageNames() async* {
+  var nextUrl = 'https://pub.dev/api/packages';
```
jonasfj, line 21: try `https://pub.dartlang.org/api/packages?compact=1` then you'll get all package names in one request.

sigurdm: Ah - nice.

PR #2387, opened 2020-03-24
Output a summary when there are newer version, but they cannot resolv…

sigurdm:
…e together

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -322,6 +322,11 @@ Future<void> _outputHuman(List<_PackageDetails> rows,
           'To update these dependencies, edit pubspec.yaml.');
     }
   }
+
+  if (notAtResolvable == 0 && upgradable == 0 && rows.isNotEmpty) {
+    log.message(
+        '\nDependencies are all on the latest mutually compatible versions');
+  }
 }
 
 Future<List<_FormattedString>> oudatedMarker(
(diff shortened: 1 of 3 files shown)


sigurdm:
cc @mit-mit

lib/src/command/outdated.dart
```diff
@@ -322,6 +322,11 @@ Future<void> _outputHuman(List<_PackageDetails> rows,
           'To update these dependencies, edit pubspec.yaml.');
     }
   }
+
+  if (notAtResolvable == 0 && upgradable == 0 && rows.isNotEmpty) {
+    log.message(
+        '\nDependencies are all on the latest mutually compatible versions');
```
jonasfj: I think we should add a period.

Or maybe write something like:
```
'\nNewer versions is available for $N dependenc(y|ies), but these are
 not mutually compatible with your other dependencies.'
```

OR
```
'\nDependencies are all on the latest mutually compatible versions'
'\nNewer versions, while available, is not mutually compatible'
```

lib/src/command/outdated.dart
```diff
@@ -325,7 +325,8 @@ Future<void> _outputHuman(List<_PackageDetails> rows,
 
   if (notAtResolvable == 0 && upgradable == 0 && rows.isNotEmpty) {
     log.message(
-        '\nDependencies are all on the latest mutually compatible versions');
+        '\nDependencies are all on the latest mutually compatible versions.'
```
mit-mit: Maybe s/Dependencies are all on the latest mutually compatible versions/Dependencies are all on the latest resolvable versions/ given we have the phrase 'mutually compatible below'?

sigurdm: Done

PR #2389, opened 2020-03-24
Include telemetry headers when sending any request to the pub site

kevmoo:


kevmoo:
Noticed we were not getting this data for `outdated` or `publish`. No reason to exclude it from any request.

Issue #2397, opened 2020-03-25
`pub publish` failing with json_serializable 


kevmoo:
pub @ ab109723d9eb782c2ea4cce68c044da40403a652

`json_serializable` @ https://github.com/dart-lang/json_serializable/commit/3ea5351cb73f1e1131a252533276e1887bb883c8

```
The method 'toList' was called on null.
Receiver: null
Tried calling: toList()
dart:core                                                         Object.noSuchMethod
package:analyzer/src/dart/analysis/file_state.dart 924:51         FileSystemState.getFilesForPath
package:analyzer/src/dart/analysis/file_tracker.dart 232:40       FileTracker.verifyApiSignature.<fn>
package:analyzer/src/dart/analysis/performance_logger.dart 34:15  PerformanceLog.run
package:analyzer/src/dart/analysis/file_tracker.dart 223:20       FileTracker.verifyApiSignature
package:analyzer/src/dart/analysis/driver.dart 924:35             AnalysisDriver.parseFileSync
package:analyzer/src/dart/analysis/session.dart 151:20            AnalysisSessionImpl.getParsedUnit
package:pub/src/dart.dart 124:49                                  AnalysisContextManager.parse
package:pub/src/validator/language_version.dart 49:39             LanguageVersionValidator.validate
package:pub/src/validator.dart 136:64                             Validator.runAll.<fn>
dart:async                                                        Future.wait
package:pub/src/validator.dart 136:19                             Validator.runAll
package:pub/src/command/lish.dart 163:32                          LishCommand._validate
package:pub/src/command/lish.dart 143:15                          LishCommand.run
package:args/command_runner.dart 197:27                           CommandRunner.runCommand
package:pub/src/command_runner.dart 176:39                        PubCommandRunner.runCommand.<fn>
dart:async                                                        new Future.sync
package:pub/src/utils.dart 113:12                                 captureErrors.wrappedCallback
dart:async                                                        runZoned
package:pub/src/utils.dart 130:5                                  captureErrors
package:pub/src/command_runner.dart 176:13                        PubCommandRunner.runCommand
```

wstrange:
Seeing the same thing.

kevmoo:
@wstrange – which version of Dart are you on?

kevmoo:
I have a PR pending here: https://github.com/dart-lang/pub/pull/2398

wstrange:
[...]

PR #2398, opened 2020-03-25
Fix NPE during language version analysis

kevmoo:
Fixes https://github.com/dart-lang/pub/issues/2397

During debugging, I noticed paths like this were flowing to getParsedUnit

```
/Users/kevmoo/github/json_serializable/json_serializable/./lib/builder.dart
/Users/kevmoo/github/json_serializable/json_serializable/./lib/type_helper.dart
/Users/kevmoo/github/json_serializable/json_serializable/./lib/json_serializable.dart
/Users/kevmoo/github/json_serializable/json_serializable/./lib/src/json_part_builder.dart
```

Adding `normalize` fixed this

### DIFF
--- a/lib/src/validator/language_version.dart
+++ b/lib/src/validator/language_version.dart
@@ -42,6 +42,7 @@ class LanguageVersionValidator extends Validator {
     for (final path in ['lib', 'bin']
         .map((path) => entrypoint.root.listFiles(beneath: path))
         .expand((files) => files)
+        .map(p.normalize)
         .map(p.absolute)
         .where((String file) => p.extension(file) == '.dart')) {
       CompilationUnit unit;


kevmoo:
No clue how to test this, but I did confirm that it works – was able to publish!

lib/src/validator/language_version.dart
```diff
@@ -42,6 +42,7 @@ class LanguageVersionValidator extends Validator {
     for (final path in ['lib', 'bin']
         .map((path) => entrypoint.root.listFiles(beneath: path))
         .expand((files) => files)
+        .map(p.normalize)
```
jonasfj: @sigurdm, do we have this issue elsewhere? Maybe in `pana`?

sigurdm: Good question
We don't normalize the path here: https://github.com/dart-lang/pana/blob/master/lib/src/tag_detection.dart#L164

On the other hand the paths there are constructed differently. Not sure it's an issue

sigurdm: I created https://github.com/dart-lang/pub/pull/2409 to make sure we don't have the problem in pub anymore

Issue #2393, opened 2020-03-25
pub outdated format option


devoncarew:
`outdated` has a `--format` option:

```
--format=<FORMAT>          Defines how the output should be formatted. Defaults to color when connected to a terminal, and no-color otherwise.
```

- the docs don't specify what we should use for `FORMAT`
- if this flag just controls using ansi colors, I would either remove it entirely and rely on ansi console detection, or, add a `--[no-]color` option to the main pub command

sigurdm:
Good point - --[no-]-color might be better

sigurdm:
There is also the --format=json option. But as you point out it was not documented.

It could have its own flag

--[-no]-color
--json

devoncarew:
Some tools use `--machine` to indicate that it will output in a json format. That's less of a ~standard though than using `--no-color` to indicate turning off ansi detection.

PR #2402, opened 2020-03-26
Simplify arguments to outdated

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/2393

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -37,7 +37,8 @@ class OutdatedCommand extends PubCommand {
             'when connected to a terminal, and no-color otherwise.');
 
     argParser.addFlag('json',
-        help: 'Outputs the results in a json formatted report');
+        help: 'Outputs the results in a json formatted report',
+        negatable: false);
 
     argParser.addFlag('up-to-date',
         defaultsTo: false,
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -32,11 +32,12 @@ class OutdatedCommand extends PubCommand {
   String get docUrl => 'https://dart.dev/tools/pub/cmd/pub-outdated';
 
   OutdatedCommand() {
-    argParser.addOption('format',
-        help: 'Defines how the output should be formatted. Defaults to color '
-            'when connected to a terminal, and no-color otherwise.',
-        valueHelp: 'FORMAT',
-        allowed: ['color', 'no-color', 'json']);
+    argParser.addFlag('color',
+        help: 'Whether to color the output. Defaults to color '
+            'when connected to a terminal, and no-color otherwise.');
+
+    argParser.addFlag('json',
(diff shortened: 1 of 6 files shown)


lib/src/command/outdated.dart
```diff
@@ -32,11 +32,12 @@ class OutdatedCommand extends PubCommand {
...
-        help: 'Defines how the output should be formatted. Defaults to color '
-            'when connected to a terminal, and no-color otherwise.',
-        valueHelp: 'FORMAT',
-        allowed: ['color', 'no-color', 'json']);
+    argParser.addFlag('color',
+        help: 'Whether to color the output. Defaults to color '
+            'when connected to a terminal, and no-color otherwise.');
+
+    argParser.addFlag('json',
+        help: 'Outputs the results in a json formatted report');
```
jonasfj: This needs `negatable: false` specifying `--no-json` makes no sense.

lib/src/command/outdated.dart
```diff
@@ -151,11 +152,11 @@ class OutdatedCommand extends PubCommand {
 
     rows.sort();
 
-    if (argResults['format'] == 'json') {
+    if (argResults['json']) {
       await _outputJson(rows);
     } else {
-      final useColors = argResults['format'] == 'color' ||
-          (!argResults.wasParsed('format') && stdin.hasTerminal);
+      final useColors = argResults['color'] ||
```
jonasfj: Does the default in fact work? Given that `addFlag` is passed `defaultsTo: false` by default. Won't this always be `false` unless called with `--color`?

Can we detect if `--color` was passed? Or if it just assumes the default value?

lib/src/command/outdated.dart
```diff
@@ -32,11 +32,12 @@ class OutdatedCommand extends PubCommand {
...
 
   OutdatedCommand() {
-    argParser.addOption('format',
-        help: 'Defines how the output should be formatted. Defaults to color '
-            'when connected to a terminal, and no-color otherwise.',
-        valueHelp: 'FORMAT',
-        allowed: ['color', 'no-color', 'json']);
+    argParser.addFlag('color',
+        help: 'Whether to color the output. Defaults to color '
+            'when connected to a terminal, and no-color otherwise.');
```
jonasfj: `defaultsTo: true` might make more sense... depending on how this works.

lib/src/command/outdated.dart
sigurdm: Good point.
Done.

lib/src/command/outdated.dart
sigurdm: No - wasParsed('color') returns false if there was no explicit color argument.

sigurdm: Can we detect if --color was passed? Or if it just assumes the default value

wasParsed('color') answers exactly that.

PR #2404, opened 2020-03-26
Handle overrides

sigurdm:
Handle the case where a dependency is overridden.

Also print the source name for packages that are not from a hosted source.

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -39,11 +39,11 @@ class OutdatedCommand extends PubCommand {
 
     argParser.addFlag('up-to-date',
         defaultsTo: false,
-        help: 'Include dependencies that are already at the latest version');
+        help: 'Include dependencies that are already at the latest version.');
 
     argParser.addFlag('pre-releases',
         defaultsTo: false,
-        help: 'Include pre-releases when reporting latest version');
+        help: 'Include pre-releases when reporting latest version.');
 
     argParser.addFlag(
       'dev-dependencies',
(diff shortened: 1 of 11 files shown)


sigurdm:
travis???

lib/src/command/outdated.dart
```diff
@@ -65,11 +64,9 @@ class OutdatedCommand extends PubCommand {
     entrypoint.assertUpToDate();
 
     final includeDevDependencies = argResults['dev-dependencies'];
-
     final upgradePubspec = includeDevDependencies
-        ? entrypoint.root.pubspec
+        ? _stripOverrides(entrypoint.root.pubspec)
         : _stripDevDependencies(entrypoint.root.pubspec);
```
jonasfj: Need to rename this `stripOverridesAndDevDeps` or we do:
`_stripDevDependencies(_stripOverrides(entrypoint.root.pubspec))`.

And fix the todo in `_stripDevDependencies` by preserving `dependency_overrides`.

sigurdm:
We decided to not handle overrides in first release.

I'll close this for now.

lib/src/command/outdated.dart
```diff
@@ -188,7 +181,16 @@ Pubspec _stripDevDependencies(Pubspec original) {
     sdkConstraints: original.sdkConstraints,
     dependencies: original.dependencies.values,
     devDependencies: [], // explicitly give empty list, to prevent lazy parsing
-    // TODO(sigurdm): consider dependency overrides.
```
jonasfj: preserve overrides

PR #2416, opened 2020-03-28
Add new command pub config

leonardoInf:
I have a added a new pub command called `pub config`. 

It allows users to provide config options which persist for future invocations of pub.

Currently only one config option (setting) is supported: `verbosity`. 
But the command is constructued in such a way that it is scalable and thus new settings (even nested ones) can be easily added without breaking the code.

On it´s core it depends on a  newly created class called `ConfigHelper`. It writes the settings to a YAML file called `bin/pub_config.yaml`

One new dependency was added to deal with the YAML: [yamlicious](https://pub.dev/packages/yamlicious).

I am willing to provide user documentation for pub.dev but I do not know how to do that yet.

Tests for this new command can be run by invoking: `pub run test test/config`.


jonasfj:
@leonardoInf, Can you elaborate the use case?

I don't envision `pub` having a lot more configuration options. I suppose something like this might be relevant for managing authorization sessions (with 3rd party pub servers).

leonardoInf:
Hello @jonasfj , actually I do.

- --[no-]precompile
- --[no-]trace
- --[no-]verbose-get
- --[no-]color

The difference to the command line options being that these settings persist over multiple invocations of pub.

And thanks for the suggestion, I will have a look at 3rd party pub server authentication

sigurdm:
@leonardoInf thanks for the pull request.

But I don't see the added complexity of adding persistent configuration as worth it.

It would make it even harder to debug the issues posted on github (we would have to ask: what are your local settings?).

It is yet another special file to keep track of and document.

I know git has config settings on many layers. Will we also need that (project local, machine local, user local).

Are there any other package managers implementing something like this?

I would rather try to make the defaults work well for most cases. And when something unusual is needed I much prefer it to be explicit on the command line invocation.

leonardoInf:
Thank you, @sigurdm 

I agree with you. I will add --enable/--disable flags to enable this feature which should prevent inexperienced users from changing the config.  

Layers are a good idea.

Every package manager I know (apt, yum, dnf, pacman, nuget) has a config file with global options.

sigurdm:
I still don't get it. What use cases are significantly improved by all this extra complexity?

leonardoInf:
I will close this for now. My initial use case was to omit the -v for easier invocation of get but apparently more functionality is needed for which I do not have time right now._

leonardoInf:
Should someone get here by accident, for me the way to go now is having `alias pub=pub .v` in my `.bashrc`

Issue #2417, opened 2020-03-30
Split transitive deps in outdated into direct/dev


mit-mit:
Per offline discussion, it would be great to split transitive deps in `pub outdated` into direct/dev:

```
[pub]$ pub outdated
Package         Current  Upgradable  Resolvable  Latest  
dependencies    
package_config  1.9.2    1.9.3       1.9.3       1.9.3  

dev_dependencies: all up-to-date

transitive dependencies: all up-to-date

transitive dev_dependencies
http_parser     3.1.3    3.1.4       3.1.4       3.1.4  

2 upgradable dependencies are locked (in pubspec.lock) to older versions.
To update these dependencies, use `pub upgrade`.
```

PR #2419, opened 2020-03-30
Split transitive dev_dependencies from other transitive dependencies

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/2417

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -74,27 +74,40 @@ class OutdatedCommand extends PubCommand {
 
     var resolvablePubspec = _stripVersionConstraints(upgradePubspec);
 
-    SolveResult upgradableSolveResult;
-    SolveResult resolvableSolveResult;
+    List<PackageId> upgradablePackages;
+    List<PackageId> resolvablePackages;
 
     await log.warningsOnlyUnlessTerminal(
       () => log.spinner(
         'Resolving',
         () async {
-          upgradableSolveResult = await resolveVersions(
+          upgradablePackages = (await resolveVersions(
             SolveType.UPGRADE,
             cache,
             Package.inMemory(upgradePubspec),
-          );
+          ))
+              .packages;
 
-          resolvableSolveResult = await resolveVersions(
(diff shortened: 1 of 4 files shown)


test/outdated/goldens/newer_versions.txt
```diff
@@ -126,6 +134,8 @@ foo           *1.2.3   *1.3.0      3.0.0       3.0.0
 transitive dependencies
 transitive    *1.2.3   2.0.0       2.0.0       2.0.0   
 
+transitive dev_dependencies
+
```
jonasfj: empty section, but it doesn't say `all up to date`

jonasfj: Also if we have  `--no-dev-dependencies` let's not show this section.

sigurdm: Hmm, both good points!

lib/src/command/outdated.dart
```diff
@@ -181,6 +194,25 @@ class OutdatedCommand extends PubCommand {
...
+      Package root, Iterable<PackageId> resolution) async {
+    final mapping =
+        Map<String, PackageId>.fromIterable(resolution, key: (id) => id.name);
+    final visited = <String>{root.name};
+    final toVisit = [...root.dependencies.keys];
+    while (toVisit.isNotEmpty) {
+      final name = toVisit.removeLast();
+      if (!visited.add(name)) continue;
+      final id = mapping[name];
+      if (id == null) print('id $name mapping $mapping');
```
jonasfj: print?

jonasfj: Throw an `AssertionError`

sigurdm: Arggg! I thought I deleted that

sigurdm: Remove

test/outdated/goldens/newer_versions.txt
sigurdm: Done, Done

PR #2422, opened 2020-03-30
Handle dependency_overrides in "pub outdated"

jonasfj:


jonasfj:
@sigurdm, this is a few hacks on top of what you originally did...

I added an option to switch between dependency overrides and not..

And I removed the bit logging what kind of override something was.

jonasfj:
Wrong branch...

PR #2423, opened 2020-03-30
Outdated overrides

jonasfj:
@sigurdm, this is a few hacks on top of what you originally did...

I added an option to switch between dependency overrides and not..

And I removed the bit logging what kind of override something was.


lib/src/command/outdated.dart
```diff
@@ -84,7 +95,7 @@ class OutdatedCommand extends PubCommand {
           upgradablePackages = (await resolveVersions(
```
sigurdm: We still need to catch the exception thrown if there is a `SolveFailure`.
Can be another CL.
I think I had that almost ready in my local branch

lib/src/command/outdated.dart
```diff
@@ -111,55 +122,83 @@ class OutdatedCommand extends PubCommand {
...
-      final latest = available.last;
-      final description = packageRef.description;
+      final current = (entrypoint.lockFile?.packages ?? {})[name];
+
+      final upgradable = upgradablePackages.firstWhere((id) => id.name == name,
+          orElse: () => null);
+      final resolvable = resolvablePackages.firstWhere((id) => id.name == name,
+          orElse: () => null);
+
+      // Find the latest version, and if it's overridden
```
sigurdm: nit: final `.`.

lib/src/command/outdated.dart
```diff
@@ -111,55 +122,83 @@ class OutdatedCommand extends PubCommand {
...
+      var latestIsOverridden = false;
+      // If present as a dependency of dev_dependency we use this ref
```
sigurdm: If present is a dependency or dev_dependency we use this ref.

lib/src/command/outdated.dart
```diff
@@ -111,55 +122,83 @@ class OutdatedCommand extends PubCommand {
...
+      var latest = await _getLatest(rootPubspec.dependencies[name]?.toRef());
+      latest ??= await _getLatest(rootPubspec.devDependencies[name]?.toRef());
+      // If not overridden in current resolution we can use this ref
+      if (!entrypoint.root.pubspec.dependencyOverrides.containsKey(name)) {
```
sigurdm: Maybe start with this, this is the simplest case...

lib/src/command/outdated.dart
```diff
@@ -111,55 +122,83 @@ class OutdatedCommand extends PubCommand {
...
+        latest ??= await _getLatest(current?.toRef());
+      }
+      // If not overridden and present in either upgradable or resolvable we
+      // use this reference to find the latest
+      if (!upgradablePubspec.dependencyOverrides.containsKey(name)) {
```
sigurdm: The dependency overrides of upgradeblePubspec and resolvablePubspec  will always be either empty or the same as for entrypoint.root?

```
if (!entrypoint.root.pubspec.dependencyOverrides.containsKey(name)) {
  latest ??= await _getLatest(current?.toRef()) ?? await _getLatest(upgradable?.toRef()) ?? await _getLatest(resolvable?.toRef());
}
```

And we will always get a version if the ref is non-null (is this true??)

```
if (!entrypoint.root.pubspec.dependencyOverrides.containsKey(name)) {
  latest ??= await _getLatest((current ?? upgradable ?? resolvable)?.toRef());
}
```
And I guess we always have at least one of the candidates:
```
if (!entrypoint.root.pubspec.dependencyOverrides.containsKey(name)) {
  latest ??= await _getLatest((current ?? upgradable ?? resolvable).toRef());
}
```

lib/src/command/outdated.dart
```diff
@@ -185,14 +224,42 @@ class OutdatedCommand extends PubCommand {
     }
   }
 
+  /// Get the latest version of [ref] given its [PackageRef].
```
sigurdm: given its source

lib/src/command/outdated.dart
```diff
@@ -185,14 +224,42 @@ class OutdatedCommand extends PubCommand {
...
+    try {
+      final available = await cache.source(ref.source).getVersions(ref);
+      if (available.isEmpty) {
+        return null;
+      }
+      available.sort(argResults['pre-releases']
+          ? (x, y) => x.version.compareTo(y.version)
+          : (x, y) => Version.prioritize(x.version, y.version));
+      return available.last;
+    } on PackageNotFoundException {
```
sigurdm: When do we get this? Can we avoid this, I think it means we are asking about a bad package.

lib/src/command/outdated.dart
```diff
@@ -111,55 +122,83 @@ class OutdatedCommand extends PubCommand {
...
+      if (!resolvablePubspec.dependencyOverrides.containsKey(name)) {
+        latest ??= await _getLatest(resolvable?.toRef());
+      }
+      // Otherwise, we might simply not have a latest, when a transitive
+      // dependency is overridden the source can depend on which versions we
+      // are picking. This is not a problem on `pub.dev` because it does not
+      // allow 3rd party pub servers, but other servers might. Hence, we choose
+      // to fallback to using the overridden source for latest.
+      if (latest == null) {
+        latest ??= await _getLatest(current?.toRef());
```
sigurdm: Again we might be able to say:
```
latest ??= await _getLatest((current ?? upgradable ?? resolvable)?.toRef());
```

lib/src/command/outdated.dart
```diff
@@ -185,14 +224,42 @@ class OutdatedCommand extends PubCommand {
     }
   }
 
+  /// Get the latest version of [ref] given its [PackageRef].
+  ///
+  /// Returns `null`, if unable to find the package.
+  Future<PackageId> _getLatest(PackageRef ref) async {
```
sigurdm: As far as I can see you always call this with `id.toRef()`.
Maybe it could be
```
Future<PackageId> _getLatestFromSameSource(PackageId id) {
  if (id == null) return null;
  ref = id.toRef();
  ...
}
```

lib/src/command/outdated.dart
jonasfj: yes, but `rootPubspec.dependencies` felt more right..

lib/src/command/outdated.dart
jonasfj: `current` can have overrides without `upgradable` and `resolvable` having overrides..

In any case, I figured it was preferable not to read the `argResult['dependency-overrides']` after creating the pubspecs for `upgradable` and `resolvable`.

Hence, I opted to derive whether or not it was overridden from the `(upgradable|resolvable)Pubspec` -- this has the benefit that if we change how overrides work in the future, or where they are applied it doesn't not affect this code.

lib/src/command/outdated.dart
jonasfj: > Can we avoid this, I think it means we are asking about a bad package.

Maybe we make an `assert(false)` here and write, this should not be possible.
But in production I would prefer to handle the case..

I think this can happen with `--no-dependency-override`, if you wrote a package that doesn't exist in the `dependencies:` section of your `pubspec.yaml` -- and then proceeded to override it.

Actually, you're probably right we should not handle this.

jonasfj:
@sigurdm, ptal

lib/src/command/outdated.dart
```diff
@@ -536,10 +570,10 @@ class _PackageDetails implements Comparable<_PackageDetails> {
   Map<String, Object> toJson() {
     return {
       'package': name,
-      'current': current?.describe,
-      'upgradable': upgradable?.describe,
-      'resolvable': resolvable?.describe,
-      'latest': latest?.describe,
+      'current': {'version': current?.describe},
```
sigurdm: Can we make a json version of `describe` that includes the override information in a structured way?

lib/src/command/outdated.dart
```diff
@@ -341,9 +440,20 @@ Future<void> _outputHuman(List<_PackageDetails> rows,
...
+    // Note. we don't care if resolvableFailed, because if it has failed then
+    // the upgradable resolution has also failed. A single mesage about what to
+    // do should be sufficient -- and running `pub upgrade --dry-run` should
+    // given the user a solid explanation.
+  }
+
   if (upgradable != 0) {
     if (upgradable == 1) {
-      log.message('1 upgradable dependency is locked (in pubspec.lock) to '
+      log.message('\n1 upgradable dependency is locked (in pubspec.lock) to '
```
sigurdm: Good catch

lib/src/command/outdated.dart
```diff
@@ -67,101 +72,123 @@ class OutdatedCommand extends PubCommand {
...
+          current,
+          entrypoint.root.pubspec.dependencyOverrides.containsKey(name),
+        ),
+        await _describeVersion(
+          upgradable,
+          upgradablePubspec.dependencyOverrides.containsKey(name),
+        ),
+        await _describeVersion(
+          resolvable,
+          resolvablePubspec.dependencyOverrides.containsKey(name),
```
sigurdm: We should name this bool parameter

lib/src/command/outdated.dart
```diff
@@ -341,9 +440,20 @@ Future<void> _outputHuman(List<_PackageDetails> rows,
           row.upgradable != row.resolvable)
       .length;
 
+  if (upgradableFailed) {
+    log.message(
+      '\nUpgradable package resolution could not be computed.\n'
+      'Use `pub upgrade --dry-run` for an explanation.',
```
sigurdm: One will need to remove overrides before this works.

lib/src/command/outdated.dart
```diff
@@ -341,9 +440,20 @@ Future<void> _outputHuman(List<_PackageDetails> rows,
           row.upgradable != row.resolvable)
       .length;
 
+  if (upgradableFailed) {
+    log.message(
+      '\nUpgradable package resolution could not be computed.\n'
```
sigurdm: How about: `'\nNo upgrade resolution with no dependency_overrides.\n'

lib/src/command/outdated.dart
```diff
@@ -263,9 +358,13 @@ Future<void> _outputJson(List<_PackageDetails> rows) async {
...
 
-Future<void> _outputHuman(List<_PackageDetails> rows,
-    Future<List<_FormattedString>> Function(_PackageDetails) marker,
-    {@required bool useColors, @required bool includeDevDependencies}) async {
+Future<void> _outputHuman(
+  List<_PackageDetails> rows,
+  Future<List<_FormattedString>> Function(_PackageDetails) marker, {
+  @required bool useColors,
+  @required bool includeDevDependencies,
+  @required bool upgradableFailed,
```
sigurdm: This seems too specific.

Perhaps we could take a list of warnings to print?

(Current solution is fine for now)

lib/src/command/outdated.dart
jonasfj: I wanted to be, but then lines will wrap :(

lib/src/command/outdated.dart
jonasfj: from chat, we agreed to drop it.

lib/src/command/outdated.dart
sigurdm: You could assign the override status to a named variable before:
```
final resolvableIsOverridden = resolvablePubspec.dependencyOverrides.containsKey(name);
...
await _describeVersion( resolvable, overridden: resolvableIsOverridden);
```

Issue #2418, opened 2020-03-30
pub outdated --json should have a map per package-version


sigurdm:
```
{
  "packages": [
    {
      "package": "bar",
      "current": "1.0.1",
      "upgradable": "2.0.0",
      "resolvable": "2.0.0",
      "latest": "2.0.0"
    },
  ]
}
```
should be
```
{
  "packages": [
    {
      "package": "bar",
      "current": {"version": "1.0.1"},
      "upgradable": {"version: "2.0.0"},
      "resolvable": {"version": "2.0.0"},
      "latest": {"version": "2.0.0"}
    },
  ]
}
```

To make it easier to add more information later (eg. source, broken, overridden, discontinued ...)

PR #2424, opened 2020-03-30
Make `pub outdated --json` version info an object

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2418

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -434,10 +434,10 @@ class _PackageDetails implements Comparable<_PackageDetails> {
   Map<String, Object> toJson() {
     return {
       'package': name,
-      'current': current?.version?.toString(),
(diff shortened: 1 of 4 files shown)


sigurdm:
cc @kevmoo: you expressed interest in the json format

kevmoo:
Insert gif: "But why?"

What does this help with? How does this improve things?

I already implemented my parser for this. I'm happy to change it, but I don't see the upside here...

jonasfj:
@kevmoo, I think we envision additional meta-data being attached in the future.

Like whether or not a given version is null-safety compatible, reported as broken, or perhaps even reported to have a vulnerability.

By creating the format as an object we can add such properties to each version without breaking existing integrations that rely on the JSON format.

kevmoo:
Totally with you there!

PR #2425, opened 2020-03-31
Normalize command names when invoked via alias

kevmoo:
Since https://github.com/dart-lang/pub/commit/ab109723d9eb782c2ea4cce68c044da40403a652
we are seeing `install` and `update` in telemetry, which is weird. (These map to `get` and `upgrade` respectively.)

This commit ensures telemetry headers only see each command with its canonical name.

### DIFF
--- a/lib/src/command.dart
+++ b/lib/src/command.dart
@@ -10,6 +10,17 @@ import 'global_packages.dart';
 import 'log.dart' as log;
 import 'system_cache.dart';
 
+/// All of the aliases used for [PubCommand] subclasses.
+///
+/// Centralized so invocations with aliases can be normalized to the primary
+/// command name when sending telemetry.
+const pubCommandAliases = {
+  'deps': ['dependencies', 'tab'],
+  'get': ['install'],
+  'publish': ['lish', 'lush'],
+  'upgrade': ['update'],
+};
+
 /// The base class for commands for the pub executable.
 ///
 /// A command may either be a "leaf" command or it may be a parent for a set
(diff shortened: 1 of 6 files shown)


kevmoo:
Thanks @jonasfj 

See also https://dart-review.googlesource.com/c/sdk/+/141900

lib/src/command.dart
```diff
@@ -10,6 +10,17 @@ import 'global_packages.dart';
 import 'log.dart' as log;
 import 'system_cache.dart';
 
+/// All of the aliases used for [PubCommand] subclasses.
+///
+/// Centralized so invocations with aliases can be normalized to the primary
+/// command name when sending telemetry.
+const pubCommandAliases = {
```
sigurdm, line 17: Nit: this table could have had the other direction:
```
cons pubCommandAliases = {
  'dependencies': 'deps',
  'tab': 'deps',
  
}
```

lib/src/command_runner.dart
```diff
@@ -45,14 +46,28 @@ class PubCommandRunner extends CommandRunner {
...
     if (_options == null) return '';
 
     var list = <String>[];
     for (var command = _options.command;
         command != null;
         command = command.command) {
-      list.add(command.name);
+      var commandName = command.name;
+
+      if (list.isEmpty) {
```
sigurdm, line 61: Nit: This could have been before the for loop - that's when the list is empty.

lib/src/command.dart
kevmoo: True. Since we need to map both ways, I figured I'd keep the existing structure so it's easy to eyeball the mapping from command -> alias

lib/src/command_runner.dart
kevmoo: True, but I wanted to keep a single place where we add the command to the list. At least I felt that made the code easier to reason about.

Issue #2428, opened 2020-04-03
pub outdated help text needs line breaks


kwalrath:
```
$ pub version
Pub 2.8.0-dev.18.0
```

OS version: macOS 10.14.6 (Mojave)

# Problem
When I run `pub outdated help`, the `--format` help is much longer than 80 characters. Other lines are also longer than 80 characters. Also, we use "prerelease" (no hyphen) in our docs. Here's the current output:

```
$ pub outdated --help
Analyze your dependencies to find which ones can be upgraded.

Usage: pub outdated [options]
-h, --help                     Print this usage information.
    --format=<FORMAT>          Defines how the output should be formatted. Defaults to color when connected to a terminal, and no-color otherwise.
                               [color, no-color, json]
    --[no-]up-to-date          Include dependencies that are already at the latest version
    --[no-]pre-releases        Include pre-releases when reporting latest version
    --[no-]dev-dependencies    When true take dev-dependencies into account when resolving.
                               (defaults to on)
    --mark=<OPTION>            Highlight packages with some property in the report.
                               [outdated (default), none]

Run "pub help" to see global options.
See https://dart.dev/tools/pub/cmd/pub-outdated for detailed documentation.
```

Here's one way to fix it (note the `--[no-]preleases` name change, which would be nice):

```
$ pub outdated --help
Analyze your dependencies to find which ones can be upgraded.

Usage: pub outdated [options]
-h, --help                     Print this usage information.
    --format=<FORMAT>          Define how the output should be formatted.
                               Default is color when connected to a terminal,
                               and no-color otherwise.
                               [color, no-color, json]
    --[no-]up-to-date          Include dependencies that are already at the
                               latest version. (Default is no.)
    --[no-]prereleases         Include prereleases when reporting the latest
                               version. (Default is no.)
    --[no-]dev-dependencies    Take dev dependencies into account (the default).
    --mark=<OPTION>            Highlight packages that have the specified
                               property.
                               [outdated (default), none]

Run "pub help" to see global options.
See https://dart.dev/tools/pub/cmd/pub-outdated for detailed documentation.
```

PR #2429, opened 2020-04-03
Improve the `pub outdated --help` text

sigurdm:
Fix of https://github.com/dart-lang/pub/issues/2428

I have made the --help text be 80 columns.

@kwalrath The text and arguments has already changed quite a bit since the version you reviewed.
Could you look it through again?

I made a golden-file test of the help text, so we will notice changes.

The golden file (test/outdated/golden/helptext.txt) in this pr can be used for reviewing this.,

I have not changed how the defaults are printed as suggested in the issue.

The (default is on) comes from package:args. And if we want to change them, we should probably change them there. The 'default default' is off, so I guess that is why package:args doesn't print anything in that case.

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -32,9 +32,7 @@ class OutdatedCommand extends PubCommand {
 
   OutdatedCommand() {
     argParser.addFlag('json',
-        help: 'Outputs the results in a json formatted\n'
-            'report.',
-        negatable: false);
+        help: 'Outputs the results using a json format.', negatable: false);
 
     argParser.addFlag('color',
         help: 'Whether to color the output.\n'
@@ -47,9 +45,7 @@ class OutdatedCommand extends PubCommand {
             'latest version.');
 
     argParser.addFlag('prereleases',
-        defaultsTo: false,
-        help: 'Include prereleases when reporting latest\n'
-            'version.');
+        defaultsTo: false, help: 'Include prereleases in latest version.');
 
     // Preserve for backwards compatibility.
     argParser.addFlag('pre-releases',
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -31,32 +31,40 @@ class OutdatedCommand extends PubCommand {
   String get docUrl => 'https://dart.dev/tools/pub/cmd/pub-outdated';
 
   OutdatedCommand() {
-    argParser.addFlag('color',
-        help: 'Whether to color the output. Defaults to color '
-            'when connected to a terminal, and no-color otherwise.');
-
     argParser.addFlag('json',
-        help: 'Outputs the results in a json formatted report',
+        help: 'Outputs the results in a json formatted\n'
+            'report.',
         negatable: false);
+
+    argParser.addFlag('color',
+        help: 'Whether to color the output.\n'
+            'Defaults to color when connected to a\n'
+            'terminal, and no-color otherwise.');
 
     argParser.addFlag('up-to-date',
         defaultsTo: false,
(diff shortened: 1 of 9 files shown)


sigurdm:
cc @mit-mit

lib/src/command/outdated.dart
```diff
@@ -31,32 +31,40 @@ class OutdatedCommand extends PubCommand {
...
-            'when connected to a terminal, and no-color otherwise.');
-
     argParser.addFlag('json',
-        help: 'Outputs the results in a json formatted report',
+        help: 'Outputs the results in a json formatted\n'
+            'report.',
         negatable: false);
 
+    argParser.addFlag('color',
+        help: 'Whether to color the output.\n'
```
mit-mit, line 40: Maybe "Whether to use color in the output.\n"?

lib/src/command/outdated.dart
```diff
@@ -31,32 +31,40 @@ class OutdatedCommand extends PubCommand {
...
         defaultsTo: false,
-        help: 'Include dependencies that are already at the latest version.');
+        help: 'Include dependencies that are already at the\n'
+            'latest version.');
 
-    argParser.addFlag('pre-releases',
+    argParser.addFlag('prereleases',
         defaultsTo: false,
-        help: 'Include pre-releases when reporting latest version.');
+        help: 'Include prereleases when reporting latest\n'
```
mit-mit, line 51: Maybe shorten to "Include pre-releases in latest version."

lib/src/command/outdated.dart
```diff
@@ -31,32 +31,40 @@ class OutdatedCommand extends PubCommand {
   String get docUrl => 'https://dart.dev/tools/pub/cmd/pub-outdated';
 
   OutdatedCommand() {
-    argParser.addFlag('color',
-        help: 'Whether to color the output. Defaults to color '
-            'when connected to a terminal, and no-color otherwise.');
-
     argParser.addFlag('json',
-        help: 'Outputs the results in a json formatted report',
+        help: 'Outputs the results in a json formatted\n'
```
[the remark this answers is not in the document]
sigurdm: Done

lib/src/command/outdated.dart
sigurdm: Done

lib/src/command/outdated.dart
sigurdm: Done

lib/src/command/outdated.dart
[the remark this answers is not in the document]
mit-mit: "in a json format" not "in a json formatted"

test/outdated/goldens/helptext.txt
```diff
@@ -0,0 +1,22 @@
+$ pub outdated --help
+Analyze your dependencies to find which ones can be upgraded.
+
+Usage: pub outdated [options]
+-h, --help                         Print this usage information.
+    --json                         Outputs the results in a json formatted
+                                   report.
```
kwalrath, line 7:
```suggestion
    --json                         Output the results in JSON format.
```

When the helps starts with a verb, it should be consistently either 3rd person (with an 's' at the end, generally) or 2nd person (no 's'). Most of these commands use 2nd person, so let's do that here.

sigurdm:
I don't think the current order is important
Alphabetical seems fine. Done

test/outdated/goldens/helptext.txt
sigurdm: Done

natebosch:
The args package supports wrapping. Why are we trying to manually line wrap these?

PR #2435, opened 2020-04-06
pub run --v2 for use in dartdev

jonasfj:
@jwren, if the first argument given to `dartdev run` matches `[<package>[:<command>]]` where `<package>` is `[a-z_][a-z0-9_]*` and `<command>` is `[^/]+` then we forward to `pub run --v2` (as introduced here).

Long term, we have to find a way to embed this properly and move the entire _run_ related logic into a single place. Maybe `pub` just exports some functions for asserting that `pubspec.yaml` is installed, and what packages are installed, and whether they are immutable.. I don't think we have to clean this up for an MVP.

### DIFF
--- a/lib/src/command/run.dart
+++ b/lib/src/command/run.dart
@@ -101,16 +101,17 @@ class RunCommand extends PubCommand {
   /// If neither `package` or `command` is given and `command` with name of
   /// the current package doesn't exist we fallback to `'main'`.
   ///
-  /// Runs `bin/<command>.dart` from package `<package>`, with snapshot stored
-  /// in `.dart_tool/pub/bin/<package>/<command>.dart.snapshot.dart2`.
+  /// Runs `bin/<command>.dart` from package `<package>`. If `<package>` is not
+  /// mutable (local root package or path-dependency) a source snapshot will be
+  /// cached in `.dart_tool/pub/bin/<package>/<command>.dart.snapshot.dart2`.
   Future _runV2() async {
     var package = entrypoint.root.name;
     var command = package;
     var args = <String>[];
 
     if (argResults.rest.isNotEmpty) {
-      if (argResults.rest[0].contains('/')) {
-        usageException('[<package>[:command]] cannot contain "/"');
+      if (argResults.rest[0].contains(RegExp(r'[/\\]'))) {
+        usageException('[<package>[:command]] cannot contain "/" or "\\"');
       }
 
       package = argResults.rest[0];
--- a/lib/src/command/run.dart
+++ b/lib/src/command/run.dart
@@ -29,10 +29,15 @@ class RunCommand extends PubCommand {
     argParser.addFlag('enable-asserts', help: 'Enable assert statements.');
     argParser.addFlag('checked', abbr: 'c', hide: true);
     argParser.addOption('mode', help: 'Deprecated option', hide: true);
+    // mode exposed for `dartdev run` to use as subprocess.
+    argParser.addFlag('v2', hide: true);
   }
 
   @override
   Future run() async {
+    if (argResults['v2']) {
+      return await _runV2();
+    }
     if (argResults.rest.isEmpty) {
       usageException('Must specify an executable to run.');
     }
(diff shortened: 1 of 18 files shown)


lib/src/command/run.dart
```diff
@@ -29,10 +29,15 @@ class RunCommand extends PubCommand {
     argParser.addFlag('enable-asserts', help: 'Enable assert statements.');
     argParser.addFlag('checked', abbr: 'c', hide: true);
     argParser.addOption('mode', help: 'Deprecated option', hide: true);
+    // mode exposed for `dartdev run` to use as subprocess.
```
natebosch, line 32: Can we find a more descriptive name? Maybe `--as-subprocess` or something?

lib/src/command/run.dart
```diff
@@ -86,4 +91,98 @@ class RunCommand extends PubCommand {
...
   }
+
+  /// Implement a v2 mode for use in `dartdev run`.
+  ///
+  /// Usage: `dartdev run [package[:command]]`
+  ///
+  /// If `package` is not given, defaults to current root package.
+  /// If `command` is not given, defaults to name of `package`.
+  /// If neither `package` or `command` is given and `command` with name of
+  /// the current package doesn't exist we fallback to `'main'`.
```
natebosch, line 102: Why do we need the magic fallback to `'main'`? Do we have templates that add a `bin/main.dart` by default?

I worry that adding something new - `bin/<package>.dart` changes the behavior of `pub run`. So if I've had a `bin/main.dart` and have been happily working, adding a new file gives new different magic behavior...

What does the error message look like if you `pub run` and there is no `bin/<package.dart` or `bin/main.dart`? Does it complain about missing `main.dart` only? That could encourage more people to add that, and if people publish those would the workflow end up looking like `pub run some_package:main` for most use cases? Or does `pub run some_package` also run` bin/main.dart` in that package?

lib/src/command/run.dart
```diff
@@ -86,4 +91,98 @@ class RunCommand extends PubCommand {
...
+  ///
+  /// Runs `bin/<command>.dart` from package `<package>`, with snapshot stored
+  /// in `.dart_tool/pub/bin/<package>/<command>.dart.snapshot.dart2`.
```
natebosch, line 105: Is this comment incomplete, or does this also want to run as a snapshot for the local package or a path dependency?

lib/src/command/run.dart
```diff
@@ -86,4 +91,98 @@ class RunCommand extends PubCommand {
...
+  Future _runV2() async {
+    var package = entrypoint.root.name;
+    var command = package;
+    var args = <String>[];
+
+    if (argResults.rest.isNotEmpty) {
+      if (argResults.rest[0].contains('/')) {
```
natebosch, line 112: there are likely many characters it can't contain - why is `/` special? Do we need to check for `\` on windows?

test/run/v2/app_can_read_from_stdin_test.dart
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
natebosch, line 1: [nit] 2020 - same in many other files

jonasfj:
> LGTM. I suspect that after it lands and you think the new capability is what you'll remove the v2 option and it will just be the new default?

The plan is to use `pub run --v2` when implementing `dartdev run`, eventually the `pub` command will go away and we will only have the `pub run --v2` behavior, as it's exposed through `dartdev run`.

lib/src/command/run.dart
jonasfj: Or we could say `--dartdev-mode`

lib/src/command/run.dart
jonasfj: It only does the fallback if `bin/main.dart` exists. @mit-mit  wanted it because he wanted `dartdev run` without any arguments to work with out existing templates. I see that as valuable, but on the other hand I agree that using `bin/main.dart` will give a bad workflow in other cases.

I think the reasoning was that if you make a console app that is not distributed on `pub.dev`, using `bin/main.dart` is what our templates have done for years, and the drawbacks to supporting that are fairly slim.
IMO, people aren't going to notice that `bin/main.dart` works, because it's only a fallback that is used when:
 * calling without arguments, and,
 * `bin/<packageName>.dart` does exist, and,
 * `bin/main.dart` does exist.

lib/src/command/run.dart
jonasfj: We could probably check more. We could disallow all characters not allowed in filenames, but that's probably not necessary as checking if a file with an illegal filename exists hopefully returns false.

But checking for slash and backslash is nice, that way we avoid going into sub-folders by accident.

jonasfj:
With this we can design `dartdev run <arg> [...]` such that if `<arg>` match `[package[:command]]` we forward the call to `pub run --v2 <arg> [...]`.

And if `<arg>` doesn't match `[package[:command]]` we assume `<arg> [...]` is meant to be a options and file-path to be forwarded to the dartvm. Thus, we would spawn `dart <arg> [...]`.

This means that `dartdev run --observe=8080 path/to/my-file.dart` works, because `--observe=8080` doesn't match the `[package[:command]]` pattern.

However, it also means that `dartdev run path/to/file.dart` will not go through `pub run --v2`, which means we will run the `dart path/to/file.dart` without checking if dependencies from `pubspec.yaml` are present.

IMO, this is fine for an MVP. But I can see how long-term we would want to:
 * check that dependencies are installed when doing `dartdev run path/to/file.dart`, and
 * allow pass-through of VM arguments (such as `--observe`) when doing `dartdev run [package[:command]]`.

lib/src/command/run.dart
natebosch: I think as long as the error message if none off the fallbacks exist mentions `bin/<packageName>.dart`, and does _not_ mention `bin/main.dart` I think it's probably OK.

I hope to avoid new things that would subtly encourage that pattern, but I'm fine with attempting to keep compatibility with the other places that were already encouraging that pattern.

mit-mit: What advantages does `bin/<packageName>.dart` have over `bin/main.dart`?

Flutter apps seem to use `lib/main.dart`, hmm...

mit-mit: @jonasfj you meant "bin/<packageName>.dart does NOT exist, and" right?

natebosch: > What advantages does `bin/<packageName>.dart` have over `bin/main.dart`?

It's what runs when you do `pub run <packageName>`, `pub run test` runs `bin/test.dart` from the test package. When publishing a package that is meant to be run this way this is the naming scheme that should be used.

> Flutter apps seem to use `lib/main.dart`, hmm...

Flutter ignored our conventions before we had an opportunity to add in hints and stuff and now we are stuck with it. FWIW `lib/main.dart` will almost always be flutter, which means it will almost never work with `pub run`.

jonasfj: > I hope to avoid new things that would subtly encourage that pattern, but I'm fine with attempting to keep compatibility with the other places that were already encouraging that pattern.

I feel exactly the same way. Which is also why `bin/main.dart` will not be used if you do `dartdev run <packageName>` -- it only works for the local root package, ie. `dartdev run` (without arguments).

mit-mit: I'm open towards changing what the recommended file name is for the entrypoint; we can always change the template if needed.

But remember that we have two different types of packages -- library packages (which are used by apps), and app packages (the CLI, web, mobile app being built). Do we think `bin/<packageName>.dart` is a better choice for apps?

jonasfj: > Do we think `bin/<packageName>.dart` is a better choice for apps?

It's the current convention for apps, example: [mono_repo](https://github.com/dart-lang/mono_repo/tree/master/mono_repo/bin), [build_runner](https://github.com/dart-lang/build/tree/master/build_runner/bin), [test](https://github.com/dart-lang/test/tree/master/pkgs/test/bin), [webdev](https://github.com/dart-lang/webdev/tree/master/webdev/bin), [dartdoc](https://github.com/dart-lang/dartdoc/tree/master/bin), [grinder](https://github.com/google/grinder.dart/tree/master/bin),  [tota](https://github.com/tota-dart/tota/tree/master/bin), [dart_up](https://github.com/regiostech/dart_up/tree/master/bin), [lefthook](https://github.com/drt-pkgz/lefthook/tree/master/bin) -- this was just looking at package that use `cli_util`.

The only counter example I found was [fvm](https://github.com/leoafarias/fvm/tree/master/bin), which used `bin/main.dart`.

mit-mit: I guess my only concern is that I don't understand *why* `<packageName>.dart` was originally selected? Using `main` as the main entry point is pretty common across a range of languages. Does it have some functional benefit that I'm not seeing?

sigurdm: I never liked/understood that convention either.
Neither for the main library (`lib/<packageName>.dart`) nor binary (`bin/packageName>.dart`).
There is too much repetition.

jonasfj: Python has `__init__.py`, node has `index.js`, rust has `mod.rs`, I don't know that `<packageName>.dart` is such a bad choice.

sigurdm: I guess this is not the right place to take this discussion.
...
But <packageName>.dart is bad (ok, let's say it has drawbacks) for at least the following reasons:

* if we rename the package we have to rename the main file, making this a more difficult operation (also making it even harder than it already is to vendor a package inside another - a use case I know you care about) 

* it made this feature (that just seems so obvious): https://github.com/dart-lang/language/issues/649 be non-trivial to specify.

I guess it comes down down to composability. - why would the name of the main library have to depend on the surrounding package?

natebosch: > Which is also why `bin/main.dart` will not be used if you do `dartdev run <packageName>` -- it only works for the local root package, ie. `dartdev run` (without arguments).

This is what I worry about, but maybe I'm imagining problems that don't exist.

What I'm worried is that `pub run` is conceptually a shorthand for "Run this package" becomes associated with `bin/main.dart` because that works. We know that `pub run <package>` is shorthand for "Run that named package". But now there is a difference in behavior between "Run this package" and "Run that package".


Regarding `lib/<package>.dart` - it's true that it's somewhat unnecessary repetition, but FWIW I think `import 'package:something/lib.dart';` would end up with complaints about weird conventions too.

I don't think at this point its worth the churn to try to change it.

test/test_pub.dart
```diff
@@ -210,6 +210,20 @@ Future<PubProcess> pubRun({bool global = false, Iterable<String> args}) async {
...
 }
 
+/// Schedules starting the "pub run --v2" process and validates the
+/// expected startup output.
+///
+/// Returns the `pub run` process.
+Future<PubProcess> pubRunV2({Iterable<String> args}) async {
+  final pub = await startPub(args: ['run', '--v2', ...args]);
+
+  // Loading sources and transformers isn't normally printed, but the pub test
```
natebosch, line 220: transformers have been removed. Was this copied from somewhere? Is it still necessary?

jonasfj: Blatantly stolen from `pubRun` above, but yes, it would be interesting to see if this is necessary.

PR #2444, opened 2020-04-09
Show summary of outdated packages after running `pub upgrade` 

### LINKED ISSUE #2434: 'pub upgrade' should show summary of outdated packages
Running `pub upgrade` it would be nice if it showed the number of packages that could be upgraded but wasn't.. and suggested the user try `pub outdated` for more information.

In other words, if there are packages that are still outdated after running `pub upgrade` we should suggest the user runs `pub outdated` to learn if these are resolvable.

```
$ pub upgrade
Resolving dependencies... (11.1s)
  _fe_analyzer_shared 2.0.0
  analyzer 0.39.5
  args 1.6.0
  async 2.4.1
  boolean_selector 2.0.0
  charcode 1.1.3
  collection 1.14.12
  convert 2.1.1
  coverage 0.13.7 (0.13.9 available)
  crypto 2.1.4
  csslib 0.16.1
  glob 1.2.0
  html 0.14.0+3
  http 0.12.0+4
  http_multi_server 2.2.0
  http_parser 3.1.4
  io 0.3.4
  js 0.6.1+1
  logging 0.11.4
  matcher 0.12.6
  meta 1.1.8
  mime 0.9.6+3
  multi_server_socket 1.0.2
  node_interop 1.0.3
  node_io 1.0.1+2
  node_preamble 1.4.8
  package_config 1.1.0 (1.9.3 available)
  package_resolver 1.0.10
  path 1.6.2 (1.6.4 available)
  pedantic 1.9.0
  pool 1.4.0
  pub_semver 1.4.4
  shelf 0.7.5
  shelf_packages_handler 1.0.4 (2.0.0 available)
  shelf_static 0.2.8
  shelf_web_socket 0.2.3
  source_map_stack_trace 1.1.5 (2.0.0 available)
  source_maps 0.10.9
  source_span 1.7.0
  stack_trace 1.9.3
  stream_channel 2.0.0
  string_scanner 1.0.5
  term_glyph 1.1.0
  test 1.13.0 (1.14.2 available)
  test_api 0.2.15
  test_core 0.3.1 (0.3.3 available)
  typed_data 1.1.6
  vm_service 3.0.0+1 (4.0.0 available)
  watcher 0.9.7+14
  web_socket_channel 1.1.0
  webkit_inspection_protocol 0.5.0+1
  yaml 2.2.0
No dependencies changed.
```

Maybe we could add two lines: (or something like that)
```
8 packages have newer versions incompatible with dependency constraints.
Try 'pub outdated' for more information.
```

ayan-b:
Which is shown if outdated packages are detected while
running upgrade.

Closes #2434

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -246,7 +246,6 @@ class Entrypoint {
 
     if (dryRun) {
       result.summarizeChanges(type, dryRun: dryRun);
-      result.showOutdated(type);
       return;
     }
 
@@ -254,7 +253,6 @@ class Entrypoint {
     _saveLockFile(result);
 
     result.summarizeChanges(type, dryRun: dryRun);
-    result.showOutdated(type);
 
     /// Build a package graph from the version solver results so we don't
     /// have to reload and reparse all the pubspecs.
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -246,6 +246,7 @@ class Entrypoint {
 
     if (dryRun) {
       result.summarizeChanges(type, dryRun: dryRun);
+      if (type == SolveType.UPGRADE) result.showOutdated(type);
       return;
     }
 
@@ -253,6 +254,7 @@ class Entrypoint {
     _saveLockFile(result);
 
     result.summarizeChanges(type, dryRun: dryRun);
+    if (type == SolveType.UPGRADE) result.showOutdated(type);
 
     /// Build a package graph from the version solver results so we don't
     /// have to reload and reparse all the pubspecs.
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -246,7 +246,7 @@ class Entrypoint {
 
     if (dryRun) {
       result.summarizeChanges(type, dryRun: dryRun);
-      if (type == SolveType.UPGRADE) result.showOutdated(type);
+      result.showOutdated(type);
       return;
     }
 
@@ -254,7 +254,7 @@ class Entrypoint {
     _saveLockFile(result);
 
     result.summarizeChanges(type, dryRun: dryRun);
-    if (type == SolveType.UPGRADE) result.showOutdated(type);
+    result.showOutdated(type);
 
     /// Build a package graph from the version solver results so we don't
     /// have to reload and reparse all the pubspecs.
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -130,6 +130,39 @@ class SolveReport {
     }
   }
 
+  /// Displays a two-line message, number of outdated packages and an
+  /// instruction to run `pub outdated` if outdated packages are detected.
+  void reportOutdated() {
+    var outdatedPackagesCount = 0;
+
+    final names = _result.packages.map((id) => id.name);
+    for (var name in names) {
+      var newId = _dependencies[name];
+
+      // See if there are any newer versions of the package that we were
+      // unable to upgrade to.
+      if (newId != null) {
+        var versions = _result.availableVersions[newId.name];
+
+        if (versions.any((v) =>
+            v > newId.version &&
+            (newId.version.isPreRelease == true || v.isPreRelease == false))) {
+          ++outdatedPackagesCount;
+        }
+      }
+    }
+    if (outdatedPackagesCount > 0) {
+      String packageCountString;
+      if (outdatedPackagesCount == 1) {
+        packageCountString = '1 package has';
+      } else {
+        packageCountString = '$outdatedPackagesCount packages have';
+      }
+      log.message('$packageCountString newer versions incompatible with '
+          'dependency constraints.\nTry `pub outdated` for more information.');
+    }
+  }
+
   /// Reports the results of the upgrade on the package named [name].
   ///
   /// If [alwaysShow] is true, the package is reported even if it didn't change,
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -133,24 +133,17 @@ class SolveReport {
   /// Displays a two-line message, number of outdated packages and an
   /// instruction to run `pub outdated` if outdated packages are detected.
   void reportOutdated() {
-    var outdatedPackagesCount = 0;
+    final outdatedPackagesCount = _result.packages.where((id) {
+      final versions = _result.availableVersions[id.name];
 
-    final names = _result.packages.map((id) => id.name);
-    for (var name in names) {
-      var newId = _dependencies[name];
-
-      // See if there are any newer versions of the package that we were
-      // unable to upgrade to.
-      if (newId != null) {
-        var versions = _result.availableVersions[newId.name];
-
-        if (versions.any((v) =>
-            v > newId.version &&
-            (newId.version.isPreRelease == true || v.isPreRelease == false))) {
-          ++outdatedPackagesCount;
-        }
+      if (versions.any((v) =>
+          v > id.version &&
+          (id.version.isPreRelease == true || v.isPreRelease == false))) {
+        return true;
       }
-    }
+      return false;
+    }).length;
+
     if (outdatedPackagesCount > 0) {
       String packageCountString;
       if (outdatedPackagesCount == 1) {
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -135,13 +135,9 @@ class SolveReport {
   void reportOutdated() {
     final outdatedPackagesCount = _result.packages.where((id) {
       final versions = _result.availableVersions[id.name];
-
-      if (versions.any((v) =>
+      return versions.any((v) =>
           v > id.version &&
-          (id.version.isPreRelease == true || v.isPreRelease == false))) {
-        return true;
-      }
-      return false;
+          (id.version.isPreRelease == true || v.isPreRelease == false));
     }).length;
 
     if (outdatedPackagesCount > 0) {
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -135,6 +135,11 @@ class SolveReport {
   void reportOutdated() {
     final outdatedPackagesCount = _result.packages.where((id) {
       final versions = _result.availableVersions[id.name];
+      // A version is counted:
+      // - if there is a newer version which is not a pre-release and current
+      // version is also not a pre-release
+      // - if the current version is pre-release then any upgraded version is
+      // considered.
       return versions.any((v) =>
           v > id.version &&
           (id.version.isPreRelease == true || v.isPreRelease == false));
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -137,12 +137,11 @@ class SolveReport {
       final versions = _result.availableVersions[id.name];
       // A version is counted:
       // - if there is a newer version which is not a pre-release and current
-      // version is also not a pre-release
+      // version is also not a pre-release or,
       // - if the current version is pre-release then any upgraded version is
       // considered.
       return versions.any((v) =>
-          v > id.version &&
-          (id.version.isPreRelease == true || v.isPreRelease == false));
+          v > id.version && (id.version.isPreRelease || !v.isPreRelease));
     }).length;
 
     if (outdatedPackagesCount > 0) {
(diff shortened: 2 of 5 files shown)


ayan-b:
It may be noted that the [CI build](https://travis-ci.org/github/dart-lang/pub/builds/672953202) actually passed, but didn't update the status here.

lib/src/solver/report.dart
```diff
@@ -130,6 +130,40 @@ class SolveReport {
...
+      // unable to upgrade to.
+      if (newId != null) {
+        var versions = _result.availableVersions[newId.name];
+
+        for (var version in versions) {
+          if (version > newId.version) {
+            ++outdatedPackagesCount;
+            break;
+          }
+        }
```
jonasfj, line 152: This can be done with something like `versions.any((v) => v > newId.version)`

But also if there is a newer version and the newer version is a pre-release we don't want to count it. Well, unless the current version is also a pre-release.

lib/src/solver/report.dart
```diff
@@ -130,6 +130,40 @@ class SolveReport {
...
+    }
+    if (outdatedPackagesCount > 0) {
+      dynamic packageCountString;
+      if (outdatedPackagesCount == 1) {
+        packageCountString = '1 package has';
+      } else {
+        packageCountString = '$outdatedPackagesCount packages have';
+      }
+      log.message('$packageCountString newer versions incompatible with '
+          "dependency constraints.\nTry 'pub outdated' for more information.");
```
jonasfj, line 163: I think it's better to say:
```
5 packages have a newer version incompatible with dependency constraints.
```

lib/src/solver/report.dart
```diff
@@ -130,6 +130,40 @@ class SolveReport {
     }
   }
 
+  /// Displays a two-line message, number of outdated packages and an
+  /// instruction to run `pub outdated` if outdated packages are detected.
+  void reportOutdated() {
+    var outdatedPackagesCount = 0;
+
+    var names = _result.packages.map((id) => id.name).toList();
```
jonasfj, line 138: No need to convert with `.toList()`.

Also avoid `var` when the variable is never reassigned, you can just use `final`.

lib/src/solver/report.dart
ayan-b: Done!

lib/src/solver/report.dart
```diff
@@ -130,6 +130,40 @@ class SolveReport {
...
+        for (var version in versions) {
+          if (version > newId.version) {
+            ++outdatedPackagesCount;
+            break;
+          }
+        }
+      }
+    }
+    if (outdatedPackagesCount > 0) {
+      dynamic packageCountString;
```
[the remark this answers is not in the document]
ayan-b: Done.

lib/src/solver/report.dart
ayan-b: Done.

lib/src/solver/report.dart
ayan-b: Done!

ayan-b:
[CI task `dartfmt`](https://travis-ci.org/github/dart-lang/pub/jobs/681084338) is failing for an unchanged file!

lib/src/solver/report.dart
```diff
@@ -130,6 +130,39 @@ class SolveReport {
     }
   }
 
+  /// Displays a two-line message, number of outdated packages and an
+  /// instruction to run `pub outdated` if outdated packages are detected.
+  void reportOutdated() {
+    var outdatedPackagesCount = 0;
+
+    final names = _result.packages.map((id) => id.name);
+    for (var name in names) {
```
jonasfj, line 139: `final name`, but why not just iterate over the `_result.packages`

lib/src/solver/report.dart
```diff
@@ -130,6 +130,39 @@ class SolveReport {
...
+      var newId = _dependencies[name];
```
jonasfj, line 140: isn't the contents of `_dependencies[name]` the same `id` as exists in `_result.packages`?

lib/src/solver/report.dart
```diff
@@ -130,6 +130,39 @@ class SolveReport {
...
+
+      // See if there are any newer versions of the package that we were
+      // unable to upgrade to.
+      if (newId != null) {
```
jonasfj, line 144: When and why can this null?

lib/src/solver/report.dart
```diff
@@ -130,6 +130,39 @@ class SolveReport {
...
+        var versions = _result.availableVersions[newId.name];
```
jonasfj, line 145: final

lib/src/solver/report.dart
```diff
@@ -130,6 +130,39 @@ class SolveReport {
     }
   }
 
+  /// Displays a two-line message, number of outdated packages and an
+  /// instruction to run `pub outdated` if outdated packages are detected.
+  void reportOutdated() {
+    var outdatedPackagesCount = 0;
```
jonasfj, line 136: Instead of counting this way we could just do:

`final outdatedPackagesCount = _result.packages.where((id) {.......}).length;`

It worked rather nicely for `numChanged` above...

lib/src/solver/report.dart
ayan-b: Iterating over `_result.packages` now.

lib/src/solver/report.dart
ayan-b: Yes, thanks.

lib/src/solver/report.dart
ayan-b: I thought the may not exist on the map, but I think I am wrong. I was previously following this logic:
https://github.com/dart-lang/pub/blob/572a23f815a5d632a8c2b259b691792674cd8b4d/lib/src/solver/report.dart#L205-L221

lib/src/solver/report.dart
ayan-b: Done.

lib/src/solver/report.dart
ayan-b: Updated the logic.

lib/src/solver/report.dart
```diff
@@ -130,6 +130,32 @@ class SolveReport {
...
+  /// Displays a two-line message, number of outdated packages and an
+  /// instruction to run `pub outdated` if outdated packages are detected.
+  void reportOutdated() {
+    final outdatedPackagesCount = _result.packages.where((id) {
+      final versions = _result.availableVersions[id.name];
+
+      if (versions.any((v) =>
+          v > id.version &&
+          (id.version.isPreRelease == true || v.isPreRelease == false))) {
+        return true;
```
ayan-b, line 142: `if` can be omitted and we can write:
```dart
(versions.any((v) =>
          v > id.version &&
          (id.version.isPreRelease == true || v.isPreRelease == false)))
```

However, I am not really sure if this is a good practice ...

lib/src/solver/report.dart
```diff
@@ -130,6 +130,28 @@ class SolveReport {
...
   }
 
+  /// Displays a two-line message, number of outdated packages and an
+  /// instruction to run `pub outdated` if outdated packages are detected.
+  void reportOutdated() {
+    final outdatedPackagesCount = _result.packages.where((id) {
+      final versions = _result.availableVersions[id.name];
+      return versions.any((v) =>
+          v > id.version &&
+          (id.version.isPreRelease == true || v.isPreRelease == false));
```
jonasfj, line 140: Would be nice with a comment documenting what this logic is supposed to mean.

ayan-b: Done!

lib/src/entrypoint.dart
```diff
@@ -246,13 +246,15 @@ class Entrypoint {
...
     if (dryRun) {
       result.summarizeChanges(type, dryRun: dryRun);
+      if (type == SolveType.UPGRADE) result.showOutdated(type);
       return;
     }
 
     await Future.wait(result.packages.map(_get));
     _saveLockFile(result);
 
     result.summarizeChanges(type, dryRun: dryRun);
```
sigurdm, line 256: Not for this PR: Just noting that dryRun should always be false here.

lib/src/solver/report.dart
```diff
@@ -130,6 +130,33 @@ class SolveReport {
...
+    final outdatedPackagesCount = _result.packages.where((id) {
+      final versions = _result.availableVersions[id.name];
+      // A version is counted:
+      // - if there is a newer version which is not a pre-release and current
+      // version is also not a pre-release
+      // - if the current version is pre-release then any upgraded version is
+      // considered.
+      return versions.any((v) =>
+          v > id.version &&
+          (id.version.isPreRelease == true || v.isPreRelease == false));
```
sigurdm, line 145: Nit: I think this line could be simplified:
```
 (id.version.isPreRelease || !v.isPreRelease));
```

lib/src/solver/report.dart
```diff
@@ -130,6 +130,33 @@ class SolveReport {
...
   }
 
+  /// Displays a two-line message, number of outdated packages and an
+  /// instruction to run `pub outdated` if outdated packages are detected.
+  void reportOutdated() {
+    final outdatedPackagesCount = _result.packages.where((id) {
+      final versions = _result.availableVersions[id.name];
+      // A version is counted:
+      // - if there is a newer version which is not a pre-release and current
+      // version is also not a pre-release
```
sigurdm, line 140: Nit: final '.' at the end of a sentence

test/test_pub.dart
```diff
@@ -81,8 +81,10 @@ String packageSpecLine(String packageName) => File(d.path(packagesFilePath))
 class RunCommand {
   static final get = RunCommand(
       'get', RegExp(r'Got dependencies!|Changed \d+ dependenc(y|ies)!'));
-  static final upgrade = RunCommand('upgrade',
-      RegExp(r'(No dependencies changed\.|Changed \d+ dependenc(y|ies)!)$'));
+  static final upgrade = RunCommand('upgrade', RegExp(r'''
+(No dependencies changed\.|Changed \d+ dependenc(y|ies)!)($|
+\d packages? (has|have) newer versions incompatible with dependency constraints.
```
sigurdm, line 86: I guess we never have a test with more than 9 newer packages, but this should be:
```
\d+ packages?
```

lib/src/solver/report.dart
ayan-b: Done!

lib/src/solver/report.dart
ayan-b: Add `or,` since `A version is counted` is continued for the next sentence as well.

test/test_pub.dart
ayan-b: Done!

lib/src/entrypoint.dart
```diff
@@ -246,13 +246,15 @@ class Entrypoint {
 
     if (dryRun) {
       result.summarizeChanges(type, dryRun: dryRun);
+      if (type == SolveType.UPGRADE) result.showOutdated(type);
```
sigurdm, line 249: One more thing I just noticed.

We should move the conditional into `SolveResult.summarizeChanges`, then we avoid repeating it.

ayan-b: Updated.

I also thought this initially but didn't implement because I thought, `result.showOutdated` could be used for some other `type` in future.

lib/src/entrypoint.dart
```diff
@@ -246,13 +246,15 @@ class Entrypoint {
 
     if (dryRun) {
       result.summarizeChanges(type, dryRun: dryRun);
+      result.showOutdated(type);
```
sigurdm, line 249: Ah, sorry I didn't explain this well enough.

I meant to change `SolveResult.summarizeChanges` like this:

```dart
  /// Displays a one-line message summarizing what changes were made (or would
  /// be made) to the lockfile.
  ///
  /// If [type] is `SolveType.UPGRADE` it also shows the number of
  /// packages that are not at the latest available version.
  ///
  /// [type] is the type of version resolution that was run.
  void summarizeChanges(SolveType type, {bool dryRun = false}) {
    final report = SolveReport(type, _sources, _root, _previousLockFile, this);
    report.summarize(dryRun: dryRun);
    if (type == SolveType.UPGRADE) {
      report.reportOutdated();
    }
  }
```

ayan-b: Ohh, sorry!

Updated the code.

PR #2448, opened 2020-04-13
Improve test.sh and testing in non-English setup

leonardoInf:
Currently there are multiple issues with testing on non-Debian, non-English systems which are addressed by this PR:

1. There is one test (test/global/active/missing_git_repo_test.dart) which evaluates git output. This fails on a non-English system because the output is in the user's locale. This is solved by setting the `LANG` environment variable in the `environment` map in `runProcess()` .

2. `tool/test.sh` used `tempfile` to generate a filename for the pub snapshot. This command is debian-specific and was replaced by `mktemp` which is generally present in all Linux distos I have tested (including Debian/Ubuntu).

3. (general): At least for myself, often tests fail just because I forgot to run `dartfmt -w . && dartanalyzer .`. I have added a new tool `lint.sh` which basically does that for you.

### DIFF
--- a/lib/src/git.dart
+++ b/lib/src/git.dart
@@ -57,7 +57,7 @@ Future<List<String>> run(List<String> args,
   log.muteProgress();
   try {
     var result = await runProcess(command, args,
-        workingDir: workingDir, environment: environment);
+        environment: {...?environment, 'LANG': 'en_GB'});
     if (!result.success) {
       throw GitException(args, result.stdout.join('\n'),
           result.stderr.join('\n'), result.exitCode);
@@ -77,7 +77,7 @@ List<String> runSync(List<String> args,
   }
 
   var result = runProcessSync(command, args,
-      workingDir: workingDir, environment: environment);
+      workingDir: workingDir, environment: {...?environment, 'LANG': 'en_GB'});
   if (!result.success) {
     throw GitException(args, result.stdout.join('\n'), result.stderr.join('\n'),
         result.exitCode);
--- a/tool/test.sh
+++ b/tool/test.sh
@@ -3,7 +3,7 @@
 ### Test wrapper script.
 # Many of the integration tests runs the `pub` command, this is slow if every
 # invocation requires the dart compiler to load all the sources. This script
-# will create a `pub.XXX.dart.snapshot.dart2` which the tests can utilize.
+# will create a `pub.XXXXXX.dart.snapshot.dart2` which the tests can utilize.
 # After creating the snapshot this script will forward arguments to
 # `pub run test`, and ensure that the snapshot is deleted after tests have been
 # run.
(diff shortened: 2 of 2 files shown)


leonardoInf:
The travis plugin seems to be somehow buggy. The CI has passed a long time ago.

lib/src/git.dart
```diff
@@ -57,7 +57,11 @@ Future<List<String>> run(List<String> args,
   log.muteProgress();
   try {
     var result = await runProcess(command, args,
-        workingDir: workingDir, environment: environment);
+
```
sigurdm, line 60: Nit: This newline should be removed

tool/lint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
sigurdm, line 1: Not sure this script belongs here, as it doesn't have anything pub-specific in it.

You could probably make a personal script that runs `dartfmt` and `dartanalyzer` in the current directory, and then run that on any dart project if it is needed for your workflow.

I personally have format-on-save in my editor to help me format.

tool/test.sh
```diff
@@ -12,8 +12,9 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 ROOT="$DIR/.."
 
+
 # PATH to a snapshot file.
-PUB_SNAPSHOT_FILE=`tempfile -p 'pub.' -s '.dart.snapshot.dart2'`;
+PUB_SNAPSHOT_FILE=`mktemp -t pub.XXXXXX.dart.snapshot.dart2`;
```
sigurdm, line 17: tempfile indeed seems to be deprecated in favor of mktemp.

tool/lint.sh
leonardoInf: Yeah, it is more of a personal script and it may not fit in everone´s workflow.

leonardoInf:
Squashed and rebased.

lib/src/git.dart
```diff
@@ -57,7 +57,10 @@ Future<List<String>> run(List<String> args,
   log.muteProgress();
   try {
     var result = await runProcess(command, args,
-        workingDir: workingDir, environment: environment);
+        /// Set git language to English to pass test/global/activate/missing_git_repo_test.dart
```
natebosch, line 60: [nit] If we keep this comment it should use 2 slashes.

Since we have a test for this I'd recommend skipping the comment - the test will prevent us from thinking we can remove it.

leonardoInf: You are right, three slashes is for public API documentation. I guess it is clear what this line of code is doing.

leonardoInf:
Ok, I will have a look at why test are failing

lib/src/git.dart
```diff
@@ -57,7 +57,7 @@ Future<List<String>> run(List<String> args,
   log.muteProgress();
   try {
     var result = await runProcess(command, args,
-        workingDir: workingDir, environment: environment);
+        environment: {...?environment, 'LANG': 'en_GB'});
```
sigurdm, line 60: This should also be changed in the 'runSync' function

sigurdm:
I don't understand the error message on travis, and it doesn't seem to reproduce for me locally ...

lib/src/git.dart
leonardoInf: Done

PR #2457, opened 2020-04-17
Use line length to wrap usage

natebosch:
Specify the `usageLineLength` in our command and command runner. Use the
terminal width or default to 80 if it is unknown.

Remove a manually placed newline in a help string with a space so that
it can wrap based on terminal width. For a width of 80 the wrapping
matches the manual newline.

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -67,7 +67,7 @@ class OutdatedCommand extends PubCommand {
 
     argParser.addFlag('up-to-date',
         defaultsTo: false,
-        help: 'Include dependencies that are already at the\n'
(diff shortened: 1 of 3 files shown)


lib/src/command/outdated.dart
```diff
@@ -33,7 +33,7 @@ class OutdatedCommand extends PubCommand {
   OutdatedCommand() {
     argParser.addFlag('color',
         help: 'Whether to color the output.\n'
-            'Defaults to color when connected to a\n'
+            'Defaults to color when connected to a '
```
sigurdm, line 36: I think there is one more in line 70.

natebosch: Fixed.

Issue #2437, opened 2020-04-07
pub outdated --mark=null-safety


jonasfj:
We need support for `pub outdated --mark=null-safety` to mark packages as null-safety compliant when doing `pub outdated`.

See: https://github.com/dart-lang/pub-dev/issues/3503 for a definition of null-safety on pub.dev.

I presume that we'll want to expose tags from pub.dev in the API, and that tags such as `is:null-safe` will be used in `pub` for marking packages. For package repositories that do not have `tags` in the API response (or sources other than `HostedSource`) we can fallback to a simplified definition asking if the package has opted into null-safety, based exclusively on SDK constraints.

/cc  @sigurdm, thoughts?

Issue #2395, opened 2020-03-25
pub outdated mark option


devoncarew:
I'm unclear what the `mark` option does.

```
    --mark=<OPTION>            Highlight packages with some property in the report.
                               [outdated (default), none]
```

Do we need an option for this? I feel like we can just determine the best presentation for the report w/o having a user flag for it.

sigurdm:
Yes this option is amputated currently. It is the plan to have options for marking null-safety and perhaps tags from pub.dev. That is planned for the first release supporting null-safety.

But we could remove it until then. @jonasfj what do you think.

jonasfj:
@sigurdm, we can just hide the flag from `--help`... `package:args` has an option for this..

PR #2461, opened 2020-04-24
pub outdated --mode=null-safety initial implementation

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/2437


lib/src/null_safety_analysis.dart
```diff
@@ -0,0 +1,165 @@
...
+  /// dependencies (transitively) have a language version >= 2.10, and no files
+  /// in lib/ of  these packages opt out to a pre-2.10 language version.
+  ///
+  /// This will do a full resolution of that package's import graph, and also
+  /// download the package and all dependencies into [cache].
+  Future<NullSafetyCompliance> _nullSafetyCompliance(
+      PackageId packageId) async {
+    return await withTempDir((tempPath) async {
+      final importingPubspec = {
+        'name': '${packageId.name}_importer',
```
jonasfj, line 62: If we use a space as separator like we do in global activate

lib/src/null_safety_analysis.dart
```diff
@@ -0,0 +1,165 @@
...
+                ? packageId.description['path']
+                : packageId.description,
+            'version': packageId.version.toString(),
+          }
+        }
+      };
+
+      writeTextFile(
+          path.join(tempPath, 'pubspec.yaml'), json.encode(importingPubspec));
+      final importingEntrypoint = Entrypoint(tempPath, _systemCache);
```
jonasfj, line 75: I think we can avoid creating a temporary directory if we really want to..

In `_installInCache` (only used by `pub global activate`) we use a `Package.inMemory` to install a package and it's dependencies into `PUB_CACHE`. We then use `await lockFile.packageConfigFile(cache)` to get the contents of `.dart_tool/package_config.json`.

I think we can copy/paste bits of the logic from `_installInCache`, such that we install dependencies into `PUB_CACHE` and obtain the contents of `.dart_tool/package_config.json` without writing any files to disk.

When we create the analysis context, we can then use a virtual overlay file system to inject the `.dart_tool/package_config.json` without writing it to disk. Checkout `AnalysisContextManager.createContextsForDirectory` where we use `OverlayResourceProvider.setOverlay` to ensure that `analysis_options.yaml` is an empty file when we run the analyzer. I think we can also overlay the file system with an empty directory and a `.dart_tool/package_config.json` file -- thus, avoiding the need for temporary directories.

lib/src/command/outdated.dart
```diff
@@ -300,6 +301,72 @@ class OutdatedCommand extends PubCommand {
...
+      return Map.fromEntries(
+        await Future.wait(
+          ids.map(
+            (id) async => MapEntry(
+              id,
+              await nullSafetyAnalyzer.nullSafetyCompliance(id),
+            ),
+          ),
+        ),
+      );
```
jonasfj, line 327: Pretty awesome, but... `NullSafetyAnalysis.nullSafetyCompliance` may install packages into `PUB_CACHE`.

I'm just now realizing that running `pub get` concurrently isn't safe, well, maybe... I'm not entirely sure. We probably could make it safe within the same process without much trouble.

If you look at this snippet from `_installInCache` used by `pub global activate`

https://github.com/dart-lang/pub/blob/6deb457048deb435009b36a4cd2d86003d107cf4/lib/src/global_packages.dart#L197-L205

For each resolution we make, we have a `result` object, where we need to iterate over `result.packages`. But we don't want to call `await source.downloadToSystemCache(id);` for the same package-version more than once.

Also if we look at `BoundHostedSource.downloadToSystemCache`:

https://github.com/dart-lang/pub/blob/6deb457048deb435009b36a4cd2d86003d107cf4/lib/src/source/hosted.dart#L258-L268

It's clear that we only download it if it does not already exist in `PUB_CACHE` this is checked by `isInSystemCache(id)`, which simply checks if the folder exists. We also create the folder before we download.
So if there is a race condition, we might download it twice, this is unlikely. However, it's fairly like that the first time we call `downloadToSystemCache` it creates the directory and starts downloading, and the next time we call it (while it's still being downloaded) `downloadToSystemCache` will return fast because the directory exists, even though the package is in fact not done being downloaded. Hence, when running analysis we may be missing some or all of the source files because they are still being downloaded.

-----------

I think we could remedy this by refactoring `NullSafetyAnalysis` such that `NullSafetyAnalysis.enqueueAnalysis(id)` is an async function that returns a Future for the result, but doesn't start processing until `NullSafetyAnalysis.populateCache()` has been called.

Inside `NullSafetyAnalysis.enqueueAnalysis(id)` we run the dependency resolution, then we add a tuple consisting of a `Completer` and a list of package ids for which we want `downloadToSystemCache` to be called.. And we await the completer being done before starting the analysis.

Then in `populateCache()` we run through the list of all tuples of completers and packages ids to be downloaded to cache, and we remove duplicate package ids and download them to cache before resolving the completer.

This is a bit crazy and there is some non-trivial synchronization going on.

lib/src/null_safety_analysis.dart
```diff
@@ -0,0 +1,165 @@
...
+
+  /// The resolution failed. Or some dart file in a dependency
+  /// doesn't parse.
+  analysisFailed,
+}
+
+class NullSafetyAnalysis {
+  final SystemCache _systemCache;
+  final Map<PackageId, Future<NullSafetyCompliance>>
+      _packageInternallyGoodCache = {};
```
jonasfj, line 42: Please document what this is... It's really cool.

lib/src/null_safety_analysis.dart
```diff
@@ -0,0 +1,165 @@
...
+          final pubspec = await boundSource.describe(dependencyId);
+          final languageVersion = _languageVersion(pubspec);
+          if (languageVersion == null ||
+              languageVersion < _firstVersionWithNullSafety) {
+            return NullSafetyCompliance.notCompliant;
+          }
+
+          if (boundSource is CachedSource) {
+            /// TODO(sigurdm): Should we set metadata here?
+            await boundSource.downloadToSystemCache(dependencyId);
```
jonasfj, line 111: Since you put a future into `_packageInternallyGoodCache`, you can't call `downloadToSystemCache´ concurrently, so maybe all of my concerns are mute :D

lib/src/command/outdated.dart
jonasfj: ah, looks like `_packageInternallyGoodCache` saves us here. It might be nice to document why this is...

lib/src/null_safety_analysis.dart
sigurdm: Yes - I considered avoiding temp directories. I just thought it wasn't worth the trouble.
But I'll look into it again if you think it matters.

lib/src/command/outdated.dart
sigurdm: > NullSafetyAnalysis.nullSafetyCompliance may install packages into PUB_CACHE.

I thought we agreed this was acceptable offline. We might be able to install unused versions somewhere else and delete them afterwards.

> downloadToSystemCache will return fast because the directory exists,

That's a good point.

I think easier will be a map keeping futures of all the packages we install. Then we can use putIfAbsent to populate it, and each package will go in only once:

```dart
class NullSafetyAnalyzer {
final Map<PackageId, Future<void>> _packagesToBeInstalled = {};
...

Future nullSafetyCompliance() {
   ...
  // instead of `await boundSource.downloadToSystemCache(dependencyId);`
  await _packagesToBeInstalled.putIfAbsent(dependencyId, () =>
   boundSource.downloadToSystemCache(dependencyId);
  ---
}
```

jonasfj: Installing into cache is fine, my concern was concurrency.

lib/src/null_safety_analysis.dart
sigurdm: Done

lib/src/null_safety_analysis.dart
sigurdm: Yes!

lib/src/null_safety_analysis.dart
sigurdm: I made this in-memory

That also uncovered a bug where relative path dependencies would be handled wrong and fail analysis!

lib/src/null_safety_analysis.dart
sigurdm: Done

lib/src/null_safety_analysis.dart
```diff
@@ -0,0 +1,178 @@
...
+        final boundSource = dependencyId.source.bind(_systemCache);
+        final pubspec = await boundSource.describe(dependencyId);
+        final languageVersion = _languageVersion(pubspec);
+        if (languageVersion == null ||
+            languageVersion < _firstVersionWithNullSafety) {
+          return NullSafetyCompliance.notCompliant;
+        }
+
+        if (boundSource is CachedSource) {
+          /// TODO(sigurdm): Should we set metadata here?
```
jonasfj, line 114: what kind of metadata?

sigurdm: The `X-Pub-Reason` header set by `withDependencyType`.

jonasfj: ah, hmm... I guess it would be nice to set this correctly. Given that we know if it's a dependency, dev_dependency or transitive dependency.

In most cases we'll probably have the response cached locally when we get here. I'm not sure I can imagine a case where we won't have that. But for consistency it would be nice to setup :)
(I'm okay leaving it as a todo if we just document what it is we mean here).

sigurdm: I clarified the TODO.

lib/src/command/outdated.dart
```diff
@@ -55,9 +61,8 @@ class OutdatedCommand extends PubCommand {
     argParser.addOption('mark',
         help: 'Highlight packages with some property in the report.',
         valueHelp: 'OPTION',
-        allowed: ['outdated', 'none'],
-        defaultsTo: 'outdated',
-        hide: true);
+        allowed: ['null-safety', 'outdated', 'none'],
```
devoncarew, line 64: It looks like the `mark` option already exists but is being un-hidden?

It may not be immediately clear to the user what the `mark` option is for. Perhaps rename to something like `mode`? It would then read: `--mode=null-safety`.

What does the `none` value do? Would user's ever send this in?

If we only ever really expect mode values of `null-safety` and `outdated`, perhaps just add a non-negatable `--null-safety` flag on the outdated command.

sigurdm: > It looks like the mark option already exists but is being un-hidden?

Yes - it was a planned feature but with very limited use until now. (At most you could disable the colorization of versions). It was hidden in https://github.com/dart-lang/pub/pull/2403 after the suggestion at https://github.com/dart-lang/pub/issues/2395 .

> It may not be immediately clear to the user what the mark option is for. Perhaps rename to something like mode? It would then read: --mode=null-safety.

`mark` is not perfect, but the mental picture is that it goes through all the versions found by outdated, and marks those that fulfill some criterium. It is not a mode of operation for the outdated command in general. (There are many modes: `--prereleases`, `--dependency-overrides`, `--up-to-date` are all what I would call modes of outdated).

> What does the none value do? Would user's ever send this in?

`none` disables highlighting. I'm not sure I have very concrete use-cases, but one could think that someone finds the colors distracting, or want to produce a 'clean' report for including somewhere.

It doesn't cost us much to provide this option.
 
If we only ever really expect mode values of

> If we only ever really expect mode values of null-safety and outdated, perhaps just add a non-negatable --null-safety flag on the outdated command (and drop the mark option altogether).

We have some other ideas for markers that could be added in the future. Possible candidates:

* Marking versions on pub.dev with a specific tag.
* Marking versions using a specific language version (not only 2.10 for null-safety).
* Marking versions that are not broken

sigurdm: After discussing this a bit with @jonasfj and @mit-mit I think I see what I got wrong here.

I thought we only had to change the presentation of the versions. But to be useful the option also has to show a different set of packages (All that are not currently null-safety enabled).

After realizing this `--mode` makes a lot more sense.

Thanks.

lib/src/command/outdated.dart
```diff
@@ -66,10 +73,16 @@ class OutdatedCommand extends PubCommand {
     argParser.addFlag('pre-releases',
         defaultsTo: false, help: 'Alias of prereleases.', hide: true);
 
+    argParser.addFlag('show-all',
+        defaultsTo: false,
+        help: 'Include dependencies that are already fullfilling --mode.');
```
jonasfj, line 78: `negatable: false`

There is no reason to have a `--no-show-all` flag, it's the default. Does it really make sense to pass this flag explicitly?

test/outdated/goldens/null_safety.txt
```diff
@@ -0,0 +1,320 @@
...
+$ pub outdated --no-color --mode=null-safety
+Dependencies                  Current  Upgradable  Resolvable  Latest  
+bar                           ✗1.0.0   ✗1.0.0      ✗1.0.0      ✓2.0.0  
+fails_analysis                ?1.0.0   ?1.0.0      ✓2.0.0      ✓2.0.0  
+fails_analysis_in_dependency  ?1.0.0   ?1.0.0      ✓2.0.0      ✓2.0.0  
+file_in_dependency_opts_out   ✗1.0.0   ✗1.0.0      ✓2.0.0      ✓2.0.0  
+file_opts_out                 ✗1.0.0   ✗1.0.0      ✓2.0.0      ✓2.0.0  
+foo                           ✗1.0.0   ✗1.0.0      ✗2.0.0      ✗2.0.0  
+
+dev_dependencies: all up-to-date
```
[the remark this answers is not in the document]
sigurdm: Good point
Done

test/outdated/goldens/no_dependencies.txt
```diff
@@ -21,3 +18,11 @@ Found no outdated packages.
 $ pub outdated --no-color --no-dependency-overrides
 Found no outdated packages.
 
+$ pub outdated --no-color --mode=null-safety
+Found no outdated packages.
```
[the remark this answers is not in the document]
sigurdm: Done

lib/src/command/outdated.dart
sigurdm: If you want to ensure the default doesn't change under you.

You could say this for several of our boolean flags. I think it is nice to be able to be explicit.

sigurdm:
> I think the tool should report in it's output something that is specific to the mode it's running in (at least for the non-default modes). E.g.:

Done

test/outdated/goldens/circular_dependencies.txt
```diff
@@ -93,3 +80,42 @@ transitive dev_dependencies: all up-to-date
 1 upgradable dependency is locked (in pubspec.lock) to an older version.
 To update it, use `pub upgrade`.
 
+$ pub outdated --no-color --mode=null-safety
+Running in 'null safety' mode; filtering versions to those that support null safety.
+Dependencies  Current  Upgradable  Resolvable  Latest  
+foo           ?1.2.3   ?1.3.0      ?1.3.0      ?1.3.0  
+
+dev_dependencies: all null safety compliant
```
sigurdm, line 88: @kwalrath Do you have ideas here. (cc @mit-mit )
Should we write 'supports null safety', 'null safety compliant' or something else.

To me 'supports null safety' sounds like it can be used with a null-safe sdk. And that is less strong than what this is.

For a package to be included it must opt-in to a null-safe language version, and all their dependencies must opt-in.

On the other hand 'compliance' might not really convey this meaning much better.

mit-mit:
> To me 'supports null safety' sounds like it can be used with a null-safe sdk. And that is less strong than what this is.

> For a package to be included it must opt-in to a null-safe language version, and all their dependencies must opt-in.

Maybe use "fully support null safety" then?

test/outdated/goldens/circular_dependencies.txt
mit-mit: > For a package to be included it must opt-in to a null-safe language version, and all their dependencies must opt-in.

Maybe use "fully support null safety" then?

sigurdm: Done

Issue #2439, opened 2020-04-07
Warn when first null-safety compliant version of a package is published


jonasfj:
When a null-safety compliant version of an existing package is published for the first time, we should provide a warning with a link to the null-safety migration guide.

There are many pitfalls one can fall into when migrating, ideally the migration guide would have a check list and we can reference this check list.

note. We want to surface a warning containing a URL to the migration guide, this way the migration guide can be adjusted as lessons are learned.

sigurdm:
This is when the sdk constraint lower bound crosses 2.10 and Not when all dependencies are null-safe etc..

 Right?

sigurdm:
More questions: 

Should we display the warning when the immediately preceding version is not-opted in or if the latest published version is not opted in.

For example for existing versions:

* 0.5.0 (not opted in)
* 1.0.0 (opted in)

Should we warn when publishing 0.7 (opted in).

Also should we give some warning if you are about to opt-out of null-safety, if the preceding version was opted in?

jonasfj:
> Also should we give some warning if you are about to opt-out of null-safety, if the preceding version was opted in?

I think that's a nice extension... I find the issue unlikely, but adding such a warning would be harmless.

jonasfj:
> Should we display the warning when the immediately preceding version is not-opted in or if the latest published version is not opted in.

This sounds reasonable to me. Also can make sure to ignore preceding pre-release versions?
If the immediately preceding non-pre-release version is not null-safety opted in, then we give a warning.

PR #2465, opened 2020-04-27
Warn first time a package version opts in to null-safety

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/2439

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -569,7 +569,9 @@ final String dartRepoRoot = (() {
 
 /// A line-by-line stream of standard input.
 final Stream<String> _stdinLines =
-    ByteStream(stdin).toStringStream().transform(const LineSplitter());
+    ByteStream(stdin).toStringStream().transform(const LineSplitter()).map((x) {
+  return x;
+});
 
 /// Displays a message and reads a yes/no confirmation from the user.
 ///
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -569,9 +569,7 @@ final String dartRepoRoot = (() {
 
 /// A line-by-line stream of standard input.
 final Stream<String> _stdinLines =
-    ByteStream(stdin).toStringStream().transform(const LineSplitter()).map((x) {
-  return x;
-});
+    ByteStream(stdin).toStringStream().transform(const LineSplitter());
 
 /// Displays a message and reads a yes/no confirmation from the user.
 ///
--- a/lib/src/validator/relative_version_numbering.dart
+++ b/lib/src/validator/relative_version_numbering.dart
@@ -17,7 +17,7 @@ import '../validator.dart';
 class RelativeVersionNumberingValidator extends Validator {
   static const String guideUrl = 'https://dart.dev/null-safety/migration-guide';
   static const String semverUrl =
-      'https://dart.dev/tools/pub/versioning#semantic-version';
+      'https://dart.dev/tools/pub/versioning#semantic-versions';
 
   final String _server;
 
@@ -49,14 +49,16 @@ class RelativeVersionNumberingValidator extends Validator {
     final previousOptedIn = _optedIntoNullSafety(previousPubspec);
 
     if (currentOptedIn && !previousOptedIn) {
-      hints.add('You are about to publish a package opting into null-safety.\n'
-          'The latest version ${previousVersion.version} has not opted in.\n'
-          'Be sure to read $guideUrl for best practices.');
+      hints.add(
+          'You\'re about to publish a package that opts into null safety.\n'
+          'The previous version (${previousVersion.version}) isn\'t opted in.\n'
+          'See $guideUrl for best practices.');
     } else if (!currentOptedIn && previousOptedIn) {
       hints.add(
           'You\'re about to publish a package that doesn\'t opt into null safety,\n'
           'but the previous version (${previousVersion.version}) was opted in.\n'
-          'This change is likely to be backwards incompatible. See $semverUrl');
+          'This change is likely to be backwards incompatible.\n'
+          'See $semverUrl for information about versioning.');
     }
   }
 
--- a/lib/src/validator/relative_version_numbering.dart
+++ b/lib/src/validator/relative_version_numbering.dart
@@ -47,12 +47,11 @@ class RelativeVersionNumberingValidator extends Validator {
     final previousOptedIn = _optedIntoNullSafety(previousPubspec);
 
     if (currentOptedIn && !previousOptedIn) {
-      warnings.add(
-          'You are about to publish a package opting into null-safety.\n'
+      hints.add('You are about to publish a package opting into null-safety.\n'
           'The latest version ${previousVersion.version} has not opted in.\n'
           'Be sure to read $guideUrl for best practices.');
     } else if (!currentOptedIn && previousOptedIn) {
-      warnings.add(
+      hints.add(
           'You are about to publish a package not opting into null-safety.\n'
           'The previous version ${previousVersion.version} was opted in.\n'
           'Be sure to read $guideUrl for best practices.');
--- a/lib/src/validator/relative_version_numbering.dart
+++ b/lib/src/validator/relative_version_numbering.dart
@@ -15,7 +15,7 @@ import '../validator.dart';
 /// Gives a warning when publishing a new version, if the latest published version
 /// was not opted into null-safety.
 class RelativeVersionNumberingValidator extends Validator {
-  static final String guideUrl =
+  static const String guideUrl =
       'http://dart.dev/null-safety-package-migration-guide';
   final String _server;
 
--- a/lib/src/validator/relative_version_numbering.dart
+++ b/lib/src/validator/relative_version_numbering.dart
@@ -14,12 +14,12 @@ import '../validator.dart';
 
 /// Gives a warning when publishing a new version, if the latest published version
 /// was not opted into null-safety.
-class FirstNullSafeValidator extends Validator {
+class RelativeVersionNumberingValidator extends Validator {
   static final String guideUrl =
       'http://dart.dev/null-safety-package-migration-guide';
   final String _server;
 
-  FirstNullSafeValidator(Entrypoint entrypoint, this._server)
+  RelativeVersionNumberingValidator(Entrypoint entrypoint, this._server)
       : super(entrypoint);
 
   @override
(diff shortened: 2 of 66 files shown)


lib/src/validator/relative_version_numbering.dart
```diff
@@ -0,0 +1,79 @@
...
+import '../exceptions.dart';
+import '../package_name.dart';
+import '../pubspec.dart';
+import '../validator.dart';
+
+/// Gives a warning when publishing a new version, if the latest published
+/// version lower to this was not opted into null-safety.
+class RelativeVersionNumberingValidator extends Validator {
+  static const String guideUrl =
+      'http://dart.dev/null-safety-package-migration-guide';
```
jonasfj, line 19: We need to check this is correct, right?

Also `https://`

lib/src/validator/relative_version_numbering.dart
```diff
@@ -0,0 +1,79 @@
...
+          .getVersions(hostedSource.refFor(entrypoint.root.name, url: _server));
+    } on PackageNotFoundException {
+      existingVersions = [];
+    }
+    existingVersions..sort((a, b) => a.version.compareTo(b.version));
+    final previousVersion = existingVersions.lastWhere(
+        (id) =>
+            !id.version.isPreRelease && id.version < entrypoint.root.version,
+        orElse: () => null);
+    if (previousVersion == null) return; // TODO(sigurdm): is this right?
```
jonasfj, line 41: Correct!

lib/src/validator/relative_version_numbering.dart
```diff
@@ -0,0 +1,79 @@
...
+    if (currentOptedIn && !previousOptedIn) {
+      warnings.add(
+          'You are about to publish a package opting into null-safety.\n'
+          'The latest version ${previousVersion.version} has not opted in.\n'
+          'Be sure to read $guideUrl for best practices.');
+    } else if (!currentOptedIn && previousOptedIn) {
+      warnings.add(
+          'You are about to publish a package not opting into null-safety.\n'
+          'The previous version ${previousVersion.version} was opted in.\n'
+          'Be sure to read $guideUrl for best practices.');
```
jonasfj, line 58: We can't reference the guide here, because there is no reverse migration guide, so instead I suggest we change the last line to:
> This will likely break semantic versioning.

lib/src/validator/relative_version_numbering.dart
```diff
@@ -0,0 +1,79 @@
...
+    /// we cannot deduce the language version.
+    ///
+    /// This will hopefully be detected elsewhere.
+    ///
+    /// A single `Version` is also a `VersionRange`.
+    if (sdkConstraint is! VersionRange) return false;
+    final constraintMin = (sdkConstraint as VersionRange).min;
+
+    return constraintMin != null &&
+        constraintMin >= _firstVersionSupportingNullSafety;
```
jonasfj, line 75: This is not the right way to check. We have to extract the SDK lower bound and check if the language version is `>= 2.10`.

Notice that SDK version `2.10.0-dev` has language version `2.10`.

lib/src/validator/relative_version_numbering.dart
kwalrath, line 58: Maybe (not sure about the last sentence):

```suggestion
          'You are about to publish a package that doesn't opt into null safety,\n'
          'but the previous version (${previousVersion.version}) was opted in.\n'
          'This change is likely to be backwards incompatible.');
```

Should we link to https://dart.dev/tools/pub/versioning#semantic-versions?

lib/src/validator/relative_version_numbering.dart
kwalrath: How about:

https://dart.dev/null-safety/migration-guide

lib/src/validator/relative_version_numbering.dart
jonasfj: Linking to https://dart.dev/tools/pub/versioning#semantic-versions seems like a good idea.

lib/src/validator/relative_version_numbering.dart
jonasfj: @kwalrath, up to you how you want to organize content on `dart.dev`. We can just setup a page that redirects, it's just important that the links we hard code here remains stable.

Looking at the current site maybe it could be `https://dart.dev/guides/libraries/null-safety` or `https://dart.dev/guides/null-safety-migration`.

kwalrath: I was considering null safety to be along the lines of Dart 2, which has a homepage at https://dart.dev/dart-2. But obvs, a redirect is always possible. I'd rather not stick this under /guides.

sigurdm: Changed to https://dart.dev/null-safety/migration-guide

lib/src/validator/relative_version_numbering.dart
sigurdm: TODO removed

lib/src/validator/relative_version_numbering.dart
sigurdm: I referenced the guide in hopes that it would explain that you should not do that.

If you have a breaking version bump it would not break anything.

I changed it.

lib/src/validator/relative_version_numbering.dart
sigurdm: Thanks!
Well spotted

sigurdm:
I had to refactor a bunch of tests because they assumed publishing would not reach the package api.

sigurdm:
I think I also fixed a test or two that would never have their expectations triggered

lib/src/validator/relative_version_numbering.dart
```diff
@@ -47,15 +49,14 @@ class RelativeVersionNumberingValidator extends Validator {
     final previousOptedIn = _optedIntoNullSafety(previousPubspec);
 
     if (currentOptedIn && !previousOptedIn) {
-      warnings.add(
-          'You are about to publish a package opting into null-safety.\n'
+      hints.add('You are about to publish a package opting into null-safety.\n'
           'The latest version ${previousVersion.version} has not opted in.\n'
           'Be sure to read $guideUrl for best practices.');
```
kwalrath:
```suggestion
      hints.add('You\'re about to publish a package that opts into null safety.\n'
          'The previous version (${previousVersion.version}) wasn\'t opted in.\n'
          'See $guideUrl for best practices.');
```

Or maybe _previous ... wasn't_ -> _current ... isn't_?

Or maybe something like this:

The package you're about to publish is opting into null safety for the first time.
Before you publish, make sure you've followed the best practices at $guideUrl.

(What exactly do we want to make sure they've done correctly before publishing their newly null-safe package?)

lib/src/validator/relative_version_numbering.dart
```diff
@@ -15,8 +15,10 @@ import '../validator.dart';
 /// Gives a warning when publishing a new version, if the latest published
 /// version lower to this was not opted into null-safety.
 class RelativeVersionNumberingValidator extends Validator {
-  static const String guideUrl =
-      'http://dart.dev/null-safety-package-migration-guide';
+  static const String guideUrl = 'https://dart.dev/null-safety/migration-guide';
+  static const String semverUrl =
+      'https://dart.dev/tools/pub/versioning#semantic-version';
```
kwalrath, line 20:
```suggestion
      'https://dart.dev/tools/pub/versioning#semantic-versions';
```

sigurdm: Thanks!

lib/src/validator/relative_version_numbering.dart
```diff
@@ -47,15 +49,14 @@ class RelativeVersionNumberingValidator extends Validator {
...
           'Be sure to read $guideUrl for best practices.');
     } else if (!currentOptedIn && previousOptedIn) {
-      warnings.add(
-          'You are about to publish a package not opting into null-safety.\n'
-          'The previous version ${previousVersion.version} was opted in.\n'
-          'Be sure to read $guideUrl for best practices.');
+      hints.add(
+          'You\'re about to publish a package that doesn\'t opt into null safety,\n'
+          'but the previous version (${previousVersion.version}) was opted in.\n'
+          'This change is likely to be backwards incompatible. See $semverUrl');
```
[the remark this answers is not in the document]
sigurdm: Better!

lib/src/validator/relative_version_numbering.dart
sigurdm: > (What exactly do we want to make sure they've done correctly before publishing their newly null-safe package?)

One thing that comes to mind is explaining about compatibility mode, and why you would want to avoid it.

I'm not sure we know everything yet. From the associated issue:

> When a null-safety compliant version of an existing package is published for the first time, we should provide a warning with a link to the null-safety migration guide.
>
> There are many pitfalls one can fall into when migrating, ideally the migration guide would have a check list and we can reference this check list.
>
> note. We want to surface a warning containing a URL to the migration guide, this way the migration guide can be adjusted as lessons are learned.

lib/src/io.dart
```diff
@@ -569,7 +569,9 @@ final String dartRepoRoot = (() {
 
 /// A line-by-line stream of standard input.
 final Stream<String> _stdinLines =
-    ByteStream(stdin).toStringStream().transform(const LineSplitter());
+    ByteStream(stdin).toStringStream().transform(const LineSplitter()).map((x) {
+  return x;
+});
```
jonasfj, line 574: Why?

lib/src/validator/relative_version_numbering.dart
```diff
@@ -0,0 +1,84 @@
...
+
+    if (constraintMin == null) return false;
+
+    final languageVersion =
+        Version(constraintMin.major, constraintMin.minor, 0);
+
+    return languageVersion >= _firstVersionSupportingNullSafety;
+  }
+
+  static final _firstVersionSupportingNullSafety = Version.parse('2.10.0');
```
jonasfj, line 85: Maybe we should make a `LanguageVersion` class that we can use to represent this... because this is a bit scary.

If we update this to something else than `x.y.0` (adding a prerelease or build version number) then we introduce a bug.
It's probably best that we don't store LanguageVersion numbers in `Version` objects.

I'm okay with refactoring this later because we're already doing this elsewhere.

lib/src/io.dart
sigurdm: That was debug code that got incorrectly reverted.

Removed.

PR #2472, opened 2020-05-01
Proposal with implementation: pub add

leonardoInf:
Hello folks,

I propose a new command for pub: `pub add <package> [--location=<src>]`.

The add command behaves like `npm install <package>` such that it resolves the latest version of `<package>`, adds the dependency to `pubspec.yaml` and runs `pub get`. An optional `--location` can be added to pull from a custom pub server. The [`PUB_HOSTED_URL`](https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md) environment variable is respected too, but `--location ` shadows it.

If the dependency already exists in `pubspec.yaml` the command terminates.

### DIFF
--- a/lib/src/command/add.dart
+++ b/lib/src/command/add.dart
@@ -1,0 +1,86 @@
+import 'dart:convert';
+import 'dart:io';
+
+import '../command.dart';
+import '../http.dart';
+import '../log.dart' as log;
+import '../pubspec.dart';
+import '../solver.dart';
+import '../source_registry.dart';
+
+class AddCommand extends PubCommand {
+  @override
+  String get name => 'add';
+  @override
+  String get description =>
+      'Retrieve a specific package and add it to pubspec.yaml';
+  @override
+  String get invocation => 'pub add <package> [--location=<URL>]';
+
+  AddCommand() {
+    argParser.addOption('location',
+        abbr: 'l',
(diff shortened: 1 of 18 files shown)


jonasfj:
I fear this is a bit too fragile.. if we tweak it a lot we might be able to handle most cases correctly.
But I fear we would get many bugs from people where it doesn't work for some reason.

I've actually proposed a Google Summer of Code project to build a YAML mutation package.
Specifically with the aim of adding a `pub add` command :smiley: 
Have a look: https://github.com/dart-lang/sdk/wiki/Dart-GSoC-2020-Project-Ideas#idea-programmatic-yaml-modification-package

Accepted projects will be announced Monday, but I expect we will accept someone to work on YAML mutation.
I had a lot of students interested in YAML mutation, with many widely different proposals for how to implement it.
Ranging from building AST, mutating a token stream, to ideas like simply looking at the `source_span`.

My hope is that with a package for mutating YAML, we can have a robust implementation for `pub add`, and maybe do other useful features too.

----------------------------
So I'm going to close this PR, because I hope we'll have a more robust solution in the works this summer.
But if you're bored and looking for contribution opportunities drop me a line `jonasfj@google.com`
I have lots of fun ideas I don't have time to play with :see_no_evil:

Issue #2467, opened 2020-04-29
Create a `pub login` command


jonasfj:
I think we could introduce:
```
# Login when I am not signed in
$ pub login
.... open this URL in your browser and sign-in ....
You are now signed in as: jonasfj@google.com

# Shows status if I'm already logged in.
$ pub login
You are already signed-in as: jonasfj@google.com
Use: `pub logout` to sign out, before you can login again.

# Allows me to log out... so I can login again.
$ pub logout
You are not signed out.
```

This would also be useful for testing login locally.

sigurdm:
I don't object - only not sure how high priority we should give this.

sigurdm:
Should this have a --server option, or always default to pub.dev?

PR #2479, opened 2020-05-01
Login command

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2467


lib/src/command/login.dart
```diff
@@ -0,0 +1,64 @@
...
+      }
+      log.warning('You are already logged in as $userInfo\n'
+          'Run `pub logout` to log out and try again.');
+    }
+  }
+
+  Future<_UserInfo> retrieveUserInfo() async {
+    return await oauth2.withClient(cache, (client) async {
+      final discovery = await httpClient
+          .get('https://accounts.google.com/.well-known/openid-configuration');
```
jonasfj, line 44: I know that `https://www.googleapis.com/oauth2/v1/userinfo` can return a user profile too.. but I can't find documentation for the API end-point :see_no_evil:

lib/src/command/login.dart
```diff
@@ -0,0 +1,64 @@
...
+    });
+  }
+}
+
+class _UserInfo {
+  final String name;
+  final String email;
+  _UserInfo(this.name, this.email);
+  @override
+  String toString() => '$name($email)';
```
jonasfj, line 63: `<$email> "$name"`

sigurdm, line 63:
```suggestion
  String toString() => '<$email> "$name"';
```

stefandevo:
Maybe this is something that can be added specifically for the official repository pub.dev?
https://github.com/dart-lang/pub/pull/2654 has added login too, but for hosted repository only.

Issue #2469, opened 2020-04-29
Output from pub outdated is confusing when using pre-release package versions


mit-mit:
Repro steps:

1. Create a `pubspec.yaml` with:
```
name: test

environment:
  sdk: '>=2.7.0 <3.0.0'

dependencies:
  angular: ^6.0.0-alpha+1
```
2. Run `pub get`
3. Run `pub outdated`

<img width="466" alt="Screenshot 2020-04-29 at 13 50 12" src="https://user-images.githubusercontent.com/13644170/80592853-6d909580-8a20-11ea-9ebe-0d0df48e7a7f.png">

It's a bit confusing that outdated lists versions in Resolveable and Latest that are older than what's currently used.

jonasfj:
I think we should compute the latest version of a package as:
```
if --include-pre-releases is given:
  latest = largest_release (sorted by version)
else:
  If current is a pre-release:
    if largest_stable_version > current:
      latest = largest_stable_version
    else:
      latest = largest_release (sorted by version, include pre-releases)
  else:
    latest = largest_stable_version
```

So that if the current version is a pre-release we will prefer to show a larger stable release as _latest_. But if no larger stable release exists, we will show the latest pre-release.

PR #2484, opened 2020-05-07
outdated: If current is a prerelease, use prereleases for latest if no newer stable

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2469

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -219,6 +219,12 @@ class OutdatedCommand extends PubCommand {
 
   /// Get the latest version of [package].
   ///
+  /// Will include prereleases in the comparison  '--prereleases' was provided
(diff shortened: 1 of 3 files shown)


lib/src/command/outdated.dart
```diff
@@ -229,12 +235,19 @@ class OutdatedCommand extends PubCommand {
     if (available.isEmpty) {
       return null;
     }
+
     final prereleases = argResults.wasParsed('prereleases')
```
jonasfj, line 239: Unrelated to this PR, but maybe give this line a comment about how we support two spellings...

sigurdm: Done in: https://github.com/dart-lang/pub/pull/2488

Issue #2487, opened 2020-05-07
pub get (flutter) changes pubspec.lock EOL to LF even when on windows


upeguiborja:
# Environment
- Flutter Pub version: `Pub 2.8.1`
- OS: `Windows 10 1909 build 18363.418`

# Problem
Pub get is not aware of platform specific EOL conventions which makes git go crazy and report changes to `pubspec.lock` even when the packages are still the same.

To reproduce:
1. Create a new flutter project (`flutter create buggy`)
2. Initialize a git repo in the project folder (`cd ./buggy & git init`)
3. Ensure core.autocrlf is set to true on git (`git config core.autocrlf true`)
4. Add and commit all files (`git add . & git commit -m "Initial commit"`)
5. Upload to github and clone the repo to a new folder
5. Run `pub get`

Then if you do `git status` a change to pubspec.lock is reported even if the content is the same (except the EOLs)

## Expected behavior
pub get should use CRLF as a EOL character when on windows.

## Actual behavior
pub get writes pubspec.lock file with LF characters.

sigurdm:
Thanks for reporting.

You can configure git to use LF for specific files (for example `pubspec.lock`).

See eg. https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings

Most windows tools today work fine with LF endings. I am not sure we are going to change this.
@jonasfj WDYT?

jonasfj:
We could mention the `.gitattribute` fix on:
https://dart.dev/tools/pub/troubleshoot

Hmm, but if this is a problem for many users we should come up with a solution. I don't actually know if windows users generally use `CRLF` or just `LF`. We should probably find out what most Windows users do, if they have `git config core.autocrlf` enabled or not.

But a possible solution would be read the existing `pubspec.lock` file and if it is valid YAML and uses `CRLF` then we write the new `pubspec.lock` using `CRLF`.

I'm not sure why `git` think there is a change, if it tries to make a commit, then after converting back to `LF` there should be no changes.

sigurdm:
I wonder what other package managers do?

sigurdm:
https://github.com/npm/npm/issues/17161 npm had the same issue.
They did detection of current newlines: https://github.com/npm/npm/pull/18943. (Well actually https://github.com/npm/npm/pull/19904 but it seems they would have preferred the other).

jonasfj:
[...]

PR #2489, opened 2020-05-12
Preserve windows line endings in pubspec.lock if they are already there

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2487


lib/src/entrypoint.dart
```diff
@@ -743,7 +743,8 @@ class Entrypoint {
 
   /// Saves a list of concrete package versions to the `pubspec.lock` file.
   void _saveLockFile(SolveResult result) {
-    _lockFile = result.lockFile;
+    _lockFile =
```
jonasfj, line 746: Can we add a comment that we use windows line-endings if the existing file already does so?

lib/src/lock_file.dart
```diff
@@ -98,6 +107,7 @@ class LockFile {
     Uri sourceUrl;
     if (filePath != null) sourceUrl = p.toUri(filePath);
     var parsed = loadYamlNode(contents, sourceUrl: sourceUrl);
+    final windowsLineEndings = contents.contains('\r\n');
```
jonasfj, line 110: If we wanted to be a bit more robust we do `count(contents, '\r\n') > count(contents, '\n')`, using:
```dart
int count(String s, Pattern p) {
  int c = 0;
  int i = -1;
  while ((i = s.indexOf(p, i + 1)) != -1) {
    c++;
  }
  return c;
}
```

So if there is more CRLF's than LFs we use CRLF, otherwise we use LF.

It's hard to imagine things getting mixed.. But maybe if someone edits the `pubspec.lock` or manually resolves a merge conflict? (I suspect most editors would normalize line endings, but just in case it might be nice).

sigurdm: We at least have to make it `count(contents, '\r\n') > count(contents, '[^\r]\n')`.

But yes, it would probably be safer.

lib/src/entrypoint.dart
sigurdm: Good idea!
Done

lib/src/lock_file.dart
sigurdm: Done

lib/src/entrypoint.dart
```diff
@@ -742,10 +743,19 @@ class Entrypoint {
...
   /// Saves a list of concrete package versions to the `pubspec.lock` file.
+  ///
+  /// Will use Windows line endings (`\r\n` if a `pubspec.lock` exists, and uses
+  /// that.
   void _saveLockFile(SolveResult result) {
-    _lockFile = result.lockFile;
-    var lockFilePath = root.path('pubspec.lock');
-    writeTextFile(lockFilePath, _lockFile.serialize(root.dir));
+    _lockFile =
+        result.lockFile(windowsLineEndings: lockFile.windowsLineEndings);
```
jonasfj, line 751: did you miss something here, didn't you remove the `windowsLineEndings` state?

sigurdm: Yes - I think I made a bad merge...

Fixed

PR #2491, opened 2020-05-12
outdated: work with no lock file

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2445

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -13,6 +13,7 @@ import 'package:path/path.dart' as path;
 
 import '../command.dart';
 import '../entrypoint.dart';
+import '../io.dart';
 import '../log.dart' as log;
 import '../null_safety_analysis.dart';
 import '../package.dart';
@@ -88,8 +89,6 @@ class OutdatedCommand extends PubCommand {
 
   @override
   Future run() async {
-    entrypoint.assertUpToDate();
-
     final includeDevDependencies = argResults['dev-dependencies'];
     final includeDependencyOverrides = argResults['dependency-overrides'];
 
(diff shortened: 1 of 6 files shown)


lib/src/command/outdated.dart
```diff
@@ -509,34 +508,40 @@ Future<void> _outputHuman(
...
     }
-  }
 
-  if (notAtResolvable == 0 && upgradable == 0 && rows.isNotEmpty) {
-    log.message('\nDependencies are all on the latest resolvable versions.'
-        '\nNewer versions, while available, are not mutually compatible.');
+    if (notAtResolvable == 0 && upgradable == 0 && rows.isNotEmpty) {
+      log.message('\nDependencies are all on the latest resolvable versions.'
+          '\nNewer versions, while available, are not mutually compatible.');
+    }
```
jonasfj, line 540: You've made these conditional on `lockFileExists`, right?

But do we really need to... I see that it makes sense to not advocate people run `pub upgrade` (because nothing is held back by `pubspec.lock`).. But newer resolvable dependencies that are held back by `pubspec.yaml` is something we could mention.

test/outdated/outdated_test.dart
```diff
@@ -54,6 +54,13 @@ Future<void> main() async {
     expectMatchesGoldenFile(
         buffer.toString(), 'test/outdated/goldens/helptext.txt');
   });
+
+  test('no lockfile', () async {
+    await d.appDir({'foo': '^1.0.0'}).create();
+    await servePackages((builder) => builder..serve('foo', '1.2.3'));
```
jonasfj, line 60: What if `2.0.0` exists, then the users should get a notice that constraint in `pubspec.yaml` can be bumped.

lib/src/command/outdated.dart
sigurdm: Done

test/outdated/outdated_test.dart
sigurdm: Done

sigurdm:
PTAL

Issue #2492, opened 2020-05-14
Use subprocess for "pub run" and allow vm options


jonasfj:
We should allow vm options, I think using a subprocess is a fine solution. Let's try it :)

sigurdm:
Should we do the vm-options something like this:?
```sh
pub run --vm-options="--enable-experiment" test
```

jakemac53:
There is no need imo for `--vm-options`, just use the position of the args instead?

This is already done for the `--enable-asserts` option (which is actually a vm option already right?).

So the usage would be `pub run <vm-options> <command> <command-options>`

sigurdm:
> There is no need imo for `--vm-options`, just use the position of the args instead?

Hmm --- not a bad idea.
I just worry it will be hard to follow what arguments goes where (eg. `--verbose` should still go to pub). Also the VM arguments are somewhat 'raw' compared to what pub usually provides, and we might want to present only the "common subset" for passing directly.

Snapshotting is also an issue. If we pass arbitrary vm arguments we probably cannot rely on a snapshot and should always run from source, but for '--enable-asserts' and some other curated subset we know that we can use a snapshot that was compiled without these flags.

jakemac53:
> I just worry it will be hard to follow what arguments goes where (eg. `--verbose` should still go to pub). 

This is already exactly how --enable-asserts works though - it has to come before the binary name you want to run. The `--vm-options` flag should similarly work the same way otherwise no pub binary can have an option with that name.

> Also the VM arguments are somewhat 'raw' compared to what pub usually provides, and we might want to present only the "common subset" for passing directly.

Imo its more confusing if some flags work and some don't. A simple pass through of the args is very easy to understand. Simplicity ftw :).

> Snapshotting is also an issue. If we pass arbitrary vm arguments we probably cannot rely on a snapshot and should always run from source, but for '--enable-asserts' and some other curated subset we know that we can use a snapshot that was compiled without these flags.

You still have access to the args before you pass them on, and could still made an educated decision to run from snapshot when a known compatible set of flags is passed.

jonasfj:
> `pub run [vm-options...] <command> [command-options...]`
I like this... we can allow an optional `--` as separator..

> `pub run [vm-options...] [--] <command> [command-options...]`

The only downside I see is that when `pub run --v2 --some-arg` or `dartdev run --some-arg` and we start the default command from the current package, then how do we know if `--some-arg` is a VM option or a command option.

[...]

Issue #2360, opened 2020-03-12
Check if "pub global activate ..." and "pub global run" works across Dart SDK upgrades


jonasfj:
See: https://github.com/dart-lang/protobuf/issues/359

It seems that if you activate with Flutter and run normally this could cause a problem.
Maybe there is something to investigate here.

sigurdm:
What would be the ideal behaviour? Should we re-`global acitivate` the package, or ask the user to do it?

sigurdm:
I checked this out, and no - it does not work across sdk upgrades.

I got different error messages than reported in: dart-lang/protobuf#359

PR #2493, opened 2020-05-15
--enable-experiments for `pub run`

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2492, https://github.com/dart-lang/pub/issues/2360, https://github.com/dart-lang/sdk/issues/41355 .

Run as a subprocess if there are vm-args, but keep using isolates in the default case due to better signal handling.

Also --vm-arg that passes arguments to the VM for `pub run`.
`pub run` now runs programs as sub-processes.

We do not run from snapshot if running with non-default vm args.

Snapshots are now saved with the sdk-version embedded in the name, this
avoids the awkward detection when the vm fails running a bad snapshot.

Contains quite some refactorings. Especially streamlining of `pub global
run`.


lib/src/global_packages.dart
```diff
@@ -605,7 +527,9 @@ class GlobalPackages {
       var script = package.pubspec.executables[executable];
 
       var previousPackage = _createBinStub(package, executable, script,
-          overwrite: overwriteBinStubs, snapshot: snapshots[script]);
+          overwrite: overwriteBinStubs,
+          snapshot: entrypoint
+              .snapshotPathOfExecutable(Executable(package.name, executable)));
```
kevmoo, line 532: Adding a trailing comma to the `_creatBinStub` args will make this format nicer 😄

lib/src/entrypoint.dart
```diff
@@ -274,59 +288,98 @@ class Entrypoint {
...
+  /// The location of the snapshot of the dart program at [path] in [package]
+  /// will be stored here.
+  ///
+  /// We use the sdk version to make sure we don't run snapshots from a
+  /// different sdk.
+  ///
+  /// [path] must be relative.
+  String snapshotPathOfExecutable(Executable executable) {
+    assert(p.isRelative(executable.relativePath));
+    final versionSuffix = Uri.encodeComponent(Platform.version);
```
jonasfj, line 369: See `hosted.dart` where we have:
```dart
  /// The system cache directory for the hosted source contains subdirectories
  /// for each separate repository URL that's used on the system.
  ///
  /// Each of these subdirectories then contains a subdirectory for each
  /// package downloaded from that site.
  @override
  String getDirectory(PackageId id) {
    var parsed = source._parseDescription(id.description);
    var dir = _urlToDirectory(parsed.last);
    return p.join(systemCacheRoot, dir, '${parsed.first}-${id.version}');
  }
```

This is probably unnecessary...

test/snapshot_test.dart
```diff
@@ -222,11 +224,11 @@ void main() {
...
         await d.appDir({'foo': '5.6.7'}).create();
 
-        await pubGet(
-            args: ['--precompile'], output: contains('Precompiled foo:hello.'));
+        await pubGet(args: ['--no-precompile']);
 
         await d.dir(p.join(appPath, '.dart_tool', 'pub', 'bin'), [
-          d.dir('foo', [d.outOfDateSnapshot('hello.dart.snapshot.dart2')])
+          d.dir('foo',
+              [d.outOfDateSnapshot('hello.dart.snapshot.2.0.0-dev.58.0')])
```
jonasfj, line 231: No `$versionSuffix`

lib/src/package.dart
```diff
@@ -34,6 +34,8 @@ class Package {
   /// The path to the directory containing the package.
   final String dir;
 
+  bool get isInMemory => dir == null;
```
jonasfj, line 37: This doesn't have to be public does it? Can we also give a single line doc-comment.

lib/src/package_graph.dart
```diff
@@ -39,7 +39,8 @@ class PackageGraph {
   /// the packages' pubspecs are already fully-parsed.
   factory PackageGraph.fromSolveResult(
       Entrypoint entrypoint, SolveResult result) {
-    var packages = Map<String, Package>.fromIterable(result.packages,
+    var packages = Map<String, Package>.fromIterable(
+        result.packages.where((id) => id.source != null),
```
jonasfj, line 43: Why is this necessary? Are we abusing notation somehow?

What does `id.source == null` imply?

jonasfj: So `root` is filtered out? why?

And why do we retain `if (id.name == entrypoint.root.name) return entrypoint.root;`.

lib/src/executable.dart
```diff
@@ -57,130 +56,152 @@ Future<int> runExecutable(Entrypoint entrypoint, String package,
...
+/// An executable in a package
+class Executable {
+  String package;
+  // The relative path to the executable inside the root of [package].
+  String relativePath;
+  Executable(this.package, String program)
+      : relativePath = adaptProgramToPath(program);
 
-  return exitCode;
+  static String adaptProgramToPath(String program) {
```
jonasfj, line 196: This can be private if we call it from the constructor...

but I wanted to avoid entangling this logic in `pub run --dart-dev`. I agree that `pub run` needs to retain this for compatibility. Can we instead of calling it from `Executable` constructor, call it before we pass to `Executable` constructor and then make it public...

Or make it private and give `Executable` two constructors.

lib/src/executable.dart
```diff
@@ -57,130 +56,152 @@ Future<int> runExecutable(Entrypoint entrypoint, String package,
...
+    // Since we don't access the package graph, this doesn't happen
+    // automatically.
+    entrypoint.assertUpToDate();
 
+    if (!fileExists(snapshotPath)) {
+      await recompile(executable);
+    }
+    if (!fileExists(snapshotPath)) {
+      // TODO XXX compilation failed, run from source??
+    }
```
jonasfj, line 90: I think `dart.snapshot` will throw `ApplicationException` if the compilation fails.. SP there is no need to handle that here.

lib/src/command/global_run.dart
```diff
@@ -57,8 +61,18 @@ class GlobalRunCommand extends PubCommand {
...
+    final vmArgs = <String>[
+      if (experiments.isNotEmpty) "--enable_experiment=${experiments.join(',')}"
+    ];
+    final globalEntrypoint = await globals.find(package);
+    final exitCode = await runExecutable(
+        globalEntrypoint, Executable(package, executable), args,
+        vmArgs: vmArgs,
+        enableAsserts: argResults['enable-asserts'] || argResults['checked'],
+        recompile: (executable) async {
+      await globalEntrypoint.precompileExecutable(executable);
```
jonasfj, line 74: `recompile: globalEntrypoint.precompileExecutable` wouldn't work? I don't think you need to wrap it.

lib/src/entrypoint.dart
```diff
@@ -122,31 +123,47 @@ class Entrypoint {
 
   PackageGraph _packageGraph;
 
+  /// Where the lock file and package configurations are to be found.
+  ///
+  /// Global packages (except those from path source)
+  /// store these in the global cache.
+  String get _configPath => root.dir == null || isCached
```
jonasfj, line 130: Should we call this `_configRoot`?

lib/src/entrypoint.dart
```diff
@@ -122,31 +123,47 @@ class Entrypoint {
...
-    var newPath = root.path('.dart_tool/pub');
-    var oldPath = root.path('.pub');
-    if (!dirExists(newPath) && dirExists(oldPath)) return oldPath;
-    return newPath;
+    if (isGlobal) {
+      return p.join(
+        cache.rootDir,
+        'global_packages',
+        root.name,
+      );
```
jonasfj, line 160: Why is this not using `_configPath`.

And why is this different for global packages installed with `pub global activate --source path`

lib/src/entrypoint.dart
jonasfj: After chatting we agree to use the version number :D

lib/src/package_graph.dart
jonasfj: Okay, let's move this special filtering hack into `_installInCache` in `global_packages.dart`.

lib/src/executable.dart
jonasfj: Let's make two constructors...

lib/src/entrypoint.dart
jonasfj: Documentation comment about why this is different would be nice.

lib/src/command/global_run.dart
```diff
@@ -57,8 +61,16 @@ class GlobalRunCommand extends PubCommand {
       log.warning('The --mode flag is deprecated and has no effect.');
     }
 
-    var exitCode = await globals.runExecutable(package, executable, args,
-        enableAsserts: argResults['enable-asserts'] || argResults['checked']);
+    final experiments = argResults['enable-experiment'] as List;
+    final vmArgs = <String>[
+      if (experiments.isNotEmpty) "--enable_experiment=${experiments.join(',')}"
```
kevmoo, line 66: Silly nit: use `enable-experiment` for the SDK. It supports both `_` and `-` for separators, but we should be consistant!

lib/src/command/run.dart
```diff
@@ -67,28 +70,16 @@ class RunCommand extends PubCommand {
       log.warning('The --mode flag is deprecated and has no effect.');
     }
 
-    // The user may pass in an executable without an extension, but the file
-    // to actually execute will always have one.
-    if (p.extension(executable) != '.dart') executable += '.dart';
+    final experiments = argResults['enable-experiment'] as List;
+    final vmArgs = <String>[
+      if (experiments.isNotEmpty) "--enable_experiment=${experiments.join(',')}"
```
kevmoo, line 75: ditto

lib/src/command/run.dart
```diff
@@ -126,41 +118,15 @@ class RunCommand extends PubCommand {
...
-
-    // Don't ever compile snapshots for mutable packages, since their code may
-    // change later on. Don't check if this the case if a snapshot already
-    // exists.
-    var useSnapshot = snapshotExists ||
-        (package != entrypoint.root.name &&
-            !entrypoint.packageGraph.isPackageMutable(package));
+    final experiments = argResults['enable-experiment'] as List;
+    final vmArgs = <String>[
+      if (experiments.isNotEmpty) "--enable_experiment=${experiments.join(',')}"
```
kevmoo, line 123: ditto

kevmoo: Seeing this logic three times, wondering if it's worth a utility 🤔

lib/src/command/global_run.dart
```diff
@@ -27,6 +28,9 @@ class GlobalRunCommand extends PubCommand {
   GlobalRunCommand() {
     argParser.addFlag('enable-asserts', help: 'Enable assert statements.');
     argParser.addFlag('checked', abbr: 'c', hide: true);
+    argParser.addMultiOption('enable-experiment',
+        help: 'Runs the executable in a VM with the given experiments enabled.',
```
kevmoo, line 32: 1) Wondering if we should include details saying that the target executable won't be precompiled, so it'll start up more slowly
2) Is there somewhere  we can point folks to get more info on experiments?

https://github.com/dart-lang/sdk/blob/master/docs/process/experimental-flags.md is not super helpful

CC @mit-mit

mit-mit: Og, maybe we should add a short page on dart.dev about experiment flags, WDYT @kwalrath ?

kevmoo: Well, I think linking to the github repo is fine – maybe we could have a redirect like flutter.dev/go/... right?

kwalrath: Sure, we could have a page on dart.dev. Where would it go? What should it cover?

lib/src/command/run.dart
sigurdm: I made a function!

lib/src/command/run.dart
```diff
@@ -101,7 +90,8 @@ class RunCommand extends PubCommand {
   ///
   /// Runs `bin/<command>.dart` from package `<package>`. If `<package>` is not
   /// mutable (local root package or path-dependency) a source snapshot will be
-  /// cached in `.dart_tool/pub/bin/<package>/<command>.dart.snapshot.dart2`.
+  /// cached in
+  /// `.dart_tool/pub/bin/<package>/<command>.dart.snapshot.$versionSuffix`.
```
jonasfj, line 94: Can we say `<sdkVersion>`?

sigurdm: Done

lib/src/command/run.dart
```diff
@@ -29,9 +29,8 @@ class RunCommand extends PubCommand {
     argParser.addFlag('enable-asserts', help: 'Enable assert statements.');
     argParser.addFlag('checked', abbr: 'c', hide: true);
     argParser.addMultiOption('enable-experiment',
-        help:
-            'Runs the executable in a VM with the given experiments enabled.\n'
-            '(Enabling experiments will disable source snapshot caching)',
+        help: 'Runs the executable in a VM with the given experiments enabled. '
+            '(Enabling experiments will disable source snapshot caching.)',
```
mit-mit, line 33: Can we say something simpler like "(Note: pub commands will run slower when using experiments)"?

sigurdm: That's very vague. Then I'll rather say nothing.

mit-mit: Then say something more specific, but describe the effect (what the user will experience), not the underlying cause of that effect.

lib/src/command/global_run.dart
mit-mit: @kwalrath i sent some notes in email

sigurdm:
Not sure why travis is not checking the last commit. It should be good to go.

kevmoo:
You could create a branch w/ this PR contents and enable travis for in in the .travis.yml file – just to sanity check!

kevmoo:
Just pushed `travis` branch and it kicked things for you! https://travis-ci.org/github/dart-lang/pub/builds/693915946

lib/src/command/run.dart
jonasfj: Maybe -> `(Experiments will skip snapshot cache, this might cause slower startup)`

lib/src/command/global_run.dart
```diff
@@ -27,6 +28,10 @@ class GlobalRunCommand extends PubCommand {
   GlobalRunCommand() {
     argParser.addFlag('enable-asserts', help: 'Enable assert statements.');
     argParser.addFlag('checked', abbr: 'c', hide: true);
+    argParser.addMultiOption('enable-experiment',
+        help: 'Runs the executable in a VM with the given experiments enabled. '
+            '(Will disable snapshotting, resulting in slower startup)',
```
jonasfj, line 33: Sounds good to me too :D

kevmoo:
Can we land this? Nowish? @sigurdm – still around?

lib/src/rate_limited_scheduler.dart
```diff
@@ -59,7 +60,7 @@ class RateLimitedScheduler<J, V> {
   final Set<J> _started = {};
 
   RateLimitedScheduler(Future<V> Function(J) runJob,
-      {maxConcurrentOperations = 10})
+      {@required int maxConcurrentOperations})
```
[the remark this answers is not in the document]
jonasfj: hehe, I guess @sigurdm opted to remove the default value in the function definition, instead of a removing the value at the callsite.. so merging the two branches I had to pick one :D

I'm not sure it matters a lot. But I suppose it's nice to explicit in this case.

PR #2517, opened 2020-06-10
Remove divisive term

mit-mit:
Related to https://github.com/dart-lang/sdk/issues/42247

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -257,7 +257,7 @@ Set<String> createFileFilter(Iterable<String> files) {
   }).toSet();
 }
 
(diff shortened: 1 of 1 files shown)


lib/src/utils.dart
```diff
@@ -257,7 +257,7 @@ Set<String> createFileFilter(Iterable<String> files) {
   }).toSet();
 }
 
-/// Given a blacklist of directory names, returns a set of patterns that can
+/// Given a denylist of directory names, returns a set of patterns that can
```
sigurdm, line 260: Lets just write 
```
/// Given a list of directory names, returns a set of patterns that can
```

I think the 'filter' part is clear from the context.

PR #2522, opened 2020-06-12
Sanitize git cache folder names

sigurdm:


lib/src/source/git.dart
```diff
@@ -580,6 +580,6 @@ class BoundGitSource extends CachedSource {
     if (name.endsWith('.git')) {
       name = name.substring(0, name.length - '.git'.length);
     }
-    return name;
+    return name.replaceAll(RegExp('[^a-zA-Z0-9.-]'), '_');
```
jonasfj, line 583:
```suggestion
    // If name contains special characters we take the suffix, if this is not empty
    final saneSuffix = name.split(RegExp('[^a-zA-Z0-9._-]')).last;
    if (saneSuffix.isNotEmpty) {
      name = saneSuffix;
    }
    // Remove special characters (if any)
    name = name.replaceAll(RegExp('[^a-zA-Z0-9._-]'), '_');
    // Shorten name to 50 chars for sanity.
    if (name.length > 50) {
      name = name.substring(0, 50);
    }
    return name;
```

We could also just do:
```suggestion
    name = name.replaceAll(RegExp('[^a-zA-Z0-9._-]'), '_');
    // Shorten name to 50 chars for sanity.
    if (name.length > 50) {
      name = name.substring(0, 50);
    }
    return name;
```

sigurdm:
I added a test. Could you review that too?

lib/src/source/git.dart
sigurdm: Done

PR #2524, opened 2020-06-15
Remove unreachable code

hainesc:
Remove the unreachable code as assignment and term can never have the different names there.


jonasfj:
Hmm, interesting..Please feel free to elaborate on the invariants that makes this deadcode.

I'm inclined to believe you :) but I might have to dig a little deeper into the solver before we land this.

hainesc:
Sorry for misunderstanding it, Closed.

PR #2532, opened 2020-06-16
Improve expectations in flutter validator test

sigurdm:
Leftovers from another PR that was cancelled

### DIFF
--- a/test/validator/flutter_plugin_format_test.dart
+++ b/test/validator/flutter_plugin_format_test.dart
@@ -17,9 +17,10 @@ Validator flutterPluginFormat(Entrypoint entrypoint) =>
 
 void main() {
   group('should consider a package valid if it', () {
-    setUp(d.validPackage.create);
-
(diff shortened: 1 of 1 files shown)


lib/src/validator/flutter_plugin_format.dart
```diff
@@ -47,18 +49,37 @@ class FlutterPluginFormatValidator extends Validator {
...
+            'not support the flutter.plugin.platforms key.\n'
+            'Please consider increasing the Flutter SDK requirement to '
+            '^1.10.0 (environment.sdk.flutter)\n\nSee $_pluginDocsUrl');
+        return;
+      }
+
+      if (flutterConstraint.allowsAny(
+        VersionRange(
+          min: Version.parse('0.0.0'),
+          max: Version.parse('1.12.0'),
```
jonasfj, line 71: 1.20.0 (next stable release)

sigurdm:
I had a few updates in the test expectations. We might as well land these while they are still fresh.

PTAL.

PR #2542, opened 2020-06-25
Pass --(no-)sound-null-safety arg through to VM.

sigurdm:
Fixes: https://github.com/dart-lang/sdk/issues/42465


lib/src/command/global_run.dart
```diff
@@ -30,8 +30,10 @@ class GlobalRunCommand extends PubCommand {
     argParser.addFlag('checked', abbr: 'c', hide: true);
     argParser.addMultiOption('enable-experiment',
         help: 'Runs the executable in a VM with the given experiments enabled. '
-            '(Will disable snapshotting, resulting in slower startup)',
+            '(Will disable snapshotting, resulting in slower startup).',
         valueHelp: 'experiment');
+    argParser.addOption('sound-null-safety',
```
jonasfj, line 35: I think it supposed to be a flag... such that it has a `--sound-null-safety` flag, and a `--no-sound-null-safety` flag.

sigurdm: Of course...

PR #2549, opened 2020-07-06
Added functionality for editing YAML

walnutdust:
@jonasfj ! For your review


lib/src/yaml_edit.dart
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
```
sigurdm, line 1: Nit: For new files use current year :)

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+    ArgumentError.checkNotNull(_yaml);
+    initialize();
+  }
+
+  void initialize() {
+    _contents = loadYamlNode(_yaml);
+    _aliases = {};
+
+    var stack = [_contents];
+    var map = <YamlNode, bool>{};
```
sigurdm, line 148: Perhaps rename or add a comment to clarify what this map does

sigurdm: And if you are always assigning `true` it should probably be a `Set`....

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+
+    while (stack.isNotEmpty) {
+      var nextElem = stack.removeLast();
```
sigurdm, line 151: nit: `nextElement`. We have a tradition for not abbreviating.

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+  bool _isWindowsEnding;
+
+  factory YamlEditor(String yaml) => YamlEditor._(yaml);
+
+  YamlEditor._(this._yaml) {
+    ArgumentError.checkNotNull(_yaml);
+    initialize();
+  }
+
+  void initialize() {
```
sigurdm, line 143: Could this be private?
Could it be inlined in the constructor?

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+  /// ```
+  YamlNode parseAt(Iterable<Object> path, {Object orElse = #noArg}) {
+    ArgumentError.checkNotNull(path, 'path');
+
+    return _traverse(path, orElse: orElse);
+  }
+
+  /// Sets [value] in the [path].
+  ///
+  /// Note that [assign] provides a different result as compared to a [remove]
```
sigurdm, line 219:
```suggestion
  /// Note that [assign] provides a different result than a [remove]
```

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+  /// followed by an [insertIntoList], because it preserves comments at the same
+  /// level.
+  ///
+  /// Throws if [path] is invalid.
```
sigurdm, line 223: Throws what? Error or Exception?

Here and below

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+  /// doc2.insertIntoList([], 1, 'test');
+  /// ```
+  ///
+  /// Expected Output:
+  /// ```yaml
+  ///   - 0
+  ///   - test
+  ///   - 2
+  /// ```
+  void assign(Iterable<Object> path, Object value) {
```
sigurdm, line 257: Perhaps rename to `update`?

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+      _performEdit(assignInMap(this, parentNode, keyOrIndex, value),
+          collectionPath, expectedMap);
+      return;
+    }
+
+    throw PathError.unexpected(
+        path, 'Scalar $parentNode does not have key $keyOrIndex');
+  }
+
+  /// Appends [value] into the list at [path].
```
sigurdm, line 300:
```suggestion
  /// Appends [value] to the list at [path].
```

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+  /// Throws if the element at the given path is not a [YamlList] or if the path
+  /// is invalid.
+  void appendToList(Iterable<Object> path, Object value) {
+    ArgumentError.checkNotNull(path, 'path');
+    final yamlList = _traverseToList(path);
+
+    insertIntoList(path, yamlList.length, value);
+  }
+
+  /// Prepends [value] into the list at [path].
```
sigurdm, line 311:
```suggestion
  /// Prepends [value] to the list at [path].
```

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+  /// Throws if the element at the given path is not a [YamlList] or if the path
+  /// is invalid.
+  void prependToList(Iterable<Object> path, Object value) {
+    ArgumentError.checkNotNull(path, 'path');
+
+    insertIntoList(path, 0, value);
+  }
+
+  /// Inserts [value] into the list at [path], only if the element at the given
+  /// path is a list.
```
sigurdm, line 322:
```suggestion
  /// Inserts [value] into the list at [path].
```

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+    final edit = insertInList(this, list, index, value);
+
+    final expectedList = updatedYamlList(
+        list, (nodes) => nodes.insert(index, wrapAsYamlNode(value)));
+
+    _performEdit(edit, path, expectedList);
+  }
+
+  /// Changes the contents of the list at [path] by removing [deleteCount] items
+  /// at [index], and inserts [values] in-place.
```
sigurdm, line 343:
```suggestion
  /// at [index], and inserting [values] in-place.
```

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+  /// [index] and [deleteCount] must be non-negative and, [index] + [deleteCount]
+  /// must be no greater than the list's length. Throws otherwise or if [path]
+  /// is invalid.
+  ///
+  /// ```dart
+  /// final doc = YamlEditor('[Jan, March, April, June]');
+  /// doc.spliceList([], 1, 0, ['Feb']); // [Jan, Feb, March, April, June]
+  /// doc.spliceList([], 4, 1, ['May']); // [Jan, Feb, March, April, May]
+  /// ```
+  Iterable<YamlNode> spliceList(Iterable<Object> path, int index,
```
sigurdm, line 354: Should this return `void` ?
Otherwise describe the returned value

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+
+    for (var i = 0; i < deleteCount; i++) {
+      remove([...path, index + values.length]);
+    }
+
+    return nodesToRemove;
+  }
+
+  /// Removes the node at [path].
+  ///
```
sigurdm, line 386: Perhaps clarify what happens to surrounding comments

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,651 @@
...
+  /// doc.assign(['YAML'], 'YAML');
+  ///
+  /// final newNode = doc.parseAt(['YAML']);
+  ///
+  /// // Note that the value does not change
+  /// print(newNode.value); // "YAML"
+  /// print(node.value); // "YAML Ain't Markup Language"
+  /// ```
+  YamlNode parseAt(Iterable<Object> path, {Object orElse = #noArg}) {
```
sigurdm, line 211: Maybe make `orElse` a function computing the result. Also don't use a symbol, they are not the right thing to use here.

Like: https://api.dart.dev/stable/2.8.4/dart-async/Stream/firstWhere.html

lib/src/yaml_edit/editor.dart
walnutdust: thanks for catching it!

lib/src/yaml_edit/editor.dart
walnutdust: I turned it private!

The reason why this was taken out as a function is because when the string is changed after a modification (in `_performEdit`), the YAML tree is reloaded, and we re-analyze the string for aliases/line endings/indentation steps!

lib/src/yaml_edit/editor.dart
walnutdust: I chose to add a return value to make it similar to the Javascript [splice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice), and also thought that doing so would make it behave similar to how a remove method works on maps.

I've added description for the return value for now, but if you think we should make it return void I can do that too!

lib/src/yaml_edit/editor.dart
walnutdust: Modified `orElse` as per your suggestion!

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,679 @@
...
+
+  /// Sets [value] in the [path].
+  ///
+  /// Note that [update] provides a different result than a [remove] followed by
+  /// an [insertIntoList], because it preserves comments at the same level.
+  ///
+  /// Throws a [PathError] if [path] is invalid.
+  ///
+  /// ```dart
+  /// final doc = YamlEditor('''
```
jonasfj, line 227:
```suggestion
  /// **Example:** (using [update])
  /// ```dart
  /// final doc = YamlEditor('''
```

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,679 @@
...
+  /// ```
+  ///
+  /// Expected Output:
+  /// ```yaml
+  ///   - 0
+  ///   - test # comment
+  ///   - 2
+  /// ```
+  ///
+  /// ```dart
```
jonasfj, line 242:
```suggestion
  /// **Example:** (using [remove] and [insertIntoList])
  /// ```dart
```

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,679 @@
...
+  /// ```
+  YamlNode parseAt(Iterable<Object> path, {YamlNode Function() orElse}) {
+    ArgumentError.checkNotNull(path, 'path');
+
+    return _traverse(path, orElse: orElse);
+  }
+
+  /// Sets [value] in the [path].
+  ///
+  /// Note that [update] provides a different result than a [remove] followed by
```
jonasfj, line 221: Instead of "note", can we just say: "There is a subtle difference between [update] and [remove] followed by [insertIntoList]........."

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,679 @@
...
+  /// print(node.value); // "YAML Ain't Markup Language"
+  /// ```
+  YamlNode parseAt(Iterable<Object> path, {YamlNode Function() orElse}) {
+    ArgumentError.checkNotNull(path, 'path');
+
+    return _traverse(path, orElse: orElse);
+  }
+
+  /// Sets [value] in the [path].
+  ///
```
jonasfj, line 220: Before we dive into explaining the subtle difference between `update` and `remove+insertIntoList` can we explain the various things that can be used as `value`.

Maybe this is something we need to explain in the documentation comment for `YamlEditor`, just as we explain `path` there... Perhaps it's best to add a generic paragraph about used as value in `YamlEditor`. This should also explain that if formatting on the YamlNode inserted can't be done, we'll just approximate it, or fallback to some formatting that always works..

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,679 @@
...
+  /// Throws a [PathError] if the element at the given path is not a [YamlList]
+  /// or if the path is invalid.
+  void prependToList(Iterable<Object> path, Object value) {
+    ArgumentError.checkNotNull(path, 'path');
+
+    insertIntoList(path, 0, value);
+  }
+
+  /// Inserts [value] into the list at [path].
+  ///
```
jonasfj, line 323: nit. example?

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,679 @@
...
+      remove([...path, index + values.length]);
+    }
+
+    return nodesToRemove;
+  }
+
+  /// Removes the node at [path]. Comments "belonging" to the node will be
+  /// removed while surrounding comments will be left untouched
+  ///
+  /// ```dart
```
jonasfj, line 391:
```suggestion
  /// **Example:**
  /// ```dart
```

lib/src/yaml_edit.dart
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
walnutdust: sounds good!

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,679 @@
...
+/// b: 2
+/// c:
+///   - 3
+///   - 4
+///   - {e: 5, f: [6, 7]}
+/// ```
+///
+/// To get to `7`, our path will be `['c', 2, 'f', 1]`. The path for the base
+/// object is the empty array `[]`. All modification methods will throw an
+/// [Error] if the path provided is invalid. Note also that that the order of
```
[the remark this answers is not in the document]
walnutdust: I might be misunderstanding something, but sigurd mentioned here: https://github.com/dart-lang/pub/pull/2549#discussion_r450910897 to specify if I was throwing an error or an exception :O

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,679 @@
...
+  YamlEditor._(this._yaml) {
+    ArgumentError.checkNotNull(_yaml);
+    _initialize();
+  }
+
+  void _initialize() {
+    _contents = loadYamlNode(_yaml);
+    _aliases = {};
+
+    var stack = [_contents];
```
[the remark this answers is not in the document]
walnutdust: Thanks for spotting that I missed out checking the keys too! Added the corresponding test for alias key nodes :)

hahaha I was trying to implement DFS with the stack, but your nested function suggestion is shorter and easier to read so I updated it - thanks!

lib/src/yaml_edit/editor.dart
```diff
@@ -0,0 +1,679 @@
...
+/// [1]: https://yaml.org/
+@sealed
+class YamlEditor {
+  /// List of [SourceEdit]s that have been applied to [_yaml] since the creation
+  /// of this instance, in chronological order. Intended to be compatible with
+  /// `package:analysis_server`.
+  final List<SourceEdit> _edits = [];
+
+  UnmodifiableListView<SourceEdit> get edits =>
+      UnmodifiableListView<SourceEdit>([..._edits]);
```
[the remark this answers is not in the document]
walnutdust: hmm... I shifted the documentation for `_edits` over to this function.

With regards to examples for how a list of source edits can be applied to a string, I think it might be better if I added the examples in the `source_edit.dart` instead!

lib/src/yaml_edit/editor.dart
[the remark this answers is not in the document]
jonasfj: Okay, then let's say we throw a `PathError`... I guess that's useful information anyways.

lib/src/yaml_edit/editor.dart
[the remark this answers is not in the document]
jonasfj: That's a good idea...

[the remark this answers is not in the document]
jonasfj: hmm, that said... it might be nice to show the purpose of this property.

Or at-least say something like: The [SourceEdit] objects can be serialized to JSON using ...., deserialized using.... and applied to a string using `[SourceEdit.apply].

lib/src/yaml_edit/README.md
```diff
@@ -0,0 +1,48 @@
...
+Golden testing (`/test/golden_test.dart`). More information on Golden testing
+and the input/output format can be found at `/test/testdata/README.md`.
+
+These tests are automatically run with `pub run test`.
+
+## Features and bugs
+
+Please file feature requests and bugs at the [issue tracker][tracker].
+
+[tracker]: https://github.com/walnutdust/yaml_edit/issues
```
jonasfj, line 33: Let's drop this section

lib/src/yaml_edit/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Known issues
```
jonasfj, line 35: This is more of a limitation than a known issue. We don't know how we want to fix it.

lib/src/yaml_edit/yaml_edit.dart
```diff
@@ -0,0 +1,28 @@
...
+///  // Expected Output:
+///  // {YAML: YAML Ain't Markup Language}
+/// }
+/// ```
+///
+/// [1]: https://yaml.org/
+library yaml_edit;
+
+export 'src/editor.dart';
+export 'src/equality.dart' show deepEquals;
```
jonasfj, line 26: What do we export this? What use does it have end-users?

lib/src/yaml_edit/src/source_edit.dart
```diff
@@ -0,0 +1,143 @@
...
+  /// The offset from the start of the string where the modification begins.
+  final int offset;
+
+  /// The length of the substring to be replaced.
+  final int length;
+
+  /// The replacement string to be used.
+  final String replacement;
+
+  factory SourceEdit(int offset, int length, String replacement) =>
```
jonasfj, line 30: Documentation comment?

lib/src/yaml_edit/src/equality.dart
```diff
@@ -0,0 +1,105 @@
...
+    /// other map might not get a hit since they may not be both using our
+    /// [deepEqualsMap].
+    final key2 = getKey(map2, key);
+
+    if (!deepEquals(map1[key], map2[key2])) {
+      return false;
+    }
+  }
+
+  return true;
```
jonasfj, line 72: Nit, but you might have fun using `Iterable.every`

```suggestion
  return map1.keys.every((key) {
    if (!containsKey(map2, key)) {
      return false;
    }

    /// Because two keys may be equal by deep equality but using one key on the
    /// other map might not get a hit since they may not be both using our
    /// [deepEqualsMap].
    final key2 = getKey(map2, key);

    if (!deepEquals(map1[key], map2[key2])) {
      return false;
    }
  });
```

lib/src/yaml_edit/src/errors.dart
```diff
@@ -0,0 +1,37 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Error thrown when a function is passed an invalid path.
+class PathError extends ArgumentError {
+  /// The path that caused the error
+  final Iterable<Object> path;
+
+  /// The last element of [path] that could be traversed.
+  Object parentNode;
```
jonasfj, line 11: Is this not always an `YamlNode`?

jonasfj: If so we could change the type, and maybe just call it `parent`.

lib/src/yaml_edit/src/errors.dart
```diff
@@ -0,0 +1,37 @@
...
+
+  PathError(this.path, Object invalidKeyOrIndex, this.parentNode,
```
jonasfj, line 13: Should also so include the index of the parent in path?

If I pass a path `['a', 'a', 'a']`, then there is no way to known which key was actually problematic...

lib/src/yaml_edit/src/errors.dart
```diff
@@ -0,0 +1,37 @@
...
+      [String message])
+      : super.value(invalidKeyOrIndex, 'path', message);
+
+  PathError.unexpected(this.path, String message) : super(message);
```
jonasfj, line 17: Any reason to make this public?

We could make private constructors... and create top-level functions like `createPathError` for creating instances..
Then in `yaml_edit.dart` we can do `export 'src/errors.dart' show PathError, AliasError;`, so we hide the constructor for end-users..

lib/src/yaml_edit/src/errors.dart
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Error thrown when a function is passed an invalid path.
+class PathError extends ArgumentError {
```
jonasfj, line 6: Let's mark this as `@sealed`

lib/src/yaml_edit/src/errors.dart
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    return 'Invalid path: $path. $message';
+  }
+}
+
+/// Error thrown when the path contains an alias along the way.
+/// Differs from [PathError] because this extends [UnsupportedError], and
+/// may be fixed in the future.
+class AliasError extends UnsupportedError {
```
jonasfj, line 32: Let's mark this as `@sealed`

lib/src/yaml_edit/src/errors.dart
```diff
@@ -0,0 +1,37 @@
...
+      return 'Invalid path: $path. Missing key or index $invalidValue in parent $parentNode.';
+    }
+
+    return 'Invalid path: $path. $message';
+  }
+}
+
+/// Error thrown when the path contains an alias along the way.
+/// Differs from [PathError] because this extends [UnsupportedError], and
+/// may be fixed in the future.
```
jonasfj, line 31: will we ever fix this?

I'm starting to think it'll always be a limitation, since I we couldn't come up with what the right thing to do would be..

note. this might be a good place to explain why we don't support this, and why..

lib/src/yaml_edit/src/errors.dart
```diff
@@ -0,0 +1,37 @@
...
+class AliasError extends UnsupportedError {
+  /// The path that caused the error
+  final Iterable<Object> path;
+
+  AliasError(this.path) : super('Encountered an alias along $path!');
```
jonasfj, line 36: A multiple message explaining what happened might be useful. Assume the user encountering this don't know that YAML has references at all :D

lib/src/yaml_edit/src/editor.dart
```diff
@@ -0,0 +1,633 @@
...
+    final pathAsList = path.toList();
+    final collectionPath = pathAsList.take(path.length - 1);
+    final keyOrIndex = pathAsList.last;
+    final parentNode = _traverse(collectionPath, checkAlias: true);
+
+    final valueNode = wrapAsYamlNode(value);
+
+    if (parentNode is YamlList) {
+      final expectedList =
+          updatedYamlList(parentNode, (nodes) => nodes[keyOrIndex] = valueNode);
```
jonasfj, line 262:
```suggestion
      final expected = wrapAsYamlNode(
        [...parentNode.nodes]..[keyOrIndex] = valueNode,
      );
```
Then you probably won't even need `updatedYamlList`, if we do similarly everywhere... I think `updatedYamlList` is only used 4 places.

lib/src/yaml_edit/src/wrap.dart
```diff
@@ -0,0 +1,186 @@
...
+  ArgumentError.checkNotNull(map, 'map');
+
+  final dummyMap = deepEqualsMap();
+  dummyMap.addAll(map.nodes);
+
+  update(dummyMap);
+  final updatedMap = {};
+
+  /// This workaround is necessary since [yamlNodeFrom] will re-wrap
+  /// [YamlNode]s, so we need to unwrap them before passing them in.
```
jonasfj, line 35: Is this still applicable now that we're using `wrapAsYamlNode`?

lib/src/yaml_edit/src/editor.dart
```diff
@@ -0,0 +1,633 @@
...
+    ArgumentError.checkNotNull(checkAlias, 'checkAlias');
+
+    if (path.isEmpty) return _contents;
+
+    var currentNode = _contents;
+
+    for (var keyOrIndex in path) {
+      if (checkAlias) {
+        if (_aliases.contains(currentNode)) throw AliasError(path);
+      }
```
jonasfj, line 472: nit,
```suggestion
      if (checkAlias && _aliases.contains(currentNode)) {
        throw AliasError(path);
      }
```

lib/src/yaml_edit/yaml_edit.dart
walnutdust: hm yeah you're right - I exported it thinking that there might be causes where users need to compare two YamlNodes for equality of their values but I guess that might not be as important. Will remove!

lib/src/yaml_edit/src/source_edit.dart
walnutdust: Should I move the comment from the `._` private constructor up here? Feels like they'll be the same :O

lib/src/yaml_edit/src/equality.dart
walnutdust: thanks!

lib/src/yaml_edit/src/errors.dart
walnutdust: hm... what would the reason to export the errors be, if we're not expecting end users to catch them :O

lib/src/yaml_edit/src/editor.dart
walnutdust: haha whoops

lib/src/yaml_edit/src/errors.dart
walnutdust: hm okay I'll update the documentation!

lib/src/yaml_edit/src/editor.dart
walnutdust: whoa thanks!

lib/src/yaml_edit/src/wrap.dart
walnutdust: nope - thanks for spotting!

lib/src/yaml_edit/src/strings.dart
```diff
@@ -0,0 +1,313 @@
...
+    return '{' + safeEntries.join(', ') + '}';
+  }
+
+  return getFlowScalar(value);
+}
+
+/// Returns [value] with the necessary formatting applied in a block context.
+///
+/// If [value] is a [YamlNode], we respect its [style] parameter.
+String getBlockString(Object value, int indentation, String lineEnding) {
```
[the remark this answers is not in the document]
walnutdust: I moved all `get__String` to `yamlEncode__` methods, and for methods like `yamlEncodeFolded`, I added a `try` prefix to indicate that they might not always result in that style!

lib/src/yaml_edit/src/strings.dart
```diff
@@ -0,0 +1,313 @@
...
+  return getFlowScalar(value);
+}
+
+/// Returns [value] with the necessary formatting applied in a flow context.
+///
+/// If [value] is a [YamlNode], we try to respect its [style] parameter where
+/// possible. Certain cases make this impossible (e.g. a plain string scalar
+/// that starts with '>', a child having a block style parameters), in which
+/// case we will produce [value] with default styling options.
+String getFlowString(Object value) {
```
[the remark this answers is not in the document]
walnutdust: ah I didn't use to wrap before passing it in - fixed it though!

So with `YamlMap.wrap` we can do something like:

```
  final doc = YamlMap.wrap({'a': File('hi.dart')});
  print(doc['a'].existsSync());
```

I'll modify `wrapAsYamlNode` to check even `YamlNodes`

lib/src/yaml_edit/src/source_edit.dart
jonasfj: Sounds good to me :D

lib/src/yaml_edit/src/errors.dart
jonasfj: Correct, indeed some people would argue that we need not export it... I would be okay with that too.

We shouldn't reference it by name in the documentation comments, elsewhere then... but we can just reference the super class instead (as it's from the Dart-SDK)...

lib/src/yaml_edit/src/strings.dart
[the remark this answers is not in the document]
jonasfj: Sweet! that's much easier to read :D

lib/src/yaml_edit/src/strings.dart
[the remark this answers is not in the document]
jonasfj: Hehe, I guess it's fine for `wrapAsYamlNode` to throw an `ArgumentError` if the value (or child of said value) passed in can't be wrapped as an `YamlNode`..

I guess it's also easier to get this error sooner rather than later..

lib/src/yaml_edit/src/errors.dart
walnutdust: I changed it such that the documentation comments for the public methods use `[ArgumentError]`, and the private methods say `[PathError]`!

lib/src/yaml_edit/src/editor.dart
```diff
@@ -0,0 +1,635 @@
...
+  /// of this instance, in chronological order. Intended to be compatible with
+  /// `package:analysis_server`.
+  ///
+  /// The [SourceEdit] objects can be serialized to JSON using the `toJSON`
+  /// function, deserialized using [SourceEdit.fromJSON], and applied to a
+  /// string using the `apply` function. Multiple [SourceEdit]s can be applied
+  /// to a string using [SourceEdit.applyAll].
+  ///
+  /// For more information, refer to the [SourceEdit] class.
+  UnmodifiableListView<SourceEdit> get edits =>
```
sigurdm, line 75: Why not just return the list if it has been copied anyway?

test/yaml_edit/test_case.dart
```diff
@@ -0,0 +1,318 @@
...
+      var entityStream = testDir.list(recursive: true, followLinks: false);
+      entityStream =
+          entityStream.where((entity) => entity.path.endsWith('.test'));
+
+      var testCasesPathStream = entityStream.map((entity) => entity.uri);
+      var testCasePaths = await testCasesPathStream.toList();
+
+      testCaseList = testCasePaths.map((inputUri) {
+        var inputName = inputUri.toFilePath(windows: false).split('/').last;
+        var inputNameWithoutExt = inputName.substring(0, inputName.length - 5);
```
jonasfj, line 34: Similarly, consider what `package:path` can already do... `pub` already has a dependency upon it.

test/yaml_edit/test_case.dart
```diff
@@ -0,0 +1,318 @@
...
+  @override
+  String toString() =>
+      'method: $method, path: $path, index: $index, value: $value, deleteCount: $deleteCount';
+}
+
+/// Returns `true` if the [text] looks like it uses windows line endings.
+///
+/// The heuristic used is to count all `\n` in the text and if stricly more than
+/// half of them are preceded by `\r` we report `true`.
+bool detectWindowsLineEndings(String text) {
```
jonasfj, line 306: Can't we import this from utils.dart which has `getLineEnding`

test/yaml_edit/test_case.dart
```diff
@@ -0,0 +1,318 @@
...
+
+      value = mod[4];
+    }
+
+    return YamlModification(method, path, index, value, deleteCount);
+  }).toList();
+}
+
+/// Gets the YAML modification method corresponding to [method]
+YamlModificationMethod getModificationMethod(String method) {
```
jonasfj, line 264: Do these need to be exported? Let's review what we need to export from this file, and hide as much as reasonably possible.

test/yaml_edit/test_case.dart
```diff
@@ -0,0 +1,318 @@
...
+    final goldenOutput = states.join('---$inputLineEndings');
+
+    goldenFile.writeAsStringSync(goldenOutput);
+    state = TestCaseStates.createdGoldenFile;
+  }
+
+  /// Tests the golden file. Ensures that the number of states are the same, and
+  /// that the individual states are the same.
+  void testGoldenFile(File goldenFile) {
+    var inputFileName = inputUri.toFilePath(windows: false).split('/').last;
```
jonasfj, line 173: package:path can help here... I think..

test/yaml_edit/test_case.dart
```diff
@@ -0,0 +1,318 @@
...
+    /// parse -> immediately dump does not affect the string.
+    states.add(yamlBuilder.toString());
+
+    _performModifications();
+  }
+
+  void _performModifications() {
+    for (var mod in modifications) {
+      _performModification(mod);
+      states.add(yamlBuilder.toString());
```
jonasfj, line 125: Just an idea:

We could consider catching `PathError` and `AliasError` and then DUMP these into `states` as `# throws PathError`....

That way we can change path-error, and alias errors into the golden tests..

Note. clearly we should check that this can only happen in the modifucation..

test/yaml_edit/append_test.dart
```diff
@@ -0,0 +1,129 @@
...
+      expect(() => doc.appendToList([], 4), throwsPathError);
+    });
+
+    test('if it is a scalar', () {
+      final doc = YamlEditor('1');
+      expect(() => doc.appendToList([], 4), throwsPathError);
+    });
+  });
+
+  group('block list', () {
```
jonasfj, line 23: Just curious, wouldn't it be easier to do this was golden tests?

Not that I mind having both :D

test/yaml_edit/fuzz.dart
```diff
@@ -0,0 +1,291 @@
...
+    for (var i = 0; i < modificationsPerRound; i++) {
+      generator.performNextModification(editor);
+    }
+
+    print(editor);
+  }
+}
+
+/// Generates the random variables we need for fuzzing.
+class Generator {
```
jonasfj, line 49: Many exported things on this class... including the class, do we need to export all this..

test/yaml_edit/fuzz.dart
```diff
@@ -0,0 +1,291 @@
...
+
+import 'package:pub/src/yaml_edit/yaml_edit.dart';
+import 'package:pub/src/yaml_edit/src/wrap.dart';
+import 'package:yaml/yaml.dart';
+
+import 'problem_strings.dart';
+import 'test_utils.dart';
+
+/// Performs naive fuzzing on an initial YAML file based on an initial seed.
+void main(List<String> args) {
```
jonasfj, line 15: Can you write a few lines explaining what it does, and maybe suggest why we don't run it as part of normal integration testing...

Could we run it with just a few rounds in a `_test.dart` file?

test/yaml_edit/test_all.dart
```diff
@@ -0,0 +1,37 @@
...
+import 'preservation_test.dart' as preservation;
+import 'remove_test.dart' as remove;
+import 'source_edit_test.dart' as source;
+import 'special_test.dart' as special;
+import 'splice_test.dart' as splice;
+import 'update_test.dart' as update;
+import 'utils_test.dart' as utils;
+import 'wrap_test.dart' as wrap;
+
+Future<void> main() async {
```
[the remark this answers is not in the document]
walnutdust: ah so sorry about this - this file was for a workaround to get CodeCov working, will remove!

test/yaml_edit/fuzz.dart
walnutdust: made the class private!

test/yaml_edit/append_test.dart
walnutdust: yeah... I'm still a little torn between understanding the division as to what tests I should include as golden tests and what tests I should be including here instead - do you have any tips?

test/yaml_edit/test_case.dart
walnutdust: Updated the file to only export the `TestCases` class!

lib/src/yaml_edit/src/editor.dart
walnutdust: Fixed it, thanks for catching it!

test/yaml_edit/test_case.dart
```diff
@@ -0,0 +1,318 @@
...
+    var testCaseList = [];
+
+    if (testDir.existsSync()) {
+      /// Recursively grab all the files in the testing directory.
+      var entityStream = testDir.list(recursive: true, followLinks: false);
+      entityStream =
+          entityStream.where((entity) => entity.path.endsWith('.test'));
+
+      var testCasesPathStream = entityStream.map((entity) => entity.uri);
+      var testCasePaths = await testCasesPathStream.toList();
```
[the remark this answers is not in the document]
walnutdust: whoa thanks!

test/yaml_edit/append_test.dart
jonasfj: I see pros/cons in both directions... I guess golden tests are more useful for large complicated mutations.

test/yaml_edit/fuzz.dart
walnutdust: modified this to `fuzz_test.dart` to run 10 rounds of 50 modifications each!

test/yaml_edit/test_case.dart
walnutdust: hm I'm not sure if a golden test is the best place to check for path errors and alias errors - the way I see it, golden tests allow us best to easily compare the output space for space, character for character, for complicated or long series of changes, whereas unit testing in the test files might be more suited for the errors?

jonasfj: Good point!

walnutdust:
Closing due to the publication of `package:yaml_edit`

PR #2556, opened 2020-07-09
Only alert about new prereleases for prereleases

sigurdm:


jonasfj:
We still have some test failures in: `test/global/activate/uncached_package_test.dart`

Issue #2563, opened 2020-07-17
Invalid argument(s) (sdkPath): Must not be null


ditman:
# Environment

- `pub version` or `flutter pub version`: **Pub 2.9.0-21.0.dev.flutter-06183779d7**
- OS version: **Linux 5.2.17-1rodete3-amd64 x86_64**
- Are you using the Chinese community mirror or a corporate firewall? **No**

# Problem

When attempting to run `pub publish --dry-run` I get the following unexpected error: **Invalid argument(s) (sdkPath): Must not be null**

This seems to happen with any package. Check all `publishable` pre-submit runs in the [flutter/plugins PRs](https://github.com/flutter/plugins/pulls).

## Expected behavior

I'd expect the command to tell me if the package is publishable or not.

## Actual behavior

The command errors out with the following error after listing all the files that are about to be "published":

```
Invalid argument(s) (sdkPath): Must not be null
dart:core                                                      ArgumentError.checkNotNull
package:analyzer/src/dart/analysis/context_builder.dart 49:19  ContextBuilderImpl.createContext
package:pub/src/dart.dart 115:36                               AnalysisContextManager.createContextsForDirectory
package:pub/src/validator/strict_dependencies.dart 27:28       new StrictDependenciesValidator
package:pub/src/validator.dart 135:7                           Validator.runAll
package:pub/src/command/lish.dart 165:21                       LishCommand._validate
package:pub/src/command/lish.dart 141:15                       LishCommand.run
package:args/command_runner.dart 197:27                        CommandRunner.runCommand
package:pub/src/command_runner.dart 193:39                     PubCommandRunner.runCommand.<fn>
dart:async                                                     new Future.sync
package:pub/src/utils.dart 113:12                              captureErrors.wrappedCallback
dart:async                                                     runZoned
package:pub/src/utils.dart 130:5                               captureErrors
package:pub/src/command_runner.dart 193:13                     PubCommandRunner.runCommand
```

## --trace output

<details>
<summary>Here's the `--trace` run:</summary>

```
Publishing extension_google_sign_in_as_googleapis_auth 1.0.0 to https://pub.dartlang.org:
|-- .gitignore
...
|-- pubspec.yaml
'-- test
    '-- extension_google_sign_in_as_googleapis_auth_test.dart

Invalid argument(s) (sdkPath): Must not be null
dart:core                                                      ArgumentError.checkNotNull
[...]

PR #2567, opened 2020-07-20
Find sdk path with cli_util

sigurdm:
Fix of https://github.com/dart-lang/pub/issues/2563

### DIFF
--- a/lib/src/dart.dart
+++ b/lib/src/dart.dart
@@ -14,6 +14,7 @@ import 'package:analyzer/dart/analysis/context_locator.dart';
 import 'package:analyzer/dart/analysis/session.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/error/error.dart';
+import 'package:cli_util/cli_util.dart';
 import 'package:path/path.dart' as p;
 
 import 'exceptions.dart';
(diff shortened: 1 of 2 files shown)


lib/src/dart.dart
```diff
@@ -112,7 +113,8 @@ class AnalysisContextManager {
       }
 
       var contextBuilder = ContextBuilder();
-      var context = contextBuilder.createContext(contextRoot: root);
+      var context = contextBuilder.createContext(
+          contextRoot: root, sdkPath: getSdkPath());
```
devoncarew, line 117: `cli_util. getSdkPath()` used to have to do a lot more, but I believe currently it's just a thin wrapper around Platform.resolvedExecutable.

sigurdm: Yes. It is just `p.dirname(p.dirname(Platform.resolvedExecutable))` still I think it is nicer to get it from cli_util, so we can update that logic in one place should it change.

Also that logic is actually not enough when you use the binary in `out/ReleaseX64/dart` from a sdk checkout, but I think it is good enough for pub.

scheglov: I would argue that one should never run `dart` binaries from non-SDK, and we should not support this.

As it was explained to me, we sometimes don't have yet all components for full SDK, but I think in this case we should place what we have in the right structure to avoid adding such arbitrary support for non-standard locations.

sigurdm: In that case we should probably work on removing those binaries from `out/ReleaseX64` and friends...

sigurdm: Please comment on: https://github.com/dart-lang/cli_util/pull/44

PR #2578, opened 2020-07-22
Configure GitHub move app

jmagman:
Configure the [GitHub move app](https://github.com/apps/move) so issues can be moved from the Flutter repo issue tracker (or any repo) to this repo.

Users with write access to the source repo and this repo will be able to use the Transfer Issue button:
88121110-95a01880-cb79-11ea-8f3a-ea770196fd83

Preferences copied from [dart-lang/sdk](https://github.com/dart-lang/sdk/blob/master/.github/move.yml).

### DIFF
--- a/.github/move.yml
+++ b/.github/move.yml
@@ -1,0 +1,21 @@
+# Configuration for move-issues - https://github.com/dessant/move-issues
+
+# Delete the command comment when it contains no other content
+deleteCommand: true
+
+# Close the source issue after moving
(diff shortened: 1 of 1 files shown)


jmagman:
Note an admin will need to install the GitHub app at https://github.com/apps/move.

jonasfj:
@mit-mit, can you enable this app when you get back :)

mit-mit:
The app is already enabled on all dartlang repos

Issue #2438, opened 2020-04-07
Warn when publishing a null-safety compatible package with non-null-safe dependencies


jonasfj:
When a package is published we should warn if:
 * The package has opted into null-safety (sdk constraint lower bound of 2.10), AND,
 * A library in the package imports a non-null-safety compliant library.

Note. The second condition can be simplified into: if the package has dependency on a non-null-safety complaint package. Tracing import-paths might be preferred, because it creates fewer false positive warnings -- but it might also be too complicated.

We need to make this warning because it is highly likely that when the dependency is published in a non-null-safety compliant version the interface will be broken (whether intentionally or not).

mit-mit:
Shouldn't this be an error?

sigurdm:
> Shouldn't this be an error?

I believe we want to support mixed-mode packages? @jonasfj ?

jakemac53:
Mixed-mode has to be supported for package cycles to be able to publish. The vast majority of package cycles are introduced due to `dev_dependencies` though so its possible if this is only taking `dependencies` into account that we could avoid the problem.

Allowing publishing of mixed mode packages in general is very problematic. We want to discourage it as much as possible without completely blocking any valid use cases that require it.

jakemac53:
I would like to propose that we add a new class of "warning" to pub, which is one that requires an explicit flag in order to bypass it.

This decreases the chance that people publish these packages by ensuring they actually read the warning message (in order to know what flag to pass to bypass it).

We should include with this warning some text along the lines of:

```
Note that by publishing with non-migrated dependencies your package may be broken at any time if one of your dependencies migrates without a breaking change release. We highly recommend that you wait until all of your dependencies have been migrated before publishing.
```

jonasfj:
> Note that by publishing with non-migrated dependencies your package may be broken
at any time if one of your dependencies migrates without a breaking change release.

While it's likely that a dependency migrating will break you, it's not necessary a given, is it?
  

--------

```
To bypass this warning and publish anyways, pass
`--unsafe-publish-with-unmigrated-dependencies`
```
This is certainly an interesting idea, but I think it's an orthogonal concern. Filed https://github.com/dart-lang/pub/issues/2535

natebosch:
> While it's likely that a dependency migrating will break you, it's not necessary a given, is it?

If you publish a migrated package with unmigrated dependencies there are a few bad things that might happen:

[...]

PR #2583, opened 2020-07-23
Warn about publishing in mixed mode

sigurdm:
https://github.com/dart-lang/pub/issues/2438

### DIFF
--- a/lib/src/null_safety_analysis.dart
+++ b/lib/src/null_safety_analysis.dart
@@ -8,10 +8,10 @@ import 'package:analyzer/dart/analysis/context_builder.dart';
 import 'package:analyzer/dart/analysis/context_locator.dart';
 import 'package:cli_util/cli_util.dart';
 
-import 'package:pub_semver/pub_semver.dart';
 import 'package:path/path.dart' as path;
 
 import 'io.dart';
+import 'language_version.dart';
 import 'package.dart';
 import 'package_name.dart';
 import 'pubspec.dart';
@@ -26,7 +26,7 @@ enum NullSafetyCompliance {
 
   /// This package opted into null safety, but some file or dependency is not
   /// opted in.
-  apiOnly,
+  mixed,
 
   /// This package did not opt-in to null safety yet.
   notCompliant,
@@ -48,9 +48,8 @@ class NullSafetyAnalysis {
   /// Furthermore by awaiting the Future stored here, we avoid race-conditions
   /// from downloading the same package-version into [_systemCache]
   /// simultaneously when doing concurrent analyses.
-  final Map<PackageId, Future<NullSafetyCompliance>>
+  final Map<PackageId, Future<NullSafetyAnalysisResult>>
       _packageInternallyGoodCache = {};
-  static final _firstVersionWithNullSafety = Version.parse('2.10.0');
 
   NullSafetyAnalysis(SystemCache systemCache) : _systemCache = systemCache;
 
@@ -67,7 +66,7 @@ class NullSafetyAnalysis {
   ///
   /// If [packageId] is a relative path dependency [containingPath] must be
   /// provided with an absolute path to resolve it against.
-  Future<NullSafetyCompliance> nullSafetyCompliance(PackageId packageId,
+  Future<NullSafetyAnalysisResult> nullSafetyCompliance(PackageId packageId,
       {String containingPath}) async {
     // A space in the name prevents clashes with other package names.
     final rootName = '${packageId.name} importer';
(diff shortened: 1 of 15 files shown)


lib/src/solver/incompatibility.dart
```diff
@@ -425,7 +425,7 @@ class Incompatibility {
       buffer.write("which doesn't exist "
           '(${(cause as PackageNotFoundCause).exception.message})');
     } else {
-      buffer.write('which is forbidden');
+      buffer.write('which is forbidden ${latter.cause}');
```
jonasfj, line 428: What does this do?

lib/src/pubspec.dart
```diff
@@ -445,6 +446,17 @@ class Pubspec {
   bool get isEmpty =>
       name == null && version == Version.none && dependencies.isEmpty;
 
+  /// The language version implied by the sdk constraint.
+  ///
+  /// Given no or unbounded constraint we assume language version 1.0.
```
jonasfj, line 451: Is this correct?

Doesn't the spec say that the fallback language version is the whatever the current SDK has?

I can see that this might cause validation errors if you have no SDK constraint..

lib/src/null_safety_analysis.dart
```diff
@@ -86,29 +85,38 @@ class NullSafetyAnalysis {
         },
         sources: _systemCache.sources));
 
+    final rootLanguageVersion =
+        (await packageId.source.bind(_systemCache).describe(packageId))
+            .languageVersion;
+    if (!rootLanguageVersion.supportsNullSafety) {
+      return NullSafetyAnalysisResult(
+          NullSafetyCompliance.notCompliant, 'Is not opting in to null safety');
```
jonasfj, line 93: Can we be more specific? Like "Package $packageId has not opted into null-safety with the environment.sdk constraint in pubspec.yaml", or something like it.

lib/src/null_safety_analysis.dart
```diff
@@ -131,50 +139,60 @@ class NullSafetyAnalysis {
...
+    return NullSafetyAnalysisResult(
+        NullSafetyCompliance.mixed, firstBadPackage.reason);
   }
 }
+
+class NullSafetyAnalysisResult {
+  final NullSafetyCompliance compliance;
+
+  /// `null` if compliance == [NullSafetyCompliance.compliant].
+  final String reason;
```
jonasfj, line 192: Love it, we might even want to output this in the JSON output from `pub outdated`.

lib/src/null_safety_analysis.dart
```diff
@@ -86,29 +85,38 @@ class NullSafetyAnalysis {
...
         final boundSource = dependencyId.source.bind(_systemCache);
         final pubspec = await boundSource.describe(dependencyId);
-        final languageVersion = _languageVersion(pubspec);
-        if (languageVersion == null ||
-            languageVersion < _firstVersionWithNullSafety) {
-          return NullSafetyCompliance.notCompliant;
+        final languageVersion = pubspec.languageVersion;
+        if (languageVersion == null || !languageVersion.supportsNullSafety) {
+          return NullSafetyAnalysisResult(NullSafetyCompliance.notCompliant,
+              'package:${dependencyId.name} is not opted into null safety');
```
jonasfj, line 119: Again, can we say something with `environment.sdk in pubspec.yaml`

lib/src/null_safety_analysis.dart
```diff
@@ -131,50 +139,60 @@ class NullSafetyAnalysis {
...
-              if (Version(languageVersionToken.major,
-                      languageVersionToken.minor, 0) <
-                  _firstVersionWithNullSafety) {
-                return NullSafetyCompliance.notCompliant;
+              final languageVersion = LanguageVersion.fromLanguageVersionToken(
+                  languageVersionToken);
+              if (!languageVersion.supportsNullSafety) {
+                return NullSafetyAnalysisResult(
+                    NullSafetyCompliance.notCompliant,
+                    '$fileUrl is opting out of null safety');
```
jonasfj, line 159:
```suggestion
                    '$fileUrl is opting out of null-safety with language version marker $languageVersionToken');
```

lib/src/solver/incompatibility.dart
sigurdm: Oh this is debug code...

jonasfj: figured there was a small chance...

lib/src/pubspec.dart
sigurdm: pub.dev treats packages with no constraint as dart 1 packages, so at least for null-safety checking I think this makes sense. We might have to revise later.

Issue #2541, opened 2020-06-25
`pub get` and unsound null safety


mit-mit:
Running pub get in a project with null safety enabled (via with lower SDK constraint), should fail if any of the resolved packages do not support null safety. It should print something like:

```
Because <name of current package> requires Dart 2.10 with null safety, 
and one or more dependencies do not support that, version solving failed.

<list of deps that don't support null safety>
```

It should be allowed to succeed if a `--no-sound-null-safety` flag is passed, but then should print a warning:

```
Warning: getting packages with unsound null safety! Consider upgrading all code to enable sound null safety.
```

mit-mit:
@jonasfj please review

sigurdm:
Last we discussed this I believed we came to the conclusion that `pub get` should merely warn if the resolution ends up in mixed mode.

The reasoning is that there is no real benefit to not just doing what the user says.

I cannot find out if we wrote this anywhere though. jonasfj@ might know when he's back from vacation.

jonasfj:
> I cannot find out if we wrote this anywhere though. jonasfj@ might know when he's back from vacation.

Yeah, I think we failed to write it down. But I recall the same thing, notably because of the implementation concerns wrt. to `pub get --no-sound-null-safety` (and the fact that it has no real benefit).

mit-mit:
We discussed this in the language team, and agree that we can just do the warning. fyi @sigurdm

leafpetersen:
Status on this?  Is it still planned for the beta release?

sigurdm:
Yes I believe so. We procrastinated on landing the PR as we learned that 2.10 was not the version to get null-safety.

mit-mit:
Its my understanding that `2.10.0-0` currently is, and will remain, the lowest SDK constraint that enables null safety. There had previously been talk about moving that marker to `2.11.0-0` once `2.10.0` (stable) ships, but I think that was changed @leafpetersen ?

mit-mit:
Re the marker being `2.10`, see https://github.com/dart-lang/sdk/issues/43031

sigurdm:
Are we settled on 2.12 being the first version with the flag on by default?
I

jonasfj:
@sigurdm yes, I think the working theory now is that `2.12` will be the first version with null-safety enabled.

PR #2590, opened 2020-08-03
Warn when mixed mode

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2541


lib/src/solver/result.dart
```diff
@@ -19,6 +23,9 @@ class SolveResult {
   /// reachable from the root.
   final List<PackageId> packages;
 
+  /// The root package of this resolution.
+  final Package root;
```
jonasfj, line 27: Why does a package resolution have a root package? is that reasonable...

I see now harm, and we do always create dummy root packages when we otherwise need to make a resolution. I'm just wondering if this is smart.

jonasfj: Maybe, I should just be asking... why are we making this public?

lib/src/null_safety_analysis.dart
```diff
@@ -113,15 +115,29 @@ class NullSafetyAnalysis {
...
 
+  /// Decides if all dependendencies (transitively) have a language version
+  /// opting in to null safety, and no files in lib/ of these packages opt out
+  /// to a pre-null-safety language version.
+  ///
+  /// This will download all dependencies into [cache].
+  ///
+  /// Assumes [mainPackage] is opted in.
+  Future<NullSafetyAnalysisResult> nullSafetyComplianceOfResolution(
+      SolveResult result) async {
```
jonasfj, line 129: We could also be passing `root` as an argument if we wanted to, instead of making it public. I don't know what's right here.

Or rather, instead of passing `root` we could be passing a `package` argument, and rename the method to `nullSafetyComplianceInResolution`.. and then say it determines if a given _package_ is _null-safety compliant_ in a given _resolution_.

----
hmm, couldn't we just pass `Iterable<PackageId>` and ask if they are null-safety compliant?

Before we used:
```dart
final boundSource = dependencyId.source.bind(_systemCache);
final pubspec = await boundSource.describe(dependencyId);
```
So we don't need `result.pubspecs`, and we don't need to pass `root` or filter out the `dummy root package`, if we filter before we pass an `Iterable<PackageId>` to `nullSafetyComplianceOfResolution`..

sigurdm: Done

lib/src/solver/result.dart
sigurdm: I managed to avoid this by passing an explicit root package.

jonasfj (approved):

PR #2605, opened 2020-08-10
Avoid double loop

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2600


bsutton:
Is there a way to determine which version of dart this will be released in.

This bug is causing crashes our our production systems as it is causing them to run out of memory.

jonasfj:
@bsutton this will ship in `2.10`. Probably in the next 2.10-dev release too.

Issue #2593, opened 2020-08-04
Running a globally activated binary after SDK upgrade - Can't load Kernel binary: Invalid SDK hash.


kevmoo:
### Reproduction

1. globally activate a package
2. run it
3. See everything fine
4. Upgrade SDK
5. Run the globally activated package again

### What should happen

Everything works fine

### What actually happens

```
Can't load Kernel binary: Invalid SDK hash.
Then app runs
```

### Suggestion

Any way we can detect this? Recompile on first run after SDK upgrade?

### Work-around

Re-run `pub global activate <package>`

...it'd be nice if we even suggested this!

kevmoo:
CC @jakemac53 @mit-mit @jonasfj @srujzs

jonasfj:
Is this new? I think we've always detected this... Maybe it just surfaced in the UI because of the refactoring of `pub global run` when we added the `--enable-experiment`.

My experience was:
```
$ dart --version
Dart SDK version: 2.9.0-21.0.dev (dev) (Wed Jul 8 10:54:08 2020 +0200) on "linux_x64"
$ mono_repo 
Can't load Kernel binary: Invalid kernel binary format version.
Precompiling executable... (1.6s)
Precompiled mono_repo:mono_repo.
Manage multiple packages in one source repository
```

Looking at `PUB_CACHE/bin/mono_repo` I find:
```bash
#!/usr/bin/env sh
# This file was created by pub v2.8.1.
# Package: mono_repo
# Version: 2.3.0
# Executable: mono_repo
# Script: mono_repo
dart "/home/jonasfj/.pub-cache/global_packages/mono_repo/bin/mono_repo.dart.snapshot.dart2" "$@"

# The VM exits with code 253 if the snapshot version is out-of-date.
# If it is, we need to delete it and run "pub global" manually.
exit_code=$?
if [ $exit_code != 253 ]; then
  exit $exit_code
fi

pub global run mono_repo:mono_repo "$@"
```

So the problem is that wrapper scripts for the old versions tries to invoke the old snapshot.

We are no longer updating this snapshot, because we figured it prudent to include the SDK version in the filename of the snapshot. I believe we came to this conclusion after realizing that the behavior of the Dart VM was undefined when given old snapshots (that might also have been fixed using a hash of the SDK in the snapshot).

-----------
[...]

Issue #2602, opened 2020-08-10
binstubs from 'pub global activate' w. pub > 2.9.0 won't snapshot with newer SDKs


jonasfj:
The following is a binstub from pub 2.9.0:
```bash
#!/usr/bin/env sh
# This file was created by pub v2.9.0.
# Package: mono_repo
# Version: 2.3.0
# Executable: mono_repo
# Script: mono_repo
if [ -f /home/jonasfj/.pub-cache/global_packages/mono_repo/bin/mono_repo.dart-2.9.0.snapshot ]; then
  dart "/home/jonasfj/.pub-cache/global_packages/mono_repo/bin/mono_repo.dart-2.9.0.snapshot" "$@"
  # The VM exits with code 253 if the snapshot version is out-of-date.	
  # If it is, we need to delete it and run "pub global" manually.	
  exit_code=$?	
  if [ $exit_code != 253 ]; then	
    exit $exit_code	
  fi	
  pub global run mono_repo:mono_repo "$@"
else
  pub global run mono_repo:mono_repo "$@"
fi
```

When I update my SDK to `2.9.1` the snapshot will no longer work with `dart`, and I'll never run a snapshot again... Unless I do `pub global activate` again.

sigurdm:
Fixed by #2610

PR #2610, opened 2020-08-10
Refresh binstubs after recompile in global run

sigurdm:
This almost fixes https://github.com/dart-lang/pub/issues/2593 and https://github.com/dart-lang/pub/issues/2602

The missing part is that the binstub will print: "Can't load Kernel binary: Invalid kernel binary format version." to stderr first time run after update.

### DIFF
--- a/bin/pub
+++ b/bin/pub
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+# for details. All rights reserved. Use of this source code is governed by a
+# BSD-style license that can be found in the LICENSE file.
+
+# Runs ./pub.dart with dart from PATH
+
+dart `dirname "$0"`/pub.dart $*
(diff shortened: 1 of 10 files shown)


bin/pub
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+# for details. All rights reserved. Use of this source code is governed by a
+# BSD-style license that can be found in the LICENSE file.
+
+# Runs ./pub.dart with dart from PATH
+
+dart `dirname "$0"`/pub.dart $*
```
jonasfj, line 8: Why this file?

sigurdm: This file is invoked by the binstub in the tests.
getEnvironment() now returns a path with bin/ in front, so these are invoked instead of pub from the sdk. (See change to  `test/global/binstubs/utils.dart`)

jonasfj: I think `test/bin` or `tool/test-bin/` might be better...

`bin/` has a special meaning in the pub world.

sigurdm: 👍

Issue #2614, opened 2020-08-12
`pub --verbose get --no-precompile` hangs on dart 2.10.0


jmagman:
From https://github.com/flutter/flutter/issues/63206 @kika:
>When it hangs `-v` shows the conflict, that is never output without `-v` and then it just sits there wasting CPU.

>Minimal repro case

<details>
  <summary>This one hangs</summary>

```yaml
name: myproject
description: MyProject

version: 1.0.1+60

environment:
  sdk: ">=2.3.0 <3.0.0"

dependencies:
  flutter:
    sdk: flutter
  flutter_localizations:
    sdk: flutter
  meta: ^1.1.6
  functional_data: ^0.3.0

dev_dependencies:
  flutter_test:
    sdk: flutter
  flutter_driver:
    sdk: flutter
  test: any
  flutter_gherkin: any
  functional_data_generator: ^0.3.3

flutter_icons:
  android: true
  ios: true
  image_path: "assets/icons/icon.png"

flutter:
  uses-material-design: true
```  
</details>

<details>
  <summary>This one errors out with conflict (as expected)</summary>

```yaml
name: myproject
description: MyProject

version: 1.0.1+60

environment:
  sdk: ">=2.3.0 <3.0.0"

dependencies:
  flutter:
    sdk: flutter
  flutter_localizations:
    sdk: flutter
  meta: ^1.2.2
  functional_data: ^0.3.0

dev_dependencies:
  flutter_test:
    sdk: flutter
  flutter_driver:
    sdk: flutter
  test: any
  flutter_gherkin: any
  functional_data_generator: ^0.3.3

flutter_icons:
  android: true
  ios: true
  image_path: "assets/icons/icon.png"

flutter:
  uses-material-design: true
```  
</details>

>The only difference is `meta: ^1.1.6` in first one and `meta: ^1.2.2` in the second

<details>
  <summary>flutter doctor -v</summary>

[✓] Flutter (Channel stable, 1.20.1, on Mac OS X 10.15.6 19G73, locale en-US)
    • Flutter version 1.20.1 at /Volumes/ExtStripe/Scratch/flutter
    • Framework revision 2ae34518b8 (2 days ago), 2020-08-05 19:53:19 -0700
    • Engine revision c8e3b94853
    • Dart version 2.9.0

[✓] Android toolchain - develop for Android devices (Android SDK version 28.0.3)
    • Android SDK at /Users/kika/Library/Android/sdk
[...]

PR #2615, opened 2020-08-12
Avoid O(2^N) complexity when reformating an incompatibility-tree of d…

jonasfj:
…epth N.

For incompatibilities with a `ConflictCause` the `reformatRanges` called itself
on each branch `ConflictCause.conflict` and `ConflictCause.other`.
Then it also `_reformatCause` which again calls `reformatRanges` for
each branch of a `ConflictCause`.

This means `reformatRanges` was called 2^N times for any node at depth N
in an `Incompatibility`-tree. Hence, a degenerate incompatibility tree
could effective cause `pub get` to hang, instead of printing an error
signaling a resolution conflict.

This bug has not affected `pub get` invocations where a resolution was
possible.

Please wait for tests, they haven't completed running locally yet :)

Fixes https://github.com/dart-lang/pub/issues/2614

### DIFF
--- a/lib/src/solver/reformat_ranges.dart
+++ b/lib/src/solver/reformat_ranges.dart
@@ -25,21 +25,16 @@ import 'term.dart';
 /// the release version (`<2.0.0`) if no pre-releases exist or with an inclusive
 /// bound on the last pre-release version that actually exists
 /// (`<=2.0.0-dev.1`).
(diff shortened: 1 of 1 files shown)


jonasfj:
I tried to reproduce this... but even with all the versions faked I could get the test case to hang... it just took a long time to run with 30k LOC... :see_no_evil: 

In any case, I don't think this is a bug we are likely to reintroduce.

Issue #2311, opened 2020-01-22
pubspec.yaml parser ignores invalid keys.


bsutton:
When parsing the pubspec.yaml invalid keys are ignored.

I recently had a problem with my dependency_overrides not working.

I eventually worked out that the problem was that I had missed spelt dependency_overrides as dependency_overides (not the missing 'r').

Errors like invalid key names should be spat out by the parser.
I can't be the first person who has spent some time trying to work out what's wrong with my pubspec.yaml.

jakemac53:
For better or worse this is an intentional decision, it allows other tools to piggyback off of the file and add their own configuration. Also any key which has ever been supported has to be supported in perpetuity.

We could probably try to catch common misspellings though and at least warn. Dependency_overrides in particular is easy to typo.

Maybe if an unknown key is within an edit distance of 1 or 2 from a know key give a warning that it is likely a misspelling?

sigurdm:
> We could probably try to catch common misspellings though and at least warn. Dependency_overrides in particular is easy to typo.

+1 I think we should do this. I hate spelling dependency_overrides myself.

bsutton:
> Also any key which has ever been supported has to be supported in perpetuity.
An approach similar to deprecation could be used to overcome that problem.

If you want to allow other apps to add their data why not force some conventions on them:

e.g.
All third party data must appear under a special section such as:
```
custom:
  my_custom_data:
```

Under the custom section each app could be responsible for its own section and the validation thereof.
pub could then fully validate the rest of the file.

The edit distance would be helpful but enforcement would be much better.

jakemac53:
It is probably worth analyzing all published packages to see how many of them violate the desired rules (no unrecognized keys) today, and in what ways (intentional, or likely spelling mistakes, etc).

It should be noted that extensibility is not the only reason that it is so permissive. For instance the format actually has to not just be backwards compatible but also forwards compatible (any version of pub has to be able to parse all previous and future pubspecs). As an example this is why the sdk version ranges don't allow ^ syntax to this day. Mostly this is during the version solve I believe, but it could quite possible have a more lax implementation in that scenario, and only enable warnings/errors for the root package or something like that.

cc @munificent do you remember if there are other concrete reasons that unrecognized keys are allowed?

sigurdm:
@jonasfj suggests we could consider implementing a lint for the analyzer that warns about keys not recognized by current pub.

Then custom data can be added with a `# ignore:...` comment.

[...]

Issue #1971, opened 2018-08-16
Consider warning about unrecognized fields when parsing pubspecs


jakemac53:
The pubpsec parser has to be lenient to allow for any old pubspec to be properly parsed, but this provides negative value at development time because simple spelling mistakes cause sections to be skipped instead of giving errors or warnings.

This came up recently with the `sdk` constraints in gitter - a user was confused about where to put the constraint, partly because [this page](https://www.dartlang.org/dart-2#migrating-packages) is misleading (it omits the `environment` key).

Another common mistake is misspelling `dev_dependencies` or `dependency_overrides`, or using the singular version of those (I have seen googlers do this several times in person).

My initial though is to emit a warning message about unrecognized fields by default (but not fail). We could also hide it behind a separate command but that would have far less value as it wouldn't be very discoverable.

We could also add a --strict flag that causes it to actually fail on unrecognized fields.

Thoughts?

natebosch:
> What should happen with additional fields supported by Flutter?

That's a good callout - we could ignore those as well but it does introduce a point of coupling that is sub-optimal...

jakemac53:
I wonder if flutter would be willing to modify pub in their version of the sdk to support parsing those fields, then they would maintain/own the logic for it at least.

natebosch:
looking again - I guess flutter already nests everything under `flutter` so we could ignore that entirely and skip the complexity of an extra configuration file...

PR #2616, opened 2020-08-13
Checks pubspec keys for potential typos

walnutdust:
@jonasfj 

Check potential typos in pubspec keys via the [Sørensen–Dice coefficient](https://en.wikipedia.org/wiki/S%C3%B8rensen%E2%80%93Dice_coefficient) for two strings. Currently implemented as a validator, which should allow it to run with the lish commands, and can be easily expanded to cover the other commands too!

Fixes: https://github.com/dart-lang/pub/issues/2311
Fixes: https://github.com/dart-lang/pub/issues/1971


lib/src/validator/pubspec_typo.dart
```diff
@@ -0,0 +1,68 @@
...
+        if (dice > bestDiceCoefficient) {
+          bestDiceCoefficient = dice;
+          closestKey = validKey;
+        }
+      }
+
+      // 0.73 is a magic value determined by looking at the most common typos
+      // in all the pubspecs on pub.dev.
+      if (bestDiceCoefficient >= 0.73 && bestDiceCoefficient < 1.0) {
+        warnings.add('$key is not a key recognizable by pub - '
```
jonasfj, line 34: Can we add an `assert(key != closestKey)` or perhaps add that as a condition in the `if` statement..

jonasfj, line 34:
```suggestion
        warnings.add('"$key" is not a key recognizable by pub - '
```

lib/src/validator/pubspec_typo.dart
```diff
@@ -0,0 +1,68 @@
...
+            'did you mean $closestKey?');
```
jonasfj, line 35:
```suggestion
            'did you mean "$closestKey?"');
```

lib/src/validator/pubspec_typo.dart
```diff
@@ -0,0 +1,68 @@
...
+      }
+
+      if (closestKey == 'author' || closestKey == 'authors') {
+        warnings.add('The "$closestKey" key is deprecated - Use a verified '
+            'publisher (https://dart.dev/tools/pub/verified-publishers) '
+            'instead.');
+      }
```
jonasfj, line 42: Don't we have another warns about `author` and `authors`?

lib/src/validator/pubspec_typo.dart
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+}
+
+/// List of keys in `pubspec.yaml` that will be recognized by pub, as well as
+/// the depreciated `author` key.
+///
+/// Retrieved from https://dart.dev/tools/pub/pubspec
```
jonasfj, line 50: Maybe we should write a tool in `tool/pubspec-field-histogram.dart` that iterates through all pubspecs on pub.dev and makes a histogram counting how frequently top-level keys are used.

We shouldn't print typo warnings for deprecated keys, and we kind of want to discover what people are using.

walnutdust: Already done :+1: 

name: 14677
description: 14677
version: 14677
homepage: 14575
environment: 13656
dependencies: 12982
dev_dependencies: 11980
flutter: 7328
authors: 851
author: 8455
repository: 1190
issue_tracker: 521
documentation: 520
executables: 367
dependency_overrides: 77
scripts: 8
ffigen: 1
verified publisher: 4
targets: 1
publish_to: 58
analyzer: 12
license: 23
linter: 4
maintainer: 82
copyright: 3
email: 27
dartdoc: 5
custom: 3
dev_dependecies: 1
publisher: 4
publish: 1
respsitory: 2
auth: 1
_dependency_overrides: 3
import_sorter: 1
homeage: 1
Repository: 1
sdk: 3
doucumentation: 1
module: 8
plugin: 2
git: 13
deprecated: 2
deprecation_message: 2
api: 1
source: 1
Site: 1
docs: 1
dev_dependencies_overrides: 1
transformers: 256
changelog -x: 1
issues: 2
protocol: 128
imc_def: 1
repostory: 2
environment_config: 1
test: 1
strong-mode: 1
latest_commit: 1
assets: 2
enable-experiment: 2
docstring: 3
dartrix: 3
contributors: 1
reposity: 1
streams_channel: 1
uploader: 3
COPYING: 1
LICENSE: 1
UNLICENSE: 1
README.MD: 1
github: 9
Reviewed-by: 1
dependecies: 2
~ pedantic: 1
publishers: 2
flutter_icons: 2
flutter_manifest: 1
auther: 1
autho: 1
pubpage: 8
originauthor: 1
originhomepage: 1
env: 1
web: 9
uses-material-design: 1
flutter_i18n: 2
about: 1
website: 3
private: 1
path: 7
anukeerthi_library: 2
http: 1
excutables: 1
dev_dependency_overrides: 4
repositiory: 1
dev-dependencies: 2
publish-to: 1
environtment: 1
executable: 4
changelogname: 2
ghooks: 1
refit_http: 1
jaguar_http: 2
dev_transformers: 1
enviroent: 1
environnment: 4
devDependencies: 2
dev_dependency: 1

walnutdust: Checking for the `author` key was suggested by @sigurdm, since it was a key that was previously recognized, and we are able to detect it and suggest the necessary corrective actions to be taken instead! I could remove it if you want though!

lib/src/validator/pubspec_typo.dart
sigurdm: Yes but not avthors

lib/src/validator/pubspec_typo.dart
sigurdm, line 34:
```suggestion
        warnings.add('$key is not a key recognized by pub - '
```

walnutdust: if `key == closestKey` then `bestDiceCoefficient = 1.0`, since the two strings should have the same bigrams in that case!

lib/src/dice_coefficient.dart
```diff
@@ -0,0 +1,89 @@
...
+/// {'da', 'ar', 'rt}), and calculates the coefficient based on twice the
+/// number of common elements over the total number of elements.
+///
+/// In this implementation, identical bigrams that appear more than once in the
+/// string will be calculated accordingly (e.g. 'aaaa' is split into {'aa', 'aa'
+/// , 'aa'}), and each them a match is made, we strike off one of the bigrams.
+/// The effect of doing so increases the meaingfulness of the coefficient
+/// produced.
+///
+/// Our implementation also works with empty strings and one-character strigns.
```
sigurdm, line 19:
```suggestion
/// Our implementation also works with empty strings and one-character strings.
```

lib/src/dice_coefficient.dart
```diff
@@ -0,0 +1,89 @@
...
+/// diceCoefficient('a', 'b')                               /// 0.0
+/// diceCoefficient('aa', 'aaa')                            /// 0.6666667
+/// diceCoefficient('aaaa', 'aaaaa')                        /// 0.857142
+/// diceCoefficient('night', 'nacht')                       /// 0.25
+/// diceCoefficient('dev_dependencies', 'dev-dependencies') /// 0.8666666
+/// diceCoefficient('dependency', 'dependencies')           /// 0.8
+/// ```
+///
+/// [1]: https://en.wikipedia.org/wiki/S%C3%B8rensen%E2%80%93Dice_coefficient
+double diceCoefficient(String string1, String string2) {
```
jonasfj, line 35: Perhaps there is an argument to be made that we should investigate if this is the best to detect likely typos.
At-least we should look at what other fuzzy matching applications exists out there.

lib/src/dice_coefficient.dart
walnutdust: Added the explanation in - thanks! :)

lib/src/dice_coefficient.dart
walnutdust: Just tested with a levenshtein distance implementation, and the results are at: https://gist.github.com/walnutdust/5d0f151025fc986c5606d87fa15d5eb5.

From this it does seem like levenshtein would be the better way to go, and I can code it up once we can confirm this!

jonasfj: I agree this is much better let's go with experiment (2) or (3) from:
https://gist.github.com/walnutdust/5d0f151025fc986c5606d87fa15d5eb5

lib/src/levenshtein.dart
```diff
@@ -0,0 +1,59 @@
...
+/// required to convert [s] to [t]. An edit operation can either be:
+///
+/// 1. Inserting a character
+/// 2. Deleting a character
+/// 3. Substituting a character.
+///
+/// This implementation is case-sensitive.
+int levenshteinDistance(String s, String t) {
+  ArgumentError.checkNotNull(s);
+  ArgumentError.checkNotNull(t);
```
jonasfj, line 15: nit:
```suggestion
  ArgumentError.checkNotNull(s, 's');
  ArgumentError.checkNotNull(t, 't');
```

lib/src/levenshtein.dart
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  return distances[b.length];
+}
+
+/// Utility function to calculate the minimum of [a], [b], and [c].
+T _min3<T extends num>(T a, T b, T c) {
+  ArgumentError.checkNotNull(a, 'a');
+  ArgumentError.checkNotNull(b, 'b');
+  ArgumentError.checkNotNull(c, 'c');
```
jonasfj, line 57: In this case that might be a bit expensive :D

nit: having this is probably fine :D

lib/src/validator/pubspec_typo.dart
```diff
@@ -0,0 +1,73 @@
...
+  @override
+  Future validate() async {
+    final fields = entrypoint.root.pubspec.fields;
+
+    if (fields == null) return;
+
+    for (final key in fields.keys) {
+      var bestLevenshteinRatio = 100.0;
+      var closestKey = '';
+
```
jonasfj, line 22: Should we make a quick check:
`if (_validPubspecKeys.contains(key)) { continue; }`

Just to save some compute time..

jonasfj: I guess we're probably not CPU bound, just seems like a really cheap thing to do :D

lib/src/validator/pubspec_typo.dart
```diff
@@ -0,0 +1,73 @@
...
+            'did you mean "$closestKey"?');
+      }
+
+      /// Differs from the above because we allow exact matches here too.
+      if (bestLevenshteinRatio < 0.21 &&
+          (closestKey == 'author' || closestKey == 'authors')) {
+        warnings.add('The "$closestKey" key is deprecated - Use a verified '
+            'publisher (https://dart.dev/tools/pub/verified-publishers) '
+            'instead.');
+      }
```
jonasfj, line 47: Don't we have another thing that warns about `authors` and `author` being deprecated?

So now users with `author` key will get 2 warnings... and users with `authoR` will get 2 warnings.
Maybe we just drop this part?

lib/src/validator/pubspec_typo.dart
```diff
@@ -0,0 +1,73 @@
...
+          closestKey = validKey;
+        }
+      }
+
+      /// 0.21 is a magic value determined by looking at the most common typos
+      /// in all the pubspecs on pub.dev.
+      if (bestLevenshteinRatio > 0 && bestLevenshteinRatio < 0.21) {
+        warnings.add('"$key" is not a key recognized by pub - '
+            'did you mean "$closestKey"?');
+      }
```
jonasfj, line 39: Maybe I suggest that we add at-most 3 typo warnings? So as to not drown all the other warnings?

We could even go as far as to say, we only have a warning for the closest typo, and not for more than one.
WDYT?

I'm mostly concerned we could be drowning all other warnings.

lib/src/levenshtein.dart
walnutdust: I could remove it, since it's a private function only used in one context and I'm pretty certain the inputs won't ever be null. Let me know!

lib/src/validator/pubspec_typo.dart
walnutdust: sounds good!

lib/src/validator/pubspec_typo.dart
walnutdust: I don't think users with `authoR` will get two warnings though, because it's checked with an exact match in `validator/deprecated_fields`, so this should only trigger the typo warning. 

I'm open to both removing this part, or adding a `bestLevenshteinRatio > 0` to ensure that it only picks up typos - let me know!

lib/src/validator/pubspec_typo.dart
walnutdust: hmm because this picks up typos, and we [experimentally validated](https://gist.github.com/walnutdust/5d0f151025fc986c5606d87fa15d5eb5) with real pubspecs that all of the other keys are fine (i.e. no false positives),  it doesn't feel like there will be too many typo warnings.

I'll do what you suggested and add at most 3 just in case!

lib/src/levenshtein.dart
jonasfj: Yeah, let's just remove it here :D

lib/src/validator/pubspec_typo.dart
```diff
@@ -0,0 +1,83 @@
...
+        warnings.add('"$key" is not a key recognized by pub - '
+            'did you mean "$closestKey"?');
+        warningCount++;
+
+        if (closestKey == 'author' || closestKey == 'authors') {
+          warnings.add('The "$closestKey" key is deprecated - Use a verified '
+              'publisher (https://dart.dev/tools/pub/verified-publishers) '
+              'instead.');
+          warningCount++;
+        }
```
jonasfj, line 54: I propose we drop this part to keep it simple.

Otherwise, we should make it, an `OR`.

So you don't get two warnings..

walnutdust: I'll drop it then!

jonasfj (approved):

PR #2617, opened 2020-08-13
Implementation of a `bump` command that bumps the current package version

walnutdust:
Updates the version of the current package in `pubspec.yaml`. By default, it bumps the version up to the nearest stable patch number. Users may pass in the `--major|--minor|--patch`, which bumps the version up to the nearest stable major/minor/patch numbers accordingly. Users may also pass in a version number as an argument, and the version number will be bumped to the argument accordingly.

**Examples**
1. `pub bump` 1.2.3 -> 1.2.4
2. `pub bump` 1.2.3-dev -> 1.2.3
3. `pub bump --major` 1.2.3 -> 2.0.0
4. `pub bump --minor` 1.2.3 -> 1.3.0
5. `pub bump 3.0.0` 1.2.3 -> 3.0.0

### DIFF
--- a/lib/src/command/bump.dart
+++ b/lib/src/command/bump.dart
@@ -1,0 +1,123 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:pub_semver/pub_semver.dart';
+import 'package:yaml_edit/yaml_edit.dart';
+
+import '../command.dart';
+import '../io.dart';
+import '../log.dart' as log;
+
+/// Handles the `bump` pub command.
+///
+/// Updates the version of the current package in `pubspec.yaml`. By default,
+/// it bumps the version up to the nearest stable patch number. Users may pass
+/// in the `--major|--minor|--patch`, which bumps the version up to the nearest
+/// stable major/minor/patch numbers accordingly. Users may also pass in a
+/// version number as an argument, and the version number will be bumped to the
+/// argument accordingly
+///
+/// **Examples**
+/// 1. `pub bump` 1.2.3 -> 1.2.4
+/// 2. `pub bump` 1.2.3-dev -> 1.2.3
+/// 3. `pub bump --major` 1.2.3 -> 2.0.0
+/// 4. `pub bump --minor` 1.2.3 -> 1.3.0
+/// 5. `pub bump 3.0.0` 1.2.3 -> 3.0.0
+class BumpCommand extends PubCommand {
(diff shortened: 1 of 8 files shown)


lib/src/command/bump.dart
```diff
@@ -0,0 +1,123 @@
...
+        /// behavior.
+
+        if (initialVersion.isPreRelease) {
+          finalVersion = Version(initialMajor, initialMinor, initialPatch);
+        } else {
+          finalVersion = Version(initialMajor, initialMinor, initialPatch + 1);
+        }
+      }
+    }
+    _updatePubspecVersion(initialVersion, finalVersion);
```
sigurdm, line 105: I wonder if we should do an acquireDependencies here. Otherwise the next `pub run` command will complain that the pubspec.yaml is newer than the pubspec lock.

Given that we know the dependencies didn't change, we could also consider rewriting the current `pubspec.lock` (and .packages and .dart_tool/package_config.json) to mark them as uop-to-date.

@jonasfj opinions?

jonasfj: > I wonder if we should do an acquireDependencies here. Otherwise the next pub run command will complain that the pubspec.yaml is newer than the pubspec lock.

Will it? Won't it just do the verification heuristic to verify that `pubspec.lock` is still a resolution that satisfies `pubspec.yaml`.

Maybe `pub run` will make some noise? I think the assertion checks the timestamps and updates them if there is nothing to be done. And if there is something to be done, then it'll cause a problem. But bumping the version should never causing anything to change (unless you have a cyclic dependency that depends on your specific version of the current package).

sigurdm: > Won't it just do the verification heuristic to verify that pubspec.lock is still a resolution that satisfies pubspec.yaml.

Ah yes, that should probably be good enough!

lib/src/command/bump.dart
```diff
@@ -0,0 +1,123 @@
...
+/// 4. `pub bump --minor` 1.2.3 -> 1.3.0
+/// 5. `pub bump 3.0.0` 1.2.3 -> 3.0.0
+class BumpCommand extends PubCommand {
+  @override
+  String get name => 'bump';
+  @override
+  String get description =>
+      'Updates the version of the current package in pubspec.yaml.';
+  @override
+  String get invocation => 'pub bump [--major|--minor|--patch]';
```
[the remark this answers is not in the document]
walnutdust: should this be a `pub bump [version | --major | --minor | --patch]` instead? I'm not sure how `<version>` will interact with the flags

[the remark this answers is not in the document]
jonasfj: Exactly, you have right :D

PR #2618, opened 2020-08-13
Adding the `pub add` command

walnutdust:
`pub add [--dev] <package>[:^<constraint>] ...`

This pull request uses `package:yaml_edit` to add the ability to add git, hosted, path, and sdk packages to `pubspec.yaml`. In the case of hosted dependencies, we allow the user to add a version constraint. After the dependencies are resolved, a new entry will be added in `pubspec.yaml`.

If no constraints are provided, (e.g. `pub add foo`), we'll do a resolution to find the latest version of foo that is compatible with the other dependencies in your pubspec.yaml, and then that as the lower-bound in a ^x.y.z constraint.

Example: "pub add retry" might give you "retry: ^3.0.1", if version 3.0.1 of retry is the latest version compatible with your other dependencies.


lib/src/command/add.dart
```diff
@@ -0,0 +1,326 @@
...
+    /// [acquireDependencies] (as opposed to calling [acquireDependencies] on
+    /// the in-memory pubspec and writing it later) because there will be
+    /// conflicts in other parts of [pub] if the pubspec is newer than
+    /// `pubspec.lock`.
+    _updatePubspec(resultPackage, packageInformation);
+
+    /// Create a new [Entrypoint] since we have to reprocess the updated
+    /// pubspec file.
+    await Entrypoint.current(cache).acquireDependencies(SolveType.GET,
+        dryRun: argResults['dry-run'], precompile: argResults['precompile']);
```
sigurdm, line 108: Do we need to do acquireDependencies if `--dry-run` was passed?

walnutdust: hm, I was thinking that this would be in line with what `pub get` does with the `dry-run` flag at the moment, showing what packages will be changed if this new package will be added.

I did mess up the implementation here though - I should probably be having two paths, with the dry run one not updating the actual pubspec and instead calling `acquireDependencies` from a new entrypoint. 

Let me know what your thoughts are on dry-run and I'll make the changes accordingly!

sigurdm: I first thought it would be enough to show the changes made to your pubspec, but you are right - other packages might change their locked version.

lib/src/command/add.dart
```diff
@@ -0,0 +1,326 @@
...
+      /// We cannot simply assign the value of version since it is likely that
+      /// [pubspecInformation] takes on the type
+      /// [Map<String, Map<String, String>>]
+      pubspecInformation = {
+        ...pubspecInformation,
+        'version': '^${resultPackage.version}'
+      };
+    }
+
+    final dependencyKey = isDevelopment ? 'dev_dependencies' : 'dependencies';
```
sigurdm, line 299: Somehow it would seem nicer to me if `isDevelopment` was passed as a parameter to `_updatePubspec`, that makes it easier to understand in isolation.

lib/src/command/add.dart
```diff
@@ -0,0 +1,326 @@
...
+  @override
+  String get name => 'add';
+  @override
+  String get description => 'Add a dependency to the current package.';
+  @override
+  String get invocation => 'pub add <package>[:<constraint>] [options]';
+  @override
+  String get docUrl => 'https://dart.dev/tools/pub/cmd/pub-add';
+
+  bool get isDevelopment => argResults['dev'];
```
sigurdm, line 36: perhaps rename to `isDev`.

jonasfj:
Should we consider splitting `pub remove` into a separate PR?

test/remove/remove_test.dart
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+// import 'package:pub/src/exit_codes.dart' as exit_codes;
```
jonasfj, line 7: Remove commented out code

test/descriptor/pubspec.dart
```diff
@@ -0,0 +1,49 @@
...
+import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart';
+import 'package:yaml_edit/yaml_edit.dart';
+
+import '../descriptor.dart';
+
+class PubspecDescriptor extends FileDescriptor {
+  final String _contents;
+
+  PubspecDescriptor(String name, this._contents) : super.protected(name);
```
jonasfj, line 20: Please leave comment revealing what format the `_contents` should be in.. I almost expect that this YAML :)

But documenting how it'll be compared it certainly nice.

walnutdust:
Got it - will create a separate PR in a bit :)

test/descriptor.dart
```diff
@@ -48,8 +49,8 @@ FileDescriptor outOfDateSnapshot(String name) =>
 ///
 /// [contents] may contain [Future]s that resolve to serializable objects,
 /// which may in turn contain [Future]s recursively.
-Descriptor pubspec(Map<String, Object> contents) =>
-    file('pubspec.yaml', yaml(contents));
+PubspecDescriptor pubspec(Map<String, Object> contents) =>
```
jonasfj, line 52: Could we just return a `Descriptor`, given that `PubspecDescriptor` is an instance of `Descriptor`?

test/descriptor/pubspec.dart
```diff
@@ -0,0 +1,49 @@
...
+
+import 'package:collection/collection.dart';
+import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart';
+import 'package:yaml_edit/yaml_edit.dart';
+
+import '../descriptor.dart';
+
+class PubspecDescriptor extends FileDescriptor {
```
jonasfj, line 17: Is this actually a `PubspecDescriptor` ? isn't it more like a `YamlFileDecriptor`? It'll work for any YAML file won't it?

test/descriptor/pubspec.dart
```diff
@@ -0,0 +1,49 @@
...
+    var fullPath = p.join(parent ?? sandbox, name);
+    if (!await File(fullPath).exists()) {
+      fail("File not found: '$fullPath'.");
+    }
+
+    var bytes = await File(fullPath).readAsBytes();
+
+    final actualContentsText = utf8.decode(bytes);
+    final actualYaml = YamlEditor(actualContentsText);
+    final expectedYaml = YamlEditor(_contents);
```
jonasfj, line 40: Shouldn't we just use `package:yaml` to parse? Is there any need to use `YamlEditor` here?

test/add/sdk/sdk_test.dart
```diff
@@ -0,0 +1,101 @@
...
+      ]),
+      d.dir('bin/cache/pkg', [
+        d.dir(
+            'baz', [d.libDir('baz', 'foo 0.0.1'), d.libPubspec('baz', '0.0.1')])
+      ]),
+      d.file('version', '1.2.3')
+    ]).create();
+  });
+
+  test("gets an SDK dependency's dependencies", () async {
```
jonasfj, line 33:
```suggestion
  test("adds an SDK dependency's dependencies", () async {
```

And same changes to other test cases :)

I assume these are more less copied from some `pub get` tests.

test/add/sdk/sdk_test.dart
```diff
@@ -0,0 +1,101 @@
...
+      })
+    ]).validate();
+  });
+
+  test(
+      "gets an SDK dependency's dependencies with version constraint specified",
+      () async {
+    await d.appDir({}).create();
+    await pubAdd(
+        args: ['foo:0.0.1', '--sdk', 'flutter'],
```
jonasfj, line 59: I think we need to forbid version constraint when specifying `--sdk <sdk>`.

When a package is installed from the SDK, it is versioned through the `environment.<sdk>` constraint in `pubspec.yaml`.
Ie. it's versioned through the SDK constraint your package/app has.

So specifying a version constraint on a package that whose source is `sdk` never makes sense.
I think we need to special case this to exit with an error message in `pub add`.

test/add/sdk/sdk_test.dart
```diff
@@ -0,0 +1,101 @@
...
+        environment: {'FLUTTER_ROOT': p.join(d.sandbox, 'flutter')});
+
+    await d.dir(appPath, [
+      d.pubspec({
+        'name': 'myapp',
+        'dependencies': {
+          'foo': {'sdk': 'flutter', 'version': '0.0.1'}
```
jonasfj, line 66: Is this actually a valid thing? Do we have other tests that does this?

I'm probably not familiar with this aspect of `pub`.

test/add/hosted/non_default_pub_server_test.dart
```diff
@@ -0,0 +1,62 @@
...
+    await d.cacheDir({'foo': '1.2.3'}, port: server.port).validate();
+    await d.appPackagesFile({'foo': '1.2.3'}).validate();
+    await d.appDir({
+      'foo': {
+        'version': '1.2.3',
+        'hosted': {'name': 'foo', 'url': url}
+      }
+    }).validate();
+  });
+
```
jonasfj, line 35: Can we add a test case with `^1.2.3` constraint?

test/add/hosted/non_default_pub_server_test.dart
```diff
@@ -0,0 +1,62 @@
...
+import '../../test_pub.dart';
+
+void main() {
+  test('adds a package from a non-default pub server', () async {
+    // Make the default server serve errors. Only the custom server should
+    // be accessed.
+    await serveErrors();
+
+    var server = await PackageServer.start((builder) {
+      builder.serve('foo', '1.2.3');
```
jonasfj, line 17: nit, can we add a few more versions to the 3rd-party pub server?

test/add/git/git_test.dart
```diff
@@ -0,0 +1,75 @@
...
+void main() {
+  test('adds a package from git', () async {
+    ensureGit();
+
+    await d.git(
+        'foo.git', [d.libDir('foo'), d.libPubspec('foo', '1.0.0')]).create();
+
+    await d.appDir({}).create();
+
+    await pubAdd(args: ['foo', '--git-url', '../foo.git']);
```
jonasfj, line 20: Can we add a test case that `pub add foo:^1.2.3 --git-url ../foo.git` won't work..

At-least I expect that one cannot have a version constraint if fetching from `git`, and that we have some special handling in place to ensure that the user gets an error message from `pub add`.

I haven't tested, but I assume that `dependencies: {'mypkg': {'git': {'url': '../foo.get', 'version': '1.2.3'}}}`  will never work in a `pubspec.yaml`, or at-least that the `version` constraint is ignored (as `pub` generally ignores unrecognized keys). I haven't tests, I'm just guessing that this is the case, we probably should test this too :D

test/add/git/ref_test.dart
```diff
@@ -0,0 +1,62 @@
...
+    await d.git('foo.git',
+        [d.libDir('foo', 'foo 2'), d.libPubspec('foo', '1.0.0')]).commit();
+
+    await d.appDir({}).create();
+
+    await pubAdd(
+        args: ['foo', '--git-url', '../foo.git', '--git-ref', 'old'],
+        error: contains('ambiguous argument \'old\': unknown revision or path '
+            'not in the working tree.'),
+        exitCode: exit_codes.UNAVAILABLE);
```
jonasfj, line 60: In this and all the other test cases where we test that `pubAdd` fails. We should probably have an assertion checking that the `pubspec.yaml` have not changed. Ideally, also that we haven't changed neither `pubspec.lock` or `.packages` or `.dart_tool/package_config.json`.

test/add/common/version_constraint_test.dart
```diff
@@ -0,0 +1,108 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+import '../../descriptor.dart' as d;
+import '../../test_pub.dart';
+
+void main() {
+  test('allows empty version constraint', () async {
+    await servePackages((builder) => builder.serve('foo', '1.2.3'));
```
jonasfj, line 12: Maybe add a few more versions to the server here..

It's worth testing a few cases where:
 * There is older versions
 * There is multiple major versions
 * There is newer prereleases (major version)
 * There is newer prereleases (minor version)

Just cover a little bit more ground for the cases where the default logic picks a good version and version constraint.

test/add/common/invalid_options.dart
```diff
@@ -0,0 +1,82 @@
...
+          '--host-url',
+          'http://localhost:${server.port}',
+          '--git-url',
+          '../foo.git'
+        ],
+        error:
+            contains('Packages must either be a git, hosted, or path package.'),
+        exitCode: exit_codes.USAGE);
+  });
+}
```
jonasfj, line 82: What about testing `pub add foo:<constraint>` in combination with flags like `--git`, `--path` or `--sdk`.
I don't think it makes sense to give a version constraint to `pub add` when the source is not a pub server.

test/add/common/version_constraint_test.dart
```diff
@@ -0,0 +1,108 @@
...
+        // The "packages" directory should not have been generated.
+        d.nothing('packages'),
+        // The ".packages" file should not have been created.
+        d.nothing('.packages'),
+      ]).validate();
+    });
+
+    test('transitive', () async {
+      await servePackages((builder) {
+        builder.serve('foo', '1.2.3', deps: {'bar': '2.0.4'});
```
jonasfj, line 84: Can we make a variant of this, where it picks an older version of `foo` because the latest version is in conflict... so it might pick `0.1.0` of `foo` instead of `1.2.3`.

test/add/sdk/sdk_test.dart
jonasfj: My bad this is for some reason a supported thing:
https://dart.dev/tools/pub/dependencies#sdk

test/add/common/invalid_options.dart
jonasfj: Nvm, looks like it's a thing... Maybe we should have tests to ensure it works... But that's a different topic.

test/add/git/git_test.dart
jonasfj: Nvm, version constraints are support even when source only has one version...

test/add/sdk/sdk_test.dart
jonasfj: This is a thing.

test/descriptor/pubspec.dart
walnutdust: Yeah you're right! I'll change it to a `YamlDescriptor`!

test/descriptor/pubspec.dart
walnutdust: changed it to use `package:yaml`!

lib/src/command/add.dart
```diff
@@ -0,0 +1,339 @@
...
+    if (isDev) {
+      final dependencyNames = dependencies.map((dependency) => dependency.name);
+
+      /// If package is originally in dependencies and we wish to add it to
+      /// dev_dependencies, this is a redundant change, and we should not
+      /// remove the package from dependencies, since it might cause the user's
+      /// code to break.
+      if (dependencyNames.contains(package.name)) {
+        usageException('${package.name} is already in dependencies. '
+            'Please remove existing entry before adding it to dev_dependencies');
```
jonasfj, line 147: Something like  `Use 'pub remove ${package.name} to remove it before adding it to 'dev_dependencies'`

lib/src/command/add.dart
```diff
@@ -0,0 +1,339 @@
...
+    } else {
+      final devDependencyNames =
+          devDependencies.map((devDependency) => devDependency.name);
+
+      /// If package is originally in dev_dependencies and we wish to add it to
+      /// dependencies, we remove the package from dev_dependencies, since it is
+      /// now redundant.
+      if (devDependencyNames.contains(package.name)) {
+        log.message('${package.name} was found in dev_dependencies. '
+            'Removing ${package.name} and adding it to dependencies instead.');
```
jonasfj, line 160:
```suggestion
        log.message('"${package.name}" was found in `dev_dependencies`. '
            'Removing "${package.name}" and adding it to `dependencies` instead.');
```

lib/src/command/add.dart
```diff
@@ -0,0 +1,339 @@
...
+
+      /// Create a new [Entrypoint] since we have to reprocess the updated
+      /// pubspec file.
+      await Entrypoint.current(cache).acquireDependencies(SolveType.GET,
+          precompile: argResults['precompile']);
+    }
+
+    if (isOffline) {
+      log.warning('Warning: Packages added when offline may not resolve to '
+          'the latest versions of your dependencies.');
```
jonasfj, line 124:
```suggestion
          'the latest compatible version available.');
```

lib/src/command/add.dart
```diff
@@ -0,0 +1,339 @@
...
+        SolveType.GET, cache, Package.inMemory(updatedPubSpec));
+
+    final resultPackage = solveResult.packages
+        .firstWhere((packageId) => packageId.name == package.name);
+
+    /// Assert that [resultPackage] is within the original user's expectations.
+    if (package.constraint != null &&
+        !package.constraint.allows(resultPackage.version)) {
+      dataError('${package.name} resolved to ${resultPackage.version} which '
+          'does not match the input ${package.constraint}! Exiting.');
```
jonasfj, line 96: We can do better... this is probably due to `dependency_overrides` can we check if `dependency_overrides` is non-empty, and then print this as a possible explanation.

Certainly we can explain better what happened.

lib/src/command/add.dart
```diff
@@ -0,0 +1,339 @@
...
+      await Entrypoint.global(newRoot, entrypoint.lockFile, cache)
+          .acquireDependencies(SolveType.GET,
+              dryRun: true, precompile: argResults['precompile']);
+    } else {
+      /// Update the pubspec. It is necessary to do this before calling
+      /// [acquireDependencies] (as opposed to calling [acquireDependencies] on
+      /// the in-memory pubspec and writing it later) because there will be
+      /// conflicts in other parts of [pub] if the pubspec is newer than
+      /// `pubspec.lock`.
+      _updatePubspec(resultPackage, packageInformation, isDev);
```
jonasfj, line 114: Will there be conflicts?

Even if not, we could a resolution in memory, and write `pubspec.yaml` and then touch timestamp on the other files.

Or we could resolve then write all 3 files in the right order.

lib/src/command/add.dart
```diff
@@ -0,0 +1,339 @@
...
+  Pair<PackageRange, dynamic> _parsePackage(String package) {
+    ArgumentError.checkNotNull(package, 'package');
+
+    if ([path != null, hasGitOptions, hasHostOptions, sdk != null]
+            .where((element) => element == true)
+            .length >=
+        2) {
+      usageException(
+          'Packages must either be a git, hosted, sdk, or path package.');
+    }
```
jonasfj, line 213: nit, it might be nice to print which flags are conflicting.

```dart
final _conflictingFlagSets = [
  ['git-url', 'git-ref', 'git-path'],
  ['hosted-url'],
  ['path'],
  ['sdk'],
];

for (final flag in _conflictingFlagSets.expand((s) => s).where(argResult.wasParsed)) {
  final conflictingFlag = _conflictingFlagSets
    .where((s) => !s.contains(flag))
    .expand((s) => s)
    .firstWhere(argResult.wasParsed, orElse: () => null);
  if (conflictingFlag != null) {
    usageException('Packages can only have one source, pub add flags "--$flag" and "--$conflictingFlag" are conflicting.');
  }
}
```

But I'm sure it can be done smarter than this too :D

lib/src/command/add.dart
```diff
@@ -0,0 +1,339 @@
...
+
+    /// There shouldn't be more than one `:` in the package information
+    if (splitPackage.length > 2) {
+      throw FormatException('Invalid package and version constraint: $package');
+    }
+
+    /// We want to allow for [constraint] to take on a `null` value here to
+    /// preserve the fact that the user did not specify a constraint.
+    final constraint = splitPackage.length == 2
+        ? VersionConstraint.parse(splitPackage[1])
```
jonasfj, line 235: Who catches it if this throws a `FormatException`, I bet we're going to need a test case for this too.

lib/src/command/add.dart
```diff
@@ -0,0 +1,339 @@
...
+      /// added to the pubspec.
+      if (gitRef == null && gitPath == null) {
+        git = gitUrl;
+      } else {
+        git = {'url': gitUrl, 'ref': gitRef, 'path': gitPath};
+        git.removeWhere((key, value) => value == null);
+      }
+
+      packageRange = cache.sources['git']
+          .parseRef(packageName, git)
```
jonasfj, line 253: Can these `parseRef` functions throw exceptions? If so who is going to catch them...

What happens if we have a `pub add foo --git-ref` but no `--git-url` or `--git-url ""` or `--git-url !` or some other invalid value?

lib/src/command/add.dart
```diff
@@ -0,0 +1,339 @@
...
+          .withConstraint(VersionConstraint.any);
```
jonasfj, line 254: Having a constraint here is actually allowed, I think, we should probably add tests for this in `pub get` if we don't already.
Same in path dependencies.. it's a bit pointless, but nonetheless allowed.

test/add/sdk/sdk_test.dart
walnutdust: haha whoops

natebosch:
> Example: "pub add retry" might give you "retry: ^3.0.1", if version 3.0.1 of retry is the latest version compatible with your other dependencies.

I don't know if we have this in a doc somewhere but the tribal knowledge has always been that it's better to leave the patch level at `0`, so if the latest `retry` is `3.0.1` the constraint we'd pick is `^3.0.0`. I'm not sure I can actually say what all the tradeoffs are, but I think it's something we should consider.

In my personal project for making it easier to update deps in an editor I chose to always use a `0` patch version. https://github.com/natebosch/dartdeps/blob/993b2075b46cb5a28135d19881e4da591a5c15b8/lib/src/locate_latest.dart#L21-L23

jonasfj:
@natebosch, hmm, I can see the argument that if you start using package `retry` version `3.2.1` then a constraint of `^3.2.0` is slightly better as:
 (A) It's unlikely new features were added between `3.2.0` and `3.2.1`, because then it shouldn't have been a patch release.
 (B) allowing your package to work with `3.2.0` allows for better compatibility with other packages.

In practice it would have little impact, as the solver will always prefer the latest possible version when possible.

The downside is that we are never testing against `3.2.0`, just stating that we assume it's compatible with this version.

Hmm, actually, I see two cases where this makes a difference.
(1) a downstream user of my package has another dependency on `retry` that pins the version to `3.2.0`.
(2) a downstream user is adding a dependency on my package, and already has `retry` in the `pubspec.lock`, said downstream user is now going to get version `3.2.0` of `retry`.

IMO, (1) is okay, because the alternative is that the downstream user would have gotten a conflict. So given that we think it's likely to work with `3.2.0`, it's probably fine to try it out. Even if we didn't test with `3.2.0`.

(2) is more unfortunate, because here an upgrade to `3.2.1` would have happened, had we had a dependency on `^3.2.1`. on the other hand unnecessarily triggering an upgrade could cause problems in other parts of the code, which is probably why the user did a `pub get` and not a `pub upgrade`.

-------
Hence, I'm leaning towards `^3.2.0` being a slightly more useful constraint, because what we're trying to do is add the widest possible constraint that we think is likely to be compatible.

lib/src/command/add.dart
walnutdust: ah maybe conflicts is not the right word... was thinking about causing `pub run` or other similar commands to have to resolve `pubspec.lock` again. Either way, because we run `acquireDependencies` after updating the `pubspec.yaml`, the timestamps on the other files should resolve correctly. :D

The purpose of the comment was more to explain why we don't do something like `acquireDependencies` before writing to the pubspec file!

lib/src/command/add.dart
walnutdust: haha I'm going to steal this :D

test/add/common/add_test.dart
```diff
@@ -0,0 +1,598 @@
...
+
+    group('overrides existing version constraint if package exists', () {
+      test('if package is added without a version constraint', () async {
+        await servePackages((builder) {
+          builder.serve('foo', '1.2.3');
+          builder.serve('foo', '1.2.2');
+        });
+
+        await d.appDir({'foo': '1.2.2'}).create();
+
```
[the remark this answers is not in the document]
walnutdust: oh, what will the goal of the test be? If it is to test whether `pub add` performs a no-op in that scenario, I could do something like:

```dart
    test('no-op if the package already exists in the pubspec', () async {
      await servePackages((builder) => builder.serve('foo', '1.2.3'));

      /// Intentionally give it a version that does not exist in the
      /// server to check that the test does not try to resolve the version,
      /// or a different error will be thrown.
      await d.appDir({'foo': '1.2.2'}).create();

      await pubAdd(
          args: ['foo:1.2.3'],
          error: contains('"foo" is already in "dependencies"!'),
          exitCode: exit_codes.USAGE);

      await d.dir(appPath, [
        d.nothing('.dart_tool/package_config.json'),
        d.nothing('pubspec.lock'),
        d.nothing('.packages'),
      ]).validate();
    });
```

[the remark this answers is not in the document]
walnutdust: actually, I think in our previous discussions we mentioned that we wanted the `pub add` to override existing version constraints, so ignore the above :O

lib/src/command/add.dart
walnutdust: hm the way we are setting things up, I don't think `parseRef` will throw an exception, because it checks mainly to see if the `git` variable is a String or `Map<String,String>` with the relevant keys, which will always be the case I think...?

Invalid values for `--git-url` are handled in the `acquireDependencies` function, the same way `pub get` does it at the moment.

I think what I'll do is to surround the `resolveVersions` call on the in-memory pubspec with a try catch, and on `GitException` I could print out a message asking the user to check the git arguments, and on `SolveFailure` (for path/hosted/sdk) I could print out a message asking the user to generally check the arguments - does this sound good?

test/add/common/add_test.dart
```diff
@@ -0,0 +1,598 @@
...
+        await pubAdd(args: ['foo']);
+
+        await d.cacheDir({'foo': '1.2.3'}).validate();
+        await d.appPackagesFile({'foo': '1.2.3'}).validate();
+        await d.appDir({'foo': '^1.2.3'}).validate();
```
jonasfj, line 89: Let's not overwrite in this case... let's instead print a message to go use  `pub upgrade --bump foo`.

jonasfj: there is a case where... we have a dependency on `foo: ^1.2.3` 
and the current version is `1.2.3` and the latest available version is `1.2.4`, in this case `pub add foo` should print that we already have a dependency on `foo`, this can be upgraded with `pub upgrade foo`.

lib/src/command/add.dart
```diff
@@ -0,0 +1,339 @@
...
+    /// take on either a string for simple version constraints or a map for
+    /// more complicated hosted/git options.
+    dynamic pubspecInformation;
+
+    final splitPackage = package.split(':');
+    final packageName = splitPackage[0];
+
+    /// There shouldn't be more than one `:` in the package information
+    if (splitPackage.length > 2) {
+      throw FormatException('Invalid package and version constraint: $package');
```
jonasfj, line 229: This is a usage exception :D

test/add/common/add_test.dart
walnutdust: I made a temporary message to use `pub upgrade foo`, with a TODO to revise it later!

lib/src/command/add.dart
```diff
@@ -170,42 +237,73 @@ class AddCommand extends PubCommand {
...
-        : null;
+    VersionConstraint constraint;
+
+    try {
+      constraint = splitPackage.length == 2
+          ? VersionConstraint.parse(splitPackage[1])
+          : null;
+    } on FormatException catch (e) {
+      usageException('${e.message} Ensure that your version constraint '
+          'satisfies pub semver requirements.');
```
jonasfj, line 295: Can't we just say something like: `Invalid version constraint: ${e.message}`

Do we have a test case for this? I don't see any containing this message.

lib/src/command/add.dart
```diff
@@ -170,42 +237,73 @@ class AddCommand extends PubCommand {
...
+    }
 
+    /// Determine the relevant [packageRange] and [pubspecInformation] depending
+    /// on the type of package.
     if (hasGitOptions) {
       dynamic git;
 
+      if (gitUrl == null) {
+        usageException('Git packages must have the --git-url option declared!');
```
jonasfj, line 304:
```suggestion
        usageException('The `--git-url` is required for git dependencies.');
```

lib/src/command/add.dart
```diff
@@ -217,46 +315,48 @@ class AddCommand extends PubCommand {
...
-          'version': constraint.toString()
-        };
+        pubspecInformation = {'hosted': hostInfo};
       }
 
       packageRange = PackageRange(packageName, cache.sources['hosted'],
           constraint ?? VersionConstraint.any, hostInfo ?? packageName);
     }
 
+    if (pubspecInformation is Map && constraint != null) {
```
jonasfj, line 344: Not sure if this is necessary?
```suggestion
    if (pubspecInformation is Map && constraint != null && constraint != VersionConstraint.any) {
```

lib/src/command/add.dart
```diff
@@ -217,46 +315,48 @@ class AddCommand extends PubCommand {
...
+    }
+
     return Pair(packageRange, pubspecInformation);
   }
 
   /// Writes the changes to the pubspec file.
-  void _updatePubspec(
-      PackageId resultPackage, Pair<PackageRange, dynamic> packageInformation) {
+  void _updatePubspec(PackageId resultPackage,
+      Pair<PackageRange, dynamic> packageInformation, bool isDevelopment) {
```
jonasfj, line 359: Do we need `isDevelopment` from what I can see you are using `isDev` anyways :D'

I don't really care either way, if you pass it or use `isDev`. Just as long as we're being consistent.

lib/src/command/add.dart
```diff
@@ -170,42 +237,73 @@ class AddCommand extends PubCommand {
...
+
+    for (final flag
+        in _conflictingFlagSets.expand((s) => s).where(argResults.wasParsed)) {
+      final conflictingFlag = _conflictingFlagSets
+          .where((s) => !s.contains(flag))
+          .expand((s) => s)
+          .firstWhere(argResults.wasParsed, orElse: () => null);
+      if (conflictingFlag != null) {
+        usageException(
+            'Packages can only have one source, pub add flags "--$flag" and '
```
jonasfj, line 263:
```suggestion
            'Packages can only have one source, `pub add` flags "--$flag" and '
```

In general, I think we tend to use ` around CLI commands and flags in messages. We should perhaps look at what other parts of pub is doing, just to be sure we are consistent.

lib/src/command/add.dart
```diff
@@ -75,63 +84,121 @@ class AddCommand extends PubCommand {
...
 
       devDependencies.add(package);
     } else {
-      final devDependencyNames =
-          devDependencies.map((devDependency) => devDependency.name);
+      /// TODO(walnut): Change the error message once pub upgrade --bump is
+      /// released
+      if (dependencyNames.contains(package.name)) {
+        usageException('"${package.name}" is already in "dependencies". '
+            'Please use "pub upgrade ${package.name}" if you wish to upgrade '
```
jonasfj, line 192:
```suggestion
            'Use "pub upgrade ${package.name}" to upgrade '
```

I'm not a messaging expert, but short and concise is usually better.

Similar in other places.. If we can skip `Please` and `if you wish` that might be good.
We can say `Use ... to ...`

lib/src/command/add.dart
jonasfj: Okay, can we update the comment to say: 
> Update the `pubspec.yaml` before calling acquireDependencies to ensure that
> the modification timestamp on `pubspec.lock` and `.dart_tool/package_config.json` is
> newer than the `pubspec.yaml`, ensuring that [entrypoint.assertUptoDate] will pass.

lib/src/command/add.dart
```diff
@@ -0,0 +1,405 @@
...
+      /// in case [package] was already a transitive dependency.
+      solveResult = await resolveVersions(
+        SolveType.UPGRADE,
+        cache,
+        Package.inMemory(updatedPubSpec),
+      );
+    } on GitException {
+      dataError('Unable to resolve package "${package.name}" with the given '
+          'git parameters.');
+    } catch (e) {
```
jonasfj, line 101: Don't catch all, when we can catch specific errors.. that way if suddenly it starts throwing something new we have bug which causes a stracktrace. Rather than having a silent bug that we are never going to discover.

lib/src/command/add.dart
```diff
@@ -0,0 +1,405 @@
...
+    SolveResult solveResult;
+
+    try {
+      /// Use [SolveType.UPGRADE] to solve for the highest version of [package]
+      /// in case [package] was already a transitive dependency.
+      solveResult = await resolveVersions(
+        SolveType.UPGRADE,
+        cache,
+        Package.inMemory(updatedPubSpec),
+      );
```
jonasfj, line 97: Shouldn't we be specifying:
```dart
        lockFile: lockFile,
        useLatest: [package.name],
```

Such that we don't unlock anything, other than the package being added.

So if I have a dependency on `foo` then adding `bar` won't upgrade `foo`, unless upgrading `foo` is required to get some version of `bar`.

jonasfj (commented):
This is mostly nits, I think we have to workout the change around `useLatest`

lib/src/command/add.dart
walnutdust: Modified the message!

There was a test `test/common/add_test.dart` at lines 263 - 284 but it tested for `${e.message}` rather than the full error message. Fixed it!

test/add/common/version_resolution_test.dart
```diff
@@ -0,0 +1,103 @@
...
+    await servePackages((builder) {
+      builder.serve('foo', '3.5.0');
+      builder.serve('foo', '3.2.1');
+      builder.serve('foo', '2.5.0');
+      builder.serve('foo', '3.1.0');
+      builder.serve('bar', '1.0.0', deps: {'foo': '^3.2.1'});
+    });
+
+    // Recreating the pubspec because it only lasts for one pub command.
+    await d.appDir({'bar': '1.0.0'}).create();
```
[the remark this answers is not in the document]
walnutdust: strange, I can't seem to trace back where I came to this conclusion. Removed it and the tests are still passing!

test/add/common/version_resolution_test.dart
```diff
@@ -0,0 +1,103 @@
...
+      builder.serve('bar', '1.0.0', deps: {'foo': '^3.2.1'});
+    });
+
+    await d.appDir({'bar': '1.0.0'}).create();
+    await pubGet();
+
+    /// foo's package creator releases a newer version of foo, and we
+    /// want to test that this is what the user gets when they run
+    /// pub add foo.
+    await servePackages((builder) {
```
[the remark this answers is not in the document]
walnutdust: ah you're right! was creating a new server without realizing that this `add` function existed - fixed!

lib/src/command/add.dart
walnutdust: Removed the `Please` and `if you wish` from everywhere I could find!

lib/src/command/add.dart
walnutdust: Got it!

lib/src/command/add.dart
walnutdust: hm I'm open to adding the condition, but because of the way we parsed it above:
```dart
constraint = splitPackage.length == 2
    ? VersionConstraint.parse(splitPackage[1])
     : null;
```
`constraint` will only take on `VersionConstraint.any` if it is explicitly passed in, so in that case it becomes a question of whether we want to preserve it if explicitly passed in. The one main place I think where it makes a difference is when adding packages hosted on a different pub server (e.g. `pub add foo:any --hosted-url https://pub.foo`), and whether we want it to reflect:

```
foo: 
  version: any
  hosted: 
    name: foo
    url: https://pub.foo
```

or 

```
foo: 
  version: ^1.2.3
  hosted: 
    name: foo
    url: https://pub.foo
```

It currently returns the former!

lib/src/command/add.dart
walnutdust: ah I'm passing it as a response to Sigurd's suggestion at https://github.com/dart-lang/pub/pull/2618#discussion_r469886569. Missed a line in the function where I accidentally used `isDev` instead, but it should be fixed now!

lib/src/command/add.dart
walnutdust: I took a look at the other parts of pub, and it seems to be a tie between double-quotes and backticks for CLI commands, and no quotes for flags, unless it seems to be part of a larger command. 

**CLI commands**

Double quotes (non-exhaustive):
1. https://github.com/dart-lang/pub/blob/master/lib/src/entrypoint.dart#L530
1. https://github.com/dart-lang/pub/blob/master/lib/src/command/deps.dart#L252
1. https://github.com/dart-lang/pub/blob/master/lib/src/command/downgrade.dart#L18

Backticks (non-exhaustive):
1. https://github.com/dart-lang/pub/blob/master/lib/src/command/outdated.dart#L535-L548
1. https://github.com/dart-lang/pub/blob/master/lib/src/sdk/flutter.dart#L27
1. https://github.com/dart-lang/pub/blob/master/lib/src/solver/report.dart#L155

**Flags**

No quotes:
1. https://github.com/dart-lang/pub/blob/master/lib/src/global_packages.dart#L632
1. https://github.com/dart-lang/pub/blob/master/lib/src/command/downgrade.dart#L43
1. https://github.com/dart-lang/pub/blob/master/lib/src/command/get.dart#L43
1. https://github.com/dart-lang/pub/blob/master/lib/src/command/global_activate.dart#L57

Backtick:
1. https://github.com/dart-lang/pub/blob/master/lib/src/validator/null_safety_mixed_mode.dart#L48

I could create a separate PR to standardize all the CLI commands and flags, just let me know!

test/add/common/version_resolution_test.dart
```diff
@@ -0,0 +1,103 @@
...
+      builder.serve('foo', '3.2.1');
+      builder.serve('foo', '2.0.0');
+      builder.serve('bar', '1.5.0', deps: {'foo': '^4.0.0'});
+      builder.serve('bar', '1.0.0', deps: {'foo': '^3.2.1'});
+    });
+
+    // Recreating the pubspec because it only lasts for one pub command.
+    await d.appDir({'bar': '^1.0.0'}).create();
+
+    await pubAdd(args: ['foo']);
```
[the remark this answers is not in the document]
walnutdust: just ran the following test which passed:

```dart
    /// The server used to only have the foo v3.2.1 as the latest,
    /// so pub get will create a pubspec.lock to foo 3.2.1
    await servePackages((builder) {
      builder.serve('foo', '3.2.1');
      builder.serve('bar', '1.0.0', deps: {'foo': '^3.2.1'});
    });

    await d.appDir({'bar': '^1.0.0'}).create();
    await pubGet();

    globalPackageServer.add((builder) {
      builder.serve('foo', '4.0.0');
      builder.serve('foo', '2.0.0');
      builder.serve('bar', '1.5.0', deps: {'foo': '^4.0.0'});
    });

    await d.appDir({'bar': '^1.0.0', 'foo': 'any'}).create();
    await d.cacheDir({'bar': '1.0.0', 'foo': '3.2.1'}).validate();
    await d.appPackagesFile({'bar': '1.0.0', 'foo': '3.2.1'}).validate();

    await pubUpgrade(args: ['foo']);

    await d.appDir({'foo': 'any', 'bar': '^1.0.0'}).validate();
    await d.cacheDir({'foo': '4.0.0', 'bar': '1.5.0'}).validate();
    await d.appPackagesFile({'foo': '4.0.0', 'bar': '1.5.0'}).validate();
  });
```

It seems like `bar` does get upgraded :O

lib/src/command/add.dart
walnutdust: To my understanding, `useLatest` forces the version of `foo` to be the latest, regardless of other constraints! Consider the following tests (tried with both `SolveType.GET` and `SolveType.UPGRADE`, and the lockfile as mentioned):

1. Dependency override.
```dart
        await servePackages((builder) {
          builder.serve('foo', '1.2.3');
          builder.serve('foo', '1.2.2');
        });

        await d.dir(appPath, [
          d.pubspec({
            'name': 'myapp',
            'dependencies': {},
            'dependency_overrides': {'foo': '1.2.2'}
          })
        ]).create();

        await pubAdd(args: ['foo']);

        await d.cacheDir({'foo': '1.2.2'}).validate();
        await d.appPackagesFile({'foo': '1.2.2'}).validate();
        await d.dir(appPath, [
          d.pubspec({
            'name': 'myapp',
            'dependencies': {'foo': '^1.2.2'},
            'dependency_overrides': {'foo': '1.2.2'}
          })
        ]).validate();
```

which currently passes, but with the addition of `useLatest`, fails with the message `Because myapp depends on foo <1.2.3-∞ or >1.2.3 but the latest version (1.2.3) is required, version solving failed.`, because it forces `foo` to be `1.2.3` even when `1.2.2` exists.

2. 

```dart
  test(
      'empty constraint allows it to choose the latest version not in conflict',
      () async {
    await servePackages((builder) {
      builder.serve('foo', '0.1.0');
      builder.serve('foo', '1.2.3', deps: {'bar': '2.0.4'});
      builder.serve('bar', '2.0.3');
      builder.serve('bar', '2.0.4');
    });

    await d.appDir({'bar': '2.0.3'}).create();

    await pubAdd(args: ['foo']);

    await d.appDir({'foo': '^0.1.0', 'bar': '2.0.3'}).validate();

    await d.cacheDir({'foo': '0.1.0', 'bar': '2.0.3'}).validate();
    await d.appPackagesFile({'foo': '0.1.0', 'bar': '2.0.3'}).validate();
  });
```

which currently passes, but with `useLatest` defined, it forces the usage of `foo:1.2.3`, which conflicts with the pubspec and results in an error:

```bash
[e] SLVR:   fact: foo >=1.2.3 depends on bar 2.0.4
[e] SLVR:   conflict: foo >=1.2.3 depends on bar 2.0.4
[e] SLVR:   ! foo >=1.2.3 is partially satisfied by not foo <1.2.3-∞ or >1.2.3
[e] SLVR:   ! which is caused by "the latest version of foo (1.2.3) is required"
[e] SLVR:   ! thus: every version of foo requires bar 2.0.4
[e] SLVR:   ! not bar 2.0.4 is satisfied by bar 2.0.3
[e] SLVR:   ! which is caused by "myapp depends on bar 2.0.3"
[e] SLVR:   ! thus: foo is forbidden
[e] SLVR:   ! foo any is satisfied by foo any
[e] SLVR:   ! which is caused by "myapp depends on foo any"
[e] SLVR:   ! thus: version solving failed
[e] SLVR: Version solving took 0:00:00.271587 seconds.
[e]     | Tried 1 solutions.
[e] ERR : Because foo >=1.2.3 depends on bar 2.0.4 and the latest version of foo (1.2.3) is required, every version of foo requires bar 2.0.4.
[e]     | So, because myapp depends on both bar 2.0.3 and foo any, version solving failed.
[e] FINE: Exception type: DataException
[e] FINE: package:pub/src/utils.dart 528:35           dataError
[e]     | package:pub/src/command/add.dart 101:7      AddCommand.run
[e]     | ===== asynchronous gap ===========================
[e]     | dart:async                                  Future.catchError
[e]     | package:pub/src/utils.dart 113:52           captureErrors.wrappedCallback
[e]     | package:stack_trace                         Chain.capture
[e]     | package:pub/src/utils.dart 126:11           captureErrors
[e]     | package:pub/src/command_runner.dart 184:13  PubCommandRunner.runCommand
```

As such, unless I'm misunderstanding something about `useLatest`, it doesn't seem like it's suitable parameter for us to set.

walnutdust: Interestingly, I just tried running this test:

```dart
  test('upgrades to the latest version not in conflict', () async {
    await servePackages((builder) {
      builder.serve('foo', '1.2.2');
      builder.serve('bar', '2.0.3');
      builder.serve('bar', '2.0.4');
    });

    await d.appDir({'foo': '^1.2.2', 'bar': '2.0.3'}).create();

    await pubGet();

    await d.appDir({'foo': '^1.2.2', 'bar': '2.0.3'}).validate();

    await d.cacheDir({'foo': '1.2.2', 'bar': '2.0.3'}).validate();
    await d.appPackagesFile({'foo': '1.2.2', 'bar': '2.0.3'}).validate();

    globalPackageServer.add((builder) {
      builder.serve('foo', '1.2.3');
      builder.serve('foo', '1.2.4', deps: {'bar': '2.0.4'});
    });

    await pubUpgrade(args: ['foo']);
  });
```

and it resulted in an error (i.e. `pub upgrade foo` did not succeed).

```
Process `pub upgrade` exited with exitCode 1. Output:
[e] FINE: Pub 0.1.2+3
    MSG : Resolving dependencies...
[e] SLVR: fact: myapp is 0.0.0
[e] SLVR: derived: myapp
[e] SLVR: fact: myapp depends on foo ^1.2.2
[e] SLVR: fact: myapp depends on bar 2.0.3
[e] SLVR:   selecting myapp
[e] SLVR:   derived: bar 2.0.3
[e] SLVR:   derived: foo ^1.2.2
[e] IO  : Get versions from http://localhost:34219/api/packages/foo.
[e] IO  : HTTP GET http://localhost:34219/api/packages/foo
[e]     | Accept: application/vnd.pub.v2+json
[e]     | X-Pub-OS: linux
[e]     | X-Pub-Command: upgrade
[e]     | X-Pub-Session-ID: 00BA9360-670E-4EF1-A632-CBF75873A3E3
[e]     | X-Pub-Environment: test-environment
[e]     | X-Pub-Reason: direct
[e]     | user-agent: Dart pub 0.1.2+3
[e] IO  : HTTP response 200 OK for GET http://localhost:34219/api/packages/foo
[e]     | took 0:00:00.053261
[e]     | transfer-encoding: chunked
[e]     | date: Fri, 21 Aug 2020 06:07:14 GMT
[e]     | x-frame-options: SAMEORIGIN
[e]     | content-type: text/plain; charset=utf-8
[e]     | x-xss-protection: 1; mode=block
[e]     | x-content-type-options: nosniff
[e]     | server: dart:io with Shelf
[e] SLVR:   fact: the latest version of foo (1.2.4) is required
[e] SLVR:   derived: not foo <1.2.4-∞ or >1.2.4
[e] IO  : Get versions from http://localhost:34219/api/packages/bar.
[e] IO  : HTTP GET http://localhost:34219/api/packages/bar
[e]     | Accept: application/vnd.pub.v2+json
[e]     | X-Pub-OS: linux
[e]     | X-Pub-Command: upgrade
[e]     | X-Pub-Session-ID: 00BA9360-670E-4EF1-A632-CBF75873A3E3
[e]     | X-Pub-Environment: test-environment
[e]     | X-Pub-Reason: direct
[e]     | user-agent: Dart pub 0.1.2+3
[e] SLVR:     selecting bar 2.0.3
[e] SLVR:     fact: foo >=1.2.4 depends on bar 2.0.4
[e] SLVR:     conflict: foo >=1.2.4 depends on bar 2.0.4
[e] SLVR:     ! foo >=1.2.4 is partially satisfied by not foo <1.2.4-∞ or >1.2.4
[e] SLVR:     ! which is caused by "the latest version of foo (1.2.4) is required"
[e] SLVR:     ! thus: every version of foo requires bar 2.0.4
[e] SLVR:     ! foo any is satisfied by foo ^1.2.2
[e] SLVR:     ! which is caused by "myapp depends on foo ^1.2.2"
[e] SLVR:     ! thus: bar 2.0.4 is required
[e] SLVR:     ! not bar 2.0.4 is satisfied by bar 2.0.3
[e] SLVR:     ! which is caused by "myapp depends on bar 2.0.3"
[e] SLVR:     ! thus: version solving failed
[e] SLVR: Version solving took 0:00:00.147784 seconds.
[e]     | Tried 1 solutions.
[e] FINE: Resolving dependencies finished (0.167s).
[e] ERR : Because foo >=1.2.4 depends on bar 2.0.4 and the latest version of foo (1.2.4) is required, every version of foo requires bar 2.0.4.
[e]     | So, because myapp depends on both foo ^1.2.2 and bar 2.0.3, version solving failed.
[e] FINE: Exception type: SolveFailure
[e] FINE: package:pub/src/solver/version_solver.dart 312:5   VersionSolver._resolveConflict
[e]     | package:pub/src/solver/version_solver.dart 133:27  VersionSolver._propagate
[e]     | package:pub/src/solver/version_solver.dart 97:11   VersionSolver.solve.<fn>
[e]     | ===== asynchronous gap ===========================
[e]     | dart:async                                         Future.catchError
[e]     | package:pub/src/utils.dart 113:52                  captureErrors.wrappedCallback
[e]     | package:stack_trace                                Chain.capture
[e]     | package:pub/src/utils.dart 126:11                  captureErrors
[e]     | package:pub/src/command_runner.dart 184:13         PubCommandRunner.runCommand
```

whereas the following succeeded:

```dart
  test('upgrades to the latest version not in conflict', () async {
    await servePackages((builder) {
      builder.serve('foo', '1.2.2');
      builder.serve('bar', '2.0.3');
      builder.serve('bar', '2.0.4');
    });

    await d.appDir({'foo': '^1.2.2', 'bar': '2.0.3'}).create();

    await pubGet();

    await d.appDir({'foo': '^1.2.2', 'bar': '2.0.3'}).validate();

    await d.cacheDir({'foo': '1.2.2', 'bar': '2.0.3'}).validate();
    await d.appPackagesFile({'foo': '1.2.2', 'bar': '2.0.3'}).validate();

    globalPackageServer.add((builder) {
      builder.serve('foo', '1.2.3');
      builder.serve('foo', '1.2.4', deps: {'bar': '2.0.4'});
    });

    await pubUpgrade();

    await d.appDir({'foo': '^1.2.2', 'bar': '2.0.3'}).validate();

    await d.cacheDir({'foo': '1.2.3', 'bar': '2.0.3'}).validate();
    await d.appPackagesFile({'foo': '1.2.3', 'bar': '2.0.3'}).validate();
  });
```

is this the intended behavior? I would have expected that if the second test passed that the first test would pass too

jonasfj (approved):

PR #2619, opened 2020-08-13
pub upgrade --major-versions

walnutdust:
`pub upgrade --bump [<packages>...]`

Updates constraints in pubspec.yaml to allow the latest resolvable version of packages for `<packages>...` specified or all packages if no specific packages are given.

This is similar to removing the upper bound from all constraints, running pub upgrade and adjusting the constraints to have upper bounds again. The motivation for this is to allow for a quick and convenient method to upgrade all the package's dependencies to their latest resolvable version as given in `pub outdated`.

Currently implemented as a hidden flag, pending confirmation on the name.

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -19,8 +19,8 @@ import '../null_safety_analysis.dart';
 import '../package.dart';
 import '../package_name.dart';
 import '../pubspec.dart';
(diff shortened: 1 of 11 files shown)


jonasfj:
Merged as part of https://github.com/dart-lang/pub/pull/2784

PR #2620, opened 2020-08-13
Adding a `pub remove` command

walnutdust:
`pub remove <package> ...`

Unlike `pub add`, this command allows you to list multiple packages for removal.

### DIFF
--- a/test/remove/remove_test.dart
+++ b/test/remove/remove_test.dart
@@ -1,0 +1,165 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+import '../descriptor.dart' as d;
+import '../test_pub.dart';
+
+void main() {
+  test('removes a package from dependencies', () async {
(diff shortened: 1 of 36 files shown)


test/remove/remove_test.dart
```diff
@@ -169,13 +167,50 @@ void main() {
...
+          foo: 1.0.0 # comment C
+        # comment D
+    ''');
+    await d.dir(appPath, [finalPubspec]).validate();
+    final fullPath = p.join(d.sandbox, appPath, 'pubspec.yaml');
+
+    expect(File(fullPath).existsSync(), true);
+
+    final contents = File(fullPath).readAsStringSync();
+    expect(contents, await finalPubspec.read());
```
jonasfj, line 216: nit: can't we make this a bit more robust by simply doing:
`expect(contents, contains('# comment A'));` and similar for B, C and D,

That way we're not relying minor whitespace changes in `package:yaml_edit`, but we are checking that comments are preserved.

jonasfj (approved):

test/remove/remove_test.dart
walnutdust: hm okay. I initially put it this way because I thought something like:
```
      name: myapp
      dependencies: # comment # comment B
          foo: 1.0.0 # comment C
        # comment D
```

could realistically happen but is not what we want, but you have a point. fixing now

PR #2622, opened 2020-08-14
Top level command

sigurdm:
Provides top-level api with a single package:args command that runs pub functionality.

Also provides a function intended for `dartdev run` to take over the functionality of `pub run` (do package resolution, and potentially snapshot).

Should make it easier to embed pub into dartdev (and also eventually into the flutter tool).

See a draft of the embedding here: https://dart-review.googlesource.com/c/sdk/+/159480#


lib/src/executable.dart
```diff
@@ -30,6 +34,21 @@ List<String> vmArgsFromArgResults(ArgResults argResults) {
...
+    Future<void> Function(Executable) recompile,
+    List<String> vmArgs = const []}) async {
+  final r = await _runExecutable(entrypoint, executable, args,
+      enableAsserts: enableAsserts,
+      packagesFile: packagesFile,
+      recompile: recompile,
+      vmArgs: vmArgs);
+  print('Returned $r');
+  return r;
+}
```
jonasfj, line 50: I'm guessing this is a debug hack.. and that `_runExecutable` should replace this.

jonasfj (approved):

jonasfj (changes requested):
Okay, I need to fix:
 * Handling of exceptions like `dataError`,
 * Closing of HTTP client to that `pub get` doesn't hang waiting for connections to close.

jonasfj (approved):
Looks like I just needed to port `pub add/remove´ commands, now it works :D

PR #2623, opened 2020-08-15
Drop upper bound instead of using "any" while resolving in "pub outdated"

### LINKED ISSUE #2584: Resolvable from "pub outdated" should not use "any", but drop the upper-bound
Today we find resolvable in "pub outdated" by taking a constraint like:
`retry: ^1.2.0` changing it to `retry: any` and resolve.

I propose we take:
`retry: ^1.2.0` change it to `retry: '>=1.2.0'` and resolve.

So that we just remove the upper-bound, and the lower bound stays.

This way, resolvable will never be an older version than what you already have.

walnutdust:
Functions are also moved to a separate file in anticipation for sharing with #2619.

There is a modification to a `pub outdated` test golden file where for the prereleases variation in the test for pub outdated ([here](https://github.com/dart-lang/pub/blob/master/test/outdated/outdated_test.dart#L326:L354)), copied below for convenience:
```dart
test(
      'latest version reported while locked on a prerelease can be a prerelease',
      () async {
    await servePackages((builder) => builder
      ..serve('foo', '0.9.0')
      ..serve('foo', '1.0.0-dev.1')
      ..serve('foo', '1.0.0-dev.2')
      ..serve('bar', '0.9.0')
      ..serve('bar', '1.0.0-dev.1')
      ..serve('bar', '1.0.0-dev.2')
      ..serve('mop', '0.10.0-dev')
      ..serve('mop', '0.10.0')
      ..serve('mop', '1.0.0-dev'));
    await d.dir(appPath, [
      d.pubspec({
        'name': 'app',
        'version': '1.0.1',
        'dependencies': {
          'foo': '1.0.0-dev.1',
          'bar': '^0.9.0',
          'mop': '0.10.0-dev'
        },
      })
    ]).create();

    await pubGet();

    await variations('prereleases');
  });
```

I wanted to check why the corresponding output:

```
Dependencies  Current       Upgradable    Resolvable  Latest       
foo           *1.0.0-dev.1  *1.0.0-dev.1  *0.9.0      1.0.0-dev.2  
mop           *0.10.0-dev   *0.10.0-dev   0.10.0      0.10.0       
```

was correct for its resolvable version - shouldn't it be 1.0.0-dev.2, which is correctly determined in this PR? Let me know if the golden file was in fact correct, and I'll investigate why the behaviour happens.

Closes #2584

### DIFF
--- a/lib/src/pubspec_utils.dart
+++ b/lib/src/pubspec_utils.dart
@@ -1,0 +1,104 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:meta/meta.dart';
+import 'package:pub_semver/pub_semver.dart';
+
+import 'package_name.dart';
+import 'pubspec.dart';
+import 'source/hosted.dart';
+
+/// Returns a new [Pubspec] without [original]'s dev_dependencies.
+Pubspec stripDevDependencies(Pubspec original) {
+  ArgumentError.checkNotNull(original, 'original pubspec');
+
+  return Pubspec(
+    original.name,
+    version: original.version,
+    sdkConstraints: original.sdkConstraints,
+    dependencies: original.dependencies.values,
+    devDependencies: [], // explicitly give empty list, to prevent lazy parsing
+    dependencyOverrides: original.dependencyOverrides.values,
+  );
+}
+
+/// Returns a new [Pubspec] without [original]'s dependency_overrides.
+Pubspec stripDependencyOverrides(Pubspec original) {
+  ArgumentError.checkNotNull(original, 'original pubspec');
+
+  return Pubspec(
+    original.name,
+    version: original.version,
+    sdkConstraints: original.sdkConstraints,
+    dependencies: original.dependencies.values,
+    devDependencies: original.devDependencies.values,
+    dependencyOverrides: [],
+  );
+}
+
+/// Returns new pubspec with the same dependencies as [original] but with the
+/// upper bounds of the constraints removed.
+///
+/// If [upgradeOnly] is provided, only the packages whose names are in
+/// [upgradeOnly] will have their upper bounds removed.
+Pubspec stripVersionUpperBounds(Pubspec original, {List<String> upgradeOnly}) {
+  ArgumentError.checkNotNull(original, 'original pubspec');
+  upgradeOnly ??= [];
+
+  List<PackageRange> _stripUpperBounds(
+    Map<String, PackageRange> constrained,
+  ) {
+    final result = <PackageRange>[];
+
+    for (final name in constrained.keys) {
+      final packageRange = constrained[name];
+      var unconstrainedRange = packageRange;
+
+      /// We only need to remove the upper bound if it is a hosted package.
+      if (packageRange.source is HostedSource &&
+          (upgradeOnly.isEmpty || upgradeOnly.contains(packageRange.name))) {
+        unconstrainedRange = PackageRange(
+            packageRange.name,
+            packageRange.source,
+            stripUpperBound(packageRange.constraint),
+            packageRange.description,
+            features: packageRange.features);
+      }
+      result.add(unconstrainedRange);
+    }
+
+    return result;
+  }
+
+  return Pubspec(
+    original.name,
+    version: original.version,
+    sdkConstraints: original.sdkConstraints,
+    dependencies: _stripUpperBounds(original.dependencies),
+    devDependencies: _stripUpperBounds(original.devDependencies),
+    dependencyOverrides: original.dependencyOverrides.values,
+  );
+}
+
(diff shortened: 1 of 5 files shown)


lib/src/pubspec_utils.dart
```diff
@@ -0,0 +1,104 @@
...
+    dependencyOverrides: [],
+  );
+}
+
+/// Returns new pubspec with the same dependencies as [original] but with the
+/// upper bounds of the constraints removed.
+///
+/// If [upgradeOnly] is provided, only the packages whose names are in
+/// [upgradeOnly] will have their upper bounds removed.
+Pubspec stripVersionUpperBounds(Pubspec original, {List<String> upgradeOnly}) {
```
jonasfj, line 45:
```suggestion
Pubspec stripVersionUpperBounds(Pubspec original, {Iterable<String> stripOnly}) {
```

lib/src/pubspec_utils.dart
```diff
@@ -0,0 +1,104 @@
...
+    dependencies: _stripUpperBounds(original.dependencies),
+    devDependencies: _stripUpperBounds(original.devDependencies),
+    dependencyOverrides: original.dependencyOverrides.values,
+  );
+}
+
+/// Removes the upper bound of [constraint]. If [constraint] is the
+/// empty version constraint, an empty version range will be returned.
+@visibleForTesting
+VersionRange stripUpperBound(VersionConstraint constraint) {
```
jonasfj, line 87: Can't we do `VersionConstraint`? is there any places where it needs to be a `VersionRange`?
```suggestion
VersionConstraint stripUpperBound(VersionConstraint constraint) {
```

For testing we can just do `as VersionRange`.

lib/src/pubspec_utils.dart
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/// Returns new pubspec with the same dependencies as [original] but with the
+/// upper bounds of the constraints removed.
+///
+/// If [upgradeOnly] is provided, only the packages whose names are in
+/// [upgradeOnly] will have their upper bounds removed.
+Pubspec stripVersionUpperBounds(Pubspec original, {List<String> upgradeOnly}) {
+  ArgumentError.checkNotNull(original, 'original pubspec');
+  upgradeOnly ??= [];
```
jonasfj, line 47: This is not what docs say, they say that if specified only those are updated.. I think it's fine to fix the docs and say that if: not specified or empty, then all packages will be upgraded.

lib/src/pubspec_utils.dart
```diff
@@ -0,0 +1,104 @@
...
+  /// or the empty [VersionConstraint].
+  if (constraint is VersionRange) {
+    return VersionRange(min: constraint.min, includeMin: constraint.includeMin);
+  }
+
+  if (constraint is VersionUnion && constraint.ranges.isNotEmpty) {
+    final firstRange = constraint.ranges.first;
+    return VersionRange(min: firstRange.min, includeMin: firstRange.includeMin);
+  }
+
```
jonasfj, line 100: An `assert(constraint == VersionConstraint.empty, 'unknown constraint type')` might be fair.

test/outdated/goldens/prereleases.txt
```diff
@@ -35,9 +35,9 @@ $ pub outdated --json
 }
 
 $ pub outdated --no-color
-Dependencies  Current       Upgradable    Resolvable  Latest       
-foo           *1.0.0-dev.1  *1.0.0-dev.1  *0.9.0      1.0.0-dev.2  
-mop           *0.10.0-dev   *0.10.0-dev   0.10.0      0.10.0       
+Dependencies  Current       Upgradable    Resolvable   Latest       
+foo           *1.0.0-dev.1  *1.0.0-dev.1  1.0.0-dev.2  1.0.0-dev.2  
+mop           *0.10.0-dev   *0.10.0-dev   0.10.0       0.10.0
```
jonasfj, line 40: @sigurdm this is the bug we're fixing.

test/pubspec_utils_test.dart
```diff
@@ -0,0 +1,74 @@
...
+      expect(removedUpperBound.max, isNull);
+    });
+
+    test(
+        'returns the empty version constraint when an empty version constraint '
+        'is provided', () {
+      final constraint = VersionConstraint.empty;
+
+      expect(stripUpperBound(constraint),
+          VersionRange(min: Version.none, max: Version.none));
```
jonasfj, line 65: Isn't this a bit more reasonable?
```suggestion
      expect(stripUpperBound(constraint), VersionConstraint.empty);
```

jonasfj (approved):
I think this just nits

lib/src/pubspec_utils.dart
```diff
@@ -0,0 +1,104 @@
...
+  /// If it gets here, [constraint] is the empty version constraint, so we
+  /// just return an empty version range
+  return VersionRange(min: Version.none, max: Version.none);
```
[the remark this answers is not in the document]
walnutdust: yeap it should be fine, I got confused with `VersionConstraint.empty`, which does not extend `VersionRange`!

PR #2639, opened 2020-08-28
Pub add/remove now remove dependencies key if it makes them empty

### LINKED ISSUE #2635: pub add <package>, can leave dev_deps empty
``` console
$ dart pub add --dev retry
$ dart pub add retry
```
Leaves us with:
```yaml
dev_dependencies: 
  {}
```

Assuming we didn't have any `dev_dependencies` before. We probably should just remove the `dev_dependencies` or `dependencies` key, if we are deleting the last entry in the dictionary.. We risk deleting a few extra comments, but that seems rare..

We could also leave it as `dev_dependencies:`, but that seems unfortunate, and I doubt `yaml_edit` can express that.

walnutdust:
Closes #2635

### DIFF
--- a/lib/src/command/add.dart
+++ b/lib/src/command/add.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:pub_semver/pub_semver.dart';
+import 'package:yaml/yaml.dart';
 
 import '../command.dart';
 import '../entrypoint.dart';
@@ -391,11 +392,18 @@ class AddCommand extends PubCommand {
 
     /// Remove the package from dev_dependencies if we are adding it to
     /// dependencies. Refer to [_addPackageToPubspec] for additional discussion.
-    if (!isDevelopment &&
-        yamlEditor.parseAt(['dev_dependencies', package.name],
-                orElse: () => null) !=
-            null) {
-      yamlEditor.remove(['dev_dependencies', package.name]);
+    if (!isDevelopment) {
+      final devDependenciesNode =
+          yamlEditor.parseAt(['dev_dependencies'], orElse: () => null);
+
+      if (devDependenciesNode is YamlMap &&
+          devDependenciesNode.containsKey(package.name)) {
+        if (devDependenciesNode.length == 1) {
+          yamlEditor.remove(['dev_dependencies']);
+        } else {
+          yamlEditor.remove(['dev_dependencies', package.name]);
+        }
+      }
     }
 
     /// Windows line endings are already handled by [yamlEditor]
--- a/lib/src/command/remove.dart
+++ b/lib/src/command/remove.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:yaml/yaml.dart';
+
 import '../command.dart';
 import '../entrypoint.dart';
 import '../io.dart';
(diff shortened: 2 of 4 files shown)


lib/src/command/add.dart
```diff
@@ -391,11 +392,18 @@ class AddCommand extends PubCommand {
...
-    if (!isDevelopment &&
-        yamlEditor.parseAt(['dev_dependencies', package.name],
-                orElse: () => null) !=
-            null) {
-      yamlEditor.remove(['dev_dependencies', package.name]);
+    if (!isDevelopment) {
+      final devDependenciesNode =
+          yamlEditor.parseAt(['dev_dependencies'], orElse: () => null);
+
+      if (devDependenciesNode is YamlMap &&
```
jonasfj, line 399: I don't think you need to import `package:yaml/yaml.dart` and use `devDependenciesNode is YamlMap`, you can just do:
```suggestion
      if (devDependenciesNode is Map &&
```

I think? :)

lib/src/command/remove.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:yaml/yaml.dart';
```
jonasfj, line 5: Same here we might not need to import this...

test/add/common/add_test.dart
```diff
@@ -50,6 +50,44 @@ void main() {
...
+          dev_dependencies:
+        ''');
+
+      await d.dir(appPath, [finalPubspec]).validate();
+      final fullPath = p.join(d.sandbox, appPath, 'pubspec.yaml');
+
+      expect(File(fullPath).existsSync(), true);
+
+      final contents = File(fullPath).readAsStringSync();
+      expect(contents, await finalPubspec.read());
```
jonasfj, line 88: Can we do substring comparison for more robust tests, like `contains('dev_dependencies:')`

test/add/common/add_test.dart
```diff
@@ -165,13 +205,21 @@ void main() {
...
+
+''');
+
+      await d.dir(appPath, [finalPubspec]).validate();
+      final fullPath = p.join(d.sandbox, appPath, 'pubspec.yaml');
+
+      expect(File(fullPath).existsSync(), true);
+
+      final contents = File(fullPath).readAsStringSync();
+      expect(contents, await finalPubspec.read());
```
jonasfj, line 222: Same here can use `not(contains('dev_dependencies:'))`?

test/remove/remove_test.dart
```diff
@@ -22,7 +22,47 @@ void main() {
 
     await d.cacheDir({}).validate();
     await d.appPackagesFile({}).validate();
-    await d.appDir({}).validate();
+    await d.appDir().validate();
+  });
+
+  test('removing a package from dependencies does not affect dev dependencies',
```
jonasfj, line 28:
```suggestion
  test('removing a package from dependencies does not affect dev_dependencies',
```

test/remove/remove_test.dart
```diff
@@ -22,7 +22,47 @@ void main() {
...
+  bar: 2.0.0
+''');
+
+    await d.dir(appPath, [finalPubspec]).validate();
+    final fullPath = p.join(d.sandbox, appPath, 'pubspec.yaml');
+
+    expect(File(fullPath).existsSync(), true);
+
+    final contents = File(fullPath).readAsStringSync();
+    expect(contents, await finalPubspec.read());
```
jonasfj, line 65: Can we use `not(contains('\ndependencies:'))`? Do we even need this check? as `{dependencies: null}` isn't the same as `{}`.

lib/src/command/add.dart
walnutdust: ah the `YamlMap` is there so that I can both do the `containsKey` check and `length` check in the subsequent lines! If I swapped it to just `Map`, the Dart linter(?) doesn't seem to resolve that the `devDependenciesNode` has the `containsKey` method and `length` property automatically. Would substituting the current code with:

```dart
      if (devDependenciesNode is Map &&
          (devDependenciesNode as Map).containsKey(package.name)) {
        if ((devDependenciesNode as Map).length == 1) {
```

be preferred?

jonasfj: interesting... I think we YamlMap is fine :D

jonasfj (approved):
just a few nits in the tests

PR #2640, opened 2020-08-31
Fixed bug in adding dependency to empty dependencies key

### LINKED ISSUE #2637: pub add <package> with empty `dependencies` key
```
name: mypkg
dependencies:
environment:
  sdk: ^2.9.0-edge
dev_dependencies: {}
```
Then run:
```
[mypkg]$pub add retry
Error on line 3, column 27: Mapping values are not allowed here. Did you miss a colon earlier?
  ╷
3 │   retry: ^3.0.1environment:
  │                           ^
  ╵
```

We have two problems here.

 1. A bug in `yaml_edit`,
 2. A bug in how `pub add` catches `AssertionError`, I assume this an internal assertion in `yaml_edit` that fails.

I think it's best if `pub add` catches `AssertionError` and shows a message where it says to report it here.
We have other places where we handle internal errors and have a message.

Side note: We should probably make methods that modify `pubspec.yaml` log contents of `pubspec.yaml` as well as the change they are doing. We can log it such it only shows up with `--verbose`

walnutdust:
See https://github.com/google/dart-neats/pull/63 for associated tests

Closes #2637

### DIFF
--- a/lib/src/command/add.dart
+++ b/lib/src/command/add.dart
@@ -21,7 +21,7 @@ import '../yaml_edit/editor.dart';
 /// the package. The user may pass in a git constraint, host url, or path as
 /// requirements. If no such options are passed in, this command will do a
 /// resolution to find the latest version of the package that is compatible with
-/// the other dependencies in `pubpsec.yaml`, and then enter that as the lower
+/// the other dependencies in `pubspec.yaml`, and then enter that as the lower
 /// bound in a ^x.y.z constraint.
 ///
 /// Currently supports only adding one dependency at a time.
@@ -377,6 +377,8 @@ class AddCommand extends PubCommand {
     final packagePath = [dependencyKey, package.name];
 
     final yamlEditor = YamlEditor(readTextFile(entrypoint.pubspecPath));
+    log.io('Reading ${entrypoint.pubspecPath}');
+    log.fine('Contents:\n$yamlEditor');
 
     /// Handle situations where the user might not have the dependencies or
     /// dev_dependencies map.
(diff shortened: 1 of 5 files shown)


lib/src/command/add.dart
```diff
@@ -389,13 +391,17 @@ class AddCommand extends PubCommand {
           packagePath, pubspecInformation ?? '^${resultPackage.version}');
     }
 
+    log.fine('Added ^${resultPackage.version} to "$dependencyKey"');
```
jonasfj, line 394: This is not what happens if adding a git dependency... etc.. I'm fine with a dumber message that just says:
`Added "${package.name}" to $dependencyKey`.

lib/src/command/add.dart
```diff
@@ -389,13 +391,17 @@ class AddCommand extends PubCommand {
...
+
     /// Remove the package from dev_dependencies if we are adding it to
     /// dependencies. Refer to [_addPackageToPubspec] for additional discussion.
     if (!isDevelopment &&
         yamlEditor.parseAt(['dev_dependencies', package.name],
                 orElse: () => null) !=
             null) {
       yamlEditor.remove(['dev_dependencies', package.name]);
+
+      log.fine('Removed ^${resultPackage.version} from "dev_dependencies".');
```
jonasfj, line 404: Again, let's just say package name and `dev_dependencies`, and avoiding mentioning the constraint we removed.. because unless we want to make a lot of logic to log the right thing.

jonasfj (approved):

PR #2646, opened 2020-09-07
pub.dartlang.org is now pub.dev.

sidrao2006:
Updated in logout description

The pub tool still displays [pub.dartlang.org] which is now [pub.dev](https://pub.dev)

Fixed this problem in logout.dart and publish.dart

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -24,9 +24,9 @@ import 'utils.dart';
 /// Headers and field names that should be censored in the log output.
 const _censoredFields = ['refresh_token', 'authorization'];
 
-/// Headers required for pub.dartlang.org API requests.
+/// Headers required for pub.dev API requests.
 ///
-/// The Accept header tells pub.dartlang.org which version of the API we're
+/// The Accept header tells pub.dev which version of the API we're
 /// expecting, so it can either serve that version or give us a 406 error if
 /// it's not supported.
 const pubApiHeaders = {'Accept': 'application/vnd.pub.v2+json'};
@@ -77,7 +77,7 @@ class _PubHttpClient extends http.BaseClient {
         return false;
       }
     } else {
-      if (request.url.origin != 'https://pub.dartlang.org') return false;
+      if (request.url.origin != 'https://pub.dev') return false;
     }
 
     return true;
@@ -212,7 +212,7 @@ class _ThrowingClient extends http.BaseClient {
     }
 
     if (status == 500 &&
-        (request.url.host == 'pub.dartlang.org' ||
+        (request.url.host == 'pub.dev' ||
             request.url.host == 'storage.googleapis.com')) {
       fail('HTTP error 500: Internal Server Error at ${request.url}.\n'
           'This is likely a transient error. Please try again later.');
@@ -268,7 +268,7 @@ final httpClient = ThrottleClient(
 http.Client get innerHttpClient => _pubClient._inner;
 set innerHttpClient(http.Client client) => _pubClient._inner = client;
 
-/// Runs [callback] in a zone where all HTTP requests sent to `pub.dartlang.org`
+/// Runs [callback] in a zone where all HTTP requests sent to `pub.dev`
 /// will indicate the [type] of the relationship between the root package and
 /// the package being requested.
 ///
@@ -279,7 +279,7 @@ Future<T> withDependencyType<T>(
   return runZoned(callback, zoneValues: {#_dependencyType: type});
 }
 
-/// Handles a successful JSON-formatted response from pub.dartlang.org.
+/// Handles a successful JSON-formatted response from pub.dev.
 ///
 /// These responses are expected to be of the form `{"success": {"message":
 /// "some message"}}`. If the format is correct, the message will be printed;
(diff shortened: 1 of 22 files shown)


jonasfj (approved):

jonasfj:
Nice catch..

sidrao2006:
Thanks but there's one more bug which I am unable to find the code for. The help info of `--services` argument of `uploader` also shows pub.dartlang.org
Does anyone know where is that part of the 

.








.


.







.

.

sigurdm:
> Does anyone know where is that part of the code?

I think that is still the server we actually use for the uploading, to provide backwards compatibility - see lib/src/source/hosted.dart around line 56.

sidrao2006:
@sigurdm , since the team has mentioned in the comments that changing it to pub.dev may cause some problems with cache and raise suspicion.
It would be great if someone in the dart team could confirm because not changing it would cause even more confusion especially for beginners as there is no pub.dev indication.

kevmoo:
The changes here just reflect the output of user-visible commands. I think these are fine

...but tests need to be fixed!

sidrao2006:
Alright @kevmoo , but what about [these changes](https://github.com/dart-lang/pub/pull/2646#issuecomment-688286006)

kevmoo:
> Alright @kevmoo , but what about [these changes](https://github.com/dart-lang/pub/pull/2646#issuecomment-688286006)

Seems like those changes were removed in a follow-up commit. I don't see them...

stefandevo:
There are more references ... just a quick search and you'll notice.

https://github.com/dart-lang/pub/blob/master/lib/src/http.dart#L80

but also in the tests.

I am wondering why not use it as a const in the code base and change everything accordingly. It's just feel awkward to have all these strings ...

sidrao2006:
Yep, got it

sidrao2006:
But what about the tests?

lib/src/source/hosted.dart
```diff
@@ -42,18 +42,7 @@ class HostedSource extends Source {
...
-    //  2. It would cause conflicts for users checking pubspec.lock into git, if using
-    //     different versions of the dart-sdk / pub client.
-    //  3. It might cause other problems (investigation needed) for pubspec.lock across
-    //     different versions of the dart-sdk / pub client.
-    //  4. It would expand the API surface we're committed to supporting long-term.
-    //
-    // Clearly, a bit of investigation is necessary before we update this to
-    // pub.dev, it might be attractive to do next time we change the server API.
-    return _defaultUrl ??= _pubHostedUrlConfig() ?? 'https://pub.dartlang.org';
+    return _defaultUrl ??= _pubHostedUrlConfig() ?? 'https://pub.dev';
```
kevmoo, line 45: This CANNOT be changed! Folks have caches that assume the old URL. This one SHOULD NOT be updated.

kevmoo:
> But what about the tests?

The tests on Travis are failing. Take a look!

stefandevo:
Important that the team confirms that pub.dev is also valid for http requests. Otherwise its better to only change visual output to the user to pub.dev

sidrao2006:
Yep, still waiting for that

lib/src/http.dart
```diff
@@ -77,7 +77,7 @@ class _PubHttpClient extends http.BaseClient {
         return false;
       }
     } else {
-      if (request.url.origin != 'https://pub.dartlang.org') return false;
+      if (request.url.origin != 'https://pub.dev') return false;
```
stefandevo, line 80: Needs to be confirmed that it is OK to change this

lib/src/http.dart
```diff
@@ -212,7 +212,7 @@ class _ThrowingClient extends http.BaseClient {
     }
 
     if (status == 500 &&
-        (request.url.host == 'pub.dartlang.org' ||
+        (request.url.host == 'pub.dev' ||
```
stefandevo, line 215: Needs to be confirmed that it is OK to change this

lib/src/oauth2.dart
```diff
@@ -207,7 +207,7 @@ Future<Client> _authorize() async {
     completer
         .complete(grant.handleAuthorizationResponse(queryToMap(queryString)));
 
-    return shelf.Response.found('https://pub.dartlang.org/authorized');
+    return shelf.Response.found('https://pub.dev/authorized');
```
stefandevo, line 210: Needs to be confirmed that it is OK to change this

stefandevo:
@kevmoo I am thinking that it's best that front facing messages to the users can be changed to `pub.dev` but all other things in the code itself should stay and not changed.

kevmoo:
> @kevmoo I am thinking that it's best that front facing messages to the users can be changed to `pub.dev` but all other things in the code itself should stay and not changed.

That's my thought, too!

sidrao2006:
Alright then, I'll revert some commits

stefandevo:
Just to inform, because I need to change some public facing values too, I added the changes in this PR https://github.com/dart-lang/pub/pull/2654

lib/src/http.dart
sigurdm: I think this could be made to send the headers to both the old and the new site (they are the same site behind the dns lookup). Just in case someone hardcodes PUB_HOSTED_URL to pub.dev

```
 // We recognize both the old and new name of the server.
 if (!['https://pub.dev', 'https://pub.dartlang.org'].contains(request.url.origin)) return false;
```
@jonasfj do you agree?

lib/src/http.dart
sigurdm: Again here I think it would be better to accept both the new and old name.

lib/src/http.dart
jonasfj: Sure, that's fine.

lib/src/oauth2.dart
jonasfj: I think this okay :D

PR #2654, opened 2020-09-13
Token Authentication for Third Party Package servers

stefandevo:
This PR provides the following pub commands:

~~`pub login <server> [--token <secret>]`~~

`pub login <server>`

The `<secret>` ~~can be set trough the command-line or if not set,~~ pub will prompt to enter a value.
~~You can use an environment variable as `secret`. In that case you need to set `'$YOUR_ENV'` (note the quotes).
Environment variables are the preferred solution. They can also be used in CI environments.~~

Alternative way to set the secret:
`echo MyCrazyToken123 | pub login https://myserver.com`
`echo '$TOKEN' | pub login https://myserver.com`

To remove a token (or all) you can:

`pub logout [<server>] [--all]`

This is an extra feature for the current `logout` command; you can specify a single server, or you can choose for the `-all` argument, which also logs you out from pub.dev.

After this all relevant commands will add an `Authorization` header with `Bearer <token>` value if
specified through the `pub login <server>` command. If not specified all calls will have
no `Authorization` header for both get and publish commands.

Also all front-facing (what the user sees) pub.dartlang.org mentions have been changed to pub.dev.

I think it is important to have this feature for pub. Now for example Github with their Packages feature can implement one for private dart/flutter packages too which is an important missing feature for Enterprise customers. This has been discussed for years now without a fully tested concrete implementation. I talked to @jonasfj about this implementation in order to have some kind of common ground.


doc/repository-spec-v2.md
```diff
@@ -22,22 +40,24 @@ repository.
 
 **Optional headers** used for popularity metrics.
 
-* `X-Pub-OS: windows | macos | linux | ...`
```
kevmoo, line 25: Please avoid unrelated changes in the PR. Do them separately. It makes it hard to review!

stefandevo: sorry for this; it's just vscode that changed the * into - for markdown. I can revert if you want.

lib/src/command/uploader.dart
```diff
@@ -28,7 +27,7 @@ class UploaderCommand extends PubCommand {
 
   UploaderCommand() {
     argParser.addOption('server',
-        defaultsTo: cache.sources.hosted.defaultUrl,
+        defaultsTo: cache.sources.hosted.frontFacingDefaultUrl,
         help: 'The package server on which the package is hosted.');
```
jonasfj, line 31: Since this has never worked before, I propose that we remove this option.

Third-party servers can simply allow users to add/remove uploaders through a web interface.
In fact, I think it's important that we do not make any assumptions about the authorization schemes employed by 3rd-party pub servers.

For corporate internal servers it seems most likely that you'll hook up access to your SSO provider, and/or manage who can publish what packages using a config file.

-----------------------

Long term, I think we are planning to deprecated the `pub uploader` command too. For pub.dev, it no-longer makes sense when the package is owned by a _verified publisher_. This work is tracked in https://github.com/dart-lang/pub-dev/issues/3381

lib/src/command/logout.dart
```diff
@@ -6,20 +6,37 @@ import 'dart:async';
...
 import '../oauth2.dart' as oauth2;
+import '../tokens.dart';
 
 /// Handles the `logout` pub command.
 class LogoutCommand extends PubCommand {
   @override
   String get name => 'logout';
   @override
-  String get description => 'Log out of pub.dartlang.org.';
+  String get description => 'Log out of pub.dev or any hosted server.';
```
jonasfj, line 16: I think we need to update terminology throughout this PR.

`pub.dev` is a _hosted source_, so using `or` makes me think it's not.

Could we say `Log out of pub.dev or any third-party pub server.`.

lib/src/command/login.dart
```diff
@@ -0,0 +1,50 @@
...
+import '../io.dart' as io;
+
+import '../tokens.dart';
+
+/// Handles the `login` pub command.
+class LoginCommand extends PubCommand {
+  @override
+  String get name => 'login';
+  @override
+  String get description => 'Log in to hosted server.';
```
jonasfj, line 17:
```suggestion
  String get description => 'Log in to third-party pub server.';
```

Let's refer to these as "third-party pub server".

doc/repository-spec-v2.md
jonasfj: I suggest we avoid canonicalization, especially when it depends on the vscode plugin in use :D

doc/repository-spec-v2.md
```diff
@@ -10,7 +9,26 @@ This can be overwritten in the `dependencies` of a `pubspec.yaml` or specified
 with the environment variable `PUB_HOSTED_URL`. In the rest of this document we
 shall refer to the base URL as `PUB_HOSTED_URL`.
 
+## Authentication
+
+All relevant commands will add an `Authorization` header with `Bearer <token>` value if
```
jonasfj, line 14: A repository specification does not speak about "commands", it speaks about the interface between _a hosted pub package repository_ and the _pub client_.

This stuff belongs in our documentation on `dart.dev`, but not in the specification for the repository API.

-----------

What we should state here is how a server signals that authorization is required, and how a client is supposed to provide said authorization.

Furthermore we should clearly state which API end-points require authorization.

Should the `archive_url` returned from `GET <PUB_HOSTED_URL>/api/packages/<PACKAGE>` require authorization. My initial inclination is no it should not. The 3rd party server can simply put a _signed URL_ in the response (though I can see that leaking in to logs).

We use this because we actually have packages hosted on a different hostname, we use a GCS storage bucket to serve packages (which is nice).

Also we should determine if authorization may be provided when following redirects. I think it's only the `archive_url` which is says it may have redirects. The others that use redirects are deprecated at this point (so the client will never use them, they are only documented because servers supporting older clients should implement them).

lib/src/http.dart
```diff
@@ -38,14 +40,24 @@ final _sessionId = createUuid();
...
   _PubHttpClient([http.Client inner]) : _inner = inner ?? http.Client();
 
   @override
   Future<http.StreamedResponse> send(http.BaseRequest request) async {
-    if (_shouldAddMetadata(request)) {
+    // check for a token in case of hosted service
+    var token = getToken(cache, request.url);
+    if (token != null) {
+      if (!request.headers.containsKey(HttpHeaders.authorizationHeader)) {
+        request.headers[HttpHeaders.authorizationHeader] = 'Bearer $token';
```
jonasfj, line 56: We need to determine if this should be stripped when following redirects, and whether or not it is.

I haven't thought this through yet, but we need to look out for vulnerabilities where one server redirects to another...

lib/src/tokens.dart
```diff
@@ -0,0 +1,144 @@
...
+    return '`server` cannot be the official package server.\n$server is invalid.';
+  }
+  return null;
+}
+
+void _save(SystemCache cache, List<TokenEntry> tokens) {
+  var tokenPath = _tokensFile(cache);
+  ensureDir(path.dirname(tokenPath));
+  writeTextFile(tokenPath, jsonEncode(tokens));
+  log.fine('Saved tokens.json');
```
jonasfj, line 100: Maybe we should call this `secrets.json` instead?

jonasfj (commented):
This is just a preliminary review, but it looks like a good start.

lib/src/io.dart
```diff
@@ -590,6 +590,22 @@ Future<bool> confirm(String message) {
...
+///
+/// This will automatically append ":" to the message, so [message]
+/// should just be a fragment like, "Enter your name".
+String prompt(String message) {
+  log.fine('Showing prompt: $message');
+  if (runningFromTest) {
+    log.message('$message: ');
+  } else {
+    stdout.write('$message: ');
+  }
```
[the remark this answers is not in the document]
stefandevo: It's exactly the same as the `Confirm` code right above; it's basically to handle the test cases. Removing it (also in Confirm), fails the tests

Confirm code currently:
```
Future<bool> confirm(String message) {
  log.fine('Showing confirm message: $message');
  if (runningFromTest) {
    log.message('$message (y/N)?');
  } else {
    stdout.write('$message (y/N)? ');
  }
  return _stdinLines.first.then(RegExp(r'^[yY]').hasMatch);
}
```

lib/src/io.dart
```diff
@@ -590,6 +590,22 @@ Future<bool> confirm(String message) {
...
 }
 
+/// Displays a message and reads a text response from the user.
+///
+/// Returns the input from stdin.
+///
+/// This will automatically append ":" to the message, so [message]
+/// should just be a fragment like, "Enter your name".
+String prompt(String message) {
+  log.fine('Showing prompt: $message');
```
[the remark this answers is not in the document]
stefandevo: Should I change the Confirm messages right above too?

```
Future<bool> confirm(String message) {
  log.fine('Showing confirm message: $message');
  if (runningFromTest) {
    log.message('$message (y/N)?');
  } else {
    stdout.write('$message (y/N)? ');
  }
  return _stdinLines.first.then(RegExp(r'^[yY]').hasMatch);
}
```

lib/src/http.dart
```diff
@@ -136,7 +148,7 @@ class _PubHttpClient extends http.BaseClient {
   /// Returns a log-formatted string for the HTTP field or header with the given
   /// [name] and [value].
   String _logField(String name, String value) {
-    if (_censoredFields.contains(name.toLowerCase())) {
+    if (!runningFromTest && _censoredFields.contains(name.toLowerCase())) {
```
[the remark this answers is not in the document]
stefandevo: Well in order to test the implementation, I needed the real token to be displayed in the log output - this is and was the way the output headers occur during testing because of the fake server; otherwise it would say 'secret' because the autohorization header is listed as a secure one - which is a good think.

lib/src/http.dart
```diff
@@ -38,14 +40,24 @@ final _sessionId = createUuid();
...
-    if (_shouldAddMetadata(request)) {
+    // check for a token in case of hosted service
+    var token = getToken(cache, request.url);
+    if (token != null) {
+      if (!request.headers.containsKey(HttpHeaders.authorizationHeader)) {
+        request.headers[HttpHeaders.authorizationHeader] = 'Bearer $token';
+      }
+    }
+
+    if (_shouldAddMetadata(request) || token != null) {
```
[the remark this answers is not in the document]
stefandevo: Until before, the metadata was only added when using the official repository. I am not sure why this was. In the `_shouldAddMetadata` there is still and edge case that during test and when `PUB_HOSTED_URL` set a false is given. Furthermore, all origins different from `https://pub.dartlang.org` are returning false.
I looked to the headers, and I believe this information can be helpful for third party servers too. 
Currently, with this code, when you using a third party server WITH authentication, the headers as sent as well.
If you decide that we don't sent headers for third party servers, I will remove the check.

lib/src/command/uploader.dart
stefandevo: I totally agree that the uploader feature is absolutely of no value for third party servers. I only added this after I saw tests where failing because of the changed notation from `pub.dartlang.org` to `pub.dev`. However, when I wanted to change this I found this code where `defaultsTo: cache.sources.hosted.defaultUrl` actually calls

```
  /// Gets the default URL for the package server for hosted dependencies.
  String get defaultUrl {
    // Changing this to pub.dev raises the following concerns:
    //
    //  1. It would blow through users caches.
    //  2. It would cause conflicts for users checking pubspec.lock into git, if using
    //     different versions of the dart-sdk / pub client.
    //  3. It might cause other problems (investigation needed) for pubspec.lock across
    //     different versions of the dart-sdk / pub client.
    //  4. It would expand the API surface we're committed to supporting long-term.
    //
    // Clearly, a bit of investigation is necessary before we update this to
    // pub.dev, it might be attractive to do next time we change the server API.
    return _defaultUrl ??= _pubHostedUrlConfig() ?? 'https://pub.dartlang.org';
  }
```

As I read the notes, I decided it was not a good idea to change this :-)

A quick fix would be that in the mentioned changed line, I just put 'https://pub.dev'. Only when the `PUB_HOSTED_URL` variable is set, this would output wrong info. 
I could change this by using the same evaluation directly as in the getter, but then I have to make the `_pubHostedUrlConfig` public to evaluate. 

Again, I understand it's an option you don't want to support for third party servers; so just setting `http:\\pub.dev` would be an option OR leave this help output to what it is today and revert the unit test where it checks for the correct url. What do you think?

lib/src/command/logout.dart
stefandevo: Great idea! Will change both login and logout commands for this.

doc/repository-spec-v2.md
stefandevo: Ok. I will revert the commit for this file until we know for sure what to include and what not.

I am not completely sure that the `archive_url` can be an unauthenticated url; I think hosts like `GitHub Packages` would require an authenticated url for the real package file as otherwise that url can become exposed. A signed URL could be a solution, but then it needs to be different for every get.

An alternative option is that we support a specific header to be returned from that call that gives an indication if we need to forward the authentication header or not (default=true).

lib/src/http.dart
stefandevo: See my comment above.

doc/repository-spec-v2.md
stefandevo: Reverted commit

stefandevo:
Feel free to resolve conversation on the open questions and my answer.
Only outstanding question as far as I can see is the redirect authorization question and what should be documented about the third party server.

lib/src/io.dart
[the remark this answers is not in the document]
jonasfj: Unless we have a reason to cargo-cult `if (runningFromTest)` I suggest that we don't :D

Does removing it make tests fail?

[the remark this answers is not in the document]
stefandevo: Yes, current tests from Confirm fail when removing it.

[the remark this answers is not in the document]
jonasfj: Okay, I'm wondering if copying it into prompt is desirable, but I'm okay keeping it.

lib/src/http.dart
jonasfj: `var token = getToken(cache, request.url);`, yeah okay, auth won't be added to requests when they are redirected.

I guess, we should think about whether or not it's reasonable that authentication may be added to the `archieve_url` returned when listing versions.

https://github.com/dart-lang/pub/blob/73f0906e152a2de98ec881523f289db04f5b3b9b/lib/src/source/hosted.dart#L371-L376

I think my concern here is that server-A could return a response for package `foo` which tricks the client into making a `GET` request against any resource on server-B with credentials for server-B.

I guess we should consider in:
https://github.com/dart-lang/pub/blob/73f0906e152a2de98ec881523f289db04f5b3b9b/lib/src/source/hosted.dart#L399

That it's only allowed to use credentials to make the request, if `versionInfo.archiveUrl` belongs to `server = source._parseDescription(id.description).last`.

Otherwise, a malicious pub server could return an archive URL and tricks the client into making an authenticated GET request against a resource of the attackers choosing.

lib/src/tokens.dart
```diff
@@ -0,0 +1,144 @@
...
+    return '`server` cannot be the official package server.\n$server is invalid.';
+  }
+  return null;
+}
+
+void _save(SystemCache cache, List<TokenEntry> tokens) {
+  var tokenPath = _tokensFile(cache);
+  ensureDir(path.dirname(tokenPath));
+  writeTextFile(tokenPath, jsonEncode(tokens));
+  log.fine('Saved secrets.json');
```
jonasfj, line 100: Should we perhaps update `_tokens`, just in case we one day decide to save a token from with-in a process that is also doing `pub get`..

lib/src/tokens.dart
```diff
@@ -0,0 +1,144 @@
...
+    return null;
+  }
+}
+
+String _tokensFile(SystemCache cache) =>
+    path.join(cache.rootDir, 'secrets.json');
+
+class TokenEntry {
+  String server;
+  String token;
```
jonasfj, line 131: Can we use `final` so these are not mutable?

lib/src/tokens.dart
```diff
@@ -0,0 +1,144 @@
...
+  } catch (e) {
+    log.error('Warning: could not load the saved tokens: $e');
+    return null;
+  }
+}
+
+String _tokensFile(SystemCache cache) =>
+    path.join(cache.rootDir, 'secrets.json');
+
+class TokenEntry {
```
jonasfj, line 129: Any reason to not make this class private? `_TokenEntry`.

jonasfj (commented):
Okay, I've proposed a few things we can de-scope and push to a follow-up PR.

We have a minor vulnerability when fetching `archive_url` in `lib/src/source/hosted.dart`.

And we have to decide what a `<server>` is?.. Can it include a path, probably needs to, but what side-effects does that have.

I'm still not done reviewing all tests, but I think these are the non-trivial concerns so far.

lib/src/http.dart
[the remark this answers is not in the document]
stefandevo: @jonasfj OK to keep the headers for third party servers?

lib/src/http.dart
stefandevo: Okay I agree. Basically, if the archive_url is not the same origin, no tokens are added, and it's the third-party responsibility to create a one-time secure url to download the package.

lib/src/tokens.dart
```diff
@@ -0,0 +1,144 @@
...
+  }
+  var tokenValue = found.token;
+
+  if (tokenValue != null && tokenValue.startsWith('\$')) {
+    var envVar = tokenValue.substring(1);
+    tokenValue = Platform.environment[envVar];
+    if (tokenValue == null) {
+      log.warning('$envVar environment variable not set');
+    }
+  }
```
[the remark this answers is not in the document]
stefandevo: I will remove the -t option for now.

lib/src/command/login.dart
```diff
@@ -0,0 +1,50 @@
...
+  String get name => 'login';
+  @override
+  String get description => 'Log in to third-party pub server.';
+  @override
+  String get invocation => 'pub login <server> [--token <secret>]';
+
+  LoginCommand() {
+    argParser.addOption('token',
+        abbr: 't',
+        help: 'Token. Environment variable can be used with \'\$YOUR_VAR\'.');
```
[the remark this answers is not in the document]
stefandevo: Noted. Will remove.

lib/src/tokens.dart
```diff
@@ -0,0 +1,144 @@
...
+    tokens.remove(found);
+    log.message('Log out $server successful.');
+  }
+  _save(cache, tokens);
+}
+
+String validateServer(String server) {
+  var uri = Uri.parse(server);
+  if (uri.scheme?.isEmpty ?? true) {
+    return '`server` must include a scheme such as "https://".\n$server is invalid.';
```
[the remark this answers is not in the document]
stefandevo: Maybe remove the last sentence all together? So not repeating the input value with the `is invalid` ?

lib/src/tokens.dart
```diff
@@ -0,0 +1,144 @@
...
+  }
+  if (!uri.hasEmptyPath) {
+    return '`server` must not have a path defined.\n$server is invalid.';
+  }
```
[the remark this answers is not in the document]
stefandevo: Okay, this needs some rework in several places. Also fetching the correct token is more than checking the origin.
This means that we can login to servers https://myserver.com/feed1 and https://myserver.com/feed2 and even https://myserver.com - so we need some evaluation per path there.

lib/src/tokens.dart
stefandevo: Noted. Will change.

lib/src/tokens.dart
stefandevo: Will do

lib/src/tokens.dart
stefandevo: TokenEntry is used in several tests so needs to be public. Should I add the `@visibleForTesting` from the meta package?

stefandevo: That doesn't work for classes I see...

lib/src/tokens.dart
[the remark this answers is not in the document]
stefandevo: @jonasfj I checked the current code base, and actually all current code have the assumption that the $server do not have a trailing slash. So I think it is better to normalize the server values where it is entered (login, logout) without a trailing slash.

stefandevo:
@jonasfj All necessary fixes were made.

I can also confirm that your concern about passing auth header to redirects is like expected, meaning, when the third party server returns an archive_url from the same known third party (for which a token exists), it will pass that token with the GET operation; when it returns another url (like from a S3 server), the auth header is not passed. Also when a third party-server package depends on a package from another server (official one), this is working as expected (also of course not passing any auth headers).

Any steps next to merge this?

lib/src/http.dart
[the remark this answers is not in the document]
stefandevo: The third party server is responsible to make a correct return of the 401. I tested this and I am returning a 401 with the message 'Token not provided or invalid. Please refer to https://mypubserver.com/tokens for more info'
This message is printed to the user.
Furthermore, the current fake server is working correctly for all tests towards official as third party server. I don't think there is a need for a special fake server in this case. The only reason for the `runningFromTest` is to get the token displayed in the log during test to evaluate the correct one.

lib/src/http.dart
stefandevo: All tested and working. Third party servers can have a path and current code fixed. Left uploader code as it is because it is not supported for third party servers.
No possibility for tricks; tokens are only added if archive_url is from same path where token exists for.

test/hosted/hosted_server_test.dart
```diff
@@ -0,0 +1,50 @@
...
+
+import 'package:pub/src/tokens.dart';
+import 'package:test/test.dart';
+
+import '../descriptor.dart' as d;
+import '../test_pub.dart';
+
+void main() {
+  test('check for auth header for hosted server', () async {
+    await servePackages((builder) {
```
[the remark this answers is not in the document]
stefandevo: Answered in other comment; dependency check is also not an issue and working well.

lib/src/tokens.dart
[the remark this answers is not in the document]
stefandevo: All of this checked and changed. Third party servers can have a path and changes made as proposed, removing the starting / from the code and checking for a trailing slash of the server variable passed. Only uploader command not changed as we don't support it for third party servers.

jonasfj (commented):
More comments, I'm not through it all yet.

.gitignore
```diff
@@ -1,10 +1,11 @@
...
 .idea
+.vscode
 .dart_tool/
 .settings/
 /build/
 packages
 /.packages
 pubspec.lock
-/bin/pub.dart.snapshot
+/bin/pub.dart.snapshot
```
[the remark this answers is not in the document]
stefandevo: Fixed in update PR

lib/src/tokens.dart
```diff
@@ -0,0 +1,147 @@
...
+    log.message('No token found for $server.');
+  } else {
+    tokens.remove(found);
+    log.message('Log out $server successful.');
+  }
+  _save(cache, tokens);
+}
+
+String validateServer(String server) {
+  var uri = Uri.parse(server);
```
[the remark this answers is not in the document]
stefandevo: Added + test in update PR

stefandevo:
> More comments, I'm not through it all yet.

All comments processed in updated PR.

lib/src/command/login.dart
```diff
@@ -0,0 +1,44 @@
...
+import '../tokens.dart';
+
+/// Handles the `login` pub command.
+class LoginCommand extends PubCommand {
+  @override
+  String get name => 'login';
+  @override
+  String get description => 'Log in to third-party pub server.';
+  @override
+  String get invocation => 'pub login <server>';
```
isoos, line 19: It is not clear for me what the `<server>` value would be: a `http://fqn.domain/` or just a hostname? I assume the former, but it would be nice if the help message would clarify it.

lib/src/command/login.dart
```diff
@@ -0,0 +1,44 @@
...
+
+  @override
+  Future run() async {
+    if (argResults.rest.isEmpty) {
+      usageException('Must specify a server to log in.');
+    }
+    var server = argResults.rest.first;
```
isoos, line 26: Let's `final` variables for values that are not changing later?

lib/src/command/logout.dart
```diff
@@ -6,20 +6,37 @@ import 'dart:async';
...
   @override
-  String get description => 'Log out of pub.dartlang.org.';
+  String get description => 'Log out of pub.dev or any third-party pub server.';
   @override
-  String get invocation => 'pub logout';
+  String get invocation => 'pub logout [<server>] [--all]';
 
-  LogoutCommand();
+  /// Whether to log out of all servers, including third-party pub servers.
+  bool get all => argResults['all'];
```
isoos, line 21: As this is only referenced in the `run()` method, it should not be exposed as public field like this.

lib/src/command_runner.dart
```diff
@@ -123,6 +124,7 @@ class PubCommandRunner extends CommandRunner {
     addCommand(UploaderCommand());
     addCommand(LogoutCommand());
     addCommand(VersionCommand());
+    addCommand(LoginCommand());
```
isoos, line 127: Maybe this could be placed before the `logout` command?

lib/src/http.dart
```diff
@@ -38,14 +40,24 @@ final _sessionId = createUuid();
 /// user-friendly error messages.
 class _PubHttpClient extends http.BaseClient {
   final _requestStopwatches = <http.BaseRequest, Stopwatch>{};
+  SystemCache get cache => _cache ??= SystemCache();
```
isoos, line 43: Why not pass in the `cache` from the `PubCommand`? This will duplicate the caching and may break a (future) assumption that the cache is a singleton.

lib/src/oauth2.dart
```diff
@@ -120,18 +122,26 @@ Future<T> withClient<T>(SystemCache cache, Future<T> Function(Client) fn) {
...
-      identifier: _identifier,
-      secret: _secret,
-      // Google's OAuth2 API doesn't support basic auth.
-      basicAuth: false,
-      httpClient: httpClient);
-  _saveCredentials(cache, client.credentials);
-  return client;
+Future<Client> _getClient(SystemCache cache, Uri requestServer) async {
+  var isThirdPartyServer =
+      !runningFromTest && requestServer.host != 'https://pub.dartlang.org';
```
isoos, line 127: @jonasfj won't this break pre-release testing? (and similarly in the `getToken` method)

isoos, line 127: @stefandevo: please extract the hard-coded hostname (and URLs) into a single place and reference only the extracted variables. It will help to find all the references that are using the same hostname.

lib/src/command/login.dart
stefandevo: I will add it to the help (validation will step in otherwise on what to expect). Any suggestion on how the message should be displayed? 
`  String get description => 'Log in to third-party pub server. Specify server as a fully qualified URL.';

lib/src/oauth2.dart
stefandevo: Ok will do; note that in the codebase there are hardcoded references already to this hostname, hence why I reused it like that. Ok that I replace it in the whole codebase? (I did get a remark before that I should limit the PR to the scope)

isoos: If @jonasfj is OK with this, let's keep it this way. We can do a follow-up PR that extracts them.

lib/src/command/login.dart
stefandevo: How about the publish command? Because there you have the server option with help text 'The package server to which to upload this package.'

I don' think the description is a good place to explain how the server value should look?
Also the 'invocation' value, is this a place to explain more?

lib/src/http.dart
stefandevo: I don't think this is possible. The _PubHttpClient is constructed through _pubClient variable, which is directly used in the RetryClient constructor, which is used in the _ThrowingClient constructor, which is only used in the httpClient = ThrottleClient which is a public field in the http.dart that is used everywhere...
So no way to pass it around I guess.

lib/src/oauth2.dart
stefandevo: Actually it won't, or at least, it is still the same behaviour as before.
Checking the `requestServer` passed in, you will see that it is set through the `server` field in the `ListCommand`
```
  /// The URL of the server to which to upload the package.
  Uri get server {
    // An explicit argument takes precedence.
    if (argResults.wasParsed('server')) {
      return Uri.parse(argResults['server']);
    }

    // Otherwise, use the one specified in the pubspec.
    if (entrypoint.root.pubspec.publishTo != null) {
      return Uri.parse(entrypoint.root.pubspec.publishTo);
    }

    // Otherwise, use the default.
    return Uri.parse(cache.sources.hosted.defaultUrl);
  }
```
The last line (default) points to
```
    return _defaultUrl ??= _pubHostedUrlConfig() ?? 'https://pub.dartlang.org';

```
in hosted.dart => so in the end it is the same compare (towards https://pub.dartlang.org).

stefandevo:
Test are failing upon `enable_experiments_test.dart`; nothing to do with this PR.

stefandevo:
@jonasfj @isoos any updates?

jonasfj (changes requested):

lib/src/command/logout.dart
```diff
@@ -6,20 +6,37 @@ import 'dart:async';
...
   @override
   Future run() async {
-    oauth2.logout(cache);
+    if (_all) {
+      oauth2.logout(cache);
+      removeToken(cache, all: true);
+    } else if (argResults.rest.isEmpty) {
+      oauth2.logout(cache);
+    } else {
+      final server = argResults.rest.first;
```
jonasfj, line 38: Should we not validate the `server` argument? I suppose that in this case `pub logout pub.dev` or `pub logout pub.dartlang.org` could trigger `oauth2.logout(cache)`.

lib/src/tokens.dart
```diff
@@ -0,0 +1,143 @@
...
+
+String validateServer(String server) {
+  Uri uri;
+  try {
+    uri = Uri.parse(server);
+  } on FormatException catch (e) {
+    return '`server` must be a valid a URL: ${e.message}';
+  }
+
+  if (uri.scheme?.isEmpty ?? true) {
```
jonasfj, line 82: Indeed we could probably say we only support `http` and `https`.

lib/src/tokens.dart
```diff
@@ -0,0 +1,143 @@
...
+  if (found == null) {
+    log.message('No token found for $server.');
+  } else {
+    tokens.remove(found);
+    log.message('Log out $server successful.');
+  }
+  _save(cache, tokens);
+}
+
+String validateServer(String server) {
```
jonasfj, line 74: Can we add a documentation comment?

This is for argument validation for commands that take a `--server` argument, right?

Maybe we could call it: `validateCustomPubServerUrl` ? or something even better?

lib/src/tokens.dart
```diff
@@ -0,0 +1,143 @@
...
+  try {
+    uri = Uri.parse(server);
+  } on FormatException catch (e) {
+    return '`server` must be a valid a URL: ${e.message}';
+  }
+
+  if (uri.scheme?.isEmpty ?? true) {
+    return '`server` must include a scheme such as "https://".\n"$server" is invalid.';
+  }
+  if (uri.hasQuery) {
```
jonasfj, line 85: nit: Probably can't have a `uri.Fragment` either :D

jonasfj: Similarly, [`uri.userInfo`](https://api.dart.dev/stable/2.10.0/dart-core/Uri/userInfo.html) probably won't work well for anybody :D

lib/src/utils.dart
```diff
@@ -626,3 +626,14 @@ Map<K2, V2> mapMap<K1, V1, K2, V2>(
       key(entry.key, entry.value): value(entry.key, entry.value),
   };
 }
+
+/// Makes sure that the given url always ends with a trailing slash
+String checkEndSlash(String url) {
+  if (url.endsWith('/')) return url;
+  return '$url/';
+}
```
jonasfj, line 634: This is better named something like `ensureTrailingSlash`

jonasfj: I've been thinking about the side effects of this...

The commands affected are:
 * `pub publish` (rarely used with 3rd party pub servers)
 * `pub uploader` (rarely used with 3rd party pub servers, and scheduled for deprecation)
 * `pub get` (commonly used by some users with `PUB_HOSTED_URL` or custom pub-server in `pubspec.yaml`).

The impact on `pub publish` and `pub uploader` is pretty breaking, but these have never really worked as intended, so I think that's fine. Well, it's mostly breaking if your custom-pub-server-url included a `/path` part, because that would have been ignored before by `.resolve('/...`

The impact on `pub get` is only that custom-pub-server-urls ending with a slash, will not get double slashes anymore.
I don't think this is too breaking, and we can probably live with that :D

stefandevo:
I am going to close this PR. Currently lack of time to implement all requested changes. 
I think it is best that there is first a RFP issue that describes it. This PR and all the comments was a learning process for you guys; in the end it was not 100% clear how you wanted it.
I still very much believe we need this functionality.
But I had the feeling I was going in circles about how this should be implemented... 
So I would love to work on it, once you have a 100% view how this should work.

sigurdm:
Yes - I think that a major stumbling block is the rather large design space here.

stefandevo:
> Yes - I think that a major stumbling block is the rather large design space here.

I saw you are working on some deprecated stuff around -server etc. This would ease the implementation later. 

I think the original design was pretty ok. 

- able to provide a bearer token through login command
- when uploading, getting checking to see host was has a beare token
- putting the bearer token in the header for the get / upload commands

jonasfj:
Yeah, I think many details to be discovered as this is implemented, because we don't want to break existing users.

PR #2665, opened 2020-09-30
[Prototype] add support for a --channel flag

jakemac53:
This implements an idea I have been considering to allow for release channels for pub packages. This would significantly improve our story around things like null safety, and pre-releases in general.

I define the "channel" for a given version to be the first component of its pre-release version, so a version like `1.2.3-nullsafety.1` is considered to be a part of the "nullsafety" channel.

I then implemented the following changes to pub:

- Add a `--channel` command line argument to all commands that do version solves.
  - **Note**: based on comments below I would revert this option in favor of configuration in the pubspec only 
- Add a required `String channel` argument to the `bestVersion` method.
- Update the `bestVersion` method to **never select pre-release versions by default** (this is technically breaking)
  - They are allowed if explicitly pinned, or if the min constraint is a pre-release version (but only for that same major/minor version) 
- Update the `bestVersion` method to treat pre-release versions with a matching channel equally to stable versions

The meat of the implementation change is [here](https://github.com/dart-lang/pub/pull/2665/files#diff-c454966c2c84f2e98c5a1941050143b0R162), the rest is just plumbing and arg parsing.

I think the configuration here would likely want to be extended to allow setting the channel permanently via the pubspec (possibly under environment), and I also think it should be configurable per package dependency so that you can opt in to the dev channel for instance on a given package only.

This should generally allow for a much more useful pre-release mechanism, which also prevents users from getting undesired pre-releases.

cc @leafpetersen @natebosch @mit-mit @munificent @jonasfj

### DIFF
--- a/lib/src/command/add.dart
+++ b/lib/src/command/add.dart
@@ -44,6 +44,7 @@ class AddCommand extends PubCommand {
   String get hostUrl => argResults['hosted-url'];
   String get path => argResults['path'];
   String get sdk => argResults['sdk'];
+  String get channel => argResults['channel'];
 
   bool get hasGitOptions => gitUrl != null || gitRef != null || gitPath != null;
   bool get hasHostOptions => hostUrl != null;
@@ -61,6 +62,9 @@ class AddCommand extends PubCommand {
     argParser.addOption('hosted-url', help: 'URL of package host server');
     argParser.addOption('path', help: 'Local path');
     argParser.addOption('sdk', help: 'SDK source for package');
+    argParser.addOption('channel',
+        help: 'Pre-release channel to allow, matches the first component of '
+            'pre-release versions');
 
     argParser.addFlag('offline',
         help: 'Use cached packages instead of accessing the network.');
@@ -95,7 +99,8 @@ class AddCommand extends PubCommand {
       /// where the user specifies a version constraint, this serves to ensure
       /// that a resolution exists before we update pubspec.yaml.
       solveResult = await resolveVersions(
-          SolveType.UPGRADE, cache, Package.inMemory(updatedPubSpec));
+          SolveType.UPGRADE, cache, Package.inMemory(updatedPubSpec),
+          channel: channel);
     } on GitException {
       dataError('Unable to resolve package "${package.name}" with the given '
           'git parameters.');
@@ -133,7 +138,9 @@ class AddCommand extends PubCommand {
       await Entrypoint.global(newRoot, entrypoint.lockFile, cache,
               solveResult: solveResult)
           .acquireDependencies(SolveType.GET,
-              dryRun: true, precompile: argResults['precompile']);
+              dryRun: true,
+              precompile: argResults['precompile'],
+              channel: channel);
     } else {
       /// Update the `pubspec.yaml` before calling [acquireDependencies] to
       /// ensure that the modification timestamp on `pubspec.lock` and
@@ -144,7 +151,7 @@ class AddCommand extends PubCommand {
       /// Create a new [Entrypoint] since we have to reprocess the updated
       /// pubspec file.
       await Entrypoint.current(cache).acquireDependencies(SolveType.GET,
-          precompile: argResults['precompile']);
+          precompile: argResults['precompile'], channel: channel);
     }
 
     if (isOffline) {
--- a/lib/src/command/downgrade.dart
+++ b/lib/src/command/downgrade.dart
@@ -24,6 +24,8 @@ class DowngradeCommand extends PubCommand {
   @override
   bool get isOffline => argResults['offline'];
 
+  String get channel => argResults['channel'];
+
   DowngradeCommand() {
     argParser.addFlag('offline',
         help: 'Use cached packages instead of accessing the network.');
@@ -32,6 +34,10 @@ class DowngradeCommand extends PubCommand {
         abbr: 'n',
         negatable: false,
         help: "Report what dependencies would change but don't change any.");
+
+    argParser.addOption('channel',
+        help: 'Pre-release channel to allow, matches the first component of '
+            'pre-release versions');
 
     argParser.addFlag('packages-dir', hide: true);
   }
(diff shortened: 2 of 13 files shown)


natebosch:
To make sure I understand, it's still possible to pick up a prerelease version without the `--channel` flag if it's the only compatible version, or if some other set of deps ends up causing it to get selected, right?

jakemac53:
> To make sure I understand, it's still possible to pick up a prerelease version without the `--channel` flag if it's the only compatible version, or if some other set of deps ends up causing it to get selected, right?

The way I have implemented this, no, but we could change that. It is breaking to no longer allow it.

I believe that you could still get them by explicit pinning, and I think we also want to allow it if the version range includes an explicit pre-release version but I don't think it current does that.

jakemac53:
Note that I am not really looking for code review at this time (but feel free) - what I would really like is to solicit feedback on the general idea.

Specifically the likely most controversial part where I attach extra meaning to the first component of pre-release versions.

jonasfj:
> Add a `--channel` command line argument to all commands that do version solves.

Flags that affect the resolution logic is generally not a good idea. Because commands such as `pub run <package>:[<command>]` asserts that all dependencies are satisfied by `pubspec.lock`, and that `.dart_tool/package_config.json` is the location of dependencies outlined in `pubspec.lock`. Commands like `pub run` generally do this checking the modification-time of `pubspec.yaml`, `pubspec.lock`, and `.dart_tool/package_config.json`, and if that fails looks at the contents of these files.

Hence, we would need to persist flags that affect resolution logic in `pubspec.lock`. And it then raises the question of what happens when the user calls `pub get` without the flag? (does it use the value from `pubspec.lock`?). You could also ask if flags that affect resolution logic should be passed to `pub run` and similar commands?

TL;DR: I would avoid flags that affect resolution logic :D

(side note: we will one-day have to solve this to make a flag like `pub get --no-dev-dependencies`, but that's less urgent)

----------------------------------------

If we put `channel` into the `pubspec.yaml` for individual dependencies, this does raise some questions, notably:
 * Can `package:foo` with dependency on `package:bar` channel `next-gen`, be published on pub.dev?
 * Will this create a source for conflicts?

---------------------------------------

I haven't fully considered the implications of this yet... We should probably also look to see what others are doing, for example [npm has distribution tags](https://docs.npmjs.com/cli/dist-tag#purpose).

jakemac53:
> Flags that affect the resolution logic is generally not a good idea.

I think you have some great points here - I do support the flag for all commands in this PR but I think you have convinced me that it would likely be best if this only lived as a configuration option in the `pubspec.yaml`. I would want that _anyways_, so I think that comes down to just removing the command line argument option.

> * Can `package:foo` with dependency on `package:bar` channel `next-gen`, be published on pub.dev?

I think we would likely want to consider requiring that `package:foo` was also a pre-release in order to do this? Possibly even a pre-release with the same channel name?

However, we don't enforce anything like that today for pre-release dependencies in general, and this isn't really any different than a normal pre-release dep.

I would at least make it a warning like we do for the SDK though.

> * Will this create a source for conflicts?

I don't know of any this would bring up, but definitely that is one of the things we need to consider :)

> I haven't fully considered the implications of this yet... We should probably also look to see what others are doing, for example [npm has distribution tags](https://docs.npmjs.com/cli/dist-tag#purpose).

Right, this is basically a way of implementing that, but one that doesn't require backend support (although possibly we might end up wanting to add backend validation of some sort).

This is essentially what I am trying to enable though. Having support for dev/beta/stable/nullsafety channels in some form I think is a big missing part of our package ecosystem right now.

sigurdm:
I think we should design a separate channel feature rather than piggy-backing on prerelease names to indicate channels.

One possible model:

on pub a package-version has a set of channels. This set can only ever grow.

A package version can then initially be published only into the 'alpha' (or perhaps 'staging') channel. And later it can be promoted by adding the 'beta' channel to it. Finally if it passes 'beta' testing you can add 'production'(the one allowed by the default pubspec). On the other hand, if errors are discovered, they will not be promoted, and a new version can start the process.

The package consumer's pubspec would have to indicate (either globally or per package) what channels it wants to pull from.

jakemac53:
Are there any specific downsides you are concerned about with inferring the channel name from the version?

There is I think nice a simplicity involved with just directly inferring it from the version - not only is the implementation trivial but also it allows for totally arbitrary channel names, and I _think_ it follows a pattern which is already quite established in the community for pre-release versions.

I would not be opposed to more explicit channels either, but I am not really sure what the extra complexity buys us?

sigurdm:
> Are there any specific downsides you are concerned about with inferring the channel name from the version?

Mainly it is a stomach feeling that we are overloading these to do more than what can be expected from them, and though it is a somewhat established pattern It seems to give us constant headaches.

But concretely when I read the semver spec, I get the feeling that the prerelease name is meant to indicate stability of the api - not necessarily stability (or other features) of the code. It is meant to be ranked alphabetically which is probably not what you want from channels.

jonasfj:
Using a single _version number_ to indicate both
 * API-compatibility, and,
 * Implementation-maturity.

It's probably a bad idea, it's not long ago that we discussed making "1.0.0 releases non-breaking" (see https://github.com/dart-lang/pub-dev/issues/3385).
This was motivated by the fact that many packages were publishing as `0.x.y` and when they reached implementation-maturity they wanted to publish a `1.0.0` release. But doing a `1.0.0`-release would also be a breaking release for the eco-system, and hence, undesired.

I think we concluded that _abusing semantic version_ was a bad idea.

---------------

If channels are intended to signal _implementation-maturity_, then there is a solid argument that it should be a concept orthogonal to _API-compatibility_.

My only concern is that a feature like channels will have little use. We already see that few package authors are publishing pre-releases before making breaking API changes.

jakemac53:
> If channels are intended to signal _implementation-maturity_, then there is a solid argument that it should be a concept orthogonal to _API-compatibility_.

Pre-release versions here already are not used to signify _API compatibility_ at all - the major/minor/patch versions preceding them should be used to indicate that. I am not quite following the concern here?

> My only concern is that a feature like channels will have little use. We already see that few package authors are publishing pre-releases before making breaking API changes.

My argument is that pre-release versions as they exist today are almost entirely useless, which is _why_ we see little to no usage. You cannot generally opt in to all pre-releases for packages, you have to explicitly pin individual ones. Nobody does that, so no package author bothers.

My entire goal here is to turn them into actually useful things, where users can opt in to the `dev` channel generally should they so choose. The hope is that this would increase usage. :)

This is also an extremely trivial feature to implement so it isn't incurring a lot of technical debt or anything.

jakemac53:
Note that this is also particularly interesting for language experiments - it allow package authors to safely publish versions of their packages that require a certain experiment, and it will be guarded behind a specific channel opt-in.

mit-mit:
Closing as stale

PR #2672, opened 2020-10-05
Fix experiments tests for Dart 2.11

kevmoo:
Tests for pub will only ever be run on the latest dev release, so
no need to skip on previous SDKs


kevmoo:
CC @jakemac53 – did you write this code?

natebosch (approved):

PR #2673, opened 2020-10-05
Use `await expectLater` in test helper

kevmoo:
Ensures better stack traces on failures

### DIFF
--- a/test/test_pub.dart
+++ b/test/test_pub.dart
@@ -312,23 +312,23 @@ Future runPub(
       args: args, workingDirectory: workingDirectory, environment: environment);
   await pub.shouldExit(exitCode);
 
-  await expectLater(() async {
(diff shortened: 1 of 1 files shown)


test/test_pub.dart
```diff
@@ -312,7 +312,7 @@ Future runPub(
       args: args, workingDirectory: workingDirectory, environment: environment);
   await pub.shouldExit(exitCode);
 
-  expect(() async {
+  await expectLater(() async {
```
natebosch, line 315: FYI As long as the future returned from `runPub` is awaited in the tests the `await expectLater(..., completes);` isn't doing anything worthwhile.

natebosch (approved):

test/test_pub.dart
kevmoo: 👍

natebosch (approved):

Issue #2695, opened 2020-10-27
Avoid stacktrace when `PUB_HOSTED_URL` is not a URL


jonasfj:
We should exit non-zero and print an error when `PUB_HOSTED_URL` is not a valid URL. We can do that by throwing the kind of exception that `pub` prints nicely.
```console
$ PUB_HOSTED_URL='not a url' pub get
Unhandled exception:
`PUB_HOSTED_URL` must include a scheme such as "https://". not a url is invalid
#0      HostedSource._pubHostedUrlConfig (package:pub/src/source/hosted.dart:66:7)
#1      HostedSource.defaultUrl (package:pub/src/source/hosted.dart:56:28)
#2      new UploaderCommand (package:pub/src/command/uploader.dart:31:42)
#3      new PubCommandRunner (package:pub/src/command_runner.dart:119:16)
#4      main (file:///b/s/w/ir/cache/builder/sdk/third_party/pkg/pub/bin/pub.dart:8:3)
#5      _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:299:32)
#6      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:168:12
```

This aims to avoid bug reports like #2693 and #1941.

PR #2697, opened 2020-10-28
Deprecate --server argument to `pub publish` and `pub uploader`.

sigurdm:
This argument does not really work optimally.

This also fixes the error message given when
PUB_HOSTED_URL has a bad scheme.

Fixes: https://github.com/dart-lang/pub/issues/2695

### DIFF
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -61,7 +61,9 @@ class LishCommand extends PubCommand {
         negatable: false,
         help: 'Publish without confirmation if there are no errors.');
     argParser.addOption('server',
-        help: 'The package server to which to upload this package.');
+        help: 'The package server to which to upload this package.\n'
+            'DEPRECATED: use `publish_to` in your pubspec.yaml or set the \n'
+            '\$PUB_HOSTED_URL environment variable.');
   }
 
   Future _publish(List<int> packageBytes) async {
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -61,9 +61,8 @@ class LishCommand extends PubCommand {
         negatable: false,
         help: 'Publish without confirmation if there are no errors.');
     argParser.addOption('server',
-        help: 'The package server to which to upload this package.\n'
-            'DEPRECATED: use `publish_to` in your pubspec.yaml or set the \n'
-            '\$PUB_HOSTED_URL environment variable.');
+        help: 'The package server to which to upload this package.',
+        hide: true);
   }
 
   Future _publish(List<int> packageBytes) async {
@@ -117,6 +116,16 @@ class LishCommand extends PubCommand {
 
   @override
   Future runProtected() async {
+    if (argResults.wasParsed('server')) {
+      await log.warningsOnlyUnlessTerminal(() {
+        log.message(
+          '''
+The --server option is deprecated. Use `publish_to` in your pubspec.yaml or set
+the \$PUB_HOSTED_URL environment variable.''',
+        );
+      });
+    }
+
     if (force && dryRun) {
       usageException('Cannot use both --force and --dry-run.');
     }
--- a/lib/src/command/uploader.dart
+++ b/lib/src/command/uploader.dart
@@ -32,7 +32,8 @@ class UploaderCommand extends PubCommand {
         defaultsTo: Platform.environment['PUB_HOSTED_URL'] ??
             'https://pub.dartlang.org',
         help: 'The package server on which the package is hosted.\n'
-            'DEPRECATED: use `publish_to` in your pubspec.yaml or set \$PUB_HOSTED_URL.');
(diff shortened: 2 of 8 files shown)


lib/src/source/hosted.dart
```diff
@@ -62,10 +62,10 @@ class HostedSource extends Source {
     var url = io.Platform.environment['PUB_HOSTED_URL'];
     if (url == null) return null;
     var uri = Uri.parse(url);
-    if (uri.scheme?.isEmpty ?? true) {
+    if (uri.scheme != 'http' && uri.scheme != 'https') {
```
kevmoo, line 65: `uri.hasScheme` is slightly more...uh...right

Handles weird casing, etc.

lib/src/command/uploader.dart
```diff
@@ -28,8 +29,10 @@ class UploaderCommand extends PubCommand {
 
   UploaderCommand() {
     argParser.addOption('server',
-        defaultsTo: cache.sources.hosted.defaultUrl,
-        help: 'The package server on which the package is hosted.');
+        defaultsTo: Platform.environment['PUB_HOSTED_URL'] ??
+            'https://pub.dartlang.org',
+        help: 'The package server on which the package is hosted.\n'
+            'DEPRECATED: use `publish_to` in your pubspec.yaml or set \$PUB_HOSTED_URL.');
```
kevmoo, line 35: Be a bit more clear here `...set the $PUB_HOSTED_URL environment variable`.

lib/src/source/hosted.dart
sigurdm: Hmm - I think hasScheme only tells us if we have a scheme or not.
We want to ensure that the scheme is one of 'http' and 'https'.

But the casing part is a good point. Fixed!

lib/src/command/uploader.dart
sigurdm: Better, thanks!

lib/src/command/uploader.dart
```diff
@@ -28,8 +29,11 @@ class UploaderCommand extends PubCommand {
 
   UploaderCommand() {
     argParser.addOption('server',
-        defaultsTo: cache.sources.hosted.defaultUrl,
-        help: 'The package server on which the package is hosted.');
+        defaultsTo: Platform.environment['PUB_HOSTED_URL'] ??
+            'https://pub.dartlang.org',
+        help: 'The package server on which the package is hosted.\n'
+            'DEPRECATED: use `publish_to` in your pubspec.yaml or set the\n'
+            '\$PUB_HOSTED_URL environment variable.');
```
jonasfj, line 36: No need to write that it's deprecated... just make it flag hidden. And then print as warning if `wasParsed('server') && isTerminal`...

lib/src/command/lish.dart
```diff
@@ -61,7 +61,9 @@ class LishCommand extends PubCommand {
         negatable: false,
         help: 'Publish without confirmation if there are no errors.');
     argParser.addOption('server',
-        help: 'The package server to which to upload this package.');
+        help: 'The package server to which to upload this package.\n'
+            'DEPRECATED: use `publish_to` in your pubspec.yaml or set the \n'
+            '\$PUB_HOSTED_URL environment variable.');
```
jonasfj, line 66: Similar here, can't we just hide the flag?

lib/src/command/uploader.dart
sigurdm: Done

lib/src/command/lish.dart
sigurdm: Done

jonasfj (approved):

lib/src/source/hosted.dart
```diff
@@ -62,10 +62,10 @@ class HostedSource extends Source {
     var url = io.Platform.environment['PUB_HOSTED_URL'];
     if (url == null) return null;
     var uri = Uri.parse(url);
-    if (uri.scheme?.isEmpty ?? true) {
+    if (!['http', 'https'].contains(uri.scheme.toLowerCase())) {
```
jonasfj, line 65: Why not `uri.hasScheme('http') || uri.hasScheme('https')`

jonasfj (approved):

lib/src/source/hosted.dart
jonasfj: I mean: `uri.isScheme('http') || uri.isScheme('https')`

jonasfj: https://api.dart.dev/stable/2.10.3/dart-core/Uri/isScheme.html

sigurdm: Ah - that works!

lib/src/source/hosted.dart
sigurdm: Ah you mean `uri.isScheme` didn't know about that!

Fixed!

jonasfj (approved):

PR #2698, opened 2020-10-28
Make `run` available (but deprecated) in the embedding

sigurdm:


test/embedding/goldens/helptext.txt
```diff
@@ -68,6 +69,7 @@ Available subcommands:
   outdated    Analyze your dependencies to find which ones can be upgraded.
   publish     Publish the current package to pub.dartlang.org.
   remove      Removes a dependency from the current package.
+  run         DEPRECATED: Use `dart run`.
```
jonasfj, line 72: https://pub.dev/documentation/args/latest/command_runner/Command/hidden.html

lib/src/command/run.dart
```diff
@@ -17,15 +17,19 @@ class RunCommand extends PubCommand {
...
+      : 'Run an executable from a package.';
   @override
   String get argumentsDescription => '<executable> [arguments...]';
   @override
   String get docUrl => 'https://dart.dev/tools/pub/cmd/pub-run';
   @override
   bool get allowTrailingOptions => false;
 
-  RunCommand() {
+  final bool deprecated;
```
jonasfj, line 30: Let's just set: https://pub.dev/documentation/args/latest/command_runner/Command/hidden.html

sigurdm: Done

test/embedding/goldens/helptext.txt
sigurdm: Done

jonasfj (approved):

natebosch:
If I understand correctly this will warn if you use `dart pub run` and recommend instead `dart run`.

The latter does not solve all the same use cases. For example running a binary from another package like `pub run build_runner`. How do we anticipate users running for that use case?

jonasfj:
@natebosch, we aim to make it so that `dart run <package>[:<command>]` will work the same way `pub run <package>[:<command>]`.

PR #2704, opened 2020-11-02
Avoid hanging if we don't use explicit `exit()`

sigurdm:


jonasfj (approved):
Can we add a comment to `alwaysUseSubprocess` explaining why we introduced it?

sigurdm:
> Can we add a comment to `alwaysUseSubprocess` explaining why we introduced it?

done

lib/src/oauth2.dart
```diff
@@ -84,16 +84,17 @@ void logout(SystemCache cache) {
...
+/// Does not close the client, since that would close the shared client. It must
+/// be closed elsewhere.
 ///
 /// This takes care of loading and saving the client's credentials, as well as
 /// prompting the user for their authorization. It will also re-authorize and
 /// re-run [fn] if a recoverable authorization error is detected.
 Future<T> withClient<T>(SystemCache cache, Future<T> Function(Client) fn) {
   return _getClient(cache).then((client) {
     return fn(client).whenComplete(() {
-      client.close();
```
jonasfj, line 96: It would be better to just use a new client... or to wrap all access to the other client with a  `withClient` function in `http.dart`.

jonasfj (approved):

lib/src/oauth2.dart
sigurdm: Added a TODO.
All client.close() does is closing the shared (inner) client. And that gets closed in the PubCommand.run.

PR #2714, opened 2020-11-04
nit: use d.file instead of YamlDescriptor

jonasfj:


jonasfj:
Closed in favor of https://github.com/dart-lang/pub/pull/2718 which also has this fix.

Issue #2707, opened 2020-11-02
Error in pub when packages don't have a lower-bound SDK constraint


jonasfj:
Outdated proposal one; for the current proposal see https://github.com/dart-lang/pub/issues/2707#issuecomment-720599514.

---

When `pubspec.yaml` does not contain a lower-bound SDK constraint (or doesn't contain an SDK constraint). We should:
 * [ ] (A) infer a _default language version_ of `2.0`, which should be written to `.dart_tool/package_config.json`, and,
 * [ ] (B) show a warning in `dart pub get`, advicing the user to specify an SDK constraint.

```
        Warning: `pubspec.yaml` has no SDK constraint lower bound; defaulting to language version 2.0.
        You should edit pubspec.yaml to add an SDK constraint, e.g.:
        environment:
          sdk: '>=2.0.0 <3.0.0'
```

**Blocked**, waiting for feedback.

jonasfj:
> IIRC if there is no SDK constraint, the current tooling assumes Dart 1.x

We do when consume from `pub.dev`, but not when reading source on the users machine. If there is no SDK constraint, we just assume the code works with the current Dart SDK that the user has installed.

mit-mit:
Alternate proposal:

Now:
  - Announce a breaking change that having a Dart SDK constraint in a pubspec is required
  - Add a warning to pub get (the dart.dev link in the warning contains docs that explain how to pick a lower constraint with/without null safety)

    ```
    Warning: pubspec.yaml has no SDK constraint lower bound; defaulting to <current SDK version>.
    You should edit pubspec.yaml to contain an SDK constraint: http://dart.dev/go/sdk-constraint
    ```

In Dart 3:
  - Make it an error in pub to not have an Dart SDK constraint

mit-mit:
While we still don't have consensus on the concrete solution, I believe we have an agreement that something has to happen here. Tagging as a beta experience blocker.

mit-mit:
@jakemac53 @natebosch any reactions to the [alternate proposal](https://github.com/dart-lang/pub/issues/2707#issuecomment-720599514) ?

jakemac53:
Yes, with the caveat of if we make this an error in Dart 3 it needs be something we ensure we can do in a way that doesn't break users on older sdks.

mit-mit:
If it's a client side error, then given pub ships in the SDK, how could it break older SDKs?

jakemac53:
[...]

PR #2718, opened 2020-11-04
Require sdk constraint

jonasfj:
Fixes https://github.com/dart-lang/pub/issues/2707#issuecomment-720599514

Associated breaking change request: https://github.com/dart-lang/sdk/issues/44072

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -170,12 +170,12 @@ class Entrypoint {
 
   /// Loads the entrypoint for the package at the current directory.
   Entrypoint.current(this.cache)
-      : root = Package.load(null, '.', cache.sources, isRootPackage: true),
+      : root = Package.load(null, '.', cache.sources),
         isGlobal = false;
 
   /// Loads the entrypoint from a package at [rootDir].
   Entrypoint(String rootDir, this.cache)
-      : root = Package.load(null, rootDir, cache.sources, isRootPackage: true),
+      : root = Package.load(null, rootDir, cache.sources),
         isGlobal = false;
 
   /// Creates an entrypoint given package and lockfile objects.
@@ -226,6 +226,21 @@ class Entrypoint {
     bool dryRun = false,
     bool precompile = false,
   }) async {
+    // We require an SDK constraint lower-bound as of Dart 2.12.0
+    final dartSdkConstraint = root.pubspec.sdkConstraints['dart'];
+    if (dartSdkConstraint is! VersionRange ||
+        (dartSdkConstraint is VersionRange && dartSdkConstraint.min == null)) {
+      throw DataException('''
+pubspec.yaml has no lower-bound SDK constraint.
+You should edit pubspec.yaml to contain an SDK constraint:
+
+environment:
+  sdk: '>=${sdk.version} <${sdk.version.nextBreaking}'
+
+See https://dart.dev/go/sdk-constraint
+''');
+    }
+
     var result = await log.progress(
       'Resolving dependencies',
       () => resolveVersions(
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -235,7 +235,7 @@ pubspec.yaml has no lower-bound SDK constraint.
 You should edit pubspec.yaml to contain an SDK constraint:
 
 environment:
-  sdk: '>=${sdk.version} <${sdk.version.nextBreaking}'
+  sdk: '>=2.10.0 <${sdk.version.nextBreaking}'
 
 See https://dart.dev/go/sdk-constraint
 ''');
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -230,12 +230,29 @@ class Entrypoint {
     final dartSdkConstraint = root.pubspec.sdkConstraints['dart'];
     if (dartSdkConstraint is! VersionRange ||
         (dartSdkConstraint is VersionRange && dartSdkConstraint.min == null)) {
+      // Suggest version range '>=2.10.0 <3.0.0', we avoid using:
+      // [CompatibleWithVersionRange] because some pub versions don't support
+      // caret syntax (e.g. '^2.10.0')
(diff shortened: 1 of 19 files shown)


lib/src/entrypoint.dart
```diff
@@ -224,6 +224,21 @@ class Entrypoint {
...
+    // We require an SDK constraint lower-bound as of Dart 2.12.0
+    final dartSdkConstraint = root.pubspec.sdkConstraints['dart'];
+    if (dartSdkConstraint is! VersionRange ||
+        (dartSdkConstraint is VersionRange && dartSdkConstraint.min == null)) {
+      throw DataException('''
+pubspec.yaml has no lower-bound SDK constraint.
+You should edit pubspec.yaml to contain an SDK constraint:
+
+environment:
+  sdk: '>=${sdk.version} <${sdk.version.nextBreaking}'
```
mit-mit, line 236: This will suggest `>=2.12.0`, right? That will cause failures for most projects that didn't have an SDK constraint to begin with. Should we consider using a fixed lower-bound?

Second, I wonder if this works for non-stable builds; doesn't it end up suggesting a stable version lower bound for a release that still might be in dev or beta only?

jonasfj: This will suggest the current SDK version as lower-bound (I think this SDK version includes prerelease information, I should add a test for that).

```
pubspec.yaml has no lower-bound SDK constraint.
You should edit pubspec.yaml to contain an SDK constraint:

environment:
  sdk: '>=${sdk.version} <${sdk.version.nextBreaking}'

See https://dart.dev/go/sdk-constraint
Use if your code is not migrated to null-safety yet.

environment:
  sdk: '>=2.10.0 <${sdk.version.nextBreaking}'

See <link-to-null-safety-migration-guide>
```

It just seemed like that was a rather lengthy message. I suspect `https://dart.dev/go/sdk-constraint` will explain opting-in/out of null-safety.

mit-mit: I suggest we fix the lower bound at `2.10`

mit-mit:
```
pubspec.yaml has no lower-bound SDK constraint.
You should edit pubspec.yaml to contain an SDK constraint:

environment:
  sdk: '>=2.10 <${sdk.version.nextBreaking}'

For more details, see https://dart.dev/go/sdk-constraint
```

lib/src/entrypoint.dart
```diff
@@ -226,6 +226,38 @@ class Entrypoint {
...
+    // We require an SDK constraint lower-bound as of Dart 2.12.0
+    final dartSdkConstraint = root.pubspec.sdkConstraints['dart'];
+    if (dartSdkConstraint is! VersionRange ||
+        (dartSdkConstraint is VersionRange && dartSdkConstraint.min == null)) {
+      // Suggest version range '>=2.10.0 <3.0.0', we avoid using:
+      // [CompatibleWithVersionRange] because some pub versions don't support
+      // caret syntax (e.g. '^2.10.0')
+      var suggestedConstraint = VersionRange(
+        min: Version.parse('2.10.0'),
+        max: Version.parse('2.10.0').nextBreaking,
```
sigurdm, line 238:
```suggestion
        min: Version.parse(Platform.version.split(' ').first),
        max: Version.parse(Platform.version.split(' ').first).nextBreaking,
```

lib/src/entrypoint.dart
```diff
@@ -226,6 +226,38 @@ class Entrypoint {
     bool dryRun = false,
     bool precompile = false,
   }) async {
+    // We require an SDK constraint lower-bound as of Dart 2.12.0
```
sigurdm, line 229: Could you move this to a helper method. acquireDependencies is already huge.

lib/src/entrypoint.dart
```diff
@@ -170,12 +170,12 @@ class Entrypoint {
 
   /// Loads the entrypoint for the package at the current directory.
   Entrypoint.current(this.cache)
-      : root = Package.load(null, '.', cache.sources, isRootPackage: true),
+      : root = Package.load(null, '.', cache.sources),
```
sigurdm, line 173: I love that we get rid of `isRootPackage`!!!

lib/src/entrypoint.dart
sigurdm, line 229: Should we move this to general validation of the pubspec? I guess that would make a difference for consuming 1.0 packages, but that doesn't really work anyway.

lib/src/entrypoint.dart
jonasfj: Yes, it's fantastic :D

lib/src/entrypoint.dart
jonasfj: > I guess that would make a difference for consuming 1.0 packages, but that doesn't really work anyway.

Except all tests have SDK version `0.1.2+4`, so it might further break those :see_no_evil:

jonasfj: Also if you try to depend on a Dart 1.x package, you won't a bug that it failed to parse the `pubspec`, instead you'll get a `SolveFailure`.

More importantly, we still need to understand Dart 1.x packages, because they still exist on pub.dev. So we need to be able to read Dart 1.x packages without an SDK constraint.

lib/src/entrypoint.dart
jonasfj: Then I could just use `sdk.version`.

For first few versions will recommend `2.10.0` as the lower-bound, we change this recommendation later...

lib/src/entrypoint.dart
sigurdm: :( isn't it a shame not promoting the latest and greatest?

lib/src/entrypoint.dart
sigurdm: Good points. I just wonder if we want to do the check for other entrypoints?

But I guess they mostly don't work before a pub get has run anyways.

sigurdm (approved):
LGTM!

Issue #2716, opened 2020-11-04
Simplify null-safety analysis for "pub outdated --mark=null-safety"


jonasfj:
We currently download package versions to look for _language version markers_.

However, some packages like `package:test` will have infrastructure libraries only used by test runners that won't be ported anytime soon. Hence, such packages will show up as _null-safety in-compatible_.

To mitigate this issue, we're going to simply our analysis to only look at the _lower-bound SDK constraint of the package_. Ignoring whether transitive dependencies have been migrated out-of-order, and whether the package contains files that are opt-out using _language version markers_.

If packages are using _language version markers_ or _migrating out-of-order_ (against recommendations), we are simply going to trust that the author knows what they are doing.
(On pub.dev will we score this negatively, but in `pub outdated --mark=null-safety` we'll accepted it).

Note:
 * This means `pub outdated --mark=null-safety` won't look at _transitive dependencies_.
 * As an added benefit `pub outdated --mark=null-safety` won't need to download package versions anymore.

---------

In a further out future, we can consider refining the definition of _null-safety compatibility_ in `pub`. Perhaps by tracing imports from the users code, to identify which packages and libraries need to be _null-safety compatible_ (before then tracing imports from those packages).

franklinyow:
@mit-mit @leafpetersen 
Beta blocker?

leafpetersen:
@franklinyow I believe so, yes.

leafpetersen:
Does this need to be pulled into the sdk and potentially cherry picked?

mit-mit:
> Does this need to be pulled into the sdk and potentially cherry picked?

We have a few more pub fixes planned for today. After those we'll roll pub (via DEPS), and then we'll need a Dart roll or a cherry pick.

jonasfj:
Fixed in https://dart-review.googlesource.com/c/sdk/+/170690

PR #2721, opened 2020-11-05
Simplify null-safety analysis in `pub outdated --mode=null-safety`

jonasfj:
Fixes https://github.com/dart-lang/pub/issues/2716

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -9,13 +9,11 @@ import 'dart:math';
 
 import 'package:pub_semver/pub_semver.dart';
 import 'package:meta/meta.dart';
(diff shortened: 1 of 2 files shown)


sigurdm (approved):
LGTM if we change the wording "fully supports null-safety" to something else.

mit-mit:
> LGTM if we change the wording "fully supports null-safety" to something else.

Can't we just remove that word, and say "dev_dependencies: all support null safety" ?

jonasfj:
> LGTM if we change the wording "fully supports null-safety" to something else.

Yes, let's do that in another PR.. but yes, I **fully** agree :D

Issue #2717, opened 2020-11-04
Remove pub get warning about partial null-safety compatibility


jonasfj:
Instead we'll rely on the analyzer to warn people when they are importing code that isn't null-safety compatible (https://github.com/dart-lang/sdk/issues/44063).

In the future, we might consider adding back this analysis, but then we'll likely have to refine the analysis to trace import-paths from the users code, such that if a dependency has a _null-safety incompatible_ library that isn't imported by the users code, this will not trigger a warning.

jonasfj:
Fixed in https://dart-review.googlesource.com/c/sdk/+/170690

PR #2723, opened 2020-11-05
Remove warning about mixed mode.

sigurdm:
The warning was hard to parse - and would be triggered too often (too noisy).

We can reconsider this later when we have a better feeling for the migration flow
or most packages are fully migrated so we would have less noise.

Fixes: https://github.com/dart-lang/pub/issues/2717


jonasfj (approved):

mit-mit:
@sigurdm looks like this needs a small tweak to remove a few imports:

```
25.88s$ dartanalyzer --fatal-infos --fatal-warnings .
Analyzing pub...
  error • Unused import: 'package:meta/meta.dart'. • lib/src/solver/result.dart:6:8 • unused_import
  error • Unused import: '../log.dart'. • lib/src/solver/result.dart:10:8 • unused_import
  error • Unused import: '../null_safety_analysis.dart'. • lib/src/solver/result.dart:11:8 • unused_import
  error • Unused import: '../system_cache.dart'. • lib/src/solver/result.dart:16:8 • unused_import
4 errors found.
The command "dartanalyzer --fatal-infos --fatal-warnings ." exited with 3.
cache.2
store build cache
```

Issue #2702, opened 2020-10-30
pub outdated --mode=null-safety output is confusing


johnpryan:
Running `dart pub outdated --mode=null-safety` on my project says that all packages fully support null safety, but `dart migrate`displays a warning.

```
dart --version
Dart SDK version: 2.12.0-edge.bf968d56270185ffd7b2600288e56f82c5bbe09d (be) (Fri Oct 30 20:59:22 2020 +0000) on "macos_x64"
```

## Steps to reproduce

1. Create a new "dart package" project in IntelliJ
```yaml
name: null_safety_migration_messages
description: A starting point for Dart libraries or applications.

environment:
  sdk: '>=2.12.0 <3.0.0'

dev_dependencies:
  test: '>=1.15.4 <2.0.0'
  pedantic: ">=1.9.2 <2.0.0"
  coverage: ">=0.13.11 <0.14.0"
```

Run `dart pub outdated --mode=null-safety`

```
dart pub outdated --mode=null-safety
Running in 'null safety' mode.
Showing packages where the current version doesn't fully support null safety.

Dependencies                Current   Upgradable  Resolvable  Latest    
all fully support null safety

dev_dependencies           
coverage                    0.14.1    -           -           0.14.1    
pedantic                    1.9.2     -           -           1.9.2     
test                        1.15.4    -           -           1.15.4    

transitive dependencies: all fully support null safety

transitive dev_dependencies
_fe_analyzer_shared         12.0.0    -           -           12.0.0    
analyzer                    0.40.5    -           -           0.40.5    
args                        1.6.0     -           -           1.6.0     
async                       2.4.2     -           -           2.4.2     
boolean_selector            2.0.0     -           -           2.0.0     
charcode                    1.1.3     -           -           1.1.3     
cli_util                    0.2.0     -           -           0.2.0     
collection                  1.14.13   -           -           1.14.13   
convert                     2.1.1     -           -           2.1.1     
crypto                      2.1.5     -           -           2.1.5     
glob                        1.2.0     -           -           1.2.0     
http                        0.12.2    -           -           0.12.2    
http_multi_server           2.2.0     -           -           2.2.0     
http_parser                 3.1.4     -           -           3.1.4     
io                          0.3.4     -           -           0.3.4     
js                          0.6.2     -           -           0.6.2     
logging                     0.11.4    -           -           0.11.4    
matcher                     0.12.9    -           -           0.12.9    
meta                        1.2.3     -           -           1.2.3     
[...]

PR #2729, opened 2020-11-06
Change message for no Latest resolution

mit-mit:
Attempt to make the message for no possible resolution for Latest versions a bit easier to read

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -501,9 +501,9 @@ Future<void> _outputHuman(
       upgradable == 0 &&
       rows.isNotEmpty) {
     log.message(
-        '\nCurrent dependencies are constrained to versions listed in Resolveable.'
+        '\nCurrent dependencies are constrained to versions listed in Resolvable.'
         '\nA resolution compatible with the Latest versions doesn\'t exist, as they are not'
-        '\nmutually compatible with one or more of your current dependencies.');
+        '\nmutually compatible.');
   }
 
   if (notAtResolvable != 0) {
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -7,8 +7,8 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:math';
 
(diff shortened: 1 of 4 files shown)


jonasfj (approved):

jonasfj:
It's a complicated scenario... and not easy to explain in 2 lines... We could also leave `dart.dev/go/...`-link and then explain it in an FAQ on dart.dev

mit-mit:
That's right. But is this attempt easier to understand than the current?

lib/src/command/outdated.dart
```diff
@@ -501,8 +501,9 @@ Future<void> _outputHuman(
       upgradable == 0 &&
       rows.isNotEmpty) {
     log.message(
-        '\nDependencies are all constrained to the latest resolvable versions.'
-        '\nNewer versions, while available, are not mutually compatible.');
+        '\nCurrent dependencies are constrained to versions listed in Resolveable.'
```
sigurdm, line 504:
```suggestion
        '\nCurrent dependencies are constrained to versions listed in Resolvable.'
```

sigurdm:
I liked @leafpetersen 's [suggestion](https://github.com/dart-lang/pub/issues/2702#issuecomment-722706204) for using fewer hard words ("mutually compatible").

I also think it is better just to say that they are not 'mutually compatible' instead of 'mutually compatible with current dependencies'.

I don't think the 'current' versions really matter that much here.

mit-mit:
@sigurdm what about this update (I just committed):

```
Current dependencies are constrained to versions listed in Resolvable.
A resolution compatible with the Latest versions doesn't exist, as they are not
mutually compatible.
```

sigurdm:
I ... think it's better. Should we ask @kwalrath ?

mit-mit:
> Should we ask @kwalrath ?

I'd like to land this now if we think it's ~better than the current, so we can start the Dart roll.

But Kathy will be testing the overall flows next week, and I'll make sure to ask her to review all UI messages.

sigurdm (approved):
LGTM

PR #2731, opened 2020-11-06
Added a dart pub outdated --transitive option

jonasfj:


test/outdated/goldens/helptext.txt
```diff
@@ -10,6 +10,9 @@ Usage: pub outdated [options]
                                    (defaults to on)
     --[no-]dev-dependencies        Take dev dependencies into account.
                                    (defaults to on)
+    --[no-]transitive              Show transitive dependencies.
+                                   (Defaults to off in --mode=null-safety).
```
sigurdm, line 14: Looks like this should be lower case to match the line below

sigurdm (approved):
LGTM

test/outdated/goldens/helptext.txt
```diff
@@ -10,6 +10,9 @@ Usage: pub outdated [options]
...
+                                   (defaults to on)
```
jonasfj, line 15: Fix this

test/outdated/goldens/newer_versions.txt
```diff
@@ -252,18 +291,10 @@ local_package  ✗0.0.1   ✗0.0.1      ✗0.0.1      ✗0.0.1
...
-transitive dev_dependencies:
-dev_trans      ✗1.0.0   -           ✗1.0.0      ✗2.0.0  
-transitive3    -        -           ✗1.0.0      ✗1.0.0  
-
-3 upgradable dependencies are locked (in pubspec.lock) to older versions.
+2 upgradable dependencies are locked (in pubspec.lock) to older versions.
 To update these dependencies, use `pub upgrade`.
 
-3  dependencies are constrained to versions that are older than a resolvable version.
+2  dependencies are constrained to versions that are older than a resolvable version.
```
jonasfj, line 297: Since we're not showing the transitive deps... I'm thinking we shouldn't count them... Which is why this is changed.

Issue #2733, opened 2020-11-06
tweaks for the pub outdated output


devoncarew:
Hi, I have some proposed tweaks to the pub outdated output:

```
Current dependencies are constrained to versions listed in Resolvable.
A resolution compatible with the Latest versions doesn't exist, as they are not
mutually compatible.
```

Above, the case is mixed (for the names of the columns). Here's a suggestion:

```
Current dependencies are constrained to versions listed in the 'Resolvable' column.

A resolution compatible with the versions listed in 'Latest' doesn't exist, as they are not
mutually compatible.
```

Also, you might want to clarify what 'they' refers to above. What are the versions in Latest not compatible with? Each other?

Perhaps, `No mutually compatible resolutions is possible with the versions listed in 'Latest'.`.

jonasfj:
> the 'Resolvable' column.

Looks like a good idea to me.

-----------------

I think it used to say:
> Dependencies are all constrained to the latest resolvable versions.
> Newer versions, while available, are not mutually compatible.

Which was also pretty good.

**Maybe we should write:**

> Dependencies are all constrained to the versions listed in the 'Resolvable' column.
> Newer versions, while available, are not obtainable by relaxing the constraints in 'pubspec.yaml'.

Then we avoid saying that they are "mutably incompatible" which may not be true if `--prereleases` is given.. It may also not be true, in the sense that the user expects, because the latest versions of the direct-dependencies may be compatible while they are incompatible with some latest transitive dependency (which we have actually seen with `vm_service`).

jonasfj:
Talk with @mit-mit, @sigurdm on GVC and reached something like:

> You are already using the newest resolvable versions listed in the 'Resolvable' column.
> Newer versions, listed in 'Latest', may not be mutually compatible.

@kwalrath any thoughts?


note:
 * we'd like to avoid saying "not mutability compatible"
 * we'd like to avoid strong wording.

sigurdm:
These messages were updated back in https://github.com/dart-lang/pub/pull/2746 .

PR #2738, opened 2020-11-09
Update outdated output

mit-mit:
Fixes https://github.com/dart-lang/pub/issues/2733

### DIFF
--- a/lib/src/command/outdated.dart
+++ b/lib/src/command/outdated.dart
@@ -536,9 +536,10 @@ Future<void> _outputHuman(
       upgradable == 0 &&
       rows.isNotEmpty) {
     log.message(
-        '\nCurrent dependencies are constrained to versions listed in Resolvable.'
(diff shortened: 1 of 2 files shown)


lib/src/command/outdated.dart
```diff
@@ -536,9 +536,10 @@ Future<void> _outputHuman(
       upgradable == 0 &&
       rows.isNotEmpty) {
     log.message(
-        '\nCurrent dependencies are constrained to versions listed in Resolvable.'
-        '\nA resolution compatible with the Latest versions doesn\'t exist, as they are not'
-        '\nmutually compatible.');
+        '\nCurrent dependencies are constrained to the versions in the \'Resolvable\' column.'
+        '\n'
+        '\nNo mutually compatible resolution is possible with the versions listed in the'
```
jonasfj, line 541: I think:
 * This is too strong a statement,
 * A resolution must always be "mutually compatible", so "no mutually compatible" is just extra words..

PR #2754, opened 2020-11-13
Enable asserts when testing pub

jonasfj:


sigurdm (approved):
LGTM

sigurdm:
We should go back t working on this. Seems like we have several hidden issues here

Issue #2756, opened 2020-11-16
Stop generating `.packages`


lrhn:
There is no need for a new SDK to generate a `.packages` file for itself. The only reason for having one is to support old third-party tools which have not started using the `.dart_tool/package_config.json` file yet.

We should identify any such tools blocking us from removing the `.packages` file, make sure they are updated, and then stop generating `.packages` files.

The sooner this is done, the better.

Would it be possible to stop generating the `.packages` file for Pub in SDK -dev builds? That should smoke out any remaining uses of the file. Maybe keep the feature available under a flag, so people really needing it can still get it that way?

jonasfj:
ironically, `pub` might be one of the tools that depends on this file :D

Or put, differently: `pub` still expects the `.packages` file to be present when testing if dependencies are up-to-date (with `pubspec.lock` and `pubspec.yaml`).

I think we could:
 1. [ ] Ensure that `pub` doesn't require the presence of `.packages` optional,
 2. [ ] Add a `--no-packages-file` flag
 3. [ ] Send out a breaking change notification that `.packages` is going away.
 4. [ ] Flip the default value of the `--no-packages-file` to `--packages-file`
 5. [ ] Hide the `--packages-file`

lrhn:
Another thing that we can do, before flipping the flag, is that every time a `pub run` of a globally activated package finds only a `.packages` file, it immediately generates a `.dart_tool/package_config.json` file too (maybe just blindly copying the data from the `.packages` file and adding language version `2.7` to every package).

mit-mit:
Hi @jonasfj we need to make progress on this; can we go ahead and add the `--no-packages-file` flag?

sigurdm:
> Hi @jonasfj we need to make progress on this; can we go ahead and add the --no-packages-file flag?

I'll start reviving the PR we have lying around.

sigurdm:
Done in https://github.com/dart-lang/pub/pull/2757

PR #2757, opened 2020-11-16
Add flag controlling creation of `.packages` file.

lrhn:
WDYT?

Related to https://github.com/dart-lang/pub/issues/2756


jonasfj:
Hmm, I think we have to update some of the code that asserts state is up-to-date.

You're suggesting, we: add a flag that allows disabling generation of `.packages` and we later flip the default for the flag.

I can see the argument for starting to remove `.packages` now that we're doing null-safety.

sigurdm:
Hmm, I think we have to update some of the code that asserts state is up-to-date.

I have landed: https://github.com/dart-lang/pub/pull/2764 that removes the assert that .packages is up to date in case it doesn't exist, it also removes several other places where we wrongly relied on .packages.

mit-mit:
@lrhn @sigurdm do we still need this?

sigurdm:
Hmm - I don't remember where we ended up last we discussed this. 

But IMO we should just continue creating this file until dart 3.
No reason to make unnecessary breaking changes before a breaking version shift.
On the other hand we could generate a comment inside the .packages file explaining that it is deprecated.

sigurdm:
I merged this with upstream and turned the flag to be off by default.

sigurdm:
ping @jonasfj

mit-mit:
@sigurdm can you clarify the flag behavoir? Is the flag `--packages-file` & off by default?

sigurdm:
> Is the flag --packages-file to create the .packages file 

yes

> it's off by default (i.e. when this lands we'll no longer create the file by default)?

yes

mit-mit:
OK. We would have to send the breaking change notice for this first then.

This is a slightly faster approach than the one discussed in https://github.com/dart-lang/pub/issues/2756 (which suggested first having a flag you must opt-in to using to skip the generation of .packages), but given how old this file is, I'm OK with moving ahead here. @lrhn WDYT?

sigurdm:
My thinking was that now we had the deprecation notice in the file we should be able to take this step.

But I'm also ok with having the .packages generation still be on by default.

lrhn:
We introduced `package_config.json` in Dart 2.8. I think almost two years should be enough warning!

lib/src/command/add.dart
```diff
@@ -92,6 +92,8 @@ class AddCommand extends PubCommand {
         help: 'Build executables in immediate dependencies.');
     argParser.addOption('directory',
         abbr: 'C', help: 'Run this in the directory <dir>.', valueHelp: 'dir');
+    argParser.addFlag('packages-file',
+        help: 'Generate the legacy ".packages" file');
```
jonasfj, line 96: Should we specify `negatable: false` ? So we don't have `--no-packages-file` option?

test/testdata/goldens/help_test/pub upgrade --help.txt
```diff
@@ -5,16 +5,18 @@ $ pub upgrade --help
...
--C, --directory=<dir>    Run this in the directory<dir>.
+-h, --help                  Print this usage information.
+    --[no-]offline          Use cached packages instead of accessing the
+                            network.
+-n, --dry-run               Report what dependencies would change but don't
+                            change any.
+    --[no-]precompile       Precompile executables in immediate dependencies.
+    --null-safety           Upgrade constraints in pubspec.yaml to null-safety
+                            versions
+    --[no-]packages-file    Generate the legacy ".packages" file
```
jonasfj, line 16: Let's just have `--packages-file`, by specifying `negatable: false`

jonasfj (approved):

lib/src/command/add.dart
sigurdm: Yeah - let's do that

test/testdata/goldens/help_test/pub upgrade --help.txt
sigurdm: Done

lib/src/command/add.dart
```diff
@@ -93,7 +93,7 @@ class AddCommand extends PubCommand {
     argParser.addOption('directory',
         abbr: 'C', help: 'Run this in the directory <dir>.', valueHelp: 'dir');
     argParser.addFlag('packages-file',
```
mit-mit, line 95: Would it make sense to cal this flag `legacy-package-file` just to be extraordinarily clear?

Issue #2758, opened 2020-11-16
If Platform.environment. containsKey('CI') – update PUB_ENVIRONMENT


kevmoo:
We want to know if pub invocations are coming from CI

Right now, we have to hard-wire configurations in each CI provider to set something.

Instead, the pub client should just sniff this environment var and use it.

See

https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
https://docs.gitlab.com/ee/ci/variables/predefined_variables.html
https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

| Provided PUB_ENVIRONMENT | Effective PUB_ENVIRONMENT |
|--------------------------|---------------------------|
| _Not set or empty_       | `CI_set`                  |
| `.some_value`            | `CI_set.some_value`       |
| `some_value`             | `CI_set.some_value`       |

Basically, the `CI_set` bit should always be first, and it should always end with `.` if there is a value after it – but don't put in two `.` if ones already there!

kevmoo:
@jonasfj – can this be ditched? We just stopped doing any telemetry if `CI` exists, right?

jonasfj:
Yes, let's close this:
https://github.com/dart-lang/pub/blob/4b5450d396a0388bf80a2990e5d21f0d6d358883/lib/src/http.dart#L88-L91

PR #2759, opened 2020-11-16
Add CI_set to X-Pub-Environment if the CI environment variable is set

kevmoo:
Allows tracking of CI environments without requiring custom code in each
environment. Setting CI=true is a convention.

Fixes https://github.com/dart-lang/pub/issues/2758

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -50,9 +50,28 @@ class _PubHttpClient extends http.BaseClient {
       request.headers['X-Pub-Command'] = PubCommand.command;
       request.headers['X-Pub-Session-ID'] = _sessionId;
 
-      var environment = Platform.environment['PUB_ENVIRONMENT'];
-      if (environment != null) {
-        request.headers['X-Pub-Environment'] = environment;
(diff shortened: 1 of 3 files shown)


kevmoo:
CC @mit-mit @athomas

athomas:
Just curious, did we give up on this idea or is there a PR/commit replacing this?

kevmoo:
@jonasfj pushed back on this. We had a discussion in chat.

natebosch:
@jonasfj - can we revisit this? Are we currently relying on hardcoding IP addresses that we know to be CI?

It would be unfortunate if we need to rely on external folks to configure their CI correctly for our metric to look right. https://github.com/dart-lang/shelf/pull/159/files#r545328401

If a `CI` variable is a pattern that exists across multiple providers it would be nice to take advantage of it.

jonasfj:
> If a CI variable is a pattern that exists across multiple providers it would be nice to take advantage of it.

I agree, filed https://github.com/dart-lang/pub/pull/2837 to just avoid sending any data when `CI=true`.

PR #2760, opened 2020-11-16
no empty environment

kevmoo:
- Add CI_set to X-Pub-Environment if the CI environment variable is set
- oops
- Ignore PUB_ENVIRONMENT that's whitespace, trim whitespace


kevmoo:
Will send for review when https://github.com/dart-lang/pub/pull/2759 is handled

Issue #2749, opened 2020-11-13
We need to handle 7-zip failures more gracefully


Hixie:
Looking at this log:
https://github.com/flutter/flutter/issues/69697#issuecomment-726406210
...it seems that we don't handle 7-zip failures well.

We should detect when 7-zip fails, and report the error. Specifically, if the failure is "access denied", we should specify what paths 7-zip tried to write to. It might even be good to detect what the failure is (e.g. look at the ACLs of the relevant directories?) and explain how to fix it (if we can do that).

Hixie:
Actually, in that bug the problems seems to be that actually running 7-zip fails, so the paths aren't really relevant there. We should detect that 7-zip has failed, abort `pub`, and report the problem as "7-zip failed to execute", including whatever information we can provide to help them fix the problem.

jonasfj:
Another reason to fix the long standing issue of doing tar in Dart https://github.com/dart-lang/pub/issues/1602

sigurdm:
But on top of that we should probably make a nicer error message if starting a sub-process fails.

sigurdm:
We are no longer using 7-zip since: https://github.com/dart-lang/pub/pull/2817

PR #2763, opened 2020-11-17
Clarify error message when spawning a subprocess fails

sigurdm:
Improves the situation on: https://github.com/dart-lang/pub/issues/2749

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -344,7 +344,7 @@ void _attempt(String description, void Function() operation) {
       var reason = getErrorReason(error);
       if (reason == null) rethrow;
 
-      log.io('Failed to $description because $reason. '
+      log.io('Pub failed to $description because $reason. '
           'Retrying in 50ms.');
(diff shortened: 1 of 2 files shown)


lib/src/io.dart
```diff
@@ -553,10 +554,15 @@ Future<PubProcessResult> runProcess(String executable, List<String> args,
...
-        environment: environment,
-        runInShell: runInShell);
+    ProcessResult result;
+    try {
+      result = await _doProcess(Process.run, executable, args,
+          workingDir: workingDir,
+          environment: environment,
+          runInShell: runInShell);
+    } on IOException catch (e) {
+      throw ApplicationException('Failed to run subprocess `$executable`: $e');
```
jonasfj, line 564: This is still an internal error...

I think we might have to tell people to file a bug, don't we?

sigurdm: Even if it is internal the stack trace won't help much.

I think the exit code has to be checked for each invocation seperately. (And I believe we do that already).

PR #2774, opened 2020-11-27
Fix pub remove --dry-run

jonasfj:
Fix case where `pub remove foo --dry-run` might update `PUB_CACHE/global_packages/`

I'm not entirely sure what happens here.. But I think think we are abusing the `Entrypoint.global`
constructor to produce the same output from `--dry-run` as you would get from mutating the
pubspec and running `Entrypoint.acquireDependenices`.

### DIFF
--- a/lib/src/command/add.dart
+++ b/lib/src/command/add.dart
@@ -134,8 +134,11 @@ class AddCommand extends PubCommand {
 
       await Entrypoint.global(newRoot, entrypoint.lockFile, cache,
               solveResult: solveResult)
-          .acquireDependencies(SolveType.GET,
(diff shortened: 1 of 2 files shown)


lib/src/command/add.dart
```diff
@@ -134,8 +134,11 @@ class AddCommand extends PubCommand {
 
       await Entrypoint.global(newRoot, entrypoint.lockFile, cache,
```
sigurdm, line 135: Could you add a TODO about not (ab)using a global entrypoint for this?

sigurdm (approved):
LGTM

PR #2778, opened 2020-11-30
Analytics

sigurdm:
Allow the embedder to provide an analytics instance, and send an analytics event about the resolution.


lib/src/command.dart
```diff
@@ -110,6 +111,8 @@ abstract class PubCommand extends Command<int> {
     return _pubEmbeddableCommand ?? (runner as PubCommandRunner);
   }
 
+  Analytics get analytics => _pubEmbeddableCommand?.analytics;
```
jonasfj, line 114: How about we make a `_NullAnalytics` and/or a `_LoggingAnalytics` implementation that just logs everything ping with log-level `fine`, I think that's useful for debugging, testing and whenever analytics is disabled.

lib/src/solver/result.dart
```diff
@@ -109,6 +122,34 @@ class SolveResult {
...
+          DependencyType.direct: 'direct',
+          DependencyType.none: 'transitive'
+        }[_root.dependencyType(package.name)];
+
+        analytics.sendEvent(
+          'pub-get',
+          package.name,
+          label: package.version.toString(),
+          value: 1,
+          parameters: {'kind': dependencyType},
```
jonasfj, line 146: Do we know what this looks like on google analytics?

I'm sure we can always do aggregations based on this using big query regardless, but it might be worth looking into.

lib/src/solver/result.dart
```diff
@@ -109,6 +122,34 @@ class SolveResult {
...
+          (runningFromTest ||
+              package.description['url'] == HostedSource.pubDevUrl)) {
+        final dependencyType = const {
+          DependencyType.dev: 'dev',
+          DependencyType.direct: 'direct',
+          DependencyType.none: 'transitive'
+        }[_root.dependencyType(package.name)];
+
+        analytics.sendEvent(
+          'pub-get',
```
jonasfj, line 142: Should this be `pub.dev hit` or `pub-hit` or `pub-package-hit` or maybe: `pub-dev-hit`, that avoids dots/space and signals that this is a `pub.dev` thing -- since we don't count hits for other package repositories.

sigurdm (commented):
>How about we make a _NullAnalytics and/or a _LoggingAnalytics implementation that just logs everything ping with log-level fine, I think that's useful for debugging, testing and whenever analytics is disabled.

I added fine logging to whenever we send analytics - this will be nice to have for user speculating what analytics are sent.

The embeeding_test has its own _LoggingAnalytics for testing.

lib/src/solver/result.dart
```diff
@@ -116,6 +130,40 @@ class SolveResult {
...
+              package.description['url'] == HostedSource.pubDevUrl)) {
+        final dependencyKind = const {
+          DependencyType.dev: 'dev',
+          DependencyType.direct: 'direct',
+          DependencyType.none: 'transitive'
+        }[_root.dependencyType(package.name)];
+        analytics.analytics.sendEvent(
+          'pub-get',
+          package.name,
+          label: package.version.toString(),
```
jonasfj, line 151: Do we need to normalize the version string encoding here?

jonasfj (approved):

jonasfj (approved):

lib/src/solver/result.dart
jonasfj: We should probably normalize using: https://github.com/dart-lang/pub_semver/pull/50

lib/src/solver/version_solver.dart
```diff
@@ -83,8 +85,7 @@ class VersionSolver {
   /// Finds a set of dependencies that match the root package's constraints, or
   /// throws an error if no such set is available.
   Future<SolveResult> solve() async {
-    var stopwatch = Stopwatch()..start();
-
+    _stopwatch.start();
```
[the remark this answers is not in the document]
sigurdm: We just need the elapsed time. Not sure how much difference it makes stopping it or not. It wasn't stopped before either...

lib/src/solver/result.dart
sigurdm: Done

PR #2780, opened 2020-12-01
Deprecate flutter constraint upper bound

sigurdm:
Ignore any flutter sdk upper bound when parsing `pubspec.yaml`s.

Warn about the flutter sdk upper bounds before publishing.

We need a better url to point to.

### DIFF
--- a/lib/src/validator/flutter_constraint.dart
+++ b/lib/src/validator/flutter_constraint.dart
@@ -14,7 +14,7 @@ class FlutterConstraintValidator extends Validator {
   FlutterConstraintValidator(Entrypoint entrypoint) : super(entrypoint);
   // TODO(sigurdm): Find a proper url for explaining this.
   static const explanationUrl =
-      'https://github.com/flutter/flutter/issues/68143';
+      'https://dart.dev/go/flutter-upper-bound-deprecation';
 
   @override
   Future validate() async {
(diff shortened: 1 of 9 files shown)


lib/src/validator/flutter_constraint.dart
```diff
@@ -0,0 +1,42 @@
...
+
+import '../entrypoint.dart';
+import '../validator.dart';
+
+/// Validates that a package's flutter constraint doesn't contain an upper bound
+class FlutterConstraintValidator extends Validator {
+  FlutterConstraintValidator(Entrypoint entrypoint) : super(entrypoint);
+  // TODO(sigurdm): Find a proper url for explaining this.
+  static const explanationUrl =
+      'https://github.com/flutter/flutter/issues/68143';
```
jonasfj, line 17: Create a `dart.dev/go/flutter-upper-bound-deprecation` in:
https://github.com/dart-lang/site-www/blob/6e22397cad4de94c935290880dc273321b35489c/firebase.json#L135

I would suggest just linking to the issue, or something (anything), so that we just don't forget to create the link and redirect it somewhere :D

jonasfj (approved):

lib/src/validator/flutter_constraint.dart
sigurdm: Done

sigurdm:
The travis failure is unrelated. It is being fixed in: https://github.com/dart-lang/pub/pull/2803

Issue #2786, opened 2020-12-01
Ignoring Files when Publishing - gitignore doesn't work 


TheHighriser:
I have searched through the issues already and I don't find anything. Either I am doing something completely wrong or this wasn't used by anyone.

I am trying to exclude certain things from publishing. So e.g. I am trying to exclude the folder "pleaseIgnoreMe/". So I will enter "pleaseIgnoreMe/" into the .gitignore file. When doing a publish --dry-run, I can still see the folder "pleaseIgnoreMe". I then tried to do a normal publish, because I though maybe this is not working for --dry-run and exclude a folder which was not needed and it still is not working. The published artifact has "pleaseIgnoreMe".

The documentation says the only way is the .gitignore file (https://dart.dev/tools/pub/publishing#what-files-are-published). Is this working for anybody?

sigurdm:
Yeah this seems broken. Played a bit with this.
I think the .gitignore is only respected for files not already checked in.

we should be able to use `git check-ignore --no-index`

sigurdm:
After thinking more about this - we cannot use the .gitignore file for this. If a file is already checked in, then .gitignore should not count for that. We should not change the semantics of that. For example if a user has ignored '*.tmp' files in .gitignore, but has checked a single 'test.tmp' file in to the test suite, it would be surprising that pub does not publish that file.

Therefore I started working on a pubignore file that will allow ignoring files for publishing only: https://github.com/dart-lang/pub/pull/2787.

Issue #2566, opened 2020-07-17
Can we ignore some files for publishing but include in the repo?


vincevargadev:
I'd like to leave some files in a package's repository on GitHub, but at the same time, make those files excluded when I publish a package on pub.dev.

The issue is that I'd like to add images or gifs to the README so that other developers immediately understand what the package is for. However, I don't want users of my package to have to download these images when they decide to add my package as a dependency.

Can we ignore some files for publishing but include in the repo?

sigurdm:
I think we should have a .pubignore or similar file.
@jonasfj what do you think?

jonasfj:
We already respect `.gitignore` I think, but I can see how that doesn't work here.

Other possible use-case in https://github.com/dart-lang/pub/issues/2222.

I imagine that if you have a `.pubignore` file, then we would not honor the `.gitignore` file.

sigurdm:
It seems we can use `git ls-files -X .pubignore --cached --others` to interpret a `.pubignore` file.

That would however put a requirement on git - which might be acceptable if we still allow having no .pubignore file.

jonasfj:
> That would however put a requirement on git - which might be acceptable if we still allow having no .pubignore file.

External dependencies provided by the system, such as `tar` haven't exactly worked out well for us before, hehe

I suppose it works better for `.gitignore`, but ideally we would write this logic in Dart.

jonasfj:
On the other hand, we can test if `git` is present, and we already use it.. so this might not be so bad..

sigurdm:
>  ideally we would write this logic in Dart.

It is quite well specified: https://git-scm.com/docs/gitignore , but not trivial to implement.

Funnily enough we hardly ever get issues around our external dependency on git. I guess there is only one implementation that everybody shares.

Most of our tar issues seem to stem from the system tar being different for different systems.

softmarshmallow:
Is there any plans on supporting .pubignore or ingore: ["", ""] in pubspec?
the suggested ways are just temporary workaround

jonasfj:
> Is there any plans on supporting .pubignore or ingore: ["", ""] in pubspec?

I don't think it's a high priority for us at the moment. But if someone wants to work on it, I'll be happy to help define the feature, review and land it.

Otherwise, please upvote, and mention any novel use-cases you might have.

Issue #2222, opened 2019-10-07
Publish an ignored file to Pub


dmdashenkov:
I'm trying to publish my first Pub package.

[The doc](https://dart.dev/tools/pub/publishing#what-files-are-published) states that the files under `.gitignore` are not published.

My question is: is there a way to work around the `.gitignore` and publish the files which are not on Git?

Background: I'm working with code generation and generating Dart from Protobuf. It's not ideal to store generated sources in Git (it looks ugly and distorts the viewers from the real source code which implements some logic).

I can always copy the required files into a separate directory and publish from there instead. However, it'be nice to avoid that.

However, the coupling between Git and Pub seems a bit odd. I'm sure that a reverse situation (when something is on Git and should now be published to Pub) is even more common.

jonasfj:
There is no workaround at this point. My suggestion would be to check the generated code into source control.

It's ugly, but it also means that you only have to re-generate the files when the input change. It makes it easier for new developers joining your project.

That said, this is something we could solve with a `.pubignore` file or an `ignores` section in `pubspec.yaml` which overrides the behavior of `.gitignore`. But I have mixed feelings about that.

dmdashenkov:
@jonasfj, thanks for the suggestion.

For now, I'll stick to the workaround of copying the required source files into a separate directory and publishing from there. With a little scripting, it's a pretty robust approach.

Though, if this feature is implemented in future, I will most certainly use it.

natebosch:
Note that within the Dart team we always track all published files in git. We also tag commits with the version number when we publish. This means that we can visit the git repo and see exactly what is published for a given version without downloading the archiver from pub to see the source files.

jonasfj:
@natebosch, yeah, I think that's a good answer to this problem.
Probably we could even score packages that match a tag in github higher, though we'd need to solve mono-repo resolution in the process :)

mveroukis:
Really? .gitignore should be just for git, not for build or publish tools. So I guess another solution is to just not use git at all. Honestly, this is terrible. There should be a .publishignore file or something, and let the developers do things their way. Shit like this that really makes me hate software development.

sigurdm:
@mveroukis please update your message to use kinder language. Even though you are frustrated we need to keep a sober tone.

sigurdm:
Also to your question there is now the possibility to put a .pubignore file in folders which override the .gitignore file in the same folder for pubs consideration.

PR #2787, opened 2020-12-04
Introduce .pubignore

sigurdm:
Implements .gitignore style files called `.pubignore`.

Files ignored by a `.pubignore` will not be published (no matter if they are checked in to git).

A `.pubignore` in a folder will override a `.gitignore` if both are present in the same folder.

pub will now no longer call out to git to have .gitignore files interpreted.
This results in some subtle changes in the package listings:

* .gitignore files are taken into account even if the root is not a git directory.
* global gitignore files are not taken into account.
* files that are ignored by a .gitignore but checked into git, would be listed by pub previously, but now they are ignored by pub.
  - Added a publish-time validator to warn against this case (it is most likely a mistake when this happens, and can always be avoided by writing more .gitignore rules).

Changed package.listFiles to always take ignore files into account - no usecases warranted the difference.

Most implementation of `ignore.dart` by jonasfj@

fixes: https://github.com/dart-lang/pub/issues/2222 
fixes: https://github.com/dart-lang/pub/issues/2786
fixes: https://github.com/dart-lang/pub/issues/2566


lib/src/ignore.dart
```diff
@@ -0,0 +1,321 @@
...
+      first = false;
+    }
+  }
+
+  var relativeToPath = false;
+  var matchesDirectoriesOnly = false;
+
+  // slashes have different significance depending on where they are in
+  // the String. Handle that here.
+  // Returns the regexp to match the just parsed slash.
```
jonasfj, line 234: It doesn't return anything

lib/src/ignore.dart
```diff
@@ -0,0 +1,321 @@
...
+        if (peekChar() == '/') {
+          current++;
+          if (current == end) {
+            expr += '.*/';
+          } else {
+            // Match nothing or a path followed by '/'
+            expr += '(?:(?:)|(?:.*/))';
+          }
+          // Handle the side effects of seeing a slash.
+          // We know it was not initial, so the return value is ignored.
```
jonasfj, line 263: HandleSlash has no return value.

lib/src/ignore.dart
```diff
@@ -0,0 +1,321 @@
...
+        expr += '[^/]*';
+      }
+    } else if (nextChar == '?') {
+      // Handle '?'
+      expr += '[^/]';
+    } else if (nextChar == '[') {
+      // Character ranges
+      final characterRange = parseCharacterRange();
+      if (characterRange == null) {
+        print('c');
```
jonasfj, line 279: Debug code?

lib/src/ignore.dart
```diff
@@ -0,0 +1,321 @@
...
+        return null;
+      }
+      expr += '[$characterRange]';
+    } else if (nextChar == '\\') {
+      // Escapes
+      final escaped = peekChar();
+      if (escaped == null) {
+        print('d');
+        return null;
```
jonasfj, line 288: I have contemplated if we should throw a FormatException and then later up just discard those... But then we could have a pedantic parse function we can use to validate the patterns... Which might be nice in some cases.

lib/src/ignore.dart
sigurdm: Oops, I changed strategy here!

lib/src/ignore.dart
sigurdm: Yeah - Thought about the same.

I'll do that.

lib/src/ignore.dart
sigurdm: Yep!

lib/src/ignore.dart
sigurdm: I opted for returning a ParseResult that tells whether the pattern was parsed. And then taking a handler.

jonasfj:
We had some discussions, consider the following options:
(A) `.pubignore` in any folder (same syntax as `.gitignore`).
(B) `files: ['glob/pattern/for/**/file.dart', ...]` in `pubspec.yaml` (if present, only files matching a pattern is included)
(C) `include: ['glob/pattern/for/**/file.dart', ...], exclude: ['glob/pattern/for/**/file.dart', ...]` in `pubspec.yaml`
(D) `exclude: ['glob/pattern/for/**/file.dart', ...]` in `pubspec.yaml`

I'm personally attracted to (A) because:
 * It's the same syntax as `.gitignore`,
 * Any glob-patterns we invent won't be as expressive as `.gitignore` (e.g. `/` has special meaning in the start of a pattern),
 * The `.pubignore` file itself should not be included when publishing the package: 
    * If user is ignoring private files they don't want to share, they might not want to share the file-patterns they are ignoring,
    * This is information that isn't needed in the package,
    * Options (B), (C), (D) needlessly increases the pubspec, which increases the size of all responses from the list-versions API.
 * Like a `.gitignore` file, you can have a multiple `.pubignore` files, you don't need to put in the root directory. If you are generating some temporary files, you can simply generate a `.pubignore` file in said directory too to avoid publishing any of them.

@sigurdm, @isoos thoughts?

sigurdm:
Yeah - the main complaint about .pubignore files is that we now have even more ("hidden") files floating around.

But for most users pubignore will not be needed.

lib/src/ignore.dart
sigurdm: fixed

lib/src/ignore.dart
```diff
@@ -0,0 +1,415 @@
...
+        onInvalidPattern(invalidResult.pattern, invalidResult.exception);
+      }
+    }
+    return MapEntry(directory,
+        parseResults.where((r) => !r.empty).map((r) => r.rule).toList());
+  });
+}
+
+GitIgnoreParseResult parseIgnorePattern(String pattern, bool ignoreCase,
+    {source}) {
```
jonasfj, line 250: Should we add some typing to `source` or pass it along when we call from `parseIgnorePatterns`.

jonasfj (approved):

lib/src/ignore.dart
```diff
@@ -0,0 +1,415 @@
...
+  /// must end in slash (`/`) if it is directory.
+  ///
+  /// **Example**:
+  /// ```dart
+  /// import 'package:ignore/ignore.dart';
+  ///
+  /// void main() {
+  ///   final ignore = Ignore([
+  ///     '*.o',
+  ///   ]);
```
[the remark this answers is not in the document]
sigurdm: Now it works again!

lib/src/ignore.dart
sigurdm: I removed source. It can just be the pattern itself.

lib/src/command/lish.dart
```diff
@@ -137,13 +137,18 @@ the \$PUB_HOSTED_URL environment variable.''',
...
 
     // Show the package contents so the user can verify they look OK.
     var package = entrypoint.root;
     log.message('Publishing ${package.name} ${package.version} to $server:\n'
         '${tree.fromFiles(files, baseDir: entrypoint.root.dir)}');
+    // Improve discoverability of the .pubignore file.
+    if (entrypoint.root.pubignores.isEmpty) {
+      log.message('''
+(To exclude files from being published, consider creating a `.pubignore`)''');
+    }
```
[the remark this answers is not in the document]
sigurdm: Ok - removed it... I still think discoverability might make it worthwhile to print something here...

lib/src/package.dart
```diff
@@ -190,26 +193,66 @@ class Package {
...
+          .toList();
+
+  Ignore _pubignores;
+
+  /// Returns an [Ignore] object representing all the [pubignoreFiles].
+  Ignore get pubignores {
+    if (_pubignores != null) {
+      return _pubignores;
+    }
+    final mapping = <String, List<String>>{};
```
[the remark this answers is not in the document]
sigurdm: No longer here

lib/src/package.dart
```diff
@@ -190,26 +193,66 @@ class Package {
...
+      final relativePath = relative.substring(
+        0,
+        max(0, relative.length - _pubignorePathEnding.length),
+      );
+      mapping[relativePath] = [readTextFile(pubignore)];
+    }
+    return _pubignores = Ignore(
+      mapping,
+      onInvalidPattern: (pattern, exception) {
+        log.warning('`pubignore` had invalid pattern: ${exception.message}');
```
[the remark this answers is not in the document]
sigurdm: Yes!

lib/src/ignore.dart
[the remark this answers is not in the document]
sigurdm: Now it does!

lib/src/ignore.dart
```diff
@@ -0,0 +1,415 @@
...
+        final result = testRulesForPrefix(path.substring(0, index));
+        if (result != null) return result;
+      }
+      return testRulesForPrefix('') == true;
+    }
+
+    for (var slashIndex = path.indexOf('/');
+        slashIndex != -1;
+        slashIndex = path.indexOf('/', slashIndex + 1)) {
+      if (ignoresSubpath(path.substring(0, slashIndex + 1))) {
```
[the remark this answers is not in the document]
sigurdm: No longer relevant... I think...

lib/src/ignore.dart
```diff
@@ -0,0 +1,415 @@
...
+        // If a folder above [path] is ignored the pattern cannot be un-ignored.
+        return true;
+      }
+    }
+    return ignoresSubpath(path);
+  }
+}
+
+class GitIgnoreParseResult {
```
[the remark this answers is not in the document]
sigurdm: Nope - made private.

lib/src/ignore.dart
```diff
@@ -0,0 +1,415 @@
...
+  final FormatException exception;
+
+  GitIgnoreParseResult(this.pattern, this.rule) : exception = null;
+  GitIgnoreParseResult.invalid(this.pattern, this.exception) : rule = null;
+  GitIgnoreParseResult.empty(this.pattern)
+      : rule = null,
+        exception = null;
+}
+
+class GitIgnoreRule {
```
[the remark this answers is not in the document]
sigurdm: Made private.

lib/src/ignore.dart
sigurdm: Removed the source. It can just be the pattern itself...

lib/src/ignore.dart
```diff
@@ -0,0 +1,560 @@
...
+  ///     }),
+  ///     ignoreForDir: (dir) {
+  ///       final f = File(resolve('dir/.gitignore'));
+  ///       return f.existsSync() ? Ignore([f.readAsStringSync()]) : null;
+  ///     },
+  ///     isDir: (dir) => Directory(resolve(dir)).existsSync(),
+  ///   ).forEach(print);
+  /// }
+  /// ```
+  static List<String> unignoredFiles({
```
jonasfj, line 215: unignored implies that there is a negative ignore pattern, doesn't it?

Why about just `listFiles`? or `listFilesNotIgnored` or `filesNotIgnored`..

jonasfj (approved):

lib/src/package.dart
```diff
@@ -202,120 +205,71 @@ class Package {
...
+                },
+              );
+      },
+      isDir: (dir) {
+        final resolved = resolve(dir);
+        final isDir = dirExists(resolved);
+        if (isDir) {
+          final canonicalized = resolveSymlinksOfDir(resolved);
+          // Avoid cycles by only visiting each location once.
+          return visitedDirs.add(canonicalized);
```
jonasfj, line 267: If we say `isDir` false won't `createTarGz` try to add it to the to a tarball with `file.openRead()`?
https://github.com/dart-lang/pub/blob/11e7b2ce977cb27b4a508963ec1ace01a56bb6e1/lib/src/io.dart#L922-L940

Also won't this mean that in:
```
/folder/
  hello.txt
/linkA -> ../folder/
/linkB -> ../folder/
```
We'll get files as:
* `[/folder/hello.txt]`, OR,
* `[/linkA/hello.txt]`, OR,
* `[/linkB/hello.txt]`.

depending on which branch we visit first.

-------

Hmm, this is going to be rather complicated to support in any predictable manner.. Maybe we should disallow symlinks to folders?

It's easy to implement something where we only follow a symlink the first time we see it.. but the outcome from such a scan easily becomes rather complicated.

The only predictable options I see are:
 * (A) don't allow symlinks to directories,
 * (B) don't allow nested symlinks (even acyclic ones) -- simply throw an error if we see the same symlink more than once.
 * (C) don't allow symlink cycles -- throw an error if we see a symlink cycle.

(C) will allow a scenario like:
```
/folder/
  hello.txt
/linkA -> folder/
/linkB -> ../linkA/
/linkC -> ../linkA/
```
But then we need to track all the symlinks seen in the current stack, and throw an error if we see the same symlink twice.. and in this case we would then include the file `hello.txt` in the tarball 4 times!

So maybe (A) or (B) isn't so bad, though (B) seems like an arbitrary level of support for symlinks.

sigurdm: You are right - it all becomes dodgy when we do this...

I would like to support 'simple' use-cases, such as sharing a directory between two packages in a mono-repo... but perhaps it is not worth the complexity.

sigurdm: Forbidding directory symlinks for now (option A). We can add it later if the pressure grows for this.

jonasfj (approved):
Let's pull this into dart-sdk ASAP, so we can get some testing in dev releases..

jonasfj (approved):

biendltb:
> * It's the same syntax as `.gitignore`,

Even though this was mentioned, the negated pattern match does not work for `.pubignore`. For example, I don't want to publish the whole `example` folder in my package, but I want to keep its `main.dart` for pub.dev's example section:
```
example/
!example/lib/main.dart
```

Hope this negated pattern match will be added soon.

biendltb:
@sigurdm thanks for pointing out this fine nature of git/pub-ignore. Although this workaround is still cumbersome to unignore files in deeper layers, it helps me strip a bunch of lines off in my pubignore for my case. Cheers! 🙌

PR #2789, opened 2020-12-08
fix tryGitCommand

jonasfj:
I'm not sure why the `git_not_installed.dart` test didn't fail before.


sigurdm (approved):
We need to rename the test

Issue #2793, opened 2020-12-08
Crash on upgrade with pinned dependency_override


kevmoo:
```
pub --trace upgrade
Resolving dependencies...
type '_EmptyVersion' is not a subtype of type 'VersionRange'
package:pub/src/pubspec.dart 619:9                                 Pubspec._parseDependencies.<fn>
dart:collection                                                    MapView.forEach
package:pub/src/pubspec.dart 601:15                                Pubspec._parseDependencies
package:pub/src/pubspec.dart 165:9                                 Pubspec.dependencies
package:pub/src/solver/package_lister.dart 211:35                  PackageLister.incompatibilitiesFor
===== asynchronous gap ===========================
dart:async                                                         Future.catchError
package:pub/src/utils.dart 113:52                                  captureErrors.wrappedCallback
package:stack_trace                                                Chain.capture
package:pub/src/utils.dart 126:11                                  captureErrors
package:pub/src/command.dart 164:13                                PubCommand.run
package:args/command_runner.dart 197:27                            CommandRunner.runCommand
package:pub/src/command_runner.dart 147:24                         PubCommandRunner.runCommand
package:pub/src/command_runner.dart 136:18                         PubCommandRunner.run
/b/s/w/ir/cache/builder/sdk/third_party/pkg/pub/bin/pub.dart 9:48  main
---- Log transcript ----
FINE: Pub 2.12.0-114.0.dev
SLVR: fact: checked_yaml is 1.0.4
SLVR: derived: checked_yaml
SLVR: Version solving took 0:00:00.020760 seconds.
    | Tried 1 solutions.
FINE: Resolving dependencies finished (0.0s).
ERR : type '_EmptyVersion' is not a subtype of type 'VersionRange'
FINE: Exception type: _TypeError
ERR : package:pub/src/pubspec.dart 619:9                                 Pubspec._parseDependencies.<fn>
    | dart:collection                                                    MapView.forEach
    | package:pub/src/pubspec.dart 601:15                                Pubspec._parseDependencies
    | package:pub/src/pubspec.dart 165:9                                 Pubspec.dependencies
    | package:pub/src/solver/package_lister.dart 211:35                  PackageLister.incompatibilitiesFor
    | ===== asynchronous gap ===========================
    | dart:async                                                         Future.catchError
    | package:pub/src/utils.dart 113:52                                  captureErrors.wrappedCallback
    | package:stack_trace                                                Chain.capture
[...]

PR #2796, opened 2020-12-09
Handle empty version constraint

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2793


jonasfj (approved):

sigurdm:
We should consider:
```
  strong-mode:
    implicit-casts: false
    implicit-dynamic: false
```
However that will require many changes

PR #2799, opened 2020-12-10
Minor refactoring of BoundSource to reduce risk of mistakes

jonasfj:
I'm not sure about `BoundSourceBase`, maybe we should make it go away completely.
Or change it into a mix-in or something else.. I just really liked using
`@visibleForOverridding`, `@protected` and `@nonVirtual` because they protect us from making mistakes.

I'm also not sure we need to use `BoundSourceBase` for `_BoundUnknownSource`, but that
might be something we can cleanup later.

### DIFF
--- a/lib/src/source.dart
+++ b/lib/src/source.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:pub_semver/pub_semver.dart';
+import 'package:meta/meta.dart';
 
 import 'exceptions.dart';
 import 'package_name.dart';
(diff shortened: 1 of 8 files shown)


lib/src/source.dart
```diff
@@ -180,21 +205,14 @@ abstract class BoundSource {
   ///
   /// This method is effectively protected: subclasses must implement it, but
   /// external code should not call this. Instead, call [getVersions].
+  @visibleForOverriding
   Future<List<PackageId>> doGetVersions(PackageRef ref);
 
   /// A cache of pubspecs described by [describe].
   final _pubspecs = <PackageId, Pubspec>{};
```
sigurdm, line 212: I would love to move this to the `SystemCache` package. We don't need a separate map of these for each Source.

sigurdm:
https://github.com/dart-lang/pub/pull/3305

Issue #2791, opened 2020-12-08
no-op `pub global activate <package>` causes latency


kenzieschmoll:
In flutter_tools/.../devtools_launcher.dart, we need to run `pub global run devtools`. At the same time, we need to make sure the devtools package is globally activated, and we need to ensure that, if it is already activated, the latest available version is in use. See https://github.com/flutter/flutter/pull/71737/files#diff-640de6fc5df0761e2783c73b6477083e9f1eeba8163f101ee0c795e0ac7e5564R50 for code and relevant discussion.

AFAIK, the only way to ensure the latest available version of devtools is activated is to run `pub global activate devtools` again. The issue is that this is creating ~4.5 seconds of latency, even when the call is a no-op (when the latest published version of package:devtools was already activated).

2 questions:
1) is there a different way to achieve the requirements of our use case (ensure devtools is activated and that the latest version is in use)? We would need a way to check the version of a globally activated package.
2) can the latency from pub be improved when the latest version of a package is already active?

sigurdm:
Hmm - I guess we could 

* make a check is the discovered version is the same as the one currently being installed.
* Have a `--no-precompile` flag as we have for `pub get`. That would make precompilation only happen when run. 

More hacky ideas for workarounds:
As mentioned in the linked issue we have `pub global list` I guess something could be hacked up on the commandline.
The latest version of a package can be queried from pub.dev with:
```curl https://pub.dev/api/packages/devtools | jq '.["latest"]["version"]'```

As a workaround: Is there any chance you could use dev_tools as a dev_dependency? That way you can use `pub get` and take advantage of the (default) --no-precompile.
Then you would run it with `pub run devtools`

kenzieschmoll:
> make a check is the discovered version is the same as the one currently being `installed.`

IMO, this seems like the most robust solution. Running `pub global activate <package>` should check that the package is outdated and return early if not, avoiding the heavy work of doing an additional install.

> As mentioned in the linked issue we have pub global list I guess something could be hacked up on the commandline.
> The latest version of a package can be queried from pub.dev with:
> curl https://pub.dev/api/packages/devtools | jq '.["latest"]["version"]'

Running `curl https://pub.dev/api/packages/devtools | jq '.["latest"]["version"]'` gives me an error because `jq` is not installed on my machine. This seems problematic in case a user also does not have `jq` installed.

[...]

PR #2810, opened 2020-12-21
Don't reprecompile if current global installed was same version

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2791

Note that we re-precompile if anything in the resolution changed.
There will still be some time spent for doing version resolution, but I suspect that is unavoidable.

This makes it slightly harder to force a recompile in case something goes wrong. But there's still 'deactivate followed by activate'.

This is a bit hacky.
Ideally we would like to align repeated `pub global activate` with `pub upgrade --precompile` and have the no-op detection in a common path. That will probably be a bigger change though.

### DIFF
--- a/lib/src/global_packages.dart
+++ b/lib/src/global_packages.dart
@@ -88,11 +88,6 @@ class GlobalPackages {
   /// when needed.
   GlobalPackages(this.cache);
 
-  Future<void> activate(SourceDescriptor sourceDescriptor,
-      {List<String> executables,
-      Map<String, FeatureDependency> features,
-      bool overwriteBinStubs}) async {}
-
   /// Caches the package located in the Git repository [repo] and makes it the
   /// active global version.
   ///
--- a/lib/src/global_packages.dart
+++ b/lib/src/global_packages.dart
@@ -29,24 +29,6 @@ import 'source/path.dart';
 import 'system_cache.dart';
 import 'utils.dart';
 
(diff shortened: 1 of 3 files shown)


lib/src/global_packages.dart
```diff
@@ -29,6 +29,24 @@ import 'source/path.dart';
...
+class HostedSourceDescriptor implements SourceDescriptor {
+  final String name;
+  final VersionConstraint constraint;
+  HostedSourceDescriptor(this.name, this.constraint);
+}
+
+class PathSourceDescriptor implements SourceDescriptor {
+  final String path;
+  PathSourceDescriptor(this.path);
+}
```
jonasfj, line 48: Any documentation comments, what are these?

lib/src/global_packages.dart
```diff
@@ -70,6 +88,11 @@ class GlobalPackages {
   /// when needed.
   GlobalPackages(this.cache);
 
+  Future<void> activate(SourceDescriptor sourceDescriptor,
+      {List<String> executables,
+      Map<String, FeatureDependency> features,
+      bool overwriteBinStubs}) async {}
```
jonasfj, line 94: Left over?

sigurdm: Yep ! removing!

jonasfj (approved):

PR #2812, opened 2020-12-21
Inline package:http_throttle

natebosch:
The package has been discontinued and will no longer be published. Copy
the one class defined into the package into the single usage site.


sigurdm (approved):
LGTM

Thanks!

PR #2817, opened 2020-12-27
Use Dart library to read and write tar files

### LINKED ISSUE #1602: Use a Dart-only library for archives instead of the 'tar' executable.
Calling the 'tar' executable makes pub vulnerable to various issues, dealing it with a Dart-only library makes it safer. This library seems to be able to deal with it adequately: https://pub.dartlang.org/packages/archive

See also: https://github.com/dart-lang/pub-dartlang-dart/issues/142

simolus3:
Uses `pkg:tar` to read and write tar files. That package is based on stream transformers and generally tries to minimize memory overhead. It supports the ustar format and extended PAX headers.

For testing, I ran `tool/extract_all_pub_dev.dart`, but only with the latest version of each package. To test backwards-compatibility, I published some packages with `pkg:tar` to a personal pub server and downloaded them with the regular pub on Linux and Windows. I don't have access to a Mac, but I can run more tests on the other OSes if needed.

Full disclosure: I wrote the tar package. 
Closes #1602 (and a bunch of related issues).

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -850,7 +850,8 @@ Future extractTarGz(Stream<List<int>> stream, String destination) async {
         }
 
         ensureDir(parentDirectory);
-        createSymlink(path.relative(resolvedTarget, from: destination), filePath);
+        createSymlink(
+            path.relative(resolvedTarget, from: parentDirectory), filePath);
         break;
       case TypeFlag.link:
         // We generate hardlinks as symlinks too, but their path needs to be
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -824,18 +824,20 @@ ByteStream createTarGz(List<String> contents, {String baseDir}) {
   baseDir = path.absolute(baseDir);
 
   final tarContents = Stream.fromIterable(contents.map((entry) {
-    entry = path.absolute(path.normalize(entry));
+    entry = path.absolute(entry);
     if (!path.isWithin(baseDir, entry)) {
       throw ArgumentError('Entry $entry is not inside $baseDir.');
     }
 
     final relative = path.relative(entry, from: baseDir);
-    final file = File(entry);
+    // On Windows, we can't open some files without normalizing them
+    final file = File(path.normalize(entry));
     final stat = file.statSync();
 
     return tar.Entry(
       tar.Header(
-        name: relative,
+        // Ensure paths in tar files use forward slashes
+        name: path.url.joinAll(path.split(relative)),
         mode: stat.mode,
         size: stat.size,
         lastModified: stat.changed,
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -8,7 +8,6 @@ import 'dart:collection';
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:async/async.dart';
 import 'package:http/http.dart' show ByteStream;
 import 'package:http_multi_server/http_multi_server.dart';
 import 'package:meta/meta.dart';
@@ -16,12 +15,12 @@ import 'package:path/path.dart' as path;
 import 'package:pedantic/pedantic.dart';
 import 'package:pool/pool.dart';
 import 'package:stack_trace/stack_trace.dart';
+import 'package:tar/tar.dart' as tar;
 
 import 'error_group.dart';
 import 'exceptions.dart';
 import 'exit_codes.dart' as exit_codes;
 import 'log.dart' as log;
-import 'sdk.dart';
 import 'utils.dart';
 
 export 'package:http/http.dart' show ByteStream;
@@ -788,102 +787,25 @@ Future<HttpServer> bindServer(String host, int port) async {
   return server;
 }
 
-String _tarPath = _findTarPath();
-
-/// Find a tar. Prefering system installed tar.
-///
-/// On linux tar should always be /bin/tar [See FHS 2.3][1]
-/// On MacOS it seems to always be /usr/bin/tar.
-///
-/// [1]: https://refspecs.linuxfoundation.org/FHS_2.3/fhs-2.3.pdf
-String _findTarPath() {
-  for (final file in ['/bin/tar', '/usr/bin/tar']) {
-    if (fileExists(file)) {
-      return file;
-    }
-  }
-  log.warning(
-      'Could not find a system `tar` installed in /bin/tar or /usr/bin/tar, '
-      'attempting to use tar from PATH');
-  return 'tar';
-}
-
 /// Extracts a `.tar.gz` file from [stream] to [destination].
 Future extractTarGz(Stream<List<int>> stream, String destination) async {
   log.fine('Extracting .tar.gz stream to $destination.');
-  final decompressed = stream.transform(GZipCodec().decoder);
+  final entries = stream.transform(gzip.decoder).transform(tar.reader);
 
-  // We used to stream directly to `tar`,  but that was fragile in certain
-  // settings.
-  final processResult = await withTempDir((tempDir) async {
-    final tarFile = path.join(tempDir, 'archive.tar');
-    try {
-      await _createFileFromStream(decompressed, tarFile);
-    } catch (e) {
-      // We don't know the error type here: https://dartbug.com/41270
-      throw FileSystemException('Could not decompress gz stream $e');
-    }
-    return (Platform.isWindows)
-        ? runProcess(_pathTo7zip, ['x', tarFile], workingDir: destination)
-        : runProcess(_tarPath, [
-            if (_noUnknownKeyword) '--warning=no-unknown-keyword',
-            '--extract',
-            '--no-same-owner',
-            '--no-same-permissions',
-            '--directory',
-            destination,
-            '--file',
-            tarFile,
-          ]);
-  });
-  if (processResult.exitCode != exit_codes.SUCCESS) {
-    throw FileSystemException(
-        'Could not un-tar (exit code ${processResult.exitCode}). Error:\n'
-        '${processResult.stdout.join("\n")}\n'
-        '${processResult.stderr.join("\n")}');
-  }
-  log.fine('Extracted .tar.gz to $destination. Exit code $exitCode.');
-}
+  destination = path.absolute(destination);
+  await for (final entry in entries) {
+    if (entry.type != tar.FileType.regular) continue;
 
-/// Whether to include "--warning=no-unknown-keyword" when invoking tar.
-///
-/// BSD tar (the default on OS X) can insert strange headers to a tarfile that
-/// GNU tar (the default on Linux) is unable to understand. This will cause GNU
-/// tar to emit a number of harmless but scary-looking warnings which are
-/// silenced by this flag.
-final bool _noUnknownKeyword = _computeNoUnknownKeyword();
-bool _computeNoUnknownKeyword() {
-  if (!Platform.isLinux) return false;
-  var result = Process.runSync(_tarPath, ['--version']);
-  if (result.exitCode != 0) {
-    throw ApplicationException(
-        'Pub failed to run tar (exit code ${result.exitCode}):\n${result.stderr}');
+    final filePath = path.join(destination, entry.name);
+    final parentDirectory = path.dirname(filePath);
+
+    deleteIfLink(filePath);
+    ensureDir(parentDirectory);
+    await _createFileFromStream(entry, filePath);
   }
 
-  var match =
-      RegExp(r'^tar \(GNU tar\) (\d+).(\d+)\n').firstMatch(result.stdout);
-  if (match == null) return false;
-
-  var major = int.parse(match[1]);
-  var minor = int.parse(match[2]);
-  return major >= 2 || (major == 1 && minor >= 23);
+  log.fine('Extracted .tar.gz to $destination.');
 }
-
-final String _pathTo7zip = (() {
-  final candidate = runningFromDartRepo
-      ? path.join(dartRepoRoot, 'third_party', '7zip', '7za.exe')
-      : path.join(
-          sdk.rootDirectory,
-          'lib',
-          '_internal',
-          'pub',
-          'asset',
-          '7zip',
-          '7za.exe',
-        );
-  if (fileExists(candidate)) return candidate;
-  throw StateError('Could not find 7zip.');
-})();
 
 /// Create a .tar.gz archive from a list of entries.
 ///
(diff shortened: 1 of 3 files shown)


jonasfj:
@simolus3, this is AWESOME! :rocket: 

@walnutdust has some work in progress in a PR with a tar-stream reader like this in https://github.com/google/dart-neats/pull/56
But we never quite finished it.. It modelled on the tar implementation from golang (not a literal translation, but tried to follow what [`archive/tar`](https://golang.org/pkg/archive/tar/) does). I think one of the bugs we've had there was a broken `ChunkedStreamIterator` which was recently fixed :D

I wonder if we could combine efforts?
 * We should at a minimum be able to reuse the test cases :tada:  
 * The approach using `ChunkedStreamIterator` is perhaps a bit more readable, so it might be worth exploring..
 * In either case, both of these implementations might need more tests wrt. special characters and the like :see_no_evil:

@walnutdust, do you recall why we went away from making transforming a `Stream<List<int>>` into a `Stream<TarEntry>`? Was it to do with cancelling or was it just because it would be harder to use?

simolus3:
I didn't know about the existing effort (and sorry for snatching the package name :D).

 > The approach using `ChunkedStreamIterator` is perhaps a bit more readable

That's true, but I like that `tar.reader` can be a `StreamTransformer` transparently handling pauses/resumes. I took the approach from [package:mime](https://github.com/dart-lang/mime/blob/master/lib/src/bound_multipart_stream.dart), IMO this is pretty convenient to use in an `await for` loop. I have some ideas to make the implementation more readable too.

I'm not sure what the best way forward is now. I can open a PR to add my writer to @walnutdust's implementation and transfer the pub package to you when that gets merged. I need a tar package that works without `dart:io` though. I also want to use it for untrusted inputs, so some guards against memory DoS attacks with huge PAX headers would be nice.
I can also continue to work on my tar implementation and add the test cases you suggested. It has the features I need, but the reader doesn't deal with sparse files or PAX Headers for modification times yet. I could copy that code from @walnutdust and give credit if that's ok.
I slightly prefer the second approach, but I'm happy to help with the other implementation too if that's easier for you.

walnutdust:
@jonasfj 

> do you recall why we went away from making transforming a Stream<List<int>> into a Stream<TarEntry>? Was it to do with cancelling or was it just because it would be harder to use?

This might have been because we were trying to move away from `StreamControllers` to async generators, and with async generators the next loop iteration is run again before the first one necessary conflicts. @simolus3 found a great solution through sync `StreamController`, and it looks like it probably works well.

@simolus3 I think making `tar.reader` be a `StreamTransformer` is pretty cool, and the additional concerns you have are valid! I would encourage possibly having more tests with regards to ensuring the transformed stream has the expected behavior even when passed on to other stream users (e.g. `StreamIterator`). We found certain bugs in `ChunkedStreamIterator`'s use of `StreamController`only when we stress tested the output stream in cases outside of an `await-for` loop aha `Stream`s are really not easy :speak_no_evil:. 

If we are certain that the relevant `Stream` logic is correct, I think the main concerns remaining are:
1. Interface preference - I personally don't have a preference
2. Maintainability + Integration into pub - @jonasfj  can better advise on this, especially with google's requirements for mirroring code.
3. What .tar formats are we supporting? This shouldn't be too much of a concern either way, since we can reference either my or golang's or other implementations to see how they process the various tar formats.
4. More test cases - feel free to grab/modify from my PR

lib/src/io.dart
```diff
@@ -788,103 +787,51 @@ Future<HttpServer> bindServer(String host, int port) async {
...
+
+    final filePath = path.joinAll([
+      destination,
+      // Tar file names always use forward slashes
+      ...path.posix.split(entry.name),
+    ]);
+    final parentDirectory = path.dirname(filePath);
+
+    switch (entry.type) {
+      case TypeFlag.reg:
```
sigurdm, line 807:
```suggestion
      case TypeFlag.dir:
         ensureDir(filePath);
         break;
      case TypeFlag.reg:
```
We need this for preserving empty directories.

lib/src/io.dart
```diff
@@ -788,103 +787,51 @@ Future<HttpServer> bindServer(String host, int port) async {
...
+        // Regular file
+        deleteIfLink(filePath);
+        ensureDir(parentDirectory);
+        await _createFileFromStream(entry.contents, filePath);
+        break;
+      case TypeFlag.symlink:
```
sigurdm, line 813: We should consider not creating the symlink if it points out of the archive or if it is absolute.

We do have instances of those in pub.dev already: `moor_generator-1.0.0.tar.gz/analysis_options.yaml -> ../analysis_options.yaml`.

I am not sure exactly how to use for an attack - but I have a feeling these could become a security issue.

In any case they are (almost?) always a mistake.

Just silently not unpacking them will keep legitimate packages with broken symlinks working. if they work currently.

We do have instances of symlinks in pub.dev that could be considered acceptable.
Eg: `ejdb2_dart-1.0.28.tar.gz/lib/libejdb2dart.so -> libejdb2dart.so.2`.

I think we should keep on unpacking those.

lib/src/io.dart
```diff
@@ -788,103 +787,51 @@ Future<HttpServer> bindServer(String host, int port) async {
...
+      ...path.posix.split(entry.name),
+    ]);
+    final parentDirectory = path.dirname(filePath);
+
+    switch (entry.type) {
+      case TypeFlag.reg:
+        // Regular file
+        deleteIfLink(filePath);
+        ensureDir(parentDirectory);
+        await _createFileFromStream(entry.contents, filePath);
```
sigurdm, line 811: We could consider running a chmod (on macos/linux) to enable the executable bit if it is set in the tar file.

sigurdm:
I have tried unpacking all archives of pub.dev with this code and the code currently in pub,.dev (on linux), then running diff -rq on the results.

I found no differences (after fixing for empty directories) except for obscure cases like `page_turn-1.0.1.tar.gz/example/test/widget_test.dart` that ends up with no permissions when unpacking with gnu tar but seems fine with this code!

lib/src/io.dart
simolus3: Good point. There actually is another vulnerability here that's much more exploitable: We could have a tar file with a regular entry named `../crypto-3.0.0-nullsafety.0/lib/crypto.dart`. Extracting that file would allow an attacker to overwrite existing files in the pub cache.

simolus3: I now fixed all three problems (invalid symlinks, invalid hardlinks, invalid file names) and added tests for them.

For invalid file names, I opted to throw an exception instead of just skipping the entry because I can't see how that could happen by mistake.

lib/src/io.dart
```diff
@@ -788,102 +791,87 @@ Future<HttpServer> bindServer(String host, int port) async {
...
+        // Link to another file in this tar, relative from this entry.
+        final resolvedTarget = path.joinAll(
+            [parentDirectory, ...path.posix.split(entry.header.linkName)]);
+        if (!path.isWithin(destination, resolvedTarget)) {
+          // Don't allow links to files outside of this tar.
+          break;
+        }
+
+        ensureDir(parentDirectory);
+        createSymlink(resolvedTarget, filePath);
```
sigurdm, line 853: I think we want the symlink targets to be relative. Something like:
```suggestion
        createSymlink(path.relative(resolvedTarget, from: destination), filePath);
```

lib/src/io.dart
```diff
@@ -788,102 +791,87 @@ Future<HttpServer> bindServer(String host, int port) async {
...
+        // Link to another file in this tar, relative from this entry.
+        final resolvedTarget = path.joinAll(
+            [parentDirectory, ...path.posix.split(entry.header.linkName)]);
+        if (!path.isWithin(destination, resolvedTarget)) {
+          // Don't allow links to files outside of this tar.
+          break;
+        }
+
+        ensureDir(parentDirectory);
+        createSymlink(path.relative(resolvedTarget, from: destination), filePath);
```
sigurdm, line 853: Sorry - my suggestion was wrong! It should be (like you wrote below):
```suggestion
        createSymlink(path.relative(resolvedTarget, from: parentDirectory), filePath);
```

Maybe it is worth extracting the link-making code that is shared between the two types of link.
Or you could even make an empty case that falls through:

```dart
    case TypeFlag.link:
        // We generate hardlinks as symlinks too, but their path needs to be
        // resolved against the root of the tar file.
    case TypeFlag.symlink:
      final resolvedTarget = path.join(parentDirectory, entry.header.linkName);
      if (!path.isWithin(destination, resolvedTarget)) {
          break; // Link points outside of the tar file.
      }
      ensureDir(parentDirectory);
      createSymlink(resolvedTarget, filePath, relative: true);
      break;
```

simolus3: > Or you could even make an empty case that falls through:

I'm not sure that this would work? For `TypeFlag.symlink`, `header.linkName` is relative to the entry. For `TypeFlag.link`, it's relative to the root of the tar file.

That's at least what GNU-tar appears to do, I couldn't find this documented anywhere:

```
mkdir -p tar/source
touch tar/source/source.txt
mkdir tar/links
cd tar/links
ln ../source/source.txt hardlink.txt
ln -s ../source/source.txt softlink.txt
cd ../..
tar --create --file out.tar tar/
```

GNU tar generates `links/softlink.txt` as a softlink with `entry.header.linkName` pointing to `../source/source.txt`. It then generates the hardlink as a regular file and `tar/source/source.txt` as a hardlink with a `linkName` of `tar/links/hardlink.txt`.

I couldn't find a way to merge the two cases without making the code more complicated.

sigurdm: Ah I was unaware!

Ok then the cases should indeed not be merged!

We could add a comment here explaining the difference.

We should also have a test that the resulting targets of both kinds of symlinks (and perhaps both for relative and absolute) are correct to avoid someone later "cleaning" this up.

simolus3: > We could add a comment here explaining the difference.

Well, we have this:

```dart
        // We generate hardlinks as symlinks too, but their path needs to be
        // resolved against the root of the tar file.
```

Clearly that didn't clear things up enough, I'll expand the comment.

> We should also have a test that the resulting targets of both kinds of symlinks

I've added a test for that here: https://github.com/dart-lang/pub/blob/4928b5b456205150ffb927eb61bb8c9643c05494/test/io_test.dart#L310-L348

Both links are in subdirectories themselves, so the test would file if either case was merged into the other. If you want a more specific test case let me know.

sigurdm: No that is fine - I somehow overlooked those!

sigurdm:
If @simolus3 and @jonasfj have nothing more I think this is ready to merge!

lib/src/io.dart
```diff
@@ -788,102 +791,88 @@ Future<HttpServer> bindServer(String host, int port) async {
...
+        // Regular file
+        deleteIfLink(filePath);
+        ensureDir(parentDirectory);
+
+        await _createFileFromStream(entry.contents, filePath);
+
+        if (Platform.isLinux || Platform.isMacOS) {
+          // Apply executable bits from tar header, but don't change r/w bits
+          // from the default
+          const defaultMode = 420; // 644 in base 8
```
jonasfj, line 834: What permissions does `await stream.pipe(File(file).openWrite());` set? Just curious if it does the right thing.

lib/src/io.dart
```diff
@@ -893,110 +882,40 @@ final String _pathTo7zip = (() {
...
+    final relative = path.relative(entry, from: baseDir);
+    // On Windows, we can't open some files without normalizing them
+    final file = File(path.normalize(entry));
+    final stat = file.statSync();
+
+    return TarEntry(
+      TarHeader(
+        // Ensure paths in tar files use forward slashes
+        name: path.url.joinAll(path.split(relative)),
+        mode: stat.mode,
```
jonasfj, line 908: Maybe we should do: `mode: defaultMode | (stat.mode & executableMask);`

lib/src/io.dart
```diff
@@ -893,110 +882,40 @@ final String _pathTo7zip = (() {
...
-      deleteEntry(tempDir);
-      rethrow;
-    }
-  })));
+    final relative = path.relative(entry, from: baseDir);
+    // On Windows, we can't open some files without normalizing them
+    final file = File(path.normalize(entry));
+    final stat = file.statSync();
+
+    return TarEntry(
```
jonasfj, line 904: Is it allowed to make file entries without first having the directory entries.

If allowed by tar format, then I think just having files is preferable.

jonasfj (approved):
Mostly nits.

I still think we should forbid symlinks, but that's something we can do on the server. That way we can roll it back if people decide we really need it.

lib/src/io.dart
```diff
@@ -893,110 +882,40 @@ final String _pathTo7zip = (() {
...
-          .transform(onDoneTransformer(() => deleteEntry(tempDir)));
-    } catch (_) {
-      deleteEntry(tempDir);
-      rethrow;
-    }
-  })));
+    final relative = path.relative(entry, from: baseDir);
+    // On Windows, we can't open some files without normalizing them
+    final file = File(path.normalize(entry));
+    final stat = file.statSync();
```
jonasfj, line 902: Can we test if this is a symlink... If so maybe we should print a warning line that the symlink will be embedded as a file, and NOT a link.

IMO, it's probably preferable to keep the format simple and avoid fancy features like symlinks in distributed packages.

lib/src/io.dart
```diff
@@ -216,6 +215,10 @@ Future<String> _createFileFromStream(Stream<List<int>> stream, String file) {
   });
 }
 
+void _chmod(int mode, String file) {
+  runProcessSync('chmod', [mode.toRadixString(8), file]);
```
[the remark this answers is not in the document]
jonasfj: We could perhaps also use `dart:ffi` to bind the syscall... But we a little care has to be taken that we don't bind symbols only exposed because the Linux build is in release mode and not product mode.

lib/src/io.dart
simolus3: Technically it's configurable. On Linux, the filemode is `666₈ & ~umask`, where `umask` defaults to `0022₈`. It can be changed though.

I wanted to avoid unnecessary overhead, but maybe we should stat the created file to know the right permissions and then apply the executable bits on that? I think it's pretty rare for users to change the default permissions.

lib/src/io.dart
simolus3: I couldn't find any documentation stating whether it is acceptable to exclude directory entries or not.

GNU tar has explicit support for it though (it recovers from `ENOENT` by creating parent directories), and I'm not aware of any tar client that doesn't support this.

lib/src/io.dart
simolus3: I couldn't find any documentation stating whether it is acceptable to exclude directory entries or not.

GNU tar has explicit support for it though (it recovers from `ENOENT` by creating parent directories), and I'm not aware of any tar client that doesn't support this.

lib/src/io.dart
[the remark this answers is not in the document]
simolus3: It seems to be `/bin/chmod` on macOS too, so that should work.

Using `dart:ffi` sounds interesting (especially because we could check the value of `umask` at runtime too). AFAIK Dart doesn't support linking libc statically, so the symbols should be available

lib/src/io.dart
jonasfj: I agree, it's pretty rare that users change default permissions. I was just curious what it did by default :D

jonasfj (approved):

Issue #2699, opened 2020-10-29
`pub publlish` is not picking symbolic links


frencojobs:
# Environment

- `pub version` or `flutter pub version`: 2.10.2
- OS version: Ubuntu Linux 20.04
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

I have a mono repo containing multiple packages that I want to publish. Since all projects share the same `README.md`, I put the `README.md` in the root and put a link in all projects. The directory structure looks like this,

```
root/
├── package1/
│   ├── README.md -> link to ../README.md
│   └── pubspec.yaml
├── package2/
│   ├── README.md -> link to ../README.md
│   └── pubspec.yaml
└── README.md
```
When I want to publish package1, I `cd` into the package1 directory, and run `pub publish`. But the command fails with an error saying the link is broken.

## Expected behavior

I want it to work well.

## Actual behavior

It shows an error like this,
```
Package archive contains a broken symlink: `README.md` -> `../README.md`.
```

## --trace output

This is a section from `--trace` output that I think is relevant to the error.
```
ERR : Package archive contains a broken symlink: `README.md` -> `../README.md`.
FINE: Exception type: ApplicationException
FINE: package:pub/src/utils.dart 513:5            fail
    | package:pub/src/http.dart 309:3             handleJsonError
    | package:pub/src/command/lish.dart 109:9     LishCommand._publish
    | ===== asynchronous gap ===========================
    | dart:async                                  Future.catchError
    | package:pub/src/utils.dart 113:52           captureErrors.wrappedCallback
    | package:stack_trace                         Chain.capture
    | package:pub/src/utils.dart 126:11           captureErrors
    | package:pub/src/command_runner.dart 182:13  PubCommandRunner.runCommand
```

jonasfj:
(A) Should multiple packages really have the same README ?
(B) Couldn't a simple script wrapping `pub publish` simply copy it?

------------------

I'm wondering if it's a good idea to allow symlinks at all?

We probably should scan packages on `pub.dev` to find out how many use symlinks, and for what purpose.

Gustl22:
@jonasfj It makes sense e.g. for having the same license in every package. `Git` supports symlinks, so I expect `pub` to support it too (maybe).

Issue #2193, opened 2019-08-27
pub publish doesn't follow file symlinks


werediver:
# Environment

- `pub version` or `flutter pub version`: Pub 2.4.0
- OS version: macOS 10.14.6
- Are you using the Chinese community mirror or a corporate firewall? No.

# Problem

`pub publish` ignores file symlinks (but follows directory symlinks).

For example, in a multi-package repo I want the top level README.md to be symlinked from some of the packages (../README.md):

https://github.com/werediver/sum_types.dart/tree/5657f9d83f26cebac98ab3a0d8d5ff7e060d0859/sum_types_generator

## Expected behavior

I expect pub to resolve file symlinks and include the target files in the package as if they were in place of the symlinks.

## Actual behavior

Pub entirely ignores file symlinks during publishing as if they do not exist.

## --trace output

Not relevant (no errors).

werediver:
No, it does. I ended up with broken symlinks, which were legitimately ignored by pub.

frencojobs:
How did you manage to solve this? I'm also having an error where the linked files and folders outside of the project base are recognized as broken symlinks.

werediver:
@frencojobs I didn't use sym-links pointing outside the package directory, so it was never a problem for me. It is likely that such links are not accepted by pub for security and/or package integrity reasons.

You may consider using hard-links, if you think you have such a need.

frencojobs:
Thanks. Even though I can't get in contact with the author of the [Retrofit](https://github.com/trevorwang/retrofit.dart) library, it seems like the library is structured a lot like what I needed, monorepo with symlinks. So, I actually think it's possible but maybe I'm doing something wrong.

I'm currently using hard-links because I'm in a hassle right now. But it doesn't look like the best idea.

PR #2827, opened 2021-01-13
Follow file symlinks too

lukaslihotzki:
Currently, `pub publish` follows directory symlinks, but not file symlinks. With this PR, file symlinks are followed too. This fixes broken symlinks if they point outside the package root and should not break anything, although it could increase package size.

The dry-run shows the directory tree with directory symlinks resolved, and without indicating symlinks for file symlinks, so it is very confusing currently.

Ideally, symlinks would only be resolved if they point outside the package root. However, that would need two or three tar runs (one with, one without `--dereference`) and using `tar --concatenate` (in the second or the final, third run). It would make the code significantly more complex.

Related issues:
- https://github.com/dart-lang/pub/issues/2193 (Why is this closed?)
- https://github.com/dart-lang/pub/issues/2699

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -893,6 +893,7 @@ ByteStream createTarGz(List<String> contents, {String baseDir}) {
         // ustar is the most recent tar format that's compatible across all
         // OSes.
         '--format=ustar',
+        '--dereference',
         '--create',
         '--gzip',
         '--directory',
--- a/test/lish/follow_dir_symlink_test.dart
+++ b/test/lish/follow_dir_symlink_test.dart
@@ -1,0 +1,76 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+import 'dart:math' as math;
+import 'dart:typed_data';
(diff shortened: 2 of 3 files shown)


sigurdm:
I think I agree this is the right behavior.

One issue here is that we don't have the same tar process for windows and linux/macos. On windows we use 7zip. On the other hand symlinks are probably much less used on windows.

Could you write a test that shows this as being fixed?

We are working on migrating to a tar implementation in dart (https://github.com/dart-lang/pub/pull/2817). But it is not clear how long it will take to land. It would be great to have the test for ensuring that the behavior stays consistent when we do the switch.

jonasfj:
> Currently, pub publish follows directory symlinks, but not file symlinks.

It does follow directory symlinks? doesn't it just package the actual symlink?

lukaslihotzki:
> It does follow directory symlinks? doesn't it just package the actual symlink?
It does follow directory symlinks, as shown by the 'publish follows directory symlinks' test in this PR. At least for symlinks pointing to excluded or out-of-package-root targets, this is the only sensible behavior.

sigurdm:
We have merged https://github.com/dart-lang/pub/pull/2817 . If you want we can land the test. I believe our implementation follows links.

sigurdm:
Closing for lack of activity.

Feel welcome to open a new PR with the test if you would like ensuring the desired behavior does not regress.

PR #2829, opened 2021-01-14
Add save-tar option in publish

lukaslihotzki:
Add an option to save the package tar file. Especially useful for dry runs, so you can check the tar file before publish.

### DIFF
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:io';
 
 import 'package:http/http.dart' as http;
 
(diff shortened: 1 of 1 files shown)


jonasfj:
Would it be more useful with a listing of files included/excluded printed to terminal?

lib/src/command/lish.dart
```diff
@@ -154,9 +161,13 @@ the \$PUB_HOSTED_URL environment variable.''',
     if (!isValid) {
       overrideExitCode(exit_codes.DATA);
       return;
-    } else if (dryRun) {
-      return;
-    } else {
+    }
+
+    if (saveTar != null) {
+      await File(saveTar).writeAsBytes(await packageBytesFuture);
```
sigurdm, line 167: If we go through with this we need a `writeBinaryFile` in `lib/src/io.dart` that logs how many bytes are written to which location (should not write the contents) analogous  to `writeTextFile`.

lukaslihotzki:
publish already prints out the list of included files. Maybe it would be useful to print excluded files too or add a flag for that, but this is independent of this option.

sigurdm:
I can kind of see how this is useful. I just cannot imagine this being used regularly by most people, and worry about adding yet another command line option... What would you use the tar file for besides checking included files?

sigurdm:
For reference npm has something similar : https://docs.npmjs.com/cli/v6/commands/npm-pack

Issue #2806, opened 2020-12-16
`package_config.json` validation fails with `pub run`


blaugold:
# Environment

- `pub version` or `flutter pub version`: Pub 2.12.0-29.10.beta
- OS version:  Ubuntu 20.04.1 LTS
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

This issue is related to this [flutter issue](https://github.com/flutter/flutter/issues/69034).

`flutter pub run build_runner build` fails when:

- l10n generation has been setup for a flutter app
  - `flutter.generate` is `true` in `pubspec.yaml`
  - `l10n.yaml` exists
  - `lib/l10n/app_en.arb` exists
- a dependency on at least one local package through `path` exists

```
➜  issue_pub_package_config flutter pub run build_runner build
The ".dart_tool/package_config.json" file is not recognized by "pub" version, please run "pub get".
pub finished with exit code 65
```

To reproduce:

1. `git clone https://github.com/blaugold/issue_pub_package_config.git`
2. `cd issue_pub_package_config`
3. `flutter pub get`
4. `flutter pub run build_runner build`

The flutter code generation process breaks `.dart_tool/package_config.json`.

To reproduce run `flutter pub get` with `generate: true`:
```json
    {
      "name": "flutter_web_plugins",
      "rootUri": "../../../../../../lib/flutter/packages/flutter_web_plugins",
      "packageUri": "lib",
      "languageVersion": "2.12"
    },
    {
      "name": "flutter_gen",
      "rootUri": "flutter_gen",
      "languageVersion": "2.8"
    }
  ]
}
```

And then again with `generate: false`:
```json
    {
      "name": "yaml",
      "rootUri": "file:///Users/gabriel/.pub-cache/hosted/pub.dartlang.org/yaml-2.2.1",
      "packageUri": "lib/",
      "languageVersion": "2.4"
    }
  ],
  "generated": "2020-11-28T17:24:30.714165Z",
  "generator": "pub",
  "generatorVersion": "2.12.0-29.10.beta"
}

```

This error seems to be related to the `packageUri` property, which is expected to always exist and have the value `lib/`:

https://github.com/dart-lang/pub/blob/c00d4b4abf5b4ff265a7ce6282b748551f1b5b1f/lib/src/entrypoint.dart#L724

This conflicts with the spec for `package_config.json`:

[...]

PR #2830, opened 2021-01-14
Allow package_config.json to contain flutter_gen

sigurdm:
Fixes: https://github.com/flutter/flutter/issues/73870
Fixes: https://github.com/dart-lang/pub/issues/2806

This is somewhat a hack but should solve the immediate issue. It would be desirable to have 

* Flutter not rely on this by generating to a synthetic package
* Some well-defined mechanism for referring to synthetic code in a specific location.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -597,11 +597,11 @@ class Entrypoint {
     // Check that [packagePathsMapping] does not contain more packages than what
     // is required. This could lead to import statements working, when they are
     // not supposed to work.
-    final hasExtraMappings = packagePathsMapping.keys.every((packageName) {
+    final hasExtraMappings = !packagePathsMapping.keys.every((packageName) {
       return packageName == root.name ||
           lockFile.packages.containsKey(packageName);
     });
-    if (!hasExtraMappings) {
+    if (hasExtraMappings) {
       return false;
     }
 
@@ -620,8 +620,7 @@ class Entrypoint {
       }
 
       final source = cache.source(lockFileId.source);
-      final lockFilePackagePath =
-          p.join(root.dir, source.getDirectory(lockFileId));
+      final lockFilePackagePath = root.path(source.getDirectory(lockFileId));
 
       // Make sure that the packagePath agrees with the lock file about the
       // path to the package.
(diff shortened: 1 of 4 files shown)


jonasfj (approved):

lib/src/entrypoint.dart
```diff
@@ -720,9 +724,22 @@ class Entrypoint {
...
+    //
+    // See https://github.com/flutter/flutter/issues/73870 .
+    final packagesToCheck =
+        cfg.packages.where((package) => package.name != 'flutter_gen');
+    for (final pkg in packagesToCheck) {
+      // Pub always sets packageUri = 'lib/' but package:package_config might
+      // change it to 'lib'.
+      if (pkg.packageUri == null ||
+          (pkg.packageUri.toString() != 'lib' &&
+              pkg.packageUri.toString() != 'lib/')) {
```
jonasfj, line 742: We could patch our `package_config.json` parser to append `/` when it is omitted, given that spec says to do so.

lib/src/entrypoint.dart
```diff
@@ -620,12 +620,16 @@ class Entrypoint {
...
 
       // Make sure that the packagePath agrees with the lock file about the
       // path to the package.
-      if (p.normalize(packagePath) != p.normalize(lockFilePackagePath)) {
+      //
+      // TODO(sigurdm): We use the absolute paths to work around:
+      // https://github.com/dart-lang/package_config/issues/99
+      // For the case where Flutter has rewritten package_config.json.
+      if (p.absolute(p.normalize(packagePath)) !=
+          p.absolute(p.normalize(lockFilePackagePath))) {
```
jonasfj, line 632: We could consider only allowing this if `flutter_gen` is present? Just pass a boolean `allowRelativePaths` to `_isPackagePathsMappingUpToDateWithLockfile`?

jonasfj (changes requested):
Let's limit the blast radius by only allowing relative paths when `flutter_gen` is present, otherwise we're going to have a harder time removing this hack in the future.

jonasfj:
@sigurdm,
Hmm, I just tried flutter master and it doesn't seem to rewrite as relative paths anymore.

jonasfj:
Talked to Flutter tools people agreed that we don't add a workaround for relative paths.. Just `flutter_gen`.

Then they'll fix the relative paths stuff in flutter.

lib/src/entrypoint.dart
sigurdm: We won't do this anyway

lib/src/entrypoint.dart
sigurdm: Yeah - that's much better!

Done

jonasfj (approved):

PR #2834, opened 2021-01-16
output of command 'dart pub upgrade' updated

### LINKED ISSUE #2801: Output of the command `dart pub upgrade` shows a deprecated command suggestion
The command `dart pub upgrade` shows a deprecated command suggestion.

```
$ dart pub upgrade
Resolving dependencies...
... 
1 package has newer versions incompatible with dependency constraints.
```

The command `pub outdated` is deprecated, instead of it the new `dart` tool `dart pub outdated` is used.
See: https://github.com/dart-lang/site-www/issues/2617

azadprajapat:
Fixes  #2801 
The output of the command `dart pub upgrade` shows a deprecated command suggestion to run 'pub outdated' which has been changed to 'dart pub outdated' present in three files.


jonasfj (approved):
r+

mit-mit:
@jonasfj one question: will this work correctly (in terms of UI string) when `dart pub` is invoked from `flutter pub`?

jonasfj:
> will this work correctly (in terms of UI string) when dart pub is invoked from flutter pub?

Nope, but that seems like an orthogonal issue. I think we might have a thing to fix it too.

PR #2835, opened 2021-01-18
Warn about discontinued packages on pub get/upgrade

sigurdm:
The responses from the version listings are cached in the pub cache, such that we don't have to re-query the server on a `pub get` where pubspec.lock already exists and the packages are in cache already.

The cached responses are stored as a json:
```
{
  ...<body>
  "_fetchedAt": Iso8601String
}
```

The cached file will be at `<PUB_CACHE>/hosted/<url>/.cache/<package-name>-versions.json`

where body is the decoded response from the server.

This gives a maximum delay of the discontinued warning of 3 days from the actual discontinuation.

To see if a cached response is too old we both check the file modified timestamp (for speed) and the json timestamp (for extra robustness).

One hope is that the cached responses can be used for speeding up `pub get` itself, if/when we have a faster way of checking if the package-listing is up-to-date...

### DIFF
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -144,7 +144,7 @@ class SolveReport {
       /// time for packages that were not changed.
       final status =
           await _cache.source(id.source).status(id, Duration(days: 3));
-      if (status.isDiscontinued) {
+      if (status.isDiscontinued ?? false) {
         final suffix = status.discontinuedReplacedBy == null
             ? ''
             : ' it has been replaced by package:${status.discontinuedReplacedBy}';
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -45,7 +45,6 @@ class SolveReport {
   Future<void> show() async {
     _reportChanges();
     _reportOverrides();
-    await _reportDiscontinued();
   }
 
   /// Displays a one-line message summarizing what changes were made (or would
@@ -133,22 +132,28 @@ class SolveReport {
     }
   }
 
-  /// Displays a warning about any discontinued packages depended upon.
-  Future<void> _reportDiscontinued() async {
-    for (final id in _result.packages) {
-      if (id.source == null) {
-        continue;
-      }
-
-      /// We allow data to be up to 3 days old to not spend too much network
-      /// time for packages that were not changed.
-      final status =
-          await _cache.source(id.source).status(id, Duration(days: 3));
-      if (status.isDiscontinued ?? false) {
+  /// Displays a warning about any discontinued packages directly depended upon.
+  Future<void> reportDiscontinued() async {
+    final directDependencies = _result.packages
+        .where((id) => _root.dependencyType(id.name) != DependencyType.none);
+    final statuses = await Future.wait(
+      directDependencies.map(
+        (id) async => Pair(
+          id,
+          // We allow data to be up to 3 days old to not spend too much network
+          // time for packages that were not unlocked.
+          await _cache.source(id.source).status(id, Duration(days: 3)),
+        ),
+      ),
+    );
+    for (final statusPair in statuses) {
+      final id = statusPair.first;
+      final status = statusPair.last;
+      if (statusPair.last.isDiscontinued ?? false) {
         final suffix = status.discontinuedReplacedBy == null
             ? ''
-            : ' it has been replaced by package:${status.discontinuedReplacedBy}';
-        log.warning('Package:${id.name} has been discontinued$suffix.');
+            : ' it has been replaced by package ${status.discontinuedReplacedBy}';
+        log.warning('Package ${id.name} has been discontinued$suffix.');
       }
     }
   }
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -9,6 +9,7 @@ import '../log.dart' as log;
 import '../package.dart';
 import '../package_name.dart';
 import '../source_registry.dart';
+import '../system_cache.dart';
 import '../utils.dart';
 import 'result.dart';
 import 'type.dart';
@@ -24,6 +25,7 @@ class SolveReport {
   final Package _root;
   final LockFile _previousLockFile;
   final SolveResult _result;
+  final SystemCache _cache;
 
   /// The dependencies in [_result], keyed by package name.
   final _dependencies = <String, PackageId>{};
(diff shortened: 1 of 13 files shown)


lib/src/source/hosted.dart
```diff
@@ -226,6 +242,82 @@ class BoundHostedSource extends CachedSource {
...
+  @override
+  Future<PackageStatus> status(PackageId id, Duration maxAge) async {
+    final ref = id.toRef();
+    // Did we already get info for this package?
+    var versionListing = _scheduler.peek(ref);
+    // Do we have a cached version response on disk?
+    versionListing ??= _cachedVersionListingResponse(ref, maxAge);
+    // Otherwise retrieve the info from the host.
+    versionListing ??= await _scheduler.schedule(ref);
+    return versionListing[id].status;
```
jonasfj, line 311: Do we need to handle the case where `id` isn't present in `versionListing`...
Maybe it's better to return an empty `PackageStatus()`, rather than crashing?

```suggestion
    assert(versionListing[id] != null);
    if (versionListing[id] == null) {
      return PackageStatus(); // empty status is better than crashing.
    }
    return versionListing[id].status;
```

jonasfj (commented):
Mostly nits, looks good.. I do propose that we only warn about discontinuation for direct-dependencies to reduce the noise to signal ratio.

lib/src/solver/report.dart
```diff
@@ -130,6 +133,26 @@ class SolveReport {
     }
   }
 
+  /// Displays a warning about any discontinued packages depended upon.
+  Future<void> _reportDiscontinued() async {
+    for (final id in _result.packages) {
```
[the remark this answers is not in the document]
jonasfj: Also maybe we should do:

```dart
await Future.wait(_result.packages.map((id) async {
  ...
}));
```

So that we get more concurrency in cases where this information isn't in the cache.

lib/src/source/hosted.dart
```diff
@@ -226,6 +242,82 @@ class BoundHostedSource extends CachedSource {
...
+  }
+
+  // The path where the response from the package-listing api is cached.
+  String _versionListingCachePath(PackageRef ref) {
+    final parsed = source._parseDescription(ref.description);
+    final dir = _urlToDirectory(parsed.last);
+    return p.join(systemCacheRoot, dir, '${ref.name}-versions.json');
```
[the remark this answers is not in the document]
jonasfj: ah, good point, I'm happy keeping it `<package>-versions.json>`.

lib/src/source/hosted.dart
jonasfj: > But I think for simplicity just returning the empty status might be relatively harmless here, and surely better than crashing.

Yeah, that's why I suggested an assertion, it shouldn't happen  -- and if it does we must be in one of those corner cases where a package has been removed, but still exists in users cache. And even that corner case is unlikely to work -- in any case, assert + avoid crash in production is probably a good balance.

jonasfj (approved):

lib/src/source/hosted.dart

lib/src/source/hosted.dart
```diff
@@ -311,17 +311,27 @@ class BoundHostedSource extends CachedSource {
...
 
   // The path where the response from the package-listing api is cached.
   String _versionListingCachePath(PackageRef ref) {
     final parsed = source._parseDescription(ref.description);
     final dir = _urlToDirectory(parsed.last);
-    return p.join(systemCacheRoot, dir, '${ref.name}-versions.json');
+    // Use a .versions/ dir because older versions of pub won't choke on that
+    // name when iterating the cache (it is not listed by [listDir]).
+    return p.join(
+        systemCacheRoot, dir, '.versions', '${ref.name}-versions.json');
```
jonasfj, line 334: nit: Do we need `-versions.json`, I guess it makes sense if we want the ability to add multiple files in the future.
In that case perhaps we should make it `.cache/<pacakge>-versions.json` instead of `.versions/<pacakge>.json`.

Either option works for me, but repeating versions twice seems unnecessary :D

lib/src/source/hosted.dart
```diff
@@ -391,7 +401,7 @@ class BoundHostedSource extends CachedSource {
         final results = <RepairResult>[];
         var packages = <Package>[];
         for (var entry in listDir(serverDir)) {
-          if (dirExists(entry)) {
+          if (_looksLikePackageDir(entry)) {
```
jonasfj, line 404: This isn't technically necessary, right?

Downside is that if you repair the cache because you delete a `pubspec.yaml` that won't work anymore :see_no_evil:

lib/src/source/hosted.dart
```diff
@@ -460,6 +469,7 @@ class BoundHostedSource extends CachedSource {
     if (!dirExists(cacheDir)) return [];
 
     return listDir(cacheDir)
+        .where(_looksLikePackageDir)
```
jonasfj, line 472: Again, we don't need `_looksLikePackageDir`, I don't know if we should remove it.. But that is correct right?

lib/src/source/hosted.dart
```diff
@@ -638,12 +648,18 @@ class _OfflineHostedSource extends BoundHostedSource {
     List<PackageId> versions;
     if (dirExists(dir)) {
       versions = listDir(dir)
+          // Only directories with a pubspec.yaml should be considered packages.
+          .where(_looksLikePackageDir)
```
jonasfj, line 652: +100, nice! :)

lib/src/source/hosted.dart
```diff
@@ -638,12 +648,18 @@ class _OfflineHostedSource extends BoundHostedSource {
...
             if (components.first != ref.name) return null;
-            return source.idFor(
-                ref.name, Version.parse(components.skip(1).join('-')),
-                url: _serverFor(ref.description));
+            try {
+              return source.idFor(
+                  ref.name, Version.parse(components.skip(1).join('-')),
+                  url: _serverFor(ref.description));
+            } on FormatException {
+              // Failing to parse a package-name is not bad.
```
jonasfj, line 661: Could also be version that fails, but I agree... Failure to parse here is best ignored

jonasfj (commented):
Don't we need to override `_OfflineHostedSource.status` ?

lib/src/source/hosted.dart
jonasfj: If our directory is called `.versions` or `.cache` then `listDir` won't return it, right?

PR #2837, opened 2021-01-21
No metadata when CI=true

jonasfj:
Follow up from https://github.com/dart-lang/pub/pull/2759, I think it's fair that we send no metadata when `CI=true`.


sigurdm (approved):

kevmoo:
grumble grumble...I like having access to the data. But I'll live.

PR #2842, opened 2021-01-22
Recommend running `flutter pub` instead of `dart pub` when running from inside flutter.

sigurdm:
Jonasfj discovered that this can be detected by `flutter pub` always setting an environment variable.

### DIFF
--- a/test/outdated/outdated_test.dart
+++ b/test/outdated/outdated_test.dart
@@ -450,7 +450,9 @@ Future<void> main() async {
 
     await variations('handles_sdk_dependencies', environment: {
       'FLUTTER_ROOT': d.path('flutter-root'),
-      '_PUB_TEST_SDK_VERSION': '2.13.0'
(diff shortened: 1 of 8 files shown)


test/outdated/outdated_test.dart
```diff
@@ -450,7 +450,9 @@ Future<void> main() async {
 
     await variations('handles_sdk_dependencies', environment: {
       'FLUTTER_ROOT': d.path('flutter-root'),
-      '_PUB_TEST_SDK_VERSION': '2.13.0'
+      '_PUB_TEST_SDK_VERSION': '2.13.0',
+      // To test that the reproduction command is reflected correctly.
+      'PUB_ENVIRONMENT': 'flutter_cli',
```
jonasfj, line 455: technically it's probably something like;
```suggestion
      'PUB_ENVIRONMENT': 'flutter_cli:get',
```

jonasfj (approved):

PR #2844, opened 2021-01-22
Show outdated and discontinued in listing in get

sigurdm:
pubspec.yaml
```yaml
name: blah 
environment:
  sdk: '>=2.7.0 <3.0.0'

dependencies: 
  package_config:
  _dummy_pkg: <0.0.4
  retry: ^2.0.0
```
Will now have this output for pub get (the second run highlights outdated and discontinued packages):
```
> dart pub get
Resolving dependencies... (1.0s)
+ _dummy_pkg 0.0.3 (discontinued)
+ charcode 1.1.3
+ package_config 1.9.3
+ path 1.7.0
+ retry 2.0.0 (3.0.1 available)
Changed 5 dependencies!
```
Second invocation (with a pubspec.lock in place):
```
> dart pub get
Resolving dependencies... 
  _dummy_pkg 0.0.3 (discontinued)
  retry 2.0.0 (3.0.1 available)
Got dependencies!
```

### DIFF
--- a/lib/src/solver/type.dart
+++ b/lib/src/solver/type.dart
@@ -2,24 +2,5 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// An enum for types of version resolution.
-class SolveType {
-  /// As few changes to the lockfile as possible to be consistent with the
-  /// pubspec.
-  static const GET = SolveType._('get');
-
-  /// Upgrade all packages or specific packages to the highest versions
-  /// possible, regardless of the lockfile.
-  static const UPGRADE = SolveType._('upgrade');
-
-  /// Downgrade all packages or specific packages to the lowest versions
-  /// possible, regardless of the lockfile.
-  static const DOWNGRADE = SolveType._('downgrade');
-
-  final String _name;
-
-  const SolveType._(this._name);
-
-  @override
-  String toString() => _name;
-}
+/// The types of version resolution.
+enum SolveType { get, upgrade, downgrade }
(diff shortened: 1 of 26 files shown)


lib/src/solver/type.dart
```diff
@@ -2,24 +2,5 @@
...
-
-  final String _name;
-
-  const SolveType._(this._name);
-
-  @override
-  String toString() => _name;
-}
+/// The types of version resolution.
+enum SolveType { get, upgrade, downgrade }
```
jonasfj: Please consider retaining the documentation comments for each of the values...

``` 
  /// Downgrade all packages or specific packages to the lowest versions
  /// possible, regardless of the lockfile.
  downgrade,
```

lib/src/solver/version_solver.dart
```diff
@@ -424,17 +425,21 @@ class VersionSolver {
...
-
-      availableVersions[package.name] = versions;
+      // If the version list was never requested, use versions from cached
+      // version listings if the package is "hosted".
+      var ids = cached ??
+          (package.source is HostedSource
+              ? (await _systemCache
+                  .source(package.source)
+                  .getVersions(package.toRef(), maxAge: Duration(days: 3)))
+              : [package]);
```
jonasfj, line 440: Does this really have to be `HostedSource` specific? Will git sources try to fetch something?

jonasfj: Since this is only place we use `maxAge`, and we still have to use a `HostedSource` specific variant, isn't it kind of fair to ask if this is the right abstraction?

jonasfj (approved):

lib/src/solver/version_solver.dart
sigurdm: Yeah - you are spot on. I added a TODO to find a better abstraction.

lib/src/solver/type.dart
sigurdm: Reverted this for now

PR #2845, opened 2021-01-25
Include a deprecation notice in .packages

sigurdm:


lib/src/lock_file.dart
```diff
@@ -219,7 +219,14 @@ class LockFile {
...
   String packagesFile(SystemCache cache, [String entrypoint]) {
-    var header = 'Generated by pub on ${DateTime.now()}.';
+    var header = '''
+This file is deprecated. Tools should instead consume 
+`.dart_tools/package_config.json`.
+
+For more info see:
+https://github.com/dart-lang/language/blob/master/accepted/future-releases/language-versioning/package-config-file-v2.md
+
+Generated by pub on ${DateTime.now()}.''';
```
jonasfj, line 229: Check the parser, I don't think older versions supports multiple lines of comments :see_no_evil: 

I might be remembering incorrectly...

We can test using a version of Dart 2.0.0 if we have to..

jonasfj: NVM, I'm wrong :D

lib/src/lock_file.dart
```diff
@@ -219,7 +219,14 @@ class LockFile {
   /// If [entrypoint] is passed, a relative entry is added for its "lib/"
   /// directory.
   String packagesFile(SystemCache cache, [String entrypoint]) {
-    var header = 'Generated by pub on ${DateTime.now()}.';
+    var header = '''
+This file is deprecated. Tools should instead consume 
+`.dart_tools/package_config.json`.
+
+For more info see:
+https://github.com/dart-lang/language/blob/master/accepted/future-releases/language-versioning/package-config-file-v2.md
```
jonasfj, line 227: How about making a `dart.dev/go/...` link? I'm not sure this URL is very stable.

jonasfj (approved):

lib/src/lock_file.dart
sigurdm: https://github.com/dart-lang/site-www/pull/2888

Issue #2839, opened 2021-01-21
pub get failed once PUB_HOSTED_URL is set


intoxicated:
# Environment

- `pub version` or `flutter pub version`: 2.10.4
- OS version: mac os x 10.15.7
- Are you using the Chinese community mirror or a corporate firewall? no

# Problem

After PUB_HOSTED_URL is set, flutter pub get failed 

## Expected behavior

successfully installed

## Actual behavior

exited with fail

## --trace output

```
ERR : The method 'map' was called on null.
    | Receiver: null
    | Tried calling: map<MapEntry<PackageId, _VersionInfo>>(Closure: (dynamic) => MapEntry<PackageId, _VersionInfo>)
FINE: Exception type: NoSuchMethodError
ERR : dart:core                                   Object.noSuchMethod
    | package:pub/src/source/hosted.dart 194:45   BoundHostedSource._fetchVersions
    | ===== asynchronous gap ===========================
    | dart:async                                  Future.catchError
    | package:pub/src/utils.dart 113:52           captureErrors.wrappedCallback
    | package:stack_trace                         Chain.capture
    | package:pub/src/utils.dart 126:11           captureErrors
    | package:pub/src/command_runner.dart 182:13  PubCommandRunner.runCommand
```

jonasfj:
What did you set it to?

Can you run with `--trace` and dump the output to a gist?

It might simply be an incorrect URL in your env var...

jonasfj:
I'm guessing we should look to improve our JSON parsing to be more robust and detect invalid responses -- because that's probably what's happening here.

PR #2847, opened 2021-01-25
Handle poor package-listing responses robustly.

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2839

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -178,20 +178,29 @@ class BoundHostedSource extends CachedSource {
 
   Map<PackageId, _VersionInfo> _versionInfoFromPackageListing(
       Map body, PackageRef ref, Uri location) {
-    final versions = body['versions'] as List;
-    return Map.fromEntries(versions.map((map) {
-      var pubspec = Pubspec.fromMap(map['pubspec'], systemCache.sources,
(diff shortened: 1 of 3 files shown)


lib/src/source/hosted.dart
```diff
@@ -565,6 +578,11 @@ class BoundHostedSource extends CachedSource {
     } else if (error is io.TlsException) {
       fail('Got TLS error trying to find package $package at $url.', error,
           stackTrace);
+    } else if (error is FormatException) {
+      throw PackageNotFoundException(
+          'Got badly formatted response trying to find package $package at $url',
```
jonasfj, line 583: Hmm, if `PUB_HOSTED_URL` is not the default value, then maybe we should include a message warning about that here -- it's common that people add an extra slash or something to their `PUB_HOSTED_URL` or they use `https://localhost...` when it should have been `http://localhost...`

That probably applies to `SocketException`, `TlsException` and `PubHttpException`...

Just an idea, that might be an orthogonal concern best addressed in a follow up PR.

jonasfj (approved):

lib/src/source/hosted.dart
sigurdm: I think the url in itself might be enough.

PR #2848, opened 2021-01-25
Fix request metadata when overriding dependencyType

jonasfj:
Running `dart pub get --verbose` in `flutter/gallery` shows:
```
IO  : Get versions from https://pub.dartlang.org/api/packages/url_launcher_platform_interface.
IO  : HTTP GET https://pub.dartlang.org/api/packages/url_launcher_platform_interface
    | Accept: application/vnd.pub.v2+json
    | X-Pub-OS: linux
    | X-Pub-Command: upgrade
    | X-Pub-Session-ID: 80D86525-9A4A-4468-AE49-23A6725FA366
    | X-Pub-Reason: direct
    | user-agent: Dart pub 2.12.0-248.0.dev
```

That's a problem... looking at the code it seems setting `dependencyType.none` does not override existing `dependencyType` in the current zone.

This has been causing issues since we introduced prefetching in Dart 2.8.

----------------

TL;DR: Dependency-type `direct` cannot be trusted since Dart 2.8.

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -56,7 +56,9 @@ class _PubHttpClient extends http.BaseClient {
       }
 
       var type = Zone.current[#_dependencyType];
-      if (type != null) request.headers['X-Pub-Reason'] = type.toString();
+      if (type != null && type != DependencyType.none) {
+        request.headers['X-Pub-Reason'] = type.toString();
+      }
     }
 
     _requestStopwatches[request] = Stopwatch()..start();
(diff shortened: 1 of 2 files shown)


jonasfj:
@kevmoo, sorry, I think we broke your metadata :disappointed: 

I'll look into cleanup, I think we just need to only look at downloads, those should still have correct metadata.

kevmoo:
You should just have taken my PR the first time! :-D

kevmoo:
Anyway to add test coverage here?

jonasfj:
> Anyway to add test coverage here?

Maybe, I was thinking of doing so in a follow up... But it's not exactly deterministic behavior because it's only when we're prefetching dependencies that the `dependencyType` is inherited from the previous zone.

I suppose one could also make the argument that we should migrate away from using zones for sending around this metadata.

sigurdm (approved):
LGTM
Good catch!

PR #2850, opened 2021-01-27
Type-safe(r) args

sigurdm:
Just an idea that came to mind when looking at enabling 
```
  strong-mode:
    implicit-casts: false
```
in analysis options.

### DIFF
--- a/lib/src/args_wrapper.dart
+++ b/lib/src/args_wrapper.dart
@@ -1,0 +1,108 @@
+import 'package:args/args.dart';
+import 'package:meta/meta.dart';
+
+abstract class Arg<T> {
+  String name;
+
+  final String help;
+  final String abbr;
+  final bool hide;
+
+  String valueHelp;
+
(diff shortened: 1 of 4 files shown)


lib/src/args_wrapper.dart
```diff
@@ -0,0 +1,112 @@
...
+      String help,
+      this.allowed,
+      this.allowedHelp,
+      this.defaultsTo,
+      bool hide})
+      : super._(name, abbr: abbr, help: help, hide: hide);
+}
+
+extension ArgResultExt on ArgResults {
+  T get<T>(Arg<T> arg) => this[arg.name] as T;
```
jonasfj, line 70: Ideally, you'd probably want to make this throw an Error of the `arg` instance wasn't registered with the associated `ArgParser`.

That requires adding and storing state on `ArgParser`.

sigurdm:
Might revisit this later...

PR #2867, opened 2021-02-02
Check SDK constraints of dependencies in global run

jonasfj:


natebosch (approved):

lib/src/global_packages.dart
```diff
@@ -378,6 +378,22 @@ To recompile executables, first run `global decativate ${dep.name}`.
...
+    // Check that the SDK constraints the lockFile says we have are honored.
+    lockFile.sdkConstraints.forEach((sdkName, constraint) {
+      var sdk = sdks[sdkName];
+      if (sdk == null) {
+        dataError('${log.bold(name)} as globally activated requires '
+            'unknown SDK "$name".');
+      } else if (sdkName == 'dart') {
+        if (constraint.allows(sdk.version)) return;
+        dataError("${log.bold(name)} as globally activated doesn't "
+            'support Dart ${sdk.version}, try to re-activate it.');
```
sigurdm, line 390: Could we give a command they could copy-paste to do that?
```
pub global activate <name> --source <source>
```
It probably requires some logic to handle  reproduction of the `source` and `name`...

I suspect they would not need to deactivate first...

sigurdm (approved):
LGTM with a suggestion for a reproduction command.

We can also leave it here

lib/src/global_packages.dart
jonasfj: I omitted the `--source` part... I see no reason we shouldn't do it. I just wanted to close this now :D

lib/src/global_packages.dart
```diff
@@ -387,7 +387,7 @@ To recompile executables, first run `global decativate ${dep.name}`.
       } else if (sdkName == 'dart') {
         if (constraint.allows(sdk.version)) return;
         dataError("${log.bold(name)} as globally activated doesn't "
-            'support Dart ${sdk.version}, try to re-activate it.');
+            'support Dart ${sdk.version}, try: dart pub global activate $name');
```
sigurdm, line 390:
```suggestion
            'support Dart ${sdk.version}, try: $topLevelProgram pub global activate $name');
```

Issue #2868, opened 2021-02-03
dart pub outdated wrong_argument -- causes no issues


jonasfj:
```
[image_picker]$dart --version
Dart SDK version: 2.12.0-205.0.dev (dev) (Thu Jan 7 20:27:37 2021 -0800) on "linux_x64"
[image_picker]$dart pub outdated wrong_argument
Showing outdated packages.
[*] indicates versions that are not the latest available.

Found no outdated packages
```

Shouldn't invalid arguments print a usage message and exit non-zero.

PR #2872, opened 2021-02-05
Don't allow outdated taking arguments

sigurdm:
Also fixes an issue where stack traces would be exposed to the user in case of some parsing errors (when using the `pub` binary - I don't think this happens with `dart pub`.
Fixes: https://github.com/dart-lang/pub/issues/2868


jonasfj (approved):
r+

Issue #2864, opened 2021-02-02
Add -C option to run pub in another directory


sigurdm:
One question is should it be 
```
> dart pub -C my/dir get
```
or

```
> dart pub get -C my/dir
```

or both.

PR #2876, opened 2021-02-08
Add --directory option

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2864

### DIFF
--- a/lib/src/command/add.dart
+++ b/lib/src/command/add.dart
@@ -74,6 +74,8 @@ class AddCommand extends PubCommand {
 
     argParser.addFlag('precompile',
         help: 'Precompile executables in immediate dependencies.');
+    argParser.addOption('directory',
+        abbr: 'C', help: 'Run this in <dir>.', valueHelp: 'dir');
   }
 
   @override
@@ -96,6 +98,7 @@ class AddCommand extends PubCommand {
       /// in case [package] was already a transitive dependency. In the case
       /// where the user specifies a version constraint, this serves to ensure
       /// that a resolution exists before we update pubspec.yaml.
+      // TODO(sigurdm): We should really use a spinner here.
       solveResult = await resolveVersions(
           SolveType.UPGRADE, cache, Package.inMemory(updatedPubSpec));
     } on GitException {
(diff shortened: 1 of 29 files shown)


lib/src/command/add.dart
```diff
@@ -74,6 +74,8 @@ class AddCommand extends PubCommand {
 
     argParser.addFlag('precompile',
         help: 'Precompile executables in immediate dependencies.');
+    argParser.addOption('directory',
+        abbr: 'C', help: 'Run this in <dir>.', valueHelp: 'dir');
```
jonasfj, line 78:
```suggestion
        abbr: 'C', help: 'Run in the directory <dir>.', valueHelp: 'dir');
```

lib/src/command_runner.dart
```diff
@@ -109,6 +112,13 @@ class PubCommandRunner extends CommandRunner<int> implements PubTopLevel {
     });
     argParser.addFlag('verbose',
         abbr: 'v', negatable: false, help: 'Shortcut for "--verbosity=all".');
+    argParser.addOption(
+      'directory',
+      abbr: 'C',
+      help: 'Run the subcommand in <dir>.',
```
jonasfj, line 118: Maybe I suggest `Run in the directory <dir>.`

lib/src/command_runner.dart
```diff
@@ -109,6 +112,13 @@ class PubCommandRunner extends CommandRunner<int> implements PubTopLevel {
     });
     argParser.addFlag('verbose',
         abbr: 'v', negatable: false, help: 'Shortcut for "--verbosity=all".');
+    argParser.addOption(
```
jonasfj, line 115: Just curious, does it have to both be a top-level argument, and an argument in all the subcommands?

What happens if it's passed twice...

jonasfj (approved):

lib/src/command_runner.dart
sigurdm: Hmm - it doesn't have to. That just made more sense to me. I would personally instinctively give it at the top-level (it is an option that changes how the following command is interpreted), but it would be annoying to have not work for each command.

We could do it either way if you think it is better.

jonasfj (approved):

Issue #2892, opened 2021-02-23
JSON output for `dart pub deps`


jonasfj:
Similar to how we allow JSON output from `outdated`.

DanTup:
This is done (#2896)?

PR #2896, opened 2021-03-02
pub deps --json

sigurdm:
https://github.com/dart-lang/pub/issues/2892


lib/src/command/deps.dart
```diff
@@ -67,23 +72,59 @@ class DepsCommand extends PubCommand {
     if (argResults['executables']) {
       _outputExecutables();
```
jonasfj, line 73: I have mixed feelings about this.

We could choose to output different JSON output when either `--dev` and/or `--executable` is given, or we could just ignore (or forbid) these flags when outputting JSON, and always output the same thing.

Since a JSON output could contain a property `"relation": "direct | dev | transitive | transitive-dev"` or something like that, and we leave it up to the reader to parse it.

Didn't we do this for `dart pub outdated --json` ?

sigurdm:
cc: @dantup do you have any comments about the json output, seen from the perspective of dart-code?

DanTup:
@sigmundch looks great! It should make both https://github.com/Dart-Code/Dart-Code/issues/3095 and https://github.com/Dart-Code/Dart-Code/issues/3202 very easy to add. Thanks! :-)

jonasfj (approved):

DanTup:
Is it intended that `package_config.json` is modified when running this command? It causes a loop for me right now - we refresh the packages tree when that file changes, but we then call this command to render the tree, so it ends up constantly refreshing. The timestamp inside the JSON doesn't change, but the file modified date changes (and the file watchers fire).

jonasfj:
@DanTup please file an issue if you can reproduce it.
```
[myapp]$dart pub get
Resolving dependencies... 
Got dependencies!
[myapp]$stat  .dart_tool/package_config.json 
  File: .dart_tool/package_config.json
  Size: 650       	Blocks: 8          IO Block: 4096   regular file
Device: fe01h/65025d	Inode: 3950172     Links: 1
Access: (0644/-rw-r--r--)  Uid: (591453/ jonasfj)   Gid: (89939/primarygroup)
Access: 2021-05-27 11:35:56.182021029 +0200
Modify: 2021-05-27 11:42:16.708804560 +0200
Change: 2021-05-27 11:42:16.708804560 +0200
 Birth: 2021-05-27 11:35:56.182021029 +0200
[myapp]$dart pub deps --json > /dev/null
[myapp]$stat  .dart_tool/package_config.json 
  File: .dart_tool/package_config.json
  Size: 650       	Blocks: 8          IO Block: 4096   regular file
Device: fe01h/65025d	Inode: 3950172     Links: 1
Access: (0644/-rw-r--r--)  Uid: (591453/ jonasfj)   Gid: (89939/primarygroup)
Access: 2021-05-27 11:35:56.182021029 +0200
Modify: 2021-05-27 11:42:16.708804560 +0200
Change: 2021-05-27 11:42:16.708804560 +0200
 Birth: 2021-05-27 11:35:56.182021029 +0200
```

I couldn't trivially reproduce it. `dart pub get` will update the mtime, even if there is no changes... we could consider changing that.

DanTup:
Hmm, I can't repro in one project but can in another:

```
danny@Dannys-MacBook-Pro dart_sample % stat  .dart_tool/package_config.json                   
16777221 47509992 -rw-r--r-- 1 danny staff 0 10044 "May 27 11:43:48 2021" "May 27 11:43:48 2021" "May 27 11:43:48 2021" "Feb 11 13:03:20 2021" 4096 24 0 .dart_tool/package_config.json

danny@Dannys-MacBook-Pro dart_sample % ~/Dev/Dart\ SDKs/nightly-2021-05-25/bin/pub deps --json > /dev/null 

danny@Dannys-MacBook-Pro dart_sample % stat  .dart_tool/package_config.json                               
16777221 47509992 -rw-r--r-- 1 danny staff 0 10044 "May 27 11:44:01 2021" "May 27 11:44:01 2021" "May 27 11:44:01 2021" "Feb 11 13:03:20 2021" 4096 24 0 .dart_tool/package_config.json
```

I'll see if I can figure out the difference and file an issue. Thanks!

DanTup:
> `dart pub get` will update the mtime, even if there is no changes... we could consider changing that.

That's not a problem for me here - the issue was just with `deps` because we call it in response to the modification, and if it triggers another, it loops. Although `pub get` would trigger the tree to refresh, the tree would not call back out to `pub get` so there would be no loop.

Issue #2521, opened 2020-06-12
Request for 'pub cache clean' in addition to Repair, Add


robinjohn:
Raising this as a new issue as I didn't find an existing one.
Related one was old one linked below and probably sidelined or down-prioritised.

**Why:** flutter SDK is set up as local, and user can switch between flutter project/application repositories and also switch between flutter channels (beta/stable) which affects the common flutter environment. This involves pub to accumulate some cache old packages which user don't want to use anymore of some package worked many months or years before. Hence instead of repairing, user could prefer a complete new pub get on the flutter projects being work after cleaning cache and running pub get.

_Related to post by @nex3 in https://github.com/dart-lang/pub/issues/1587#issuecomment-331013040_

sigurdm:
This should probably exist for completeness, but all it would do is equivalent to `rm -r $PUB_CACHE`, right?

sigurdm:
Or should it be able to prune, leaving the dependencies of a specified number of projects behind...

robinjohn:
Ideally it should be an equivalent of rm -r at this stage, just because most other pub commands don't take effect per project but rather to users development environment in general. 
(Sorry for my late response, didn't get notified here).

robinjohn:
Linked Issue:
Relates to [https://github.com/flutter/flutter/issues/77029]

PR #2904, opened 2021-03-08
`cache clean`

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2521

We could consider letting the command only delete some files (for example leaving credentials.json), but I'm not sure the complexity is worth it - it is not that hard to log in again.

### DIFF
--- a/lib/src/command/cache_clean.dart
+++ b/lib/src/command/cache_clean.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import '../command.dart';
+import '../command_runner.dart';
 import '../io.dart';
 import '../io.dart';
 import '../log.dart' as log;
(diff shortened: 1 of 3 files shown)


robinjohn:
LGTM. Please await internal Thumbs-up too. Thanks @sigurdm

lib/src/command/cache_clean.dart
```diff
@@ -0,0 +1,26 @@
...
+  @override
+  String get name => 'clean';
+  @override
+  String get description => 'Clears the entire system cache.';
+  @override
+  bool get takesArguments => false;
+
+  @override
+  Future<void> runProtected() async {
+    if (entryExists(cache.rootDir)) {
```
jonasfj, line 19: I'm tempted to suggest the user should confirm this action.

Maybe we have:
 * `dart pub cache clean --force`, no prompt.
 * `dart pub cache clean`, prompts user on the CLI.

**Reasoning:** deleting the entire cache means that anything installed will stop working.
What about flutter? will it recover from this?

sigurdm: Yeah - probably better do this!

jonasfj (approved):

lib/src/command/cache_clean.dart
sigurdm: I changed the code, so the (now empty) directory will be created again. That will make sure flutter keeps using the internal cache if it is already doing so.

PR #2908, opened 2021-03-10
pub upgrade command to show number of discontinued packages

### LINKED ISSUE #2865: Upgrade summary should show a count of discontinued packages
```
pub upgrade
Resolving dependencies... (2.4s)
+ _dummy_pkg 0.0.122+oiiri (discontinued)
  pub_semver 1.4.4
+ retry 2.0.0 (3.0.1 available)
...
Downloading _dummy_pkg 0.0.122+oiiri...
Changed 2 dependencies!
3 packages have newer versions incompatible with dependency constraints.
1 package is discontinued <--- this line is missing!
```

tusharojha:
## Description

I have added a message showing the number of discontinued packages for pub upgrade command.

Fixes #2865

### DIFF
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -153,10 +153,12 @@ class SolveReport {
           await _cache.source(id.source).status(id, Duration(days: 3));
       if (status.isDiscontinued) numDiscontinued++;
     }
-    if (numDiscontinued > 1) {
-      log.message('$numDiscontinued packages are discontinued.');
-    } else if (numDiscontinued == 1) {
-      log.message('1 package is discontinued.');
+    if (numDiscontinued > 0) {
+      if (numDiscontinued == 1) {
+        log.message('1 package is discontinued.');
+      } else {
+        log.message('$numDiscontinued packages are discontinued.');
+      }
     }
   }
 
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -143,6 +143,23 @@ class SolveReport {
     }
   }
 
+  /// Displays a single-line message, number of discontinued packages
+  /// if discontinued packages are detected.
+  Future<void> reportDiscontinued() async {
+    var numDiscontinued = 0;
+    for (var id in _result.packages) {
+      if (id.source == null) continue;
+      final status =
+          await _cache.source(id.source).status(id, Duration(days: 3));
+      if (status.isDiscontinued) numDiscontinued++;
+    }
+    if (numDiscontinued > 1) {
+      log.message('$numDiscontinued packages are discontinued.');
+    } else if (numDiscontinued == 1) {
+      log.message('1 package is discontinued.');
+    }
+  }
+
   /// Displays a two-line message, number of outdated packages and an
   /// instruction to run `pub outdated` if outdated packages are detected.
   void reportOutdated() {
--- a/lib/src/solver/result.dart
+++ b/lib/src/solver/result.dart
@@ -109,6 +109,7 @@ class SolveResult {
         SolveReport(type, _sources, _root, _previousLockFile, this, cache);
     report.summarize(dryRun: dryRun);
     if (type == SolveType.UPGRADE) {
+      await report.reportDiscontinued();
       report.reportOutdated();
     }
   }
--- a/test/upgrade/dry_run_does_not_apply_changes_test.dart
+++ b/test/upgrade/dry_run_does_not_apply_changes_test.dart
@@ -50,50 +50,6 @@ void main() {
     ]).validate();
   });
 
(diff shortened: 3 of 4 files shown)


tusharojha:
@sigurdm PTAL.

lib/src/solver/report.dart
```diff
@@ -143,6 +143,23 @@ class SolveReport {
     }
   }
 
+  /// Displays a single-line message, number of discontinued packages
+  /// if discontinued packages are detected.
+  Future<void> reportDiscontinued() async {
+    var numDiscontinued = 0;
```
sigurdm, line 149: Minor nit: follow the style from reportOutdated.
```
final discontinuedCount = result.packages.where((pkg) => ...).length;
```

lib/src/solver/result.dart
```diff
@@ -109,6 +109,7 @@ class SolveResult {
         SolveReport(type, _sources, _root, _previousLockFile, this, cache);
     report.summarize(dryRun: dryRun);
     if (type == SolveType.UPGRADE) {
+      await report.reportDiscontinued();
```
sigurdm, line 112: Hmm - question is if we only want to do this on `upgrade` or also on `get`? I think we would like to show this always...

@jonasfj WDYT?

sigurdm (commented):
Nice! Thanks for the contribution!

We need a test to test this behavior.
See https://github.com/dart-lang/pub/blob/5aadb70e0a59017968a4898b383798d7d507152e/test/get/hosted/warn_about_discontinued_test.dart for inspiration on how to set up a discontinued package in the test framework.

lib/src/solver/report.dart
tusharojha: Yes, I tried to do it using `where` but the method needs to async which is not possible with it.

tusharojha: Yes, I tried to do it using `where` but the method needs to async which is not possible with it.

tusharojha:
@sigurdm I have added a test case. PTAL.

lib/src/solver/report.dart
sigurdm: Ah! - good point. Did not think of that. Then your solution is  fine.

test/upgrade/dry_run_does_not_apply_changes_test.dart
```diff
@@ -50,6 +50,50 @@ void main() {
     ]).validate();
   });
 
+  test('--dry-run: shows count of discontinued packages', () async {
```
sigurdm, line 53:
```suggestion
  test('Shows count of discontinued packages', () async {
```

sigurdm: I think this test belongs better in for example test/upgrade/report/describes_change_test.dart . It does not really relate to dry-run.

test/upgrade/dry_run_does_not_apply_changes_test.dart
```diff
@@ -50,6 +50,50 @@ void main() {
...
+      args: ['--dry-run'],
+      output: contains('1 package is discontinued.'),
+    );
+
+    await d.dir(appPath, [
+      // The lockfile should be unmodified.
+      d.file('pubspec.lock', contains('2.0.0')),
+      // The ".dart_tool" directory should not have been regenerated.
+      d.nothing('.dart_tool')
+    ]).validate();
```
sigurdm, line 84: I don't think we need this validation in this test. We want to test that we print the count, not how --dry-run behaves.

lib/src/solver/report.dart
```diff
@@ -143,6 +143,23 @@ class SolveReport {
...
+      if (id.source == null) continue;
+      final status =
+          await _cache.source(id.source).status(id, Duration(days: 3));
+      if (status.isDiscontinued) numDiscontinued++;
+    }
+    if (numDiscontinued > 1) {
+      log.message('$numDiscontinued packages are discontinued.');
+    } else if (numDiscontinued == 1) {
+      log.message('1 package is discontinued.');
+    }
```
sigurdm, line 160: I think this would be slightly easier to follow:
```suggestion
    if (numDiscontinued > 0) {
      if (numDiscontinued == 1) {
        log.message('1 package is discontinued.');
      } else {
        log.message('$numDiscontinued packages are discontinued.');
      }
    }
```

tusharojha:
@sigurdm Thanks for the review! 
I have updated the code as per your suggestions.

lib/src/solver/result.dart
sigurdm: We decided for now this is good!

sigurdm (approved):
LGTM

Thanks!

PR #2918, opened 2021-03-15
Added null check for name in UserInfo class

### LINKED ISSUE #2897: userInfo inside pub login command doesn't contain user's name
## Description

I am new here and looking to contribute. Therefore, I stumbled upon the `pub login` command which uses oauth2 to authenticate the user and store their email & name into userInfo. But I tried with two different accounts and both resulted in a `null` for the name field.

Here is the response coming from `userinfo_endpoint`:

```dart
{
sub: 103068153864256870934, 
picture: https://lh3.googleusercontent.com/a-/AOh14GgynQMyqolmPXQ87dTgmpG5PQaLk75IlNlzEf-qUA=s96-c, 
email: tusharojha2001@gmail.com,
email_verified: true
}
``` 
Either we can remove the name property from the `UserInfo` class or the API should respond with the user's name.

tusharojha:
## Description

pub login command which uses oauth2 to authenticate the user and store their email & name into userInfo sometimes returns null for the name.

Fixes #2897

### DIFF
--- a/lib/src/command/login.dart
+++ b/lib/src/command/login.dart
@@ -60,5 +60,5 @@ class _UserInfo {
   final String email;
   _UserInfo(this.name, this.email);
   @override
-  String toString() => '<$email> "$name"';
+  String toString() => '<$email> ${name ?? ''}';
 }
--- a/lib/src/command/login.dart
+++ b/lib/src/command/login.dart
@@ -43,7 +43,7 @@ class LoginCommand extends PubCommand {
       final discovery = await httpClient.get(Uri.https(
           'accounts.google.com', '/.well-known/openid-configuration'));
       final userInfoEndpoint = json.decode(discovery.body)['userinfo_endpoint'];
-      final userInfoRequest = await client.get(userInfoEndpoint);
+      final userInfoRequest = await client.get(Uri.parse(userInfoEndpoint));
       if (userInfoRequest.statusCode != 200) return null;
       try {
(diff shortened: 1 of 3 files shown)


tusharojha:
@sigurdm I have resolved some linter issues in the last commit. PTAL.

lib/src/ignore.dart
```diff
@@ -532,7 +532,7 @@ class _IgnorePrefixPair {
   @override
   String toString() {
     // TODO: implement toString
```
sigurdm, line 534: Let's delete this stray todo
```suggestion
```

lib/src/command/login.dart
```diff
@@ -60,5 +60,5 @@ class _UserInfo {
   final String email;
   _UserInfo(this.name, this.email);
   @override
-  String toString() => '<$email> "$name"';
+  String toString() => '<$email> ${name ?? ''}';
```
sigurdm, line 63: I don't like the extra space we get when name is null. Perhaps:

```suggestion
  String toString() => ['<$email>', if (name != null) name].join(' ');
```

sigurdm (approved):
Looks fine - only we should fix the extra space.

lib/src/ignore.dart
tusharojha: Done.

sigurdm:
Thanks a lot for the contribution!

PR #2919, opened 2021-03-16
Handle relative git-url-paths correctly when --directory

sigurdm:


jonasfj (approved):

lib/src/source/git.dart
```diff
@@ -249,34 +283,43 @@ class BoundGitSource extends CachedSource {
...
           .run(['show', '$revision:$pubspecPath'], workingDir: repoPath);
     } on git.GitException catch (_) {
       fail('Could not find a file named "$pubspecPath" in '
-          '${ref.description['url']} $revision.');
+          '$actualPath $revision.');
     }
 
     return Pubspec.parse(lines.join('\n'), systemCache.sources,
-        expectedName: ref.name);
+        expectedName: ref.name, location: Uri.parse(actualPath));
```
jonasfj, line 322: Let's not add this: `, location: Uri.parse(actualPath)`

lib/src/command/add.dart
```diff
@@ -322,7 +322,7 @@ class AddCommand extends PubCommand {
       }
 
       packageRange = cache.sources['git']
-          .parseRef(packageName, git)
+          .parseRef(packageName, git, containingPath: entrypoint.pubspecPath)
           .withConstraint(constraint ?? VersionConstraint.any);
       pubspecInformation = {'git': git};
```
jonasfj, line 327: `git` contains `'url'` which must be relative to `entrypoint.pubspecPath`.

jonasfj: We have the same problem in `path`.

sigurdm (commented):
Could you PTAL - especially the add-part?

lib/src/global_packages.dart
```diff
@@ -90,9 +90,14 @@ class GlobalPackages {
     // dependencies. Their executables shouldn't be cached, and there should
     // be a mechanism for redoing dependency resolution if a path pubspec has
     // changed (see also issue 20499).
+    PackageRef ref;
+    try {
+      ref = cache.git.source.parseRef(name, {'url': repo}, containingPath: '.');
```
jonasfj, line 95: Technically, we're going to write the `pubspec.yaml` somewhere other than `.`, so the `containingPath` should be something like `$PUB_CACHE/global_packages/<package>/`, shouldn't it?

I'm not sure it matters here?

jonasfj (approved):

lib/src/global_packages.dart
sigurdm: We don't write it to disk (yet), so it should not matter...

PR #2924, opened 2021-03-18
Minor cleanup to reduce risk of using path.current

jonasfj:
This shouldn't have any side-effect, as nobody is calling without `baseDir`, this just stops people from using said footgun in the future.


sigurdm (commented):
LGTM

PR #2927, opened 2021-03-19
Add command to fail if extra arguments are passed

### LINKED ISSUE #2923: "dart pub add" ignores extra options
Example of extra input:
 * `dart pub add --dev test ^1.16.0` this should not work.

The correct way to do this is:
 * `dart pub add --dev test:^1.16.0`

We should probably fail if the constraint is not provided correctly. If we don't understand additional options/arguments we should always fail.
That way we can add additional options/arguments in the future.

tusharojha:
## Description

Added usage exception if extra arguments are passed.

Fixes #2923 

cc @sigurdm @jonasfj 

I am bit unsure whether to put the test case into `add_test.dart` or `invalid_options.dart`. So, as of now it's in `add_test` let me know if its needs to be changed.

### DIFF
--- a/lib/src/command/add.dart
+++ b/lib/src/command/add.dart
@@ -82,6 +82,8 @@ class AddCommand extends PubCommand {
   Future<void> runProtected() async {
     if (argResults.rest.isEmpty) {
       usageException('Must specify a package to be added.');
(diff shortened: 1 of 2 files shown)


sigurdm (approved):
LGTM thanks!

PR #2932, opened 2021-03-23
Vendor package:tar and package:chunked_stream

sigurdm:


lib/src/third_party/chunked_stream/METADATA
```diff
@@ -0,0 +1,17 @@
...
+    "Utilities for working with chunked streams, such as byte streams which is"
+    "often given as a stream of byte chunks with type `Stream<List<int>>`."
+
+third_party {
+  url {
+    type: GIT
+    value: "https://github.com/google/dart-neats/"
+  }
+  version: "1.4.0"
+  last_upgrade_date { year: 2014 month: 10 day: 20 }
```
jonasfj, line 12: Should we remove this?

lib/src/third_party/tar/.github/workflows/dart.yml
```diff
@@ -0,0 +1,44 @@
+name: build
```
jonasfj, line 1: I think we can skip this file...

jonasfj (approved):

lib/src/third_party/chunked_stream/METADATA
sigurdm: Or update it :)

lib/src/third_party/tar/.github/workflows/dart.yml
sigurdm: ++

PR #2935, opened 2021-03-24
Handle package:tar cancellations

jonasfj:
Rolling `pub` into the Dart SDK was reverted, due to an intermittent failure.

I dug into it, and found that `pub get` fails to fetch all dependencies if there is more than 16.
I'm not exactly sure what is going on, I suspect we got some concurrency wrong.

Probably because we don't allow more than 16 concurrent downloads.
Still not sure what caused the queuing to not work as it used to, it's not like we changed anything there.

This PR adds a test that `pub get` can fetch more than 16 dependencies. That's always nice to test, especially when 16 is a magic number wrt. to limiting concurrent downloads.

Culprit from `git bisect`.
```
9a70949e25e9407423c68af5aed0f52b0655fe09 is the first bad commit
commit 9a70949e25e9407423c68af5aed0f52b0655fe09
Date:   Fri Feb 5 13:51:03 2021 +0100

    Use Dart library to read and write tar files (#2817)

 lib/src/io.dart   | 318 ++++++++++++++++++++++--------------------------------
 pubspec.yaml      |   1 +
 test/io_test.dart | 248 ++++++++++++++++++++++++++++++++----------
 3 files changed, 322 insertions(+), 245 deletions(-)
bisect run success
```

FYI @simolus3

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -10,6 +10,7 @@ import 'dart:math' as math;
 
 import 'package:http/http.dart' as http;
 import 'package:http_retry/http_retry.dart';
+import 'package:pedantic/pedantic.dart';
 import 'package:pool/pool.dart';
 import 'package:stack_trace/stack_trace.dart';
 
@@ -382,13 +383,10 @@ class _ThrottleClient extends http.BaseClient {
       rethrow;
     }
 
-    var stream = response.stream.transform(
-        StreamTransformer<List<int>, List<int>>.fromHandlers(
-            handleDone: (sink) {
-      resource.release();
-      sink.close();
-    }));
-    return http.StreamedResponse(stream, response.statusCode,
+    final responseController = StreamController<List<int>>(sync: true);
+    unawaited(response.stream.pipe(responseController));
+    unawaited(responseController.done.then((_) => resource.release()));
+    return http.StreamedResponse(responseController.stream, response.statusCode,
         contentLength: response.contentLength,
         request: response.request,
         headers: response.headers,
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -196,16 +196,6 @@ void writeTextFile(String file, String contents,
 }
 
 /// Creates [file] and writes [contents] to it.
-Future<void> writeBinaryFileFromStream(
-  String file,
-  Stream<List<int>> contents,
-) async {
-  deleteIfLink(file);
-  log.fine('Writing text file $file.');
-  await File(file).openWrite().addStream(contents);
-}
-
-/// Creates [file] and writes [contents] to it.
 ///
 /// If [dontLogContents] is `true`, the contents of the file will never be
 /// logged.
@@ -227,7 +217,7 @@ Future<void> writeTextFileAsync(String file, String contents,
 ///
 /// Replaces any file already at that path. Completes when the file is done
 /// being written.
-Future<String> _createFileFromStream(Stream<List<int>> stream, String file) {
+Future<String> createFileFromStream(Stream<List<int>> stream, String file) {
   // TODO(nweiz): remove extra logging when we figure out the windows bot issue.
   log.io('Creating $file from stream.');
 
@@ -866,7 +856,7 @@ Future extractTarGz(Stream<List<int>> stream, String destination) async {
         // Regular file
         deleteIfLink(filePath);
         ensureDir(parentDirectory);
-        await _createFileFromStream(entry.contents, filePath);
+        await createFileFromStream(entry.contents, filePath);
 
         if (Platform.isLinux || Platform.isMacOS) {
           // Apply executable bits from tar header, but don't change r/w bits
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -170,6 +170,13 @@ List<int> readBinaryFile(String file) {
   return contents;
 }
 
+/// Reads the contents of the binary file [file] as a [Stream].
+Stream<List<int>> readBinaryFileAsSream(String file) {
+  log.io('Reading binary file $file.');
+  var contents = File(file).openRead();
+  return contents;
+}
+
 /// Creates [file] and writes [contents] to it.
 ///
 /// If [dontLogContents] is `true`, the contents of the file will never be
@@ -186,6 +193,16 @@ void writeTextFile(String file, String contents,
 
   deleteIfLink(file);
   File(file).writeAsStringSync(contents, encoding: encoding);
+}
+
+/// Creates [file] and writes [contents] to it.
+Future<void> writeBinaryFileFromStream(
+  String file,
+  Stream<List<int>> contents,
+) async {
+  deleteIfLink(file);
+  log.fine('Writing text file $file.');
+  await File(file).openWrite().addStream(contents);
 }
 
 /// Creates [file] and writes [contents] to it.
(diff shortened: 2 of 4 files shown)


jonasfj:
Traced this to a problem closing streams.. the "done" message is not properly consumed.

In `package:tar` we have logic like this:
```dart
      nextHeader = await _readHeader(rawHeader);
      if (nextHeader == null) {
        if (eofAcceptable) {
          await cancel();
          return false;
        } else {
          _unexpectedEof();
        }
      }
```

We see two empty zero blocks and stops reading. That's because two empty blocks indicate end-of-archive.
Then we cancel, which gets propagated to `StreamIterator` used by `ChunkedStreamIterator`.
But when `StreamIterator` only need to deliver the "done" message, it fails to cancel properly, or so it seems to me at-least.

However, we also should not stop reading the HTTP stream, just because the tar stream contains two zero blocks.
We should not ignore additional data.

We have two options:
 * Parse the remaining data as tar.
 * Throw an error.

In strict mode a tar reader should throw an error, but in lenient mode a tar reader should assume that a tar archive can be a concatenation of two tar files. I think I saw somewhere that this is how other tar readers do. I don't recall the source on that :see_no_evil: 

But we probably have to investigate what to do here.

simolus3:
Thanks for the information! It sounds to me that the cancellation issue comes from `chunked_stream`, but I'll try to reproduce this with a minimal sample (piping three zero blocks into a `TarReader` should cancel after the second) and see at which point it goes wrong.

> However, we also should not stop reading the HTTP stream, just because the tar stream contains two zero blocks.
We should not ignore additional data.

If that's the primary cause here, why does it only happen when there are more than 16 dependencies? As far as I can remember, the comparison script extracting everything with `package:tar` and the system's tar was concurrent as well.

We could implement both workarounds (treat the rest as another archive, emit error) in `package:tar`, but we could also guard against that with a stream transformer in pub. IMHO closing the stream after two zero blocks where a header is expected sounds like the expected behavior from a tar reader to me. It signals the end of an archive, so we should stop listening.

simolus3:
Actually I think `package:tar` should properly cancel the stream? We have [a test](https://github.com/simolus3/tar/blob/b1fe3153da80c3808c3a5a8007613615c1c450eb/test/reader_test.dart#L69-L81) that I think should cover this. I didn't look into the detailed failure of pub, can this be reproduced by just extracting more than 16 tars from a concurrency-limiting client? Please let me know if there's anything in `package:tar` that I could look into.

simolus3:
This can be fixed with the appended patch. Whether that's a good solution or not is not my call.

I think the `StreamTransformer.fromHandlers` might not catch cancellations properly? What's probably happening here is

1. We receive two empty null blocks where a header was expected
2. We then close the reader (during local debugging, I found that this does cancel the subscription)
3. The response stream is then done, but that event gets lost since there isn't any subscription (this is speculation)

We can use a `StreamController` to work around the last issue.

```patch
diff --git a/lib/src/http.dart b/lib/src/http.dart
index 29b35258..ee8b6bd2 100644
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -10,6 +10,7 @@ import 'dart:math' as math;
 
 import 'package:http/http.dart' as http;
 import 'package:http_retry/http_retry.dart';
+import 'package:pedantic/pedantic.dart';
 import 'package:pool/pool.dart';
 import 'package:stack_trace/stack_trace.dart';
 
@@ -382,13 +383,11 @@ class _ThrottleClient extends http.BaseClient {
       rethrow;
     }
 
-    var stream = response.stream.transform(
-        StreamTransformer<List<int>, List<int>>.fromHandlers(
-            handleDone: (sink) {
-      resource.release();
-      sink.close();
-    }));
-    return http.StreamedResponse(stream, response.statusCode,
+    final responseController = StreamController<List<int>>(sync: true);
+    unawaited(response.stream.pipe(responseController));
+    unawaited(responseController.done.then((_) => resource.release()));
+
+    return http.StreamedResponse(responseController.stream, response.statusCode,
         contentLength: response.contentLength,
         request: response.request,
         headers: response.headers,
```

jonasfj:
Clearly, it's also a problem if `StreamTransformer.fromHandlers` doesn't handle cancellations.

jonasfj:
> IMHO closing the stream after two zero blocks where a header is expected sounds like the expected behavior from a tar reader to me. It signals the end of an archive, so we should stop listening.

From [GNU Tar Manual: Basic Tar Format](https://www.gnu.org/software/tar/manual/html_node/Standard.html):

> When reading an archive, a reasonable system should properly handle an archive whose last record is shorter than the rest, or which contains garbage records after a zero block.

So yes, I agree, it is reasonable to close the stream and stop listening.
However, if the two blocks happens right before end-of-stream, it's certainly "nicer" to read the end-of-stream and let the stream close correctly. When streaming over HTTP this is often preferable, as cancelling the response (if supported) probably involves terminating the TCP connection. Hence, the connection pool won't be able to reuse the underlying TCP connection for subsequent requests.

If after the two zero blocks there is a lot of garbage, then canceling is probably desired.

simolus3:
> Clearly, it's also a problem if `StreamTransformer.fromHandlers` doesn't handle cancellations.

This can be reproduced with a very minimal sample, maybe you can ask around if that's supposed to happen or if there's a better workaround:

```dart
import 'dart:async';

void main() {
  final controller = StreamController<int>(onCancel: () => print('cancelled'));
  final stream = controller.stream
      .transform<int>(StreamTransformer.fromHandlers(handleDone: (sink) {
    print('done'); // this isn't called
    sink.close();
  }));

  late StreamSubscription<int> sub;
  sub = stream.listen((event) {
    sub.cancel();
  });

  controller.add(3);
}
```

> However, if the two blocks happens right before end-of-stream, it's certainly "nicer" to read the end-of-stream and let the stream close correctly

I agree, but it's hard to come up with a general solution for this. What if we have

1. Two zero blocks
2. No event for 2 seconds
3. End of stream / further data

Should `TarReader.hasNext` take two seconds to complete here, when it could just return false after 1? If so, should we add a general timeout for those scenarios? Something like "wait for one event loop iteration" sounds like a reasonable default, but it doesn't feel right either.
But I definitely agree that we shouldn't close ready TCP sockets just because of this, in most cases we're unlikely to have any delay since the size is known from the response.

sigurdm (approved):

lib/src/http.dart
```diff
@@ -382,13 +383,10 @@ class _ThrottleClient extends http.BaseClient {
...
-    var stream = response.stream.transform(
-        StreamTransformer<List<int>, List<int>>.fromHandlers(
-            handleDone: (sink) {
-      resource.release();
-      sink.close();
-    }));
-    return http.StreamedResponse(stream, response.statusCode,
+    final responseController = StreamController<List<int>>(sync: true);
+    unawaited(response.stream.pipe(responseController));
+    unawaited(responseController.done.then((_) => resource.release()));
```
lrhn, line 388: Consider using `.whenComplete(resource.release)`. It works even if the `done` event contains an error.

lrhn:
About the done event, the behavior is correct. You get no events after calling `cancel`, and the done event is an event, so the `onDone` handler of a subscription won't (and *must not*) be called after you have called `cancel` on the subscription.
The done event represents a *completion* of the stream. It is only sent if the stream has reached its natural end.

PR #2959, opened 2021-04-13
Use relative imports

cazci:
I've changed the import as Dart prefers relative imports for files in lib. Hope this helps, :)


jonasfj (approved):

sigurdm:
Thanks!

PR #2963, opened 2021-04-16
forbid pushing .pub_cache

n0npax:
return error if the user tries to push package which contains `.pub_cache`


jonasfj:
I think we already exclude dot-files/folders by default.

Issue #2962, opened 2021-04-15
Improve invalidation of precompiled executable snapshots


natebosch:
Currently (I think) we invalidate snapshots on every `pub get`, which is pessimistic. If a user adds a dependency on some package which is not a transitive dep on `test`, it would be nice if they don't need to wait again for the `Precompiling` step the next time they run tests.

We could potentially calculate the versions of the transitive deps of the specific package with the executable into a hash, and store a manifest with that hash. After a pub get we can recompute the hash and throw away the snapshot only if it has changed.

cc @mit-mit @grouma

mit-mit:
Currently you don't even need to edit your pubspec and add a new dep. -- simply running `pub get` seems ti invalidate:

```console
$ dart pub get
Resolving dependencies...
Got dependencies!

$ dart test
Precompiling executable... (10.5s)
Precompiled test:test.
00:01 +1: All tests passed!

$ dart pub get
Resolving dependencies...
Got dependencies!

$ dart test
Precompiling executable... (10.4s)
Precompiled test:test.
00:01 +1: All tests passed!
```

mit-mit:
cc @sigurdm @jonasfj

mit-mit:
This seems to have been resolved (I'm using `dart` from todays Flutter master):

```
mit-macbookpro4:tmp mit$ dart create -t console-full c1
Creating c1 using template console-full...
...
mit-macbookpro4:tmp mit$ cd c1
mit-macbookpro4:c1 mit$ dart test
Precompiling executable... (8.9s)
Precompiled test:test.
00:01 +1: All tests passed!
mit-macbookpro4:c1 mit$ dart test
00:01 +1: All tests passed!
mit-macbookpro4:c1 mit$ dart test
00:01 +1: All tests passed!

```

Was this resolved as part of https://github.com/dart-lang/pub/pull/2968, @jakemac53  ?

jakemac53:
In order to trigger the new behavior you need to run a pub get in between test runs - that is what triggers a recompile. It still does trigger a recompile but it will be much faster now. So this issue isn't directly solved but it is mitigated significantly enough that I would say we can close it personally.

mit-mit:
Got it. That is still pretty fast:

```
$ dart test
Precompiling executable... (1.9s)
Precompiled test:test.
00:01 +1: All tests passed!

$ dart pub get
Resolving dependencies...
  analyzer 1.7.0 (1.7.1 available)
Got dependencies!

$ dart test
[...]

PR #2968, opened 2021-04-20
Use the frontend server to compile pub executables

jakemac53:
Opening for discussion, I wanted to try this out to see if we can speed up the rebuilds of pub executables - it does appear to be much faster (basically immediate if there is no change).  Surprisingly, it is almost twice as fast on the initial precompile as well for me. I think that probably means that the vm snapshot functionality is doing something additional - I don't know what the implications of using kernel instead might be.

Specifically this:

- Creates a new dir `.dart_tool/pub/incremental` where it stores the previous dill outputs for each executable.
- Uses the frontend server to compile to that directory (which will use those files automatically to initialize from).
- Copies the output from there to the actual executable path if successful.

Previous snapshot behavior:
snapshot

New fe server behavior, first build:
kernel_first

New behavior, pub get and run again (no changes):
kernel_incremental

Note that in the rebuild case with no changes, it didn't actually log the time at all which I believe means it was <1 second.

cc @mit-mit

### DIFF
--- a/lib/src/dart.dart
+++ b/lib/src/dart.dart
@@ -16,6 +16,7 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/error/error.dart';
 import 'package:cli_util/cli_util.dart';
 import 'package:frontend_server_client/frontend_server_client.dart';
+import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 
 import 'exceptions.dart';
@@ -34,41 +35,6 @@ bool isEntrypoint(CompilationUnit dart) {
         node.name.name == 'main' &&
         node.functionExpression.parameters.parameters.length <= 2;
   });
-}
-
-/// Snapshots the Dart executable at [executablePath] to a snapshot at
-/// [snapshotPath].
-///
-/// If [packageConfigFile] is passed, it's used to resolve `package:` URIs in
-/// the executable. Otherwise, a `packages/` directory or a package spec is
-/// inferred from the executable's location.
-///
-/// If [name] is passed, it is used to describe the executable in logs and error
-/// messages.
-Future snapshot(
-  String executablePath,
-  String snapshotPath, {
-  String packageConfigFile,
-  String name,
-}) async {
-  final args = [
-    if (packageConfigFile != null) '--packages=$packageConfigFile',
-    '--snapshot=$snapshotPath',
-    p.toUri(executablePath).toString()
-  ];
-
-  final result = await runProcess(Platform.executable, args);
-  final highlightedName = name = log.bold(name ?? executablePath.toString());
-  if (result.success) {
-    log.message('Precompiled $highlightedName.');
-  } else {
-    // Don't leave partial results.
-    deleteEntry(snapshotPath);
-
-    throw ApplicationException(
-        log.yellow('Failed to precompile $highlightedName:\n') +
-            result.stderr.join('\n'));
-  }
 }
 
 class AnalysisContextManager {
@@ -176,9 +142,23 @@ class AnalyzerErrorGroup implements Exception {
   String toString() => errors.join('\n');
 }
 
-Future<void> precompile(
-    String executablePath, String outputPath, String incrementalDillOutputPath,
-    {String packageConfigFile, String name}) async {
+/// Precompiles the Dart executable at [executablePath] to a kernel file at
+/// [outputPath].
+///
+/// This file is also cached at [incrementalDillOutputPath] which is used to
+/// initialize the compiler on future runs.
+///
+/// The [packageConfigPath] should point at the package config file to be used
+/// for `package:` uri resolution.
+///
+/// The [name] is used to describe the executable in logs and error messages.
+Future<void> precompile({
+  @required String executablePath,
+  @required String incrementalDillOutputPath,
+  @required String name,
+  @required String outputPath,
+  @required String packageConfigPath,
+}) async {
   ensureDir(p.dirname(outputPath));
   ensureDir(p.dirname(incrementalDillOutputPath));
   const platformDill = 'lib/_internal/vm_platform_strong.dill';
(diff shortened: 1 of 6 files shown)


lib/src/entrypoint.dart
```diff
@@ -373,6 +380,15 @@ class Entrypoint {
             '${p.basename(executable.relativePath)}-$versionSuffix.snapshot');
   }
 
+  String incrementalDillPathOfExecutable(Executable executable) {
+    assert(p.isRelative(executable.relativePath));
```
kevmoo, line 384: Seems like this should be an assert *in* `Executable`

jakemac53:
Possible resolution to https://github.com/dart-lang/pub/issues/2962 as well - in the case of no changes the precompile step is now much faster.

jakemac53:
I will look more into the windows failures and do additional cleanup if we want to push forward with this change.

jakemac53:
cc @a-siva would you be able to speak to the implications of running from kernel versus snapshot?

a-siva:
Could you clarify what kind of 'snapshot' you are referring to here? Compile Dart to various formats.

Usage: dart compile <subcommand> [arguments]

Available subcommands:
  aot-snapshot   Compile Dart to an AOT snapshot.
  exe            Compile Dart to a self-contained executable.
  jit-snapshot   Compile Dart to a JIT snapshot.
  js             Compile Dart to JavaScript.
  kernel         Compile Dart to a kernel snapshot.
(aot-snapshot, jit-snapshot or just plain kernel which is also sometimes referred to as a snapshot).

jakemac53:
> Could you clarify what kind of 'snapshot' you are referring to here?

It is just `--snapshot` - that must be an alias for some other listed flag? See https://github.com/dart-lang/pub/pull/2968/files#diff-32ea3c3bc788ef583688a17f090b6addf54968f08c16ff62b510c1166ae52aa3R56

a-siva:
if you are just using the --snapshot option and not passing in the --snapshot-kind option then the default is kernel
```
--snapshot-kind=<snapshot_kind>
--snapshot=<file_name>
  These snapshot options are used to generate a snapshot of the loaded
  Dart script:
    <snapshot-kind> controls the kind of snapshot, it could be
                    kernel(default) or app-jit
    <file_name> specifies the file into which the snapshot is written
```
which means you are already only producing a kernel file in this step.

jakemac53:
Ok thanks @a-siva so there should be no difference in terms of the result then.

I do see some slight difference in output, the md5 hash is different and the new result is a bit smaller - (18016104 bytes versus 18633360 bytes).

a-siva:
Is it possible that the version of CFE used in the two cases is different.

jakemac53:
@jonasfj @sigurdm anything you are waiting for me on here? If I can get a general affirmative I will spend some more time cleaning things up (get tests passing, delete the old snapshot function, etc)

jonasfj:
@jakemac53 looks great to me.. I'm not familiar with `FrontendServerClient`, but I assume it incremental compilation infrastructure...

Only concern is how this will interact when users switch SDK versions forwards and backwards... But that might not be so bad.. also it might affect global packages too in the same way.

jakemac53:
> @jakemac53 looks great to me.. I'm not familiar with `FrontendServerClient`, but I assume it incremental compilation infrastructure...

Yes, this is the same tool that flutter uses for incremental recompilation.

> Only concern is how this will interact when users switch SDK versions forwards and backwards... But that might not be so bad.. also it might affect global packages too in the same way.

The way I have implemented this it should all just work. I haven't changed the invalidation logic around the actual runnable executables, so they still get invalidated (deleted) in the same way as before.

When running from a cached dill file, the frontend server automatically throws it away if its kernel version doesn't match afaik. At least in some way it does handle this case and does a full recompile.

jakemac53:
@jonasfj the remaining test failure is here https://github.com/dart-lang/pub/pull/2968/checks?check_run_id=2442925564#step:7:415 (same failing test on the 3 bots).

```
SocketException: Failed host lookup: 'arg.com' (OS Error: Temporary failure in name resolution, errno = -3)
```

Do you have any more details/info on this test?

jakemac53:
This is ready for review now @jonasfj @sigurdm

lib/src/dart.dart
```diff
@@ -174,3 +141,52 @@ class AnalyzerErrorGroup implements Exception {
...
+    packagesJson: packageConfigPath,
+    printIncrementalDependencies: false,
+  );
+  try {
+    var result = await client.compile();
+
+    final highlightedName = log.bold(name);
+    if (result?.errorCount == 0) {
+      log.message('Precompiled $highlightedName.');
+      await File(incrementalDillOutputPath).copy(outputPath);
```
jonasfj, line 180: So `dart run test:test` will create `.dart_tool/pub/incremental/test/test.dart.incremental.dill` and then that file will be copied to `.dart_tool/pub/bin/test/test.dart-x.y.z.snapshot` ?

I guess that's fine, we could also consider running the recompilation step whenever `.dart_tool/package_config.json` has a modification time newer than `.dart_tool/pub/incremental/test/test.dart-x.y.z.dill`.
Then we won't have to store things in `.dart_tool/pub/bin/` anymore.

lib/src/entrypoint.dart
```diff
@@ -373,6 +379,15 @@ class Entrypoint {
...
+    return isGlobal
+        ? p.join(_incrementalDillsPath,
+            '${p.basename(executable.relativePath)}.incremental.dill')
+        : p.join(_incrementalDillsPath, executable.package,
+            '${p.basename(executable.relativePath)}.incremental.dill');
```
jonasfj, line 388: Are we sure we don't want to inject SDK version number into the filename... like `.dart-a.b.c.`?

How will behave with SDK upgrades, or downgrades?

jakemac53: When upgrading/downgrading the sdk it still does a full recompile internally (I believe this is detected by the kernel version of the incremental dill) - cc @jonahwilliams can maybe confirm?

jonasfj: But will it throw an error? or just overwrite the old invalid `.dill` file?

jakemac53: We will just overwrite the old file with a new one, it handles it gracefully.

jakemac53: It would be ideal to have some sort of test around this to ensure it continues to work as expected.

I am not sure what the easiest way would be to actually write that test (how would we detect "failure" consistently?).

We could keep files around with sdk version numbers in them but we would want some way of cleaning up old ones in that case. Maybe just remember the previous sdk version (store it in a file?) and then bash away the entire `incremental` dir if it changes?

jonasfj: I think such tests would be fun to do... but we don't have that at the moment, so let's do that in a separate PR another day.

jonasfj (approved):

jakemac53:
@jonasfj This is ready to land from my perspective, but I will let you hit the button unless you want me to :D.

PR #2975, opened 2021-04-26
Update to analyzer 1.5.0, migrate from deprecated AnalysisSession.getParsedUnit()

scheglov:


scheglov:
There are failing tests, but I'm surprised that my changes caused them, e.g. `SocketException: Failed host lookup: 'arg.com'`. Are there pre-existing failures in testing, or can you think of a way my changes affected it?

natebosch:
@jonasfj @sigurdm - do you have background on the failing tests?

natebosch:
This does look like it's failing on master too so I think it's probably safe to land with a failing CI, but I'll defer to @jonasfj for that decision.

scheglov:
It seems that the tests are green now.

jonasfj (approved):

Issue #2939, opened 2021-03-30
Roll pub into dart-sdk


jonasfj:
We've recently added dependency on `package:tar` to avoid using system tar implementation.
This means migrating into Dart SDK will involve a bit of moving around with dependencies.

Specifically:
 * [x] Land `ChunkedStreamReader` in `package:async` (https://github.com/dart-lang/async/pull/161)
 * [x] Migrate `package:tar` to use `ChunkedStreamReader` from `package:async`
 * [x] Vendor `https://github.com/simolus3/tar/` in `pub`
 * [x] Re-land pub in Dart SDK ([sdk/+/192304](https://dart-review.googlesource.com/c/sdk/+/192304))
 * [x] Stop packaging 7zip with Dart SDK on Windows.

simolus3:
I've just published (and tagged) version 0.3.3 of `package:tar`. It only depends on the `async`, `meta` and `typed_data` packages, which are available in the SDK repo.

jonasfj:
Filed https://dart-review.googlesource.com/c/sdk/+/199040

jonasfj:
https://dart-review.googlesource.com/c/sdk/+/199250

jonasfj:
Starting 2.14 Dart should no longer include 7zip on windows, and not system tar on linux :D

With a little luck it'll be in the first 2.14 beta :rocket:

PR #2987, opened 2021-05-06
Vendor tar

jonasfj:
Vendoring `package:tar` in line with: https://opensource.google/docs/thirdparty/non-google3/#i-need-to-include-thirdparty-source-code-in-a-dart-package

This avoids external dependencies in Dart SDK, and allows us to move forward on https://github.com/dart-lang/pub/issues/2939

### DIFF
--- a/bin/pub.dart
+++ b/bin/pub.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// @dart=2.10
+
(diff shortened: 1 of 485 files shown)


athomas (approved):
LGTM, a bit weird to batch these two things together, perhaps we shouldn't squash this when we land it.

jonasfj:
Split out the partial null-safety migration into https://github.com/dart-lang/pub/pull/2988 -- I agree, it's better for history if we keep the separate.

Issue #3000, opened 2021-05-19
Avoid using `pub` from binstubs


jonasfj:
Stubs in `PUB_CACHE/bin/` currently call into `pub`, when it would be better to use `dart pub`.

I think we did this because the migration to `dart pub` wasn't complete at the time.
If we do this, it'll make migrating back to older Dart SDKs harder, as it might break the bin stubs, but that seems acceptable since I hope people aren't downgrading to `2.9.x` anymore.

This would fix https://github.com/dart-lang/sdk/issues/42530

jonasfj:
Also fixes https://github.com/dart-lang/pub/issues/2981

PR #3002, opened 2021-05-19
New binstubs for global activate using `dart pub`

jonasfj:
Fixes https://github.com/dart-lang/pub/issues/3000, downside is that if you want to downgrade from Dart 2.14.x to Dart 2.9.x, then packages activated with `dart pub global activate <package>` won't work anymore.

Upside is that now `dart pub global activate <package>` will work for users who have installed Dart from the Flutter SDK.

----

It's still be messy if you have two Dart SDK installations as Flutter in some configurations still use a different `PUB_CACHE`.

### DIFF
--- a/lib/src/global_packages.dart
+++ b/lib/src/global_packages.dart
@@ -756,9 +756,9 @@ if exist "$snapshot" (
   if not errorlevel 253 (
     goto error
   )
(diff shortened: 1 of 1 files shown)


athomas (approved):
LGTM

Issue #2999, opened 2021-05-19
Do not store credentials in PUB_CACHE


jonasfj:
The `PUB_CACHE` folder is something that it makes sense to cache in CI, so storing credentials in `PUB_CACHE/credentials.json` might not be ideal.

On windows we prefer to use a non-roaming folder for `PUB_CACHE`, but configuration options such as `credentials.json` should probably be roaming with the user profile.

See also: https://github.com/dart-lang/sdk/issues/41560

jonasfj:
Counter argument do like `dartdev` does and store it in `~/.dart/`, as determined in:
https://github.com/dart-lang/sdk/blob/5caa4365d75a82ba2f3a702c5f426a719256f9de/pkg/dartdev/lib/src/analytics.dart#L97-L110

This is messy, and IMO we should clean it up. But until we do, perhaps it's better to just expand the mess rather than storing files in more locations.

PR #3007, opened 2021-05-20
Third party hosted pub registry authentication

### LINKED ISSUE #1381: Support authenticating pub against 3rd-party servers

themisir:
Closes #1381

This PR adds support to pub CLI to authenticate with third party hosted pub servers. Before that only official pub servers: pub.dev (aka: pub.dartlang.org) could be authenticated and the authentication was only applied when publishing packages. We usually need to share some pieces of code across company projects. In other ecosystems like npm, nuget, docker this could be done using private repositories / servers. Private repositories does needs authentication for both pulling and pushing data to it - unlike pub's current behavior which only does authenticates requests for pushing new packages.

## CLI Interface

```shell
dart pub login                      # Will initiate login-flow for pub.dev
dart pub logout                     # Will delete credential.json

dart pub token list                # Will list servers for which a token exists
dart pub token add <hosted_url>    # Will prompt for a token for <hosted_url> and store both
dart pub token remove <hosted_url> # Will remove token for <hosted_url>
dart pub token remove --all        # Will delete all tokens stored
```


lib/src/http.dart
```diff
@@ -241,7 +244,13 @@ class _ThrowingClient extends http.BaseClient {
 /// The HTTP client to use for all HTTP requests.
 final httpClient = _ThrottleClient(
     16,
-    _ThrowingClient(RetryClient(_pubClient,
+    _ThrowingClient(RetryClient(
+        AuthenticationClient(
+          _pubClient,
+          // TODO(themisir): Figure out a way to resolve system cache from
+          // somewhere else
+          credentialStore: CredentialStore(SystemCache()),
```
TheMisir, line 252: @jonasfj The easiest and DRY way of adding authentication layer to current implementation is by adding another layer to httpClient.

But the issue is I need to access `SystemCache` instance in order to get `SystemCache.rootDir` which later used to locate `tokens.json` file. For now I'm just instantiating a new SystemCache instance just to make sure things works. But later on this might cause some issues, especially during testing. 

Do you have any ideas on that?

lib/src/command/logout.dart
```diff
@@ -16,12 +16,16 @@ class LogoutCommand extends PubCommand {
...
+  bool get takesArguments => true;
 
   LogoutCommand();
 
   @override
   Future<void> runProtected() async {
-    oauth2.logout(cache);
+    if (argResults.rest.isEmpty) {
+      oauth2.logout(cache);
+    } else {
```
jonasfj, line 27: You can't just take arguments and then throw them away... If there is multiple arguments, but you only accept one then we have to throw a usage error..

TheMisir: Actually I thought about that, but then forgot to write it 😅 Anyways done.

lib/src/http.dart
jonasfj: You have other potential problems... maybe... What happens if `pub get` for a `pubspec.yaml` like this:
```yaml
dependencies:
  foo:
    hosted:
      name: foo
      url: https://some-pub-server.definitely-not-evil.com
    version: ^1.2.3
```

Then pub will make a request to `https://some-pub-server.definitely-not-evil.com/api/packages/foo` which will return a list of package versions. This list will ALSO include `archive_url` from which a package can be downloaded.

Now if `archive_url: https://your-internal-private-pub-server.com/bar-1.0.0.tar.gz`, then the innocent looking `https://some-pub-server.definitely-not-evil.com` have tricked you into using the credentials you have for `https://your-internal-private-pub-server.com` to request a file and download it to your machine.

Sure, nothing has been leaked, and the credential is still safe, the `https://some-pub-server.definitely-not-evil.com` server doesn't even have access to neither the credential or the package that you downloaded -- BUT: your credential was used in a surprising manner, that's a security vulnerability. Who knows how someone could possibly leverage that in the future.

So we might need to re-think this. I think we started digging into it in:
https://github.com/dart-lang/pub/pull/2588

Maybe instead of exposing a `httpClient` client from `http.dart` we should expose something much much simpler, like:
`fetch(String method, String url, String hostedServerUrlContext, List<int> body)`
Or something like that... We could have maybe `fetch` and `fetchStream`, and maybe that's all we need...?

In the example above `hostedServerUrlContext` is an indicator that we're in a "trust context" that belongs to a specific server.. that trust context should NEVER be allowed to trick us into using a token that belongs to a different server, is that not reasonable? Maybe there is a better way of expressing _trust context_.

How do we avoid that these tokens are exploited by third-party servers to do something unexpected? Perhaps it's safer and simpler to wrapper all HTTP usage with 2 methods or so... And perhaps this is something we need to do first? because it's essentially refactoring to avoid using `http.Client`.

doc/authentication-proposal.md
```diff
@@ -0,0 +1,114 @@
...
+asks for it using a modified **pub login** command. By default if no argument is
+presented, **pub login** will behave as previously: authenticate with Google.
+But if you you specify server to login, it will try to authenticate with the
+given server instead.
+
+```plain
+pub login https://myspuberver.dev
+```
+
+To discover authentication method, client will send **GET** request to `/login`
```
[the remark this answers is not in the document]
TheMisir: But what if `/publish` endpoint utilizes some sort of resources? Everytime we send request to that endpoint, we can create some garbage for the repository.

For example a server might add an entry to the "UploadRequests" table on every request which might be used to generate unique URL for the upload using models auto increment ID.

lib/src/authentication/credential.dart
```diff
@@ -0,0 +1,42 @@
...
+    return <String, dynamic>{
+      'method': authenticationType,
+      'credentials': toMapInternal(),
+    };
+  }
+
+  /// Authentication type of this credential.
+  String get authenticationType;
+
+  /// Add required details for authentication to [request].
```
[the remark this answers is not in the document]
TheMisir: We can. But this will limit some possibilities like modifying custom request headers for new credential types in future (I don't know what that might be, but it's there just in case we need it).

If you think that's unnecessary I'll modify it to something like `Future<String> getAuthenticationHeader()` which just returns header value. (Future because some credentials might resolve header value asynchronously, eg: by requesting from special endpoint that generates unique token for that request: like OAuth2 token refresh) .

lib/src/command/login.dart
```diff
@@ -40,7 +73,7 @@ class LoginCommand extends PubCommand {
     }
   }
 
-  Future<_UserInfo> retrieveUserInfo() async {
+  Future<_UserInfo> _retrieveUserInfo() async {
```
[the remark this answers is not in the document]
TheMisir: > Are you just making this private because it's not used anywhere else?

Actually I saw other methods are private so I thought you might forget to make this one too.

lib/src/http.dart
TheMisir: Actually the credentials only added to requests which url matches to registered server url. I'm checking it there: https://github.com/dart-lang/pub/pull/3007/files#diff-1b5bb007bd95e90db7c52eb49f5e0e15069c373e8350ab97c8d2bc2ccda2319bR87

jonasfj: That insufficient, because one server can return urls that point somewhere else... Look at archive_url in API spec..

lib/src/command/logout.dart
```diff
@@ -16,12 +16,18 @@ class LogoutCommand extends PubCommand {
...
 
   LogoutCommand();
 
   @override
   Future<void> runProtected() async {
-    oauth2.logout(cache);
+    if (argResults.rest.isEmpty) {
+      oauth2.logout(cache);
+    } else if (argResults.rest.length > 1) {
+      usageException('Takes only a single argument.');
```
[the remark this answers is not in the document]
TheMisir: > It's also possible we should consider making it dart pub login/logout --server <server> just because it might read better in the CLI documentation when probing with --help.

Done ✅

lib/src/command/login.dart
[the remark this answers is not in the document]
TheMisir: Done ✅

https://github.com/dart-lang/pub/pull/3010

lib/src/command/login.dart
```diff
@@ -21,16 +23,47 @@ class LoginCommand extends PubCommand {
...
 
-  LoginCommand();
+  String get token => argResults['token'];
+  bool get tokenStdin => argResults['token-stdin'];
+
+  LoginCommand() {
+    argParser.addOption('token', help: 'Authorization token for the server');
+
+    argParser.addFlag('token-stdin',
+        help: 'Read authorization token from stdin stream');
```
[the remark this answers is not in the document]
TheMisir: Didn't get that part actually? Should we prompt for token when `pub get`'ing from unauthenticated repository?

I added `--token-stdin` flag to make sure it could be used for piping credentials to pub without exposing to terminal emulator. I think it's mostly useful for CI/CD systems.

Check out: https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin

lib/src/authentication/bearer.dart
```diff
@@ -0,0 +1,27 @@
+import 'dart:io';
+
+import 'package:http/http.dart' as http;
+
+import 'credential.dart';
+
+class BearerCredential extends Credential {
```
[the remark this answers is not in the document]
TheMisir: The class itself is now documented. The members overrides Credential members which is already documented.

lib/src/authentication/authentication_client.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:http/http.dart' as http;
+import 'credential_store.dart';
+
+class AuthenticationClient extends http.BaseClient {
```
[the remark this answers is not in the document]
TheMisir: ✅ Done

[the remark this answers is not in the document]
TheMisir: Didn't get this. Isn't it implemented same as clients like this one? :

Screen Shot 2021-05-22 at 2 12 03 AM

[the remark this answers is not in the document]
TheMisir: ✅ Done

doc/authentication-proposal.md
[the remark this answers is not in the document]
jonasfj: hmm, what if there was a `/server-info.json` file, and it contains the information (or something like this).

It's easy to write a JSON file to an S3 bucket (or GCS bucket)... that could point contain how to authenticate and how to publish in a programmatic form.

I imagine that if you're writing a custom server it might be attractive to simply use an S3 bucket, and point the `dart pub` client at an S3 bucket. This is not fully conceived yet. But adding a `/login` end-point would make this harder.

--------

I need to find a little time to figure this out..

lib/src/authentication/client.dart
```diff
@@ -0,0 +1,66 @@
...
+}
+
+Future<T> withAuthenticatedClient<T>(
+  SystemCache systemCache,
+  String server,
+  Future<T> Function(http.Client) fn, {
+  List<String>? alsoMatches,
+}) {
+  final store = CredentialStore(systemCache);
+  final match = store.getCredential(server, alsoMatches: alsoMatches);
```
jonasfj, line 49:
```suggestion
  final credential = store.getCredential(server, alsoMatches: alsoMatches);
```

lib/src/authentication/client.dart
```diff
@@ -0,0 +1,66 @@
...
+
+  @override
+  void close() => _inner.close();
+}
+
+Future<T> withAuthenticatedClient<T>(
+  SystemCache systemCache,
+  String server,
+  Future<T> Function(http.Client) fn, {
+  List<String>? alsoMatches,
```
jonasfj, line 46: This probably needs a better name, I'm not following?

lib/src/authentication/credential_store.dart
```diff
@@ -0,0 +1,105 @@
...
+  void removeServer(String url) {
+    serverCredentials.removeWhere((key, value) => serverKeyMatches(key, url));
+    _save();
+  }
+
+  /// Returns pair of credential and server key for server that [url] and
+  /// [alsoMatches] matches to server key.
+  Pair<String, Credential>? getCredential(String url,
+      {List<String>? alsoMatches}) {
+    for (final key in serverCredentials.keys) {
```
jonasfj, line 45: I think words like `key` and `serverKey` are unfortunate here.. might be best to call it `serverUrl`, `serverBaseUrl` or something like that.

In a "credential store" I associate "key" with a secret private/public key. In this case it's more like _server identifier_.

lib/src/source/hosted.dart
```diff
@@ -557,22 +564,28 @@ class BoundHostedSource extends CachedSource {
...
       var archivePath =
           p.join(tempDirForArchive, '$packageName-$version.tar.gz');
-      var response = await httpClient.send(http.Request('GET', url));
+      var response = await withAuthenticatedClient(
+        systemCache,
+        server,
+        (client) {
+          return client.send(http.Request('GET', url));
+        },
+        alsoMatches: [archivePath],
```
jonasfj, line 587: What is this?

lib/src/source/hosted.dart
```diff
@@ -557,22 +564,28 @@ class BoundHostedSource extends CachedSource {
...
     await withTempDir((tempDirForArchive) async {
       var archivePath =
           p.join(tempDirForArchive, '$packageName-$version.tar.gz');
-      var response = await httpClient.send(http.Request('GET', url));
+      var response = await withAuthenticatedClient(
+        systemCache,
+        server,
+        (client) {
+          return client.send(http.Request('GET', url));
+        },
```
jonasfj, line 586: nit: :smile: 
```suggestion
        (client) => client.send(http.Request('GET', url)),
```

lib/src/source/hosted.dart
TheMisir: I'll probably replace `alsoMatches` with just `serverUrls` argument that accepts list of strings. It will be used to choose a credential that matches all the server urls. It's there to ensure credentials are not leaked to 3rd parties when archives are not hosted on the same authenticated server.

..And I just noticed that, `archivePath` has to be `url` instead, sorry for that.

lib/src/authentication/client.dart
TheMisir: Removed `alsoMatches` argument in favour of adding a check for `serverBaseUrl` matching with request.url [here](https://github.com/dart-lang/pub/pull/3007/files#diff-38bbc4e7b857842267a32eefea9dce429375c7045d6509d89437b15bbd70cbffR37).

lib/src/authentication/client.dart
```diff
@@ -0,0 +1,70 @@
...
+
+import '../http.dart';
+import '../system_cache.dart';
+import 'credential.dart';
+import 'credential_store.dart';
+
+/// This client automatically modifies request to contain required credentials
+/// in request. For example some credentials might add `Authentication` header
+/// to request.
+class _AuthenticationClient extends http.BaseClient {
```
jonasfj, line 17:
```suggestion
class _AuthenticatedClient extends http.BaseClient {
```

lib/src/authentication/credential.dart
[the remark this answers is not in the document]
jonasfj: There is a strong argument to be made that authentication should never be done by means other than an `Authentication` header.

I can see the argument that you might want to have access to the HTTP method and URL being requested... If the authentication mechanism is signing a JWT or HMAC (like AWS signatures v4).

But attaching other headers or using a custom `x-key` header is a very bad idea, because proxies and other tooling is often aware of the `authentication` header, and won't write these to logs, error reports, etc.

lib/src/source/hosted.dart
```diff
@@ -220,7 +221,11 @@ class BoundHostedSource extends CachedSource {
     try {
       // TODO(sigurdm): Implement cancellation of requests. This probably
       // requires resolution of: https://github.com/dart-lang/sdk/issues/22265.
-      bodyText = await httpClient.read(url, headers: pubApiHeaders);
+      bodyText = await withAuthenticatedClient(
+        systemCache,
```
jonasfj, line 225: We need to refactor this a bit.. what if the `CredentialsStore` is a property on `SystemCache`?

Here we're loading a file and parsing JSON for every single outgoing request :see_no_evil:

lib/src/source/hosted.dart
```diff
@@ -220,7 +221,11 @@ class BoundHostedSource extends CachedSource {
...
+        url.toString(),
+        (client) => client.read(url, headers: pubApiHeaders),
+      );
```
jonasfj, line 228: So if you use token-authentication for `pub.dev` we'll be adding a user identifiable authentication token to all requests..

Authenticating requests that don't need it could be undesirable.. for multiple reasons:
 * Respecting users privacy by not sending unnecessary user identifiable tokens,
 * Attaching secret tokens to requests that don't need it can only increase the risk we leak them accidentally,
 * We could imagine that some tokens like the oauth stuff used in pub.dev needs to be refreshed, using tokens when not needed could reduce performance,
 * Attaching an unnecessary `Authorization` header is often an effective way to blow through CDN or other web cache... reducing performance further.

I realize that in it's current form `withAuthenticatedClient` isn't intended for use when authenticating to `pub.dartlang.org`, but I imagine that we'll want to support authenticating with tokens here too.

-------------------------------

Maybe, it's acceptable for 3rd party pub servers, since most of them will likely require a token for fetching packages anyways.

In that case we can just specialize `pub.dartlang.org` if/when we add support for `pub.dartlang.org`.

lib/src/authentication/client.dart
```diff
@@ -0,0 +1,76 @@
...
+class _AuthenticatedClient extends http.BaseClient {
+  _AuthenticatedClient(
+    this._inner, {
+    required this.credential,
+    required this.serverBaseUrl,
+  });
+
+  final http.BaseClient _inner;
+  final Credential credential;
+  final String serverBaseUrl;
```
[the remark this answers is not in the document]
TheMisir: > Question: do we normalize serverBaseUrl to end in slash or not?

Yes, we do! I'm also normalizing it before matching it with `request.url` to make sure it's normalized if "tokens.json" was modified manually.

doc/authentication-proposal.md
```diff
@@ -0,0 +1,114 @@
+Authenticating with Hosted Pub Repositories Proposal
```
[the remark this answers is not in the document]
TheMisir: I am not so good at documentation. Maybe someone else might help out to merge this with repository specs. For now I'll be using this file as specs drafts.

lib/src/command/login.dart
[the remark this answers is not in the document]
jonasfj: > Should we prompt for token when pub get'ing from unauthenticated repository?

I think that's a thing we can consider supporting in the future.

------------------

But in general, when I want to sign-on my development machine. Isn't it better than I use:
*  (A) `dart pub login <server>` rather than, using
*  (B) `dart pub login <server> --token <secret>`.

Even if we are on my machine, when I do (B) the `<secret>` is liable to end up in my `~/.bash_history`, which is unfortunate.

If someday my laptop is compromised by malicious code the first thing it would reasonably steal is `~/.bash_history`.

So I'm suggesting we only support (A), and never implement (B). We simply require that the token is always entered over stdin. If a user really doesn't want to do that, then the user can do `echo <secret>  | dart pub login <server>`.

lib/src/authentication/client.dart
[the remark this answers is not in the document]
jonasfj: Okay, hmm, maybe we should do this normalization in:
 * (A) `HostedSource._parseDescription`, and,
 * (B) `HostedSource._pubHostedUrlConfig`, too?

Indeed, correct me if I'm wrong... if we normalize in (A) and (B), then we won't need to normalize anywhere else, right?

Reading (B) it seems like we're already normalizing to **not** have trailing slash. Unless, we have a good reason to change that I propose that we normalize to not have a trailing slash. If we change this will cause unnecessary changes in all `pubspec.lock` files, and they'll swap formatting when switching between Dart SDK versions.

This is perhaps something we should do in a separate PR.

Investigating this I also discovered a bug -> https://github.com/dart-lang/pub/issues/3024.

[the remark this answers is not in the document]
TheMisir: > Reading (B) it seems like we're already normalizing to not have trailing slash.

Actually I don't agree with that.

Wouldn't this might make invalid url matching possible? If pub already authenticated to "http://pub.myserver.com". 
"http://pub.myserver.com.attacker.com" could get token for previously mentioned server because we're matching URLs using `url.startsWith(normalizedBaseUrl)`. Having trailing slash ensures that hostname could not be modified. I don't know how to explain this actually.

You can check this demo regarding my comment: https://dartpad.dev/342b9230acbc6f3527fe4acdd960bab4?null_safety=true

lib/src/source/hosted.dart
TheMisir: We can add scopes to authentication schemes which might be used to determine which request should be / could be authenticated using given scheme. It could be added in future without making breaking change.

lib/src/command/login.dart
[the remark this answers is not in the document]
TheMisir: Hey @jonasfj,

Fixed ✅

- `pub login` - Login to pub.dev
- `pub login --server <server>` - Login to given server

lib/src/authentication/scheme.dart
```diff
@@ -0,0 +1,77 @@
...
+// scheme could define scopes as [AuthenticationScope.publish] which means it
+// only requires authentication for publish command.
+
+/// Abstract interface for authentication scheme. It's used for validating
+/// request URLs and contains [credential] which will be used for
+/// authentication.
+abstract class AuthenticationScheme {
+  /// Authentication credential which used to resolve `Authorization` header
+  /// value.
+  Credential get credential;
```
jonasfj, line 14: I've been digging into this a bit the past few days... My two cents is that having abstractions:
 (A) `AuthenticationScheme`, and,
 (B) `Credential`,
is perhaps a bit much.

I don't have a solution. But I'm tiny bit concern we're making this a lot more complicated than it needs to be.

I'll play with this a bit more later.

robbecker-wf:
@themisir Thank you for working on adding this functionality!

lib/src/authentication/credential_store.dart
```diff
@@ -0,0 +1,152 @@
...
+
+import 'package:path/path.dart' as path;
+
+import '../io.dart';
+import '../log.dart' as log;
+import 'credential.dart';
+import 'scheme.dart';
+
+/// Stores and manages authentication credentials.
+class CredentialStore {
```
jonasfj, line 17: How about we simplify all of this a bit:
 * Do away with `Credential` as an abstraction
 * Rename `CredentialStore` to `TokenStore`

doc/authentication-proposal.md
```diff
@@ -0,0 +1,86 @@
...
+## Storing credentials
+
+Hosted Pub Repository authentication credentials will be stored on json file
+named `tokens.json` located in cache root directory. Authentication details will
+be stored at this file as json values while their keys will be URLs of the
+server (`PUB_HOSTED_URL`).
+
+```json
+{
+  "version": "1.0",
```
jonasfj, line 59: I would suggest simplifying this down to:
```suggestion
  "version": 1,
```

We are really only going to care about major version changes anyways. This an internal version number after all.

doc/authentication-proposal.md
```diff
@@ -0,0 +1,86 @@
...
+  "hosted": [
+    {
+      "url": "https://myserver.com",
+      "credential": {
+        "kind": "Basic",
+        "username": "bob",
+        "password": "password"
+      }
+    },
```
jonasfj, line 68: Let's remove this and only do `Bearer` authentication.

If someone wants to use basic authentication they can just use `token = "<user>:<password>"` in a bearer-token authentication scheme (it's less pretty, but much simpler on the client).

jonasfj:
We had a discussion around the CLI interface, to ensure that we keep it easy to use, and landed on:
```
dart pub login                      # Will initiate login-flow for pub.dev
dart pub logout                     # Will delete credential.json

dart pub token list                # Will list servers for which a token exists
dart pub token add <hosted_url>    # Will prompt for a token for <hosted_url> and store both
dart pub token remove <hosted_url> # Will remove token for <hosted_url>
dart pub token remove --all        # Will delete all tokens stored
```

Feedback is welcome, but unless we have something much better I suggest we go with this :D

We also have to move the location where we store tokens, but let's do that in a separate PR fixing https://github.com/dart-lang/pub/issues/2999

themisir:
I think the PR is ready for review..

I've tested the feature locally with a pub server that requires authentication. There still the e2e testing might needed for testing get and lish commands with the server - but I don't exactly know how to setup the tests for those scenarios.

/cc @jonasfj @sigurdm 

**Notes:**

1. I have a pub server for doing local tests which I can give you server url if you need for testing.
2. I think the failed CI test is not caused by changes I've made, might be race condition or something cause it never happened before.

lib/src/authentication/token_store.dart
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  /// Deletes tokens.json file from the disk.
+  void deleteTokensFile() {
+    deleteEntry(_tokensFile);
+    log.message('tokens.json is deleted.');
+  }
+
+  /// Full path to the "tokens.json" file.
+  String get _tokensFile => path.join(cacheRootDir, 'tokens.json');
```
sigurdm, line 155: We should not store these in the pub cache.

When (hopefully) https://github.com/dart-lang/pub/pull/3092 has landed we can use `dartConfigDir/tokens.json`.

lib/src/authentication/token_store.dart
```diff
@@ -0,0 +1,156 @@
...
+
+    return result;
+  }
+
+  /// Writes [tokens] into "tokens.json".
+  void _saveTokens(List<Token> tokens) {
+    writeTextFile(
+        _tokensFile,
+        jsonEncode(<String, dynamic>{
+          'version': 1,
```
sigurdm, line 79: Should we make this a map from url->token. That would make duplicate tokens for the same host a non-issue...

lib/src/io.dart
```diff
@@ -568,6 +568,16 @@ Future<bool> confirm(String message) {
   return _stdinLines.first.then(RegExp(r'^[yY]').hasMatch);
 }
 
+/// Reads a line from stdin stream.
+Future<String> readLine(String message) {
+  if (runningFromTest) {
+    log.message('$message:');
+  } else {
+    stdout.write('$message: ');
```
sigurdm, line 576: What happens if you always use log.message?

lib/src/authentication/token_store.dart
themisir: Yes I know, we've decided to do this in a separate PR because credentials.json (for pub.dartlang.org) is also stored there.

lib/src/command/token_add.dart
```diff
@@ -0,0 +1,59 @@
...
+          .timeout(const Duration(minutes: 5));
+
+      if (token.isEmpty) {
+        usageException('Token is not provided.');
+      }
+
+      tokenStore.addToken(Token.bearer(hostedUrl, token));
+      log.message('You are now logged in to $hostedUrl using bearer token.');
+    } on FormatException catch (_) {
+      usageException('Invalid or malformed server URL provided.');
```
jonasfj, line 50: This is a usage exception, but also include `e.message` from the `FormatException`. This is useful and will tell you want went wrong.

Probably a message along the lines of:
```
Invalid [hosted-url]: "${argResults.rest.first}"
${e.message}
```

Or something like that. We might have to add some tests using `expectMatchesGoldenFile`, or just manually try it..

lib/src/io.dart
```diff
@@ -568,6 +568,16 @@ Future<bool> confirm(String message) {
   return _stdinLines.first.then(RegExp(r'^[yY]').hasMatch);
 }
 
+/// Reads a line from stdin stream.
+Future<String> readLine(String message) {
+  if (runningFromTest) {
```
sigurdm, line 573: Please also update `confirm` to make use of this.

lib/src/command/token_add.dart
```diff
@@ -0,0 +1,59 @@
...
+    } else if (argResults.rest.length > 1) {
+      usageException('Takes only a single argument.');
+    }
+
+    try {
+      var hostedUrl = validateAndNormalizeHostedUrl(argResults.rest.first);
+
+      if (hostedUrl.isScheme('HTTP')) {
+        // TODO(themisir): Improve the following message.
+        usageException('Unsecure pub server could not be added.');
```
jonasfj, line 37: This is not a `usageException`, you are using the command-line arguments correctly, but we are forbidding you from doing what you're trying to do.

Maybe a `DataException` is more appropriate :D

sigurdm (commented):
Here is a first round of general comments.
Mainly it is requests to simplify code that is more general than what we need yet.

lib/src/command/token_list.dart
```diff
@@ -0,0 +1,29 @@
...
+  @override
+  String get invocation => 'pub token list';
+
+  @override
+  Future<void> runProtected() async {
+    // TODO(themisir): The output interface could be improved even more with
+    // additional details line, token preview, token kind, and instructions
+    // to remove a token.
+    log.message('Found ${cache.tokenStore.tokens.length} entries.');
+    for (final scheme in cache.tokenStore.tokens) {
```
jonasfj, line 25: Replace `scheme` with `token`.

lib/src/command/token_list.dart
```diff
@@ -0,0 +1,29 @@
...
+  String get description => 'List servers for which a token exists.';
+  @override
+  String get invocation => 'pub token list';
+
+  @override
+  Future<void> runProtected() async {
+    // TODO(themisir): The output interface could be improved even more with
+    // additional details line, token preview, token kind, and instructions
+    // to remove a token.
+    log.message('Found ${cache.tokenStore.tokens.length} entries.');
```
jonasfj, line 24: Let's try with a message like:
> You have secret tokens for 2 package repositories:
> https://some-pub-server.com/
> https://other-pub-server.com/

lib/src/authentication/token_store.dart
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  /// Removes tokens with matching [url] from store. Returns whether or not
+  /// there's a stored token with matching url.
+  bool removeMatchingTokens(String url) {
+    final normalizedUrl = validateAndNormalizeHostedUrl(url);
+
+    var i = 0;
+    var found = false;
+    while (i < tokens.length) {
```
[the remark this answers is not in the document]
themisir: We could, but `removeWhere` doesn't returns whether or not there's matching item in the list so I can customize printed message according to result like 'Token removed' or 'No token available for $url'.

lib/src/command/token_add.dart
```diff
@@ -0,0 +1,59 @@
...
+      }
+
+      final token = await readLine('Please enter bearer token')
```
jonasfj, line 40: Let's use the term: "secret token".

Maybe, it enough to simply say:
> Enter secret token:

lib/src/command/token_add.dart
```diff
@@ -0,0 +1,59 @@
...
+          .timeout(const Duration(minutes: 5));
+
+      if (token.isEmpty) {
+        usageException('Token is not provided.');
+      }
+
+      tokenStore.addToken(Token.bearer(hostedUrl, token));
+      log.message('You are now logged in to $hostedUrl using bearer token.');
```
jonasfj, line 48: Message like:
> Requests to $hostedUrl will now be authenticated using the secret token.

lib/src/authentication/token_store.dart
```diff
@@ -0,0 +1,156 @@
...
+      if (tokens[i].url == normalizedUrl) {
+        tokens.removeAt(i);
+        found = true;
+      } else {
+        i++;
```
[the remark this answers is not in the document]
themisir: removeAt removes current item so in next iteration the item at `i` will be item at `i+1` of previous iteration. It's hard to explain 😅 but [1,2,3] when we remove 1st item from this list index of the next item will become 1.

The code looks a bit weird because it's usually not allowed to modify list while iterating using foreach loop. I'm not sure if dart allows this case or not, at least other languages doesn't allows so I decided to do in this way.

lib/src/authentication/token_store.dart
themisir: I've discussed this with @jonasfj and we came into conclusion that using array is more future proof if we want to store more than 1 entry for the same URL in future.

lib/src/command/token_remove.dart
```diff
@@ -0,0 +1,47 @@
...
+
+import '../command.dart';
+import '../log.dart' as log;
+
+/// Handles the `token remove` pub command.
+class TokenRemoveCommand extends PubCommand {
+  @override
+  String get name => 'remove';
+  @override
+  String get description => 'Remove token for server.';
```
jonasfj, line 15: Something like:
> Remove secret token for package repository.

(or whatever terminology we settle on)

lib/src/command/token_remove.dart
```diff
@@ -0,0 +1,47 @@
...
+  @override
+  String get invocation => 'pub token remove';
+
+  bool get isAll => argResults['all'];
+
+  TokenRemoveCommand() {
+    argParser.addFlag('all',
+        help: 'Removes all saved tokens from token store.');
```
jonasfj, line 23: Something like:
> Remove secret tokens for all package repositories.

Or:
> Remove all secret tokens.

lib/src/command/token_remove.dart
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @override
+  Future<void> runProtected() async {
+    if (isAll) {
+      return tokenStore.deleteTokensFile();
+    }
+
+    if (argResults.rest.isEmpty) {
+      usageException('Must specify a package to be added.');
```
jonasfj, line 33: Something like:
> The [hosted-url] for a package repository must be specified.

lib/src/command/token_remove.dart
```diff
@@ -0,0 +1,47 @@
...
+    } else if (argResults.rest.length > 1) {
+      usageException('Takes only a single argument.');
+    }
+
+    final hostedUrl = argResults.rest.first;
+    final found = tokenStore.removeMatchingTokens(hostedUrl);
+
+    if (found) {
+      log.message('Token removed for server $hostedUrl.');
```
jonasfj, line 42: Something like:
> Removed secret token for package repository: $hostedUrl

lib/src/command/token_remove.dart
```diff
@@ -0,0 +1,47 @@
...
+      return tokenStore.deleteTokensFile();
+    }
+
+    if (argResults.rest.isEmpty) {
+      usageException('Must specify a package to be added.');
+    } else if (argResults.rest.length > 1) {
+      usageException('Takes only a single argument.');
+    }
+
+    final hostedUrl = argResults.rest.first;
```
jonasfj, line 38: Should we normalize here?

lib/src/authentication/client.dart
```diff
@@ -0,0 +1,112 @@
...
+import '../http.dart';
+import '../log.dart' as log;
+import '../system_cache.dart';
+import 'token.dart';
+
+/// This client authenticates requests by injecting `Authentication` header to
+/// requests.
+///
+/// Requests to URLs not under [serverBaseUrl] will not be authenticated.
+class _AuthenticatedClient extends http.BaseClient {
```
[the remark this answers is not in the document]
themisir: > I imagine the Future<http.StreamedResponse> send(http.BaseRequest request) method would add the token header if the host of the request has a token, and if we have a 401 we log the error message.

Pub server 'https://serverA.com/' might redirect to 'https://serverB.com/download/pkg.tar.gz'. In this case authentication header should not be attached to the request because of security reasons as described in [specs document](https://docs.google.com/document/d/1cJygfURx3hpX4oengdiDmpYuSWRQ3w3CnTAvGBnJSlY/edit?resourcekey=0-E30NU2oL_-F0nnIA6--Yzw#) "Client / Server authentication Protocol" section:

> - The requested URL is invoked for listing versions or fetching tarball from the server for which the credential is registered.
>   - Example: If a tokenA is registered for https://serverA.com and the client makes a request to list versions from serverB which says to download the tarballs to https://serverA.com/... then tokenA must not be attached to the request.

lib/src/authentication/token_store.dart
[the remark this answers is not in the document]
sigurdm: Good point.

lib/src/io.dart
themisir: I don't know 😅 I've just copied the code from `confirm(String message)` implementation.

lib/src/authentication/token.dart
```diff
@@ -0,0 +1,102 @@
...
+///   }
+/// }
+/// ```
+class Token {
+  Token({required this.url, required this.kind, required this.token});
+
+  /// Create [Token] instance with the `'Bearer'` kind.
+  Token.bearer(this.url, this.token) : kind = 'Bearer';
+
+  /// Deserialize [json] into [Token] type.
```
jonasfj, line 31: Suggest adding:
```
///
/// Throws [FormatException] if [json] is not a valid [Token].
```

This is just a nit thing, since it's an internal method.
But it's nice to try to document which exceptions can be thrown, only `Exception`s, not
`Error`s as callers are not expected to handle them.

In this case, it can be important that the caller handles `FormatException`s somehow.

lib/src/authentication/token.dart
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  /// Returns future that resolves "Authorization" header value used for
+  /// authenticating.
+  ///
+  /// This method returns future to make sure in future we could use the [Token]
+  /// interface for OAuth2.0 authentication too - which requires token rotation
+  /// (refresh) that's async job.
+  Future<String> getAuthorizationHeaderValue() {
+    return Future.value('$kind $token');
```
jonasfj, line 90: Let's just hardcode `Bearer` here, anything else could be dangerous.

lib/src/io.dart
sigurdm: haha :) Then I guess the question is on us - probably the log.message prints a newline...

lib/src/command/token_add.dart
```diff
@@ -0,0 +1,59 @@
...
+import '../io.dart';
+import '../log.dart' as log;
+import '../source/hosted.dart';
+
+/// Handles the `token add` pub command.
+class TokenAddCommand extends PubCommand {
+  @override
+  String get name => 'add';
+  @override
+  String get description => 'Add token for a server.';
```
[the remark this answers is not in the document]
sigurdm: I prefer 'package repository'...

lib/src/authentication/client.dart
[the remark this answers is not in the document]
sigurdm: Discussed this with @jonasfj and I think this is for another day.

Your solution is good for now!

lib/src/command/token_add.dart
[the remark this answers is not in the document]
themisir: > I prefer 'package repository'...

I prefer too, but this might cause conflict with packages hosted on git repository since 'repository' is a term in git ecosystem too..

lib/src/authentication/token_store.dart
jonasfj: Yes, let's do this separately.

lib/src/authentication/token_store.dart
```diff
@@ -0,0 +1,156 @@
...
+      }
+    } on FormatException catch (error, stackTrace) {
+      log.error('Failed to load tokens.json', error, stackTrace);
+
+      // When an invalid, damaged or not compatible version of token.json is
+      // found, we remove it after showing error message. Otherwise the error
+      // message will be displayed on each pub command.
+      // Or instead we could write instructrions to execute
+      //`pub token remove --all` if user couldn't solve the issue.
+      deleteEntry(_tokensFile);
```
jonasfj, line 68: yes, let's not delete the file.

Then it'll be overwritten when  the user does: `dart pub token add`, and user will get a warning in log saying that it failed to load tokens.json

lib/src/authentication/token.dart
```diff
@@ -0,0 +1,102 @@
...
+/// this:
+///
+/// ```json
+/// {
+///   "url": "https://example.com/",
+///   "credential": {
+///     "kind": "Bearer",
+///     "token": "gjrjo7Tm2F0u64cTsECDq4jBNZYhco"
+///   }
+/// }
```
jonasfj, line 23: As a general guideline it's best to [avoid unnecessary nesting](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Flattened_data_vs_Structured_Hierarchy#Flattened_data_vs_Structured_Hierarchy).

Perhaps, we should just do:
```js
{
  "url": "https://example.com",
  "token": "<secret-token>"
},
```

If we want to introduce additional authentication mechanisms in the future, we can add them like:
```js
{
  "url": "https://example.com",
  "environmentToken": "NAME_OF_ENV_VAR_CONTAINING_TOKEN"
},
// Or maybe:
{
  "url": "https://example.com",
  "serviceAccountName": "<email of service account>"
},
```
( don't think we'll do service accounts, but if we ever decide to try something like that we could do this way).

lib/src/authentication/token.dart
```diff
@@ -0,0 +1,102 @@
...
+/// ```
+class Token {
```
jonasfj, line 25: @sigurdm suggests we call this a `Credential` (sorry, I probably suggest a name change to token).

But thinking about it, it makes sense because: it's a pair of _package repository_ and _secret token_.

lib/src/authentication/token_store.dart
```diff
@@ -0,0 +1,156 @@
...
+
+import 'package:path/path.dart' as path;
+
+import '../io.dart';
+import '../log.dart' as log;
+import '../source/hosted.dart';
+import 'token.dart';
+
+/// Stores and manages authentication credentials.
+class TokenStore {
```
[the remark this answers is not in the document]
jonasfj: This might be too simple....

lib/src/authentication/token_store.dart
```diff
@@ -0,0 +1,156 @@
...
+          throw FormatException(
+              'tokens.json format is invalid or not supported.');
+        }
+
+        result.addAll((json['hosted'] as List)
+            .cast<Map<String, dynamic>>()
+            .map((it) => Token.fromJson(it)));
+      }
+    } on FormatException catch (error, stackTrace) {
+      log.error('Failed to load tokens.json', error, stackTrace);
```
jonasfj, line 61: Let's not log the stack trace, but we can do:
```
log.warning('Failed to load tokens.json: ${error.message}');
```

lib/src/authentication/token_store.dart
```diff
@@ -0,0 +1,156 @@
...
+  /// [Token].
+  List<Token> _loadTokens() {
+    final result = List<Token>.empty(growable: true);
+    final path = _tokensFile;
+    if (!fileExists(path)) {
+      return result;
+    }
+
+    try {
+      final json = jsonDecode(readTextFile(path));
```
jonasfj, line 41:
```suggestion
     try {
      final json = jsonDecode(readTextFile(path));
       } on FormatException {
          throw FormatException('$path is not valid JSON');
       }
```

This is probably better than a message saying: "did not expect ," :see_no_evil:

lib/src/authentication/token_store.dart
```diff
@@ -0,0 +1,156 @@
...
+
+      if (json.containsKey('hosted')) {
+        if (json['hosted'] is! List) {
+          throw FormatException(
+              'tokens.json format is invalid or not supported.');
+        }
+
+        result.addAll((json['hosted'] as List)
+            .cast<Map<String, dynamic>>()
+            .map((it) => Token.fromJson(it)));
```
jonasfj, line 58: If we fail to load an entry from `hosted`, then it's probably because the credential type is from the future.

We should print a warning that: `we failed to load credentials for ${it['url']}`, and then continue loading the rest of the credentials.

lib/src/authentication/token.dart
themisir: Yes nesting is unnecessary. But I think we might still need to store kind or something else for storing token type otherwise what we'll to if the token is like that:

```
{
  "url": "https://example.com",
  "token": "<secret-token>",
  "environmentToken": "NAME_OF_ENV_VAR"
}
```

Which one we'll going to use in this case?

themisir: If we need we could add it in future anyways. So removing it for now..

lib/src/authentication/token_store.dart
```diff
@@ -0,0 +1,156 @@
...
+    return result;
+  }
+
+  /// Writes [tokens] into "tokens.json".
+  void _saveTokens(List<Token> tokens) {
+    writeTextFile(
+        _tokensFile,
+        jsonEncode(<String, dynamic>{
+          'version': 1,
+          'hosted': tokens.map((it) => it.toJson()).toList(),
```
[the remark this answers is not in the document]
themisir: Which way would you suggest?

A. Opening & json de-serializing the file again and modifying `hosted` field.
B. Caching json contents on `_loadTokens` and using cached value to modify `hosted` field.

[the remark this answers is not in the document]
jonasfj: I think either solution would work. I think the only concern is race conditions, when the user is running two pub commands concurrently.

Perhaps (A) is best: when modifying `tokens.json` we just load it, do the modification and save it back (retaining unknown fields on objects that are not modified).

Example:
```js
// If modifying token for "https://some-pub-server.com", then
{
  "version": 1,
  "hosted": [
    {
      "url": "https://some-pub-server.com"
      "token": "..."
      "unknownField": ..., // delete this unknown field because we're modifying this object
    },
    {
      "url": "https://other-pub-server.com"
      "token": "..."
      "unknownFieldB": ...,  // preserve this unknown field because we're not modifying this object.
    },
  ],
}
```

Doesn't that sound like behavior that will allow us to be reasonably future proof?

test/token/remove_token_test.dart
```diff
@@ -0,0 +1,62 @@
...
+    await d.tokensFile({
+      'version': 1,
+      'hosted': [
+        {'url': 'https://server.demo', 'token': 'auth-token'}
+      ]
+    }).create();
+
+    await runPub(
+      args: ['token', 'remove', 'https://another-server.demo'],
+      output:
```
sigurdm, line 36: I would expect this to exit with a non-zero exit-code.

sigurdm (approved):
I think this starts looking good.
LGTM assuming the last comments are addressed

themisir:
> LGTM assuming the last comments are addressed

Expect [the last comment by jonasfj](https://github.com/dart-lang/pub/pull/3007#discussion_r701059242). Trying to figure out the implementation.

lib/src/command/lish.dart
```diff
@@ -138,6 +141,32 @@ class LishCommand extends PubCommand {
...
+        // If there's a saved credential for the server, publish using
+        // httpClient which should authenticate with the server using
+        // AuthenticationClient.
+        await withAuthenticatedClient(cache, server, (client) {
+          return _publishUsingClient(packageBytes, client);
+        });
+      } else {
+        // If user had not yet logged in into the server, use OAuth2 credentials
+        // for authentication.
+        await oauth2.withClient(cache, (client) {
```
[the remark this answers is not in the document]
themisir: > @jonasfj should we limit this to pub.dartlang.org/pub.dev?

I think some third party servers uses google id token to authenticate publish commands. So, limiting domain might cause incompatibility with those servers. But In other case this will add additional security / privacy (since id_token includes identifiers: email, id) layer 🤔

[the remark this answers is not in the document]
jonasfj: Third-party servers should never be given access to the users `id_token` as created by `oauth2.dart`.

That `id_token` is intended for use with `pub.dev`, hence, a third-party server would possibly be able to use this to make requests to pub.dev on behalf of a users. This would be a security vulnerability.

lib/src/authentication/token_store.dart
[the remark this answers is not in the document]
jonasfj: It's also possible that what I proposed is too complicated.

I think what I'm suggesting is that we think about how new features can be added to `tokens.json`, and how old clients should behave when this happens.

Ideally, we could leave a comment in here outlining:
 * What old clients should do when seeing a new feature,
 * How new clients can introduce new features without breaking old clients horribly.

This is worth thinking about because users sometimes switches back and forth between two different SDK versions. And if we don't consider upgrade-strategy upfront we might have painted ourselves into a corner where it's hard to add new features.

lib/src/command/lish.dart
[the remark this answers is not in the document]
themisir: But AFAIK it's currently possible for 3rd party servers to know id_tokens cause when I did my research prior to working on this PR I found out that exists pub clones are authenticating using pub.dev tokens when uploading archives. You can check out [here](https://github.com/bytedance/unpub/blob/master/unpub/lib/src/app.dart#L79-L101) for example.

lib/src/authentication/client.dart
```diff
@@ -0,0 +1,113 @@
...
+          'Insufficient permissions to the resource in hosted server at: '
+          '$hostedUrl\n'
+          'You can use the following command to update token for the server:\n'
+          '\n    pub token add $hostedUrl\n',
+        );
+      }
+
+      if (serverMessage?.isNotEmpty == true) {
+        // TODO(themisir): Sanitize and truncate serverMessage when needed.
+        log.error(serverMessage);
```
jonasfj, line 108: I propose that we limit this at:
 * 1024 bytes
 * ascii only

```dart
final truncatedMessage = String.fromCharCodes(
  // Only allow printable ASCII, map anything else to whitespace, take at-most 1024 characters
  serverMessage.runes.map((r) => 32 >= r && <= 127 ? r : 32).take(1024)
);
```

lib/src/authentication/credential.dart
```diff
@@ -0,0 +1,95 @@
...
+/// {
+///   "url": "https://example.com/",
+///   "token": "gjrjo7Tm2F0u64cTsECDq4jBNZYhco"
+/// }
+/// ```
+class Credential {
+  Credential({required this.url, required this.token, this.unknownFields});
+
+  /// Create [Credential] instance for bearer tokens.
+  Credential.bearer(this.url, this.token) : unknownFields = null;
```
jonasfj, line 27: Let's call this `token`, maybe in the future we have one called `tokenFromEnvironmentVariable` or something like that :D
```suggestion
  Credential.token(this.url, this.token) : unknownFields = null;
```

lib/src/authentication/credential.dart
```diff
@@ -0,0 +1,95 @@
...
+  /// Server url which this token authenticates.
+  final Uri url;
+
+  /// Authentication token value
+  final String? token;
+
+  /// Unknown fields found in tokens.json. The fields might be created by the
+  /// future version of pub tool. We don't want to override them when using the
+  /// old SDK.
+  final Map<String, dynamic>? unknownFields;
```
jonasfj, line 60: Instead of making this nullable, why don't we just use an empty map?

lib/src/authentication/credential.dart
```diff
@@ -0,0 +1,95 @@
...
+  /// Deserialize [json] into [Credential] type.
+  ///
+  /// Throws [FormatException] if [json] is not a valid [Credential].
+  factory Credential.fromJson(Map<String, dynamic> json) {
+    if (json['url'] is! String) {
+      throw FormatException('Url is not provided for the token');
+    }
+
+    final hostedUrl = validateAndNormalizeHostedUrl(json['url'] as String);
+    final token = json['token'] is String ? json['token'] as String : null;
```
jonasfj, line 38: Let's throw a `FormatException` if `token == null`, then we can remove the nullability...

And if we throw `FormatException` here, that's fine, because in `List<Credential> _loadCredentials() {` we will appropriately log: `Failed to load credentials for ${element['url']}: `

lib/src/authentication/credential.dart
```diff
@@ -0,0 +1,95 @@
...
+  /// authenticating.
+  // This method returns future to make sure in future we could use the
+  // [Credential] interface for OAuth2.0 authentication too - which requires
+  // token rotation (refresh) that's async job.
+  Future<String> getAuthorizationHeaderValue() {
+    if (token == null) {
+      throw DataException(
+        'Saved credential for $url pub repository is not supported by current '
+        'version of Dart SDK.',
+      );
```
jonasfj, line 81: Let's make this a format exception when loading. Then it'll get logged as a warning every time the user uses an old pub version with a newer config file.

Hmm, I can see that being less smart, but I also don't like dragging around nullability.

@sigurdm any thoughts here?

jonasfj (approved):
Just a few nits... the part about `truncatedMessage` is probably important.

An attacker could otherwise make some very confusing output in the terminal :see_no_evil:

lib/src/authentication/credential.dart
themisir: The issue is current saving mechanism will remove entries that doesn't contains 'token' property because it'll not be able to deserialize it. So they'll not added to the credential list and when saving those entries will be removed.

themisir: I wanted to encapsulate json serialization part to `Credential` class itself to reduce code duplication for serialization & validation, otherwise the code would look like much confusing I think.

sigurdm:
> fyi @sigurdm adding limitation to use OAuth credentials only for pub.dev official servers caused some tests to fail (probably because they're using localhost server to test the feature).
> 
> Should I revert those changes back or what would you suggest?
> 
> Maybe we could override "official pub server" url using environment variable for the sake of tests. Any suggestions?

Yes - I think that is the way to go!
We have `runningFromTest` in `lib/src/io.dart` - I suggest if that is `true` then we consider the host of `PUB_HOSTED_URL` to be the "official" host that gets the credentials. 

(PS When landing: https://github.com/dart-lang/pub/pull/3092this will be more safe, because runningFromTest will no longer only depend on the environment.)

themisir:
> We have `runningFromTest` in `lib/src/io.dart` - I suggest if that is `true` then we consider the host of `PUB_HOSTED_URL` to be the "official" host that gets the credentials.


Yes, used this method - it works. In future when we create integration tests for using token authentication for publishing we might want to declare another environment variable to provide whether given `PUB_HOSTED_URL` should use oauth2 authentication (or: considered as an official pub server) or uses token authentication. But for now it works and aside from testing I don't think this might cause any issue for production.

themisir:
Do I have to write integration tests for server integration or we could do that in a separate PR later?

sigurdm:
Thanks @themisir for all the hard work on this!

lib/src/command/token_add.dart
```diff
@@ -0,0 +1,64 @@
...
+        'token.',
+      );
+    } on FormatException catch (e) {
+      usageException('Invalid [hosted-url]: "${argResults.rest.first}"\n'
+          '${e.message}');
+    } on TimeoutException catch (_) {
+      // Timeout is added to readLine call to make sure automated jobs doesn't
+      // get stuck on noop state if user forget to pipe token to the 'token add'
+      // command. This behavior might be removed.
+      throw ApplicationException('Token is not provided within 5 minutes.');
```
mnordine, line 61: @themisir @sigurdm Says 5 minutes here, but the code above says 15 minutes.

lib/src/command/token_add.dart
```diff
@@ -0,0 +1,64 @@
...
+      usageException(
+          'The [hosted-url] for a package repository must be given.');
+    } else if (argResults.rest.length > 1) {
+      usageException('Takes only a single argument.');
+    }
+
+    try {
+      var hostedUrl = validateAndNormalizeHostedUrl(argResults.rest.first);
+      if (hostedUrl.isScheme('HTTP')) {
+        throw DataException('Unsecure package repository could not be added.');
```
mnordine, line 40: Unsecure => Insecure?

lib/src/command/token_add.dart
themisir: Yes, I forget to change message after changing timeout duration.

jonasfj: Feel free to file as many follow-up PRs as you'd like... smaller PRs are easy to merge :D

themisir: > Feel free to file as many follow-up PRs as you'd like... smaller PRs are easy to merge :D

Yep, filed #3110

PR #3008, opened 2021-05-20
Use the frontend server to compile pub executables

jonasfj:
Reverts dart-lang/pub#3006 which reverted https://github.com/dart-lang/pub/pull/2968, hence, introducing: Use the frontend server to compile pub executables.

cc @jakemac53 

We can merge this once we've figured out how to use the `frontend_server_client` in the Dart SDK.


jakemac53:
Interesting, I assumed this had been worked out because the test package does use it... but maybe it hasn't been updated either yet or something. I can look into it.

jakemac53:
https://dart-review.googlesource.com/c/sdk/+/200920

PR #3009, opened 2021-05-21
Make `.gitignore` and `.pubignore` case-insensitive on Windows / MacOS

### LINKED ISSUE #3003: `dart pub publish` includes extraneous binary files
I attempted to upload my package to the pub server, but it complained that it was too large after attempting to upload:

```
[C:\src\win32] dart pub publish
Publishing win32 2.1.1 to https://pub.dartlang.org:
|-- AUTHORS
|-- CHANGELOG.md
...
|-- example
|   |-- appcontainer.dart
|   |-- bin
|   |   |-- README.md
|   |   '-- taskdialog.exe.manifest
|   |-- bluetooth.dart
|   | (39 more...)
|   |-- window.dart
|   |-- winrt_picker.dart
|   '-- wmi.dart
...
|-- test
|   |-- api_test.dart
|   |-- btooth_test.dart
|   |-- com
|   |   |-- IApplicationActivationManager_test.dart
|   |   |-- IAppxFactory_test.dart
|   |   |-- IAppxFile_test.dart
|   |   | (67 more...)
|   |   |-- IWbemContext_test.dart
|   |   |-- IWbemLocator_test.dart
|   |   '-- IWbemServices_test.dart
|   | (17 more...)
|   |-- window_test.dart
|   |-- winrt_calendar_test.dart
|   '-- winrt_test.dart
'-- tool
    |-- README.md
...
    '-- struct_sizes.cpp

Do you want to publish win32 2.1.1 (y/N)? y
Uploading...
Uncompressed package archive is too large (size > 104857600).
```

On inspecting further, I discover that `dart pub publish` is attempting to upload various executables and symbol files that are artifacts of the build process. For example, 

- `build\test_cache\build\c075001b96339384a97db4862b8ab8db.cache.dill.track.dill`
- `example\explorer\windows\flutter\ephemeral\flutter_windows.dll.pdb` (170MB)

These are files in my `.gitignore`, and I'm surprised that they're being included. I am also surprised that they didn't cause a failure in the past, which makes me wonder if this might be a change of behavior in new builds...

```
[C:\src\win32] dart --version
Dart SDK version: 2.14.0-119.0.dev (dev) (Mon May 17 05:41:11 2021 -0700) on "windows_x64"
```

Full log: https://gist.github.com/timsneath/769172f3938c2a143c9f2d8366743820

jonasfj:
Fixes #3003

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -272,6 +272,41 @@ class Package {
                   log.warning(
                       '$ignoreFile had invalid pattern $pattern. ${exception.message}');
                 },
+                // Ignore case on MacOs and Windows, because `git clone` and
+                // `git init` will set `core.ignoreCase = true` in the local
+                // local `.git/config` file for the repository.
+                //
+                // So on Windows and MacOS most users will have case-insensitive
+                // behavior with `.gitignore`, hence, it seems reasonable to do
+                // the same when we interpret `.gitignore` and `.pubignore`.
+                //
+                // There are cases where a user may have case-sensitive behavior
+                // with `.gitignore` on Windows and MacOS:
+                //
+                //  (A) The user has manually overwritten the repository
+                //      configuration setting `core.ignoreCase = false`.
+                //
+                //  (B) The git-clone or git-init command that create the
+                //      repository did not deem `core.ignoreCase = true` to be
+                //      appropriate. Documentation for [git-config]][1] implies
+                //      this might depend on whether or not the filesystem is
+                //      case sensitive:
+                //      > If true, this option enables various workarounds to
+                //      > enable Git to work better on filesystems that are not
+                //      > case sensitive, like FAT.
+                //      > ...
+                //      > The default is false, except git-clone[1] or
+                //      > git-init[1] will probe and set core.ignoreCase true
+                //      > if appropriate when the repository is created.
+                //
+                // In either case, it seems likely that users on Windows and
+                // MacOS will prefer case-insensitive matching. We specifically
+                // know that some tooling will generate `.PDB` files instead of
+                // `.pdb`, see: [#3003][2]
+                //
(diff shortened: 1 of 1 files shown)


lib/src/package.dart
```diff
@@ -272,6 +272,41 @@ class Package {
...
+                //      > if appropriate when the repository is created.
+                //
+                // In either case, it seems likely that users on Windows and
+                // MacOS will prefer case-insensitive matching. We specifically
+                // know that some tooling will generate `.PDB` files instead of
+                // `.pdb`, see: [#3003][2]
+                //
+                // [1]: https://git-scm.com/docs/git-config/2.14.6#Documentation/git-config.txt-coreignoreCase
+                // [2]: https://github.com/dart-lang/pub/issues/3003
+                ignoreCase: Platform.isMacOS || Platform.isWindows,
```
athomas, line 309: What if I override that in my own config? Wouldn't a better way be to check if `git config core.ignorecase` is `true`?

athomas: Perhaps keep the platform defaults if we aren't in a git repo?

jonasfj: I could be convinced that way... the downside of reading `git config core.ignorecase` is that now git config determines our interpretation of `.pubignore`.

So documenting how this works becomes really hard.

When documenting how `dart pub publish` interprets `.gitignore` and `.pubignore` it might be best to say: it's like `.gitignore` case sensitive on Linux, case-insensitive on Mac / Windows.

jonasfj: But I could be convinced to go the other direction... :D

athomas: It's probably rare, and upon re-reading the documentation comment it already states that this may happen.

athomas (approved):

PR #3014, opened 2021-05-26
Change terminology for precompilation to building

mit-mit:
Fixes https://github.com/dart-lang/pub/issues/2951

PR #3017, opened 2021-05-27
Building

jonasfj:
Fixing tests for https://github.com/dart-lang/pub/pull/3014


jonasfj:
Moving forwards with my own r+ from https://github.com/dart-lang/pub/pull/3014

PR #3027, opened 2021-06-05
Downgraded does_not_include_dot_file.dart to 2.10

TheMisir:
All other tests files contains `@dart=2.10` annotation expect does_not_include_dot_file.dart file.


jonasfj (approved):

jonasfj:
Thanks, that's a great catch! :rocket:

Issue #3024, opened 2021-06-02
HostedSource should check that `url` is valid!


jonasfj:
`HostedSource._parseDescription` should probably ensure that `description['url']`:
 * Can be parsed using `Uri.parse()` then catch and throw new `FormatException`, and,
 * Check that the parsed `Uri` has a scheme that is either `http` or `https`.

A `pubspec.yaml` like:
```yaml
name: myapp
environment:
  sdk: '>=2.8.1 <3.0.0'
dependencies:
  retry:
    hosted:
      name: retry
      url: pub.dev  # <-- missing scheme
      version: ^3.0.0
```
causes:
```
Resolving dependencies... 
Bad state: Cannot use origin without a scheme: pub.dev/api/packages/retry
package:pub/src/source/hosted.dart 634:7           BoundHostedSource._throwFriendlyError
package:pub/src/source/hosted.dart 233:7           BoundHostedSource._fetchVersionsNoPrefetching
===== asynchronous gap ===========================
package:pub/src/source/hosted.dart 273:20          BoundHostedSource._fetchVersions
===== asynchronous gap ===========================
package:pub/src/rate_limited_scheduler.dart 89:11  RateLimitedScheduler._processNextTask.runJob
===== asynchronous gap ===========================
package:pub/src/rate_limited_scheduler.dart 98:5   RateLimitedScheduler._processNextTask
===== asynchronous gap ===========================
package:pool/pool.dart 127:14                      Pool.withResource
This is an unexpected error. Please run

    dart pub --trace get 

and include the logs in an issue on https://github.com/dart-lang/pub/issues/new 
```

Given that invalid URLs doesn't seem to work, I don't think it's a breaking change to add these checks.

Issue #3028, opened 2021-06-05
Bad state: [BUG] firebase_messaging ^8.0.0-dev.7 is not satisfied.


mshamsi502:
when i run `flutter pub get `, I get this result :
```
unning "flutter pub get" in mobile_jeanswest_app_android...     
Bad state: [BUG] firebase_messaging ^8.0.0-dev.7 is not satisfied.

package:pub/src/solver/partial_solution.dart 157:5  PartialSolution.satisfier

package:pub/src/solver/version_solver.dart 252:27   VersionSolver._resolveConflict

package:pub/src/solver/version_solver.dart 129:27   VersionSolver._propagate

package:pub/src/solver/version_solver.dart 93:11    VersionSolver.solve.<fn>

===== asynchronous gap ===========================

package:pub/src/source/hosted.dart 635:14           BoundHostedSource.withPrefetching.<fn>

===== asynchronous gap ===========================

package:pub/src/rate_limited_scheduler.dart 110:14  RateLimitedScheduler.withPrescheduling

===== asynchronous gap ===========================

package:pub/src/source/hosted.dart 634:12           BoundHostedSource.withPrefetching

===== asynchronous gap ===========================

package:pub/src/solver/version_solver.dart 90:14    VersionSolver.solve

===== asynchronous gap ===========================

package:pub/src/entrypoint.dart 233:18              Entrypoint.acquireDependencies

This is an unexpected error. Please run



    dart pub --trace get --no-precompile



and include the logs in an issue on https://github.com/dart-lang/pub/issues/new
pub get failed (1; and include the logs in an issue on https://github.com/dart-lang/pub/issues/new)
exit code 1
```

jonasfj:
@mshamsi502 thanks, this could be very interesting.
Can you share the minimal `pubspec.yaml` for reproducing this?

mshamsi502:
@jonasfj  of course, thanks
my `pubspec.yaml` :
```

version: 1.0.0+1

environment:
  sdk: ">=2.7.0 <3.0.0"

dependencies:
  flutter:
    sdk: flutter


  equatable: ^2.0.0
  flutter_bloc: ^7.0.0 
  easy_localization: ^3.0.0
  flutter_easyloading: ^3.0.0 
  flutter_keyboard_visibility: ^4.0.0
  flutter_phoenix: ^1.0.0
  retrofit: ^1.3.4+1
  dio: ^3.0.10
  get_it: ^6.0.0
[...]

PR #3030, opened 2021-06-08
Validate and normalize hosted url.

jonasfj:
Fixes https://github.com/dart-lang/pub/issues/3024

We normalize the URL for a _hosted pub server_ to have no slash if the
server is just a bare domain like `https://example.com`, but if the URL
contains a path like `https://example.com/my-folder` then we always
normalize to `https://example.com/my-folder/`.

The reason for normalizing the URL is to improve consistency in
`pubspec.lock` and make it easier to implement authentication without
risks of being tricked by incorrect prefixes.

Additionally, be normalizing to no slash for empty paths, and paths
always ending in a slash when path is non-empty, we gain the benefit
that relative URLs can always be constructed correctly using
`hostedUrl.resolve('api/packages/$package')`.

This additionally forbids a few edge cases such as:
 * querystring in the hosted URL (`https://server.com/?query`),
 * fragment in the hosted URL (`https://server.com/#hash`),
 * user-info in the hosted URL (`https://user:pwd@server.com`).

These may have worked with previous versions of the `pub` client, but
most likely the _querystring_ or _fragment_ would cause URLs to be garbled.
Any user-info would likely have been ignored, this was not tested, any
usage of these options is considered unlikely.

Previously, `dart pub publish` would ignore the path in the hosted URL
and always upload to `/api/packages/new`. This commit fixes this issue.

### DIFF
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -10,11 +10,13 @@ import 'package:http/http.dart' as http;
 
 import '../ascii_tree.dart' as tree;
 import '../command.dart';
+import '../exceptions.dart' show DataException;
 import '../exit_codes.dart' as exit_codes;
 import '../http.dart';
 import '../io.dart';
 import '../log.dart' as log;
 import '../oauth2.dart' as oauth2;
+import '../source/hosted.dart' show validateAndNormalizeHostedUrl;
 import '../utils.dart';
 import '../validator.dart';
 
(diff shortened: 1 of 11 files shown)


jonasfj:
@TheMisir I'm hoping this will make authentication easier... Let me know what you think.

TheMisir:
> @TheMisir I'm hoping this will make authentication easier... Let me know what you think.

1. Is there was a normalization function already existed? I would like to use it for normalizing URLs when adding a new credentials using `dart login`.
2. I planned to also use custom normalization for matching request URLs to credential URLs. (already discussed this on #hackers-pub)

I am a bit busy for a few days. I will continue working on authentication soon 🤞

jonasfj:
> I am a bit busy for a few days. I will continue working on authentication soon

That's fine. I'm busy too finding [bugs in the solver](https://github.com/dart-lang/pub/issues/3028).

As long as we get back to it, there is no rush :D

lib/src/command/lish.dart
```diff
@@ -77,7 +95,7 @@ class LishCommand extends PubCommand {
         return log.progress('Uploading', () async {
           // TODO(nweiz): Cloud Storage can provide an XML-formatted error. We
```
athomas, line 96: This comment looks like it applies to the POST request, perhaps it should be moved there or deleted.

lib/src/command/lish.dart
```diff
@@ -33,20 +35,36 @@ class LishCommand extends PubCommand {
...
+      }
     }
 
-    // Otherwise, use the default.
-    return Uri.parse(cache.sources.hosted.defaultUrl);
+    // Use the default server if nothing else is specified
+    return _server = cache.sources.hosted.defaultUrl;
   }
 
+  /// Cache value for [server].
```
athomas, line 65: Assuming that a publish command pretty much always needs to look at the server property, it seems easier to just have a `final Uri server = (){ return ...; }();` property and get rid of the lazy loading.

athomas (approved):
LGTM, comments are nits.

lib/src/command/lish.dart
jonasfj: if only members like `argResults` can't be accessed in an initializer :sob:

PR #3038, opened 2021-06-27
Add failing test for pre-release and backtracking behavior

mpizenberg:
Hi, I have been spending a lot of time on a [Rust port of pubgrub](https://github.com/pubgrub-rs/pubgrub) and stumbled upon a tricky case regarding handling of pre-releases and backtracking behavior. In particular, due to the creation of `NoVersions` incompatibilities potentially preventing the selection of a pre-release version after backtracking.

So I've checked how pub handles this by adding a test and it seems that pub fails to solve dependencies in that situation. Here is the test I've added:

```dart
  test('unable to backtrack to pre-release after NoVersions', () async {
    await servePackages((builder) {
      builder.serve('a', '1.1.0', deps: {'b': '^1.0.0'});
      builder.serve('b', '1.0.0', deps: {'c': '^1.0.0'});
      builder.serve('c', '0.9.0');
      builder.serve('b', '1.1.0-alpha');
      builder.serve('a', '1.0.0', deps: {'b': '^1.1.0-alpha'});
    });

    await d.appDir({
      'a': '^1.0.0',
    }).create();
    await expectResolves();
    // fails though I would have expected it to succeed with either
    // a 1.0.0, b 1.1.0-alpha or
    // a 1.1.0, b 1.1.0-alpha
  });
```

It consists of putting the solver on the wrong track by making it pick `a 1.1.0` followed by `b 1.0.0` then realizing that there is no compatible `c`. At that moment, it should backtrack and remember the wrong decision with `NoVersions` incompatibilities. The issue is that it seems to create knowledge preventing it to pick the `b 1.1.0-alpha` pre-release while it's a totally valid version after backtracking.

What are your thoughts on this?

### DIFF
--- a/test/version_solver_test.dart
+++ b/test/version_solver_test.dart
@@ -1760,6 +1760,41 @@ void prerelease() {
       'c': '1.0.0',
     });
   });
+
+  test('existence of pre-release prevent backtracking', () async {
+    await servePackages((builder) {
+      builder.serve('a', '1.1.0', deps: {'b': '^1.0.0'});
+      builder.serve('a', '1.0.0', deps: {'b': '^1.1.0-alpha'});
+      builder.serve('b', '1.1.0-alpha');
+    });
+
+    await d.appDir({
+      'a': '^1.0.0',
+    }).create();
+    await expectResolves(result: {
+      'a': '1.1.0',       // would have expected 1.0.0 here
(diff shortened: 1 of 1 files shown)


sigurdm:
I included this test in: https://github.com/dart-lang/pub/pull/3078

Issue #2307, opened 2020-01-17
make the logic to locate the Flutter SDK more robust


devoncarew:
When running `pub get` on a Flutter project, I see:

```
Resolving dependencies... 
Because mp_chart depends on flutter_test any from sdk which doesn't exist (the Flutter SDK is not available), version solving failed.
```

This is when using the pub client that is embedded in the flutter sdk (`flutter/bin/cache/dart-sdk/bin/pub`). Currently the pub client only locates the Flutter sdk when the FLUTTER_SDK environment variable is set. We should expand this to also check if the specific instance of the pub tool itself is part of the Flutter sdk (something like using `Platform.resolvedExecutable`, and verifying it's location).

jakemac53:
Note that build_runner also needed to know if you were running from the flutter sdk or the dart sdk - you can check the `Platform.version` for the `flutter` string but only in newer versions.

We have an outstanding issue to have a fallback for older sdks that does some file system spelunking to figure it out.

It would be good to establish a common way of determining if you are running from the flutter sdk (and its location).

https://github.com/dart-lang/build/issues/2528 is the issue for the build repo

sigurdm:
I'd like to get a bit more context - what is the use-case of calling `pub get` directly for a Flutter project? Could you just use `flutter pub get` (or set FLUTTER_SDK manually)?

It is not clear to me that `pub` in `flutter/bin/cache/dart-sdk/pub` is publicly exposed, but is there to be embedded by the `flutter` tool, and the right way to get dependencies of a Flutter project is `flutter pub get` that would set FLUTTER_SDK correctly.

Personally I'd like to put as little flutter-specific logic in the pub client as possible. But I think it would make sense to give a better error message explaining to use `flutter get` instead.

jakemac53:
> It is not clear to me that `pub` in `flutter/bin/cache/dart-sdk/pub` is publicly exposed, but is there to be embedded by the `flutter` tool, and the right way to get dependencies of a Flutter project is `flutter pub get` that would set FLUTTER_SDK correctly.

It is easier if you are a flutter developer to not bother installing the dart sdk twice, and instead add the one from flutter to your PATH. I don't have an idea of how common it actually is though.

devoncarew:
> It is not clear to me that pub in flutter/bin/cache/dart-sdk/pub is publicly exposed

People do use the dart sdk embedded in the Flutter SDK, and @mit-mit has a proposal to make this an officially supported use case.

[...]

PR #3045, opened 2021-07-15
Detect Flutter SDK location relative to Dart SDK

jonasfj:
Fix https://github.com/dart-lang/pub/issues/2307, tested on Windows and Linux using a few hacks.

If the Dart SDK is present inside the Flutter SDK in sub-folder
`bin/cache/dart-sdk/`, then we no-longer require the environment
variable `FLUTTER_ROOT` to be specified. Instead we simply derive
the location of the Flutter SDK from the location of the Dart SDK.

We still allow the environment variable `FLUTTER_ROOT` to override
the otherwise automatically detected Flutter SDK location.

### DIFF
--- a/lib/src/sdk/flutter.dart
+++ b/lib/src/sdk/flutter.dart
@@ -20,13 +20,45 @@ class FlutterSdk extends Sdk {
   @override
   Version get firstPubVersion => Version.parse('1.19.0');
 
-  static final bool _isAvailable =
-      Platform.environment.containsKey('FLUTTER_ROOT');
-  static final String _rootDirectory = Platform.environment['FLUTTER_ROOT'];
+  static final bool _isAvailable = _rootDirectory != null;
+  static final String _rootDirectory = () {
+    // If FLUTTER_ROOT is specified, then this always points to the Flutter SDK
+    if (Platform.environment.containsKey('FLUTTER_ROOT')) {
+      return Platform.environment['FLUTTER_ROOT'];
+    }
+
+    // We can try to find the Flutter SDK relative to the Dart SDK.
+    // We know that the Dart SDK is always present, this is found relative to
+    // the `dart` executable, for details see: lib/src/sdk/dart.dart
+    //
+    // Once we have the location of the Dart SDK, we can look at its parent
+    // directories, if going 3 levels-up and down `bin/cache/dart-sdk/` is equal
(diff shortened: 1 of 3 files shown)


jonasfj:
@natebosch feel free to unassign review if you're busy, I'm just a little short on reviewers this summer :sun_with_face:

lib/src/sdk/flutter.dart
```diff
@@ -20,13 +20,45 @@ class FlutterSdk extends Sdk {
...
+    // Once we have the location of the Dart SDK, we can look at its parent
+    // directories, if going 3 levels-up and down `bin/cache/dart-sdk/` is equal
+    // to the Dart SDK root, then it's probably because we are located inside
+    // the Flutter SDK, at: `$FLUTTER_ROOT/bin/cache/dart-sdk`
+    final parts = p.split(sdk.rootDirectory);
+    if (parts.length > 3) {
+      // Go 3-levels up from the Dart SDK root
+      final flutterSdk = p.joinAll(parts.take(parts.length - 3));
+      // If going down 'bin/cache/dart-sdk/' yields the same path as the Dart
+      // SDK has, then it's probably because the Dart SDK is located inside
```
natebosch, line 43: Are there any other flutter files we can look for to double check this assumption? I wouldn't expect anyone to pick this directory name accidentally I suppose...

lib/src/sdk/flutter.dart
```diff
@@ -20,13 +20,45 @@ class FlutterSdk extends Sdk {
...
+      }
+    }
+
+    return null;
+  }();
 
   @override
   String get installMessage =>
-      'Flutter users should run `flutter pub get` instead of `pub get`.';
+      'Flutter users should run `flutter pub get` instead of `dart pub get`.';
```
natebosch, line 61: Is this still true?

jonasfj: Yes, because `flutter pub get` has some hooks generating code for plugins.

But I think you'll only really see this message if the Flutter SDK isn't found.

lib/src/sdk/flutter.dart
jonasfj: Hmm, let's read the `version` file, I think it's being read further down anyways.

Buf if we gracefully fail to find it here, then missing the flutter SDK is a better error message.

natebosch (approved):

PR #3048, opened 2021-07-19
Add executables to allErrors

jonasfj:
Just discovered this while hacking something else, figure it might be worth fixing -- added some tests too, let's see if they pass.

### DIFF
--- a/test/validator/pubspec_field_test.dart
+++ b/test/validator/pubspec_field_test.dart
@@ -59,6 +59,7 @@ void main() {
       var pkg = packageMap('test_pkg', '1.0.0');
       pkg['executables'] = <String, String>{
         'test_pkg': null,
+        'test_pkg_helper': 'helper',
       };
       await d.dir(appPath, [d.pubspec(pkg)]).create();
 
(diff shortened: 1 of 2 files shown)


test/validator/pubspec_field_test.dart
```diff
@@ -117,5 +135,23 @@ void main() {
...
+
+    test('has invalid executables', () async {
+      var pkg = packageMap('test_pkg', '1.0.0');
+      pkg['executables'] = <String>['wrong-thing'];
+      await d.dir(appPath, [d.pubspec(pkg)]).create();
+
+      await expectValidation(pubspecField, errors: isNotEmpty);
+    });
+
+    test('has invalid executables (2)', () async {
```
athomas, line 147: Perhaps describe how it's invalid instead?

test/validator/pubspec_field_test.dart
```diff
@@ -46,6 +46,24 @@ void main() {
...
+      pkg['executables'] = <String, String>{};
+      await d.dir(appPath, [d.pubspec(pkg)]).create();
+
+      await expectValidation(pubspecField);
+    });
+
+    test('has executables', () async {
+      var pkg = packageMap('test_pkg', '1.0.0');
+      pkg['executables'] = <String, String>{
+        'test_pkg': null,
```
athomas, line 61: Shouldn't there also be a case for an executable key with a string value?

https://dart.dev/tools/pub/pubspec#executables has this example:
```yaml
executables:
  slidy: main
  fvm:
```

athomas (approved):
LGTM

PR #3055, opened 2021-07-27
Remove duplicate global invocation on Windows

### LINKED ISSUE #2934: [Windows]: Programs installed with pub global activate run twice
# Environment

- `pub version` or `flutter pub version`: `Pub 2.12.2`
- OS version: Windows 10
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

When running an app installed with `pub global activate`, the program runs twice, except if executed with `pub global run`. 

## Expected behavior

```
> pub global activate pana
> pana Could not find an option or flag "-h".

Usage: pana [<options>] --hosted <published package name> [<version>]
       pana [<options>] <local directory>

Options:
      --flutter-sdk     The directory of the Flutter SDK.
  -j, --json            Output log records and full report as JSON.
      --hosted-url      The server that hosts <package>.
                        (defaults to "https://pub.dev")
  -l, --line-length     The line length to use with dartfmt.
      --hosted          Download and analyze a hosted package (from https://pub.dev).
      --[no-]warning    Shows the warning message before potentially destructive operation.
                        (defaults to on)

```

## Actual behavior
```
> pub global activate pana
> pana Could not find an option or flag "-h".

Usage: pana [<options>] --hosted <published package name> [<version>]
       pana [<options>] <local directory>

Options:
      --flutter-sdk     The directory of the Flutter SDK.
  -j, --json            Output log records and full report as JSON.
      --hosted-url      The server that hosts <package>.
                        (defaults to "https://pub.dev")
  -l, --line-length     The line length to use with dartfmt.
      --hosted          Download and analyze a hosted package (from https://pub.dev).
      --[no-]warning    Shows the warning message before potentially destructive operation.
                        (defaults to on)
Could not find an option or flag "-h".

Usage: pana [<options>] --hosted <published package name> [<version>]
       pana [<options>] <local directory>

Options:
      --flutter-sdk     The directory of the Flutter SDK.
  -j, --json            Output log records and full report as JSON.
      --hosted-url      The server that hosts <package>.
                        (defaults to "https://pub.dev")
  -l, --line-length     The line length to use with dartfmt.
      --hosted          Download and analyze a hosted package (from https://pub.dev).
      --[no-]warning    Shows the warning message before potentially destructive operation.
                        (defaults to on)-h  

```

## The problem: 

Using `pana` as an example (it happens with every package), I can find the location of the batch file using `where`: 

```
C:\Users\user\AppData\Roaming\Pub\Cache\bin\pana.bat
```
Inspecting the file, I found the call was made twice: 
```batch
@echo off
rem This file was created by pub v2.12.2.
rem Package: pana
rem Version: 0.15.4
rem Executable: pana
rem Script: pana
if exist "C:\Users\levi\AppData\Roaming\Pub\Cache\global_packages\pana\bin\pana.dart-2.12.2.snapshot" (
  dart "C:\Users\levi\AppData\Roaming\Pub\Cache\global_packages\pana\bin\pana.dart-2.12.2.snapshot" %*
  rem The VM exits with code 253 if the snapshot version is out-of-date.
  rem If it is, we need to delete it and run "pub global" manually.
  if not errorlevel 253 (
    goto error
  )
  pub global run pana:pana %* 
) else (
  pub global run pana:pana %*
)
goto eof
:error
exit /b %errorlevel%
:eof
```
As you can see, it calls `dart <path_to_pana>`, and then if there's an error, it has a `goto`. So the immediate call to `pub` should not be there. I removed it and it worked just fine.

Levi-Lesches:
~~In the Batch file defined in `_createBinStub`, the script attempts to call the package using the dart snapshot. If it fails, the Batch script uses `dart pub global` directly. However, the `dart pub global` command is still present even when using the snapshot works. This PR removes the second call so that package is only run once. This is especially important/urgent because a lot of executable packages focus on project management, and make sensitive changes to the files in a user's project. Having those changes run twice can cause errant behavior.~~

**EDIT**: Adding a `call` before the dart command in the generated binstub fixes an error where, if the user has `flutter\bin\dart.bat` ahead of `flutter\bin\cache\dart-sdk\bin\dart.exe` in their PATH, the script gets called twice. I tried tracking down the cause but couldn't. All I know is that _something_ happens in flutter's `dart.bat` that _really_ messes with the binstub, and won't even let the simple `goto error` statement run properly. Using `call` fixes that... somehow.

Fixes #2934

### DIFF
--- a/lib/src/global_packages.dart
+++ b/lib/src/global_packages.dart
@@ -750,7 +750,7 @@ To recompile executables, first run `global deactivate ${dep.name}`.
         assert(p.isAbsolute(snapshot));
         invocation = '''
 if exist "$snapshot" (
-  dart "$snapshot" %*
+  call dart "$snapshot" %*
   rem The VM exits with code 253 if the snapshot version is out-of-date.
   rem If it is, we need to delete it and run "pub global" manually.
   if not errorlevel 253 (
--- a/lib/src/global_packages.dart
+++ b/lib/src/global_packages.dart
@@ -756,6 +756,7 @@ if exist "$snapshot" (
   if not errorlevel 253 (
     goto error
   )
+  dart pub global run ${package.name}:$script %*
 ) else (
   dart pub global run ${package.name}:$script %*
 )
--- a/test/global/binstubs/binstub_runs_executable_test.dart
+++ b/test/global/binstubs/binstub_runs_executable_test.dart
@@ -31,6 +31,7 @@ void main() {
         environment: getEnvironment());
 
     expect(process.stdout, emits('ok [arg1, arg2]'));
+    expect(process.stdout, neverEmits('ok [arg1, arg2]'));
     await process.shouldExit();
   });
 
--- a/test/global/binstubs/runs_once_even_when_dart_is_batch_test.dart
+++ b/test/global/binstubs/runs_once_even_when_dart_is_batch_test.dart
@@ -1,0 +1,55 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// @dart=2.10
+
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
+
+import '../../descriptor.dart' as d;
+import '../../test_pub.dart';
+
+void main() {
+  test(
+      'runs only once even when dart on path is a batch file (as in flutter/bin)',
+      () async {
+    await servePackages((builder) {
+      builder.serve(
+        'foo',
+        '1.0.0',
+        contents: [
+          d.dir('bin', [d.file('script.dart', 'main(args) => print(args);')]),
+        ],
+        pubspec: {
+          'executables': {'script': 'script'},
+        },
+      );
+    });
+
+    await runPub(args: ['global', 'activate', 'foo']);
+
+    await d.dir(
+      'bin',
+      [
+        d.file('dart.bat', '''
+@echo off
+${Platform.resolvedExecutable} %*
+'''),
+      ],
+    ).create();
+
+    var process =
+        await Process.run(p.join(d.sandbox, cachePath, 'bin', 'script.bat'), [
+      'hi'
+    ], environment: {
+      'PATH': [p.join(d.sandbox, 'bin'), p.dirname(Platform.resolvedExecutable)]
+          .join(';'),
+      ...getPubTestEnvironment(),
+    });
+    expect((process.stdout as String).trim(), '[hi]');
+    expect(process.exitCode, 0);
+  }, skip: !Platform.isWindows);
+}


Levi-Lesches:
https://github.com/dart-lang/pub/blob/26576f6c0648dd24ea8a1faf4988d2df2cf6bcdd/lib/src/global_packages.dart#L752-L765

Adding the `CALL` command in line 753 instead of deleting the line I did also fixes this issue. I'll look into which one is preferable.

Levi-Lesches:
Okay, I figured it out. The error is not the duplicate `dart pub` line, it's the need for the `call` command in Dart. I added that now.

lib/src/global_packages.dart
```diff
@@ -750,7 +750,7 @@ To recompile executables, first run `global deactivate ${dep.name}`.
         assert(p.isAbsolute(snapshot));
         invocation = '''
 if exist "$snapshot" (
-  dart "$snapshot" %*
+  call dart "$snapshot" %*
```
jonasfj, line 753: If `dart` from `PATH` is actually `dart.bat` then using [`call`](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/call) makes a difference, because invoking a bat script from another bat script will stop the original bat script from working.

How that triggers this to be run twice I don't understand.

hmm, can this cause problems when `dart` is not a `dart.bat` file, which is the case when installed from the Dart SDK.

Levi-Lesches: I have no clue how but from several hours of investigating, I found that even when the if branch of the `errorlevel 253` if statement runs, the `goto error` line doesn't run. I put an echo before and after and both are printed -- it's almost like the goto is skipped. Using `call` somehow prevents this from happening. My only guess is that something in `dart.bat`, or one of the batch files it calls, must be changing something or setting a conflicting label, and that using `call` tells CMD to use Dart in a separate context. I couldn't find anything concrete though.

When `dart` is really `dart.exe` (in `flutter/bin/cache/dart-sdk/bin/dart.exe`, the place `dart.bat` forwards to, `call` still works. I ran all the binstub tests for both versions of PATH, making sure to close cmd.exe.

Levi-Lesches: To clarify, it isn't the `dart snapshot` that runs twice. Instead, the `goto error` line is skipped, even when the if statement is run. That causes the first `dart pub global run` command to run as well, effectively running the executable twice. 

That's why so many fixes out there (and my initial draft for this PR) simply deleted the first `dart pub global run` line. But that broke some tests, this is the proper solution.

Levi-Lesches:
The review for this change is here: 

https://dart-review.googlesource.com/c/pub/+/208620

Levi-Lesches:
@jonasfj, @sigurdm, is this PR good? There are many issues across many repos (almost any packages you'd install with `dart pub global activate`) that would be fixed by this.

sigurdm (approved):
This LGTM - @jonasfj do you have anything more?

jonasfj (approved):
LGTM

kevmoo:
Are we going to land this?

sigurdm:
I added a specific regression test.

PR #3062, opened 2021-08-01
Fix a typo

whym:
I believe "instead" should be outside of the quotation.

> Use `dart run instead`
> Use `dart run` instead.


sigurdm (approved):
Thanks! LGTM

Issue #3012, opened 2021-05-24
pubspec.lock paths are incompatible on Windows


creativecreatorormaybenot:
# Environment

```
Pub 2.14.0-edge.a527411e5100a0a4f48c4009087a1b988aa784af
Dart SDK version: 2.14.0-edge.a527411e5100a0a4f48c4009087a1b988aa784af (be) (Fri May 14 15:47:38 2021) on "windows_x64"
```

# Problem

When using `dart pub run build_runner build` (using the [`build_runner` package](https://pub.dev/packages/build_runner)), the command immediately crashed because:

```
Could not find a file named "pubspec.yaml" in "Pub\Cache\git\foo-<git-hash>\packages%5Cbar".
```

The `pubspec.yaml` for this package looks like this:

```yaml
dev_dependencies:
  foo:
    git:
      url: git@github.com:user/repo.git
      path: packages/bar
      ref: some-tag
```

The **generated** `pubspec.lock` for this package looks like this:

```
  foo:
    dependency: "direct dev"
    description:
      path: "packages\\bar"
      ref: "some-tag"
      resolved-ref: <git-tag>
      url: "git@github.com:user/repo.git"
    source: git
    version: "0.1.1"
```

And here it was immediately obvious to me: `packages\\bar` was the culprit.

## Expected behavior

1. The `path` in the `pubspec.lock` should be a forward slash (`packages/bar`) as the lock file should be version controlled for apps, which means that non-Windows users should use the same lock file.
2. The end result should be that `flutter pub run build_runner build` works :)

## Actual behavior

```
Could not find a file named "pubspec.yaml" in "Pub\Cache\git\foo-<git-hash>\packages%5Cbar".
```

## Workaround

I figured out a **workaround solution**:

1. **Manually** change the path from `packages\\bar` to `packages/bar` in the **`pubspec.lock`** file.
2. Success (`flutter pub run build_runner build` runs successfully).

technolion:
Please also see https://github.com/dart-lang/build/issues/3137

ZeroOneTV:
@creativecreatorormaybenot
I understand how you explained it and it really applies to my case. However, when I run the command: ` flutter pub run build_runner build` , it says that the file has changed and that I should run pub get again.
How do I get around this problem?

creativecreatorormaybenot:
@ZeroOneTV For me, it is enough to run `pub get` get again and then *only* change these lines (or 2 characters).  
This does not seem to change the integrity of the file.

[...]

PR #3063, opened 2021-08-02
Use uris for paths in git source descriptions

sigurdm:
Sometimes they would be handled as local file-paths leading to broken behviour.

Fixes: https://github.com/dart-lang/pub/issues/3012

### DIFF
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -142,6 +142,8 @@ class GitSource extends Source {
     return {'relative': relative, 'url': url};
   }
 
+  /// Returns [path] normalized.
+  ///
   /// Throws a [FormatException] if [path] isn't a relative url or null.
   String _validatedPath(dynamic path) {
     path ??= '.';
@@ -161,7 +163,7 @@ class GitSource extends Source {
           "The 'path' field of the description must not reach outside the "
           'repository.');
     }
-    return p.normalize(parsed.toString());
+    return p.url.normalize(parsed.toString());
   }
 
   /// If [description] has a resolved ref, print it out in short-form.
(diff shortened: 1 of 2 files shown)


natebosch (approved):

jonasfj (approved):

lib/src/source/git.dart
```diff
@@ -142,6 +142,8 @@ class GitSource extends Source {
     return {'relative': relative, 'url': url};
   }
 
+  /// Returns [path] normalized.
```
lrhn, line 145: I'd document it as `/// Normalizes [path].` And name it `_normalizePath`, and say that [path] must be a string or `null`.

lib/src/source/git.dart
```diff
@@ -142,6 +142,8 @@ class GitSource extends Source {
...
+  ///
```
lrhn, line 146: Also, say what a "relative URL" is. It's not clear from context, and not a well-defined term.

The `Uri.isAbsolute` checks for a scheme and no fragment, so `http://example.com/foo#bar` is *not* considered absolute (it has a fragment!)
It's very unlikely to ever be what you want.

You may want `uri.hasAbsolutePath` instead, and check that `url.hasScheme` and `url.hasAuthority` are both false. (Because `scheme:relative/path#fragment` is not `isAbsolute` and has a relative path, but is probably not what you'd consider a releative URI reference).

lrhn (commented):
Drive-by comments!

/L

lib/src/source/git.dart
sigurdm: Ah I thought a relative uri was any non-absolute one.

Thanks, will try to see if I can come up with something...

Issue #3073, opened 2021-08-10
Use incremental compilation when running from source/git


jakemac53:
Now that we do incremental compilation, we should pre-compile all binaries even when running with a source or git dependency.

We would still re-compile on every run in this scenario, but it would be an incremental recompile which will be much faster.

jakemac53:
One thing to note here is that we usually log a `Building package executable...` message when compiling. But that can break some interactions with other executables, and it would now log on every run to any executable run via source...

PR #3074, opened 2021-08-10
Always precompile scripts before running them

jakemac53:
Closes https://github.com/dart-lang/pub/issues/3073

The idea is that since we have incremental compilation, its worth actually compiling the scripts and running that compiled script. For mutable packages we still always re-compile before each run, but it will be much faster to do so, as each build will be incremental.

This makes running pub executables from source much faster, and it also means you will get some more immediate feedback (the `Building package executable...` message). Previously it was just a blank screen for several seconds (depending on size of the app).

Note that this does mean we will always emit some log messages about building scripts for each run. This could break some scripts that rely on the stdout of pub executables. This was always the case for the first run of an immutable package ran with `pub run`, but would come up more often now. We could consider moving that output to `stderr`, or suppressing it, but I do think it is actually nice, when running interactively.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -300,7 +300,7 @@ class Entrypoint {
   ///
   /// Except globally activated packages they should precompile executables from
   /// the package itself if they are immutable.
-  List<Executable> get builtExecutables {
+  List<Executable> get _builtExecutables {
     if (isGlobal) {
       if (isCached) {
         return root.executablePaths
@@ -311,9 +311,6 @@ class Entrypoint {
       }
     }
     final r = root.immediateDependencies.keys.expand((packageName) {
-      if (packageGraph.isPackageMutable(packageName)) {
-        return <Executable>[];
-      }
       final package = packageGraph.packages[packageName];
       return package.executablePaths
           .map((path) => Executable(packageName, path));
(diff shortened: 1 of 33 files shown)


sigurdm (approved):
LGTM
Nice! - this is awesome!

sigurdm:
> Note that this does mean we will always emit some log messages about building scripts for each run. This could break some scripts that rely on the stdout of pub executables.

I'm sure you're aware, but we do have the warningsOnlyUnlessTerminal guard that prevents the output if run non-interactively.

Question is if the information is relevant to the user - maybe we should only show anything if the wait is longer than ~2 seconds...

jakemac53:
> I'm sure you're aware, but we do have the warningsOnlyUnlessTerminal guard that prevents the output if run non-interactively.

Oh! I was not aware this was a thing :).

> Question is if the information is relevant to the user - maybe we should only show anything if the wait is longer than ~2 seconds...

I don't have any strong opinions here, other than we should probably be consistent between the modes :). Today we always show this message when pre-compiling.

jakemac53:
Fwiw this is ready to merge from my perspective - not sure if you also wanted @jonasfj to take a look? Or if you wanted to iterate on the UX (namely the build message?).

jonasfj (approved):
r+

PR #3075, opened 2021-08-10
use incremental compilation in the tool/test.dart script

jakemac53:
This updates the `tool/test.dart` script to do incremental compilation of the `bin/pub.dart` script using the utilities already available in pub.

This makes running tests using `tool/test.dart` significantly faster - since we get incremental builds of pub itself.

### DIFF
--- a/tool/test.dart
+++ b/tool/test.dart
@@ -15,26 +15,21 @@
 import 'dart:io';
 import 'package:path/path.dart' as path;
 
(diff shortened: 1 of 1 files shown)


sigurdm (approved):
Super nice!

I wonder if we can be without this script at all when `dart run` does incremental precompilation always.

But for now this is good.

Issue #3057, opened 2021-07-30
Infinite loop in the solver


simolus3:
This was found by @jeroen-meijer. I'm on `Dart SDK version: 2.13.4 (stable)`, but this also happens on `2.14.0-341.0.dev`.
```yaml
name: flutter_pub_get_loop_example
description: Dependencies go brrrrrrrrrrrrt
version: 0.0.1

environment:
  sdk: ">=2.12.0 <3.0.0"

dependencies:
  freezed_annotation: ^0.12.0

dev_dependencies:
  build_runner:
  freezed:
  mocktail: ^0.1.3
```

The solver quickly goes into this loop from which it doesn't recover:

```
SLVR:                                                                                         conflict: freezed ^0.13.0-nullsafety.0 depends on freezed_annotation >=0.13.0-nullsafety.0 <2.0.0
SLVR:                                                                                         ! freezed ^0.13.0-nullsafety.0 is partially satisfied by not freezed <0.13.0-nullsafety.0
SLVR:                                                                                         ! which is caused by "freezed <0.13.0-nullsafety.0 depends on build >=0.12.6 <2.0.0"
SLVR:                                                                                         ! thus: freezed <0.14.0 requires freezed_annotation >=0.13.0-nullsafety.0 <2.0.0 or build >=0.12.6 <2.0.0
SLVR:                                                                                         ! freezed <0.14.0 is partially satisfied by not freezed <0.13.0-nullsafety.0
SLVR:                                                                                         ! which is caused by "freezed <0.13.0-nullsafety.0 depends on build >=0.12.6 <2.0.0"
SLVR:                                                                                         ! thus: freezed <0.14.0 requires freezed_annotation >=0.13.0-nullsafety.0 <2.0.0 or build >=0.12.6 <2.0.0
SLVR:                                                                                         ! freezed <0.14.0 is partially satisfied by not freezed <0.13.0-nullsafety.0
SLVR:                                                                                         ! which is caused by "freezed <0.13.0-nullsafety.0 depends on build >=0.12.6 <2.0.0"
SLVR:                                                                                         ! thus: freezed <0.14.0 requires freezed_annotation >=0.13.0-nullsafety.0 <2.0.0 or build >=0.12.6 <2.0.0
SLVR:                                                                                         ! freezed <0.14.0 is partially satisfied by not freezed <0.13.0-nullsafety.0
[...]

PR #3078, opened 2021-08-16
Fix broken simplification of prerelease constraints

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3057
Fixes: https://github.com/dart-lang/pub/pull/3038
Fixes: https://github.com/dart-lang/pub/issues/3028

I am not sure what was exactly meant by having the case `(range.min.isPreRelease && range.max == range.min.nextBreaking)` but it seems not. quite right. 

`>=0.9.0-1 < 0.10.0` would get translated to `^0.9.0-1`. But `^0.9.0-1` in pub_semver is the same as `>=0.9.0-1 < 0.10.0-0`, and that is what gets captured by the first part of the condition: `range.max == range.min.nextBreaking.firstPreRelease`.

### DIFF
--- a/lib/src/package_name.dart
+++ b/lib/src/package_name.dart
@@ -268,8 +268,7 @@ class PackageRange extends PackageName {
     if (!range.includeMin) return this;
     if (range.includeMax) return this;
     if (range.min == null) return this;
-    if (range.max == range.min.nextBreaking.firstPreRelease ||
-        (range.min.isPreRelease && range.max == range.min.nextBreaking)) {
+    if (range.max == range.min.nextBreaking.firstPreRelease) {
       return withConstraint(VersionConstraint.compatibleWith(range.min));
     } else {
       return this;
--- a/test/version_solver_test.dart
+++ b/test/version_solver_test.dart
@@ -1759,6 +1759,44 @@ void prerelease() {
       'c': '1.0.0',
     });
   });
+
+  test('https://github.com/dart-lang/pub/issues/3057 regression', () async {
+    // This used to cause an infinite loop.
+    await servePackages((builder) {
+      builder.serve('a', '0.12.0', deps: {});
+      builder.serve('b', '0.1.0', deps: {'c': '2.0.0'});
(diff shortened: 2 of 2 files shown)


lib/src/package_name.dart
```diff
@@ -268,8 +268,7 @@ class PackageRange extends PackageName {
     if (!range.includeMin) return this;
     if (range.includeMax) return this;
     if (range.min == null) return this;
-    if (range.max == range.min.nextBreaking.firstPreRelease ||
-        (range.min.isPreRelease && range.max == range.min.nextBreaking)) {
+    if (range.max == range.min.nextBreaking.firstPreRelease) {
```
jonasfj, line 271:
```suggestion
    if (range.max == range.min.nextBreaking.firstPreRelease ||
        (range.min.isPreRelease && range.max == range.min.nextBreaking)) {
        assert(range.includeMax == false);
```

But wouldn't you believe this to be correct if `range.includeMax == false`?

Which is always the case because of the preceeding `if (range.includeMax) return this;`

jonasfj: nvm... that still wrong.

jonasfj:
> `>=0.9.0-1 < 0.10.0` would get translated to `^0.9.0-1`. But `^0.9.0-1` in pub_semver is the same as `>=0.9.0-1 < 0.10.0-0`, and that is what gets captured by the first part of the condition: `range.max == range.min.nextBreaking.firstPreRelease`.


I imagine this was intended to catch the case where: `>= 0.10.0-1 <0.10.0` can be translated to `^0.10.0-1`.

I'm guessing we'll need something like: (requiring us to add `Version.withoutPreRelease` or auxiliary function)
```dart
    if (range.max == range.min.nextBreaking.firstPreRelease ||
        (range.min.isPreRelease && range.max == range.min.withoutPreRelease)) {
```
@sigurdm Does that make sense?

jonasfj (approved):
I think this is correct, but that we could probably fix the bug that is the original intent here, if we wanted to -- see previous comments.

PR #3087, opened 2021-08-27
Tweak strict dependencies error message for `_validateBenchmarkTestTool`

guidezpl:
I saw this error today `* line 7, column 1 of test/dynamic_colors_plugin_test.dart: This package does not have meta in the dependencies section of pubspec.yaml.`, which isn't quite accurate. 

This PR differentiates the case when a dependency should be added to `dependencies` from the case where it should be added to either `dependencies` or `dev_dependencies`.

There are no tests currently checking the error message, but I could add a couple if needed.


sigurdm:
Really nice.

A test of the changed behavior would be great!

guidezpl:
Done!

sigurdm (approved):
LGTM

PR #3092, opened 2021-09-02
Store credentials in a config directory

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2999


lib/src/oauth2.dart
```diff
@@ -80,6 +80,12 @@ void logout(SystemCache cache) {
     log.message('Logging out of pub.dartlang.org.');
     log.message('Deleting $credentialsFile');
     _clearCredentials(cache);
+    // Test if we also have a legacy credentials file.
+    final legacyCredentialsFile = _legacyCredentialsFile(cache);
+    if (entryExists(legacyCredentialsFile)) {
+      log.message('Also deleting legacy credentials at $legacyCredentialsFile');
+      deleteEntry(legacyCredentialsFile);
```
themisir, line 87: Wouldn't it make more sense if we move legacy credentials file into new path instead of deleting? It would be backwards-compatible with CI jobs that uses credentials.json for authentication & publishing packages.

sigurdm: Here we are in the `dart pub logout` code - it should delete any credentials, so I don't think it makes sense to keep the old ones.

In the code for `_credentialsFile` below we use the credentials file from the legacy location if it exists there. We could consider moving it, but that would make it hard(er) to switch between older and newer versions of pub.

WDYT?

themisir: Oh sorry, 😅 haven't noticed it's for logout command.

jonasfj (approved):

jonasfj (approved):

lib/src/system_cache.dart
```diff
@@ -89,7 +89,7 @@ class SystemCache {
   /// Defaults to `false`.
   SystemCache({String rootDir, bool isOffline = false})
       : rootDir = rootDir ?? SystemCache.defaultDir,
-        tokenStore = TokenStore(rootDir ?? SystemCache.defaultDir) {
+        tokenStore = TokenStore(dartConfigDir) {
```
jonasfj, line 92: Let's also rename `tokens.json` to `pub-tokens.json`.

sigurdm: Yes!
Done!

PR #3115, opened 2021-09-14
Environment credentials support

### LINKED ISSUE #3113: introduce: `dart token add <hosted-url> --env-var ENV_VAR`
Proposal to introduce:
```
dart token add <hosted-url> --env-var <ENV_VAR>
```
Which would store a credential with:
```js
{
  "url": "<hosted-url>",
  "env": "<ENV_VAR>"
}
``` 

The idea being that when a request to `<hosted-url>` is made the authentication token attached is loaded from the environment variable `ENV_VAR`.

I'm open to ideas for how to define the command, maybe using `dart token add <hosted-url> --from-env <ENV_VAR>` is better.

## Motivation
In many environments the easiest way to inject a secret is through environment variables. This allows a single line to configure `dart pub` to read the secret from a known environment variable. We'll even store the name of the environment variable in our config, instead of storing the actual secret.

## Corner cases

 * If environment variable doesn't exist when `dart token add <hosted-url> --env-var <ENV_VAR>` is called, we should print a warning.
 * If environment variable doesn't exist when `dart pub get` is making a request, we should either:
    * (A) Print a warning and proceed,
    * (B) Exit non-zero with an error message explaining that the environment variable configured is not present.

Given that (A) will likely trigger a 401, maybe (B) is preferable. Or maybe we want to do (A) and then if the request fails we want to augment the error message with information about the fact the environment variable from which we should pickup the secret was not configured.

 * If the server respond 401, we can't delete the configuration from `pub-tokens.json` like we otherwise would. Instead we have to print an error message explaining that the token we read from `ENV_VAR` was not accepted by the server.

--------

/cc @themisir

themisir:
Fixes #3113

### DIFF
--- a/lib/src/authentication/credential.dart
+++ b/lib/src/authentication/credential.dart
@@ -4,6 +4,8 @@
 
 // ignore_for_file: import_of_legacy_library_into_null_safe
 
+import 'dart:io';
+
 import '../exceptions.dart';
 import '../source/hosted.dart';
 
@@ -24,13 +26,21 @@ class Credential {
   /// Internal constructor that's only used by [fromJson].
   Credential._internal({
     required this.url,
-    required this.token,
     required this.unknownFields,
+    required this.token,
+    required this.env,
   });
 
-  /// Create a new [Credential].
+  /// Create credential that stores clear text token.
   Credential.token(this.url, this.token)
-      : unknownFields = const <String, dynamic>{};
+      : env = null,
+        unknownFields = const <String, dynamic>{};
+
+  /// Create credential that stores environment variable name that stores token
+  /// value.
+  Credential.env(this.url, this.env)
+      : token = null,
+        unknownFields = const <String, dynamic>{};
 
   /// Deserialize [json] into [Credential] type.
   ///
@@ -42,14 +52,19 @@ class Credential {
 
     final hostedUrl = validateAndNormalizeHostedUrl(json['url'] as String);
 
-    const knownKeys = {'url', 'token'};
+    const knownKeys = {'url', 'token', 'env'};
     final unknownFields = Map.fromEntries(
         json.entries.where((kv) => !knownKeys.contains(kv.key)));
 
+    String? _optional(String key) {
+      return json[key] is String ? json[key] as String : null;
+    }
+
     return Credential._internal(
       url: hostedUrl,
-      token: json['token'] is String ? json['token'] as String : null,
       unknownFields: unknownFields,
+      token: _optional('token'),
+      env: _optional('env'),
     );
   }
 
@@ -58,6 +73,9 @@ class Credential {
 
   /// Authentication token value
   final String? token;
+
+  /// Environment variable name that stores token value
+  final String? env;
 
   /// Unknown fields found in tokens.json. The fields might be created by the
   /// future version of pub tool. We don't want to override them when using the
@@ -69,6 +87,7 @@ class Credential {
     return <String, dynamic>{
       'url': url.toString(),
       if (token != null) 'token': token,
+      if (env != null) 'env': env,
       ...unknownFields,
     };
   }
@@ -88,6 +107,17 @@ class Credential {
       );
     }
 
+    if (env != null) {
+      final value = Platform.environment[env];
+      if (value == null) {
+        throw DataException(
+          'Saved credential for $url pub repository requires environment '
+          'variable named $env but not defined.',
+        );
+      }
+      return Future.value('Bearer $value');
+    }
+
     return Future.value('Bearer $token');
   }
 
@@ -101,7 +131,7 @@ class Credential {
   ///
   /// This method might return `false` when a `tokens.json` file created by
   /// future SDK used by pub tool from old SDK.
-  bool isValid() => token != null;
+  bool isValid() => token != null || env != null;
 
   static String _normalizeUrl(String url) {
     return (url.endsWith('/') ? url : '$url/').toLowerCase();
--- a/lib/src/authentication/credential.dart
+++ b/lib/src/authentication/credential.dart
@@ -123,8 +123,8 @@ class Credential {
       final value = Platform.environment[env];
       if (value == null) {
         throw DataException(
-          'Saved credential for $url pub repository requires environment '
-          'variable named $env but not defined.',
+          'Saved credential for "$url" pub repository requires environment '
+          'variable named "$env" but not defined.',
         );
       }
       return Future.value('Bearer $value');
--- a/lib/src/authentication/credential.dart
+++ b/lib/src/authentication/credential.dart
@@ -4,9 +4,9 @@
 
 // @dart=2.11
 
-import 'package:meta/meta.dart';
-
 import 'dart:io';
+
+import 'package:meta/meta.dart';
 
 import '../exceptions.dart';
 import '../source/hosted.dart';
@@ -77,7 +77,7 @@ class Credential {
   final String token;
 
   /// Environment variable name that stores token value
-  final String? env;
+  final String env;
 
   /// Unknown fields found in pub-tokens.json. The fields might be created by the
   /// future version of pub tool. We don't want to override them when using the
--- a/lib/src/authentication/credential.dart
+++ b/lib/src/authentication/credential.dart
@@ -58,15 +58,25 @@ class Credential {
     final unknownFields = Map.fromEntries(
         json.entries.where((kv) => !knownKeys.contains(kv.key)));
 
-    String _optional(String key) {
-      return json[key] is String ? json[key] as String : null;
+    /// Returns [String] value from [json] at [key] index or `null` if [json]
+    /// doesn't contains [key].
+    ///
+    /// Throws [FormatException] if value type is not [String].
+    String _string(String key) {
+      if (json.containsKey(key)) {
+        if (json[key] is! String) {
+          throw FormatException('Provided $key value should be string');
+        }
+        return json[key] as String;
+      }
+      return null;
     }
 
     return Credential._internal(
       url: hostedUrl,
       unknownFields: unknownFields,
-      token: _optional('token'),
-      env: _optional('env'),
+      token: _string('token'),
+      env: _string('env'),
     );
   }
 
@@ -133,7 +143,8 @@ class Credential {
   ///
   /// This method might return `false` when a `pub-tokens.json` file created by
   /// future SDK used by pub tool from old SDK.
-  bool isValid() => token != null || env != null;
+  // Either [token] or [env] should be defined to be valid.
+  bool isValid() => (token == null) ^ (env == null);
 
   static String _normalizeUrl(String url) {
     return (url.endsWith('/') ? url : '$url/').toLowerCase();
--- a/lib/src/command/token_add.dart
+++ b/lib/src/command/token_add.dart
@@ -79,7 +79,7 @@ class TokenAddCommand extends PubCommand {
     );
   }
 
-  Future<void> _addTokenFromEnv(Uri hostedUrl) async {
+  Future<void> _addEnvVarToken (Uri hostedUrl) async {
     if (envVar.isEmpty) {
       throw DataException('Cannot use the empty string as --env-var');
     }
--- a/lib/src/command/token_add.dart
+++ b/lib/src/command/token_add.dart
@@ -51,8 +51,7 @@ class TokenAddCommand extends PubCommand {
       if (envVar == null) {
         await _addTokenFromStdin(hostedUrl);
       } else {
-        await _addTokenFromEnvVar(hostedUrl);
-      }
+        await _addEnvVarToken(hostedUrl);
       }
     } on FormatException catch (e) {
       usageException('Invalid [hosted-url]: "${argResults.rest.first}"\n'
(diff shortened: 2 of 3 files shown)


themisir:
I think this is ready for review.

/cc @jonasfj @sigurdm

lib/src/command/token_add.dart
```diff
@@ -61,4 +63,33 @@ class TokenAddCommand extends PubCommand {
...
+      'Requests to $hostedUrl will now be authenticated using the secret '
+      'token.',
+    );
+  }
+
+  Future<void> _addTokenFromEnv(Uri hostedUrl) async {
+    tokenStore.addCredential(Credential.env(hostedUrl, envVar));
+    log.message(
+      'Requests to $hostedUrl will now be authenticated using the secret '
+      'token stored in environment variable.',
```
sigurdm, line 85:
```suggestion
      'token stored in the environment variable `$envVar`.',
```

lib/src/command/token_add.dart
```diff
@@ -61,4 +63,33 @@ class TokenAddCommand extends PubCommand {
...
+    );
+
+    if (!Platform.environment.containsKey(envVar)) {
+      // If environment variable doesn't exist when
+      // pub token add <hosted-url> --env-var <ENV_VAR> is called, we should
+      // print a warning.
+      log.warning('Environment variable $envVar is not defined.');
```
sigurdm, line 92: I'm not sure we should warn here. There is no reason the variable should be defined already - instead it should be defined when using.

@jonasfj do you agree?

lib/src/command/token_add.dart
```diff
@@ -61,4 +63,33 @@ class TokenAddCommand extends PubCommand {
...
+    }
+
+    tokenStore.addCredential(Credential.token(hostedUrl, token));
+    log.message(
+      'Requests to $hostedUrl will now be authenticated using the secret '
+      'token.',
+    );
+  }
+
+  Future<void> _addTokenFromEnv(Uri hostedUrl) async {
```
sigurdm, line 81: I would call it:
```suggestion
  Future<void> _addEnvVarToken (Uri hostedUrl) async {
```

lib/src/command/token_add.dart
```diff
@@ -40,17 +48,11 @@ class TokenAddCommand extends PubCommand {
...
       }
 
-      final token = await stdinPrompt('Enter secret token:')
-          .timeout(const Duration(minutes: 15));
-      if (token.isEmpty) {
-        usageException('Token is not provided.');
+      if (envVar?.isNotEmpty == true) {
+        await _addTokenFromEnv(hostedUrl);
+      } else {
+        await _addTokenFromStdin(hostedUrl);
```
sigurdm, line 54: I think it should be an error to have an empty envVar, but probably better to check it inside `_addTokenFromEnvVar`.
```suggestion
      if (envVar == null) {
        await _addTokenFromStdin(hostedUrl);
      } else {
        await _addTokenFromEnvVar(hostedUrl);
      }
```

lib/src/command/token_add.dart
```diff
@@ -61,4 +63,33 @@ class TokenAddCommand extends PubCommand {
...
+
+    tokenStore.addCredential(Credential.token(hostedUrl, token));
+    log.message(
+      'Requests to $hostedUrl will now be authenticated using the secret '
+      'token.',
+    );
+  }
+
+  Future<void> _addTokenFromEnv(Uri hostedUrl) async {
+    tokenStore.addCredential(Credential.env(hostedUrl, envVar));
```
sigurdm, line 82: Here we can check the format eg:
```suggestion
    if (envVar.isEmpty) {
      throw DataException('Cannot use the empty string as --env-var');
    }
    tokenStore.addCredential(Credential.env(hostedUrl, envVar));
```

lib/src/command/token_add.dart
```diff
@@ -25,6 +26,13 @@ class TokenAddCommand extends PubCommand {
   @override
   String get argumentsDescription => '[hosted-url]';
 
+  String get envVar => argResults['env-var'];
+
+  TokenAddCommand() {
+    argParser.addOption('env-var',
+        help: 'Environment variable name that stores secret token');
```
sigurdm, line 33:
```suggestion
        help: 'Use this environment variable to fetch the secret token.');
```

test/token/add_token_test.dart
```diff
@@ -34,6 +34,52 @@ void main() {
...
+        error: isNot(contains('Environment variable TOKEN is not defined.')),
+      );
+
+      await d.tokensFile({
+        'version': 1,
+        'hosted': [
+          {'url': 'https://example.com', 'env': 'TOKEN'},
+        ]
+      }).validate();
+    });
```
sigurdm, line 80: We would also need a test that the credential is actually read from the envvar.

sigurdm (commented):
Nice - I've got a few comments

lib/src/authentication/credential.dart
```diff
@@ -44,14 +54,19 @@ class Credential {
 
     final hostedUrl = validateAndNormalizeHostedUrl(json['url'] as String);
 
-    const knownKeys = {'url', 'token'};
+    const knownKeys = {'url', 'token', 'env'};
     final unknownFields = Map.fromEntries(
         json.entries.where((kv) => !knownKeys.contains(kv.key)));
 
+    String _optional(String key) {
+      return json[key] is String ? json[key] as String : null;
```
sigurdm, line 62: I **think** this can be implemented as just: 
```suggestion
      return json[key] as String;
```

lib/src/authentication/credential.dart
```diff
@@ -44,14 +54,19 @@ class Credential {
...
+    }
+
     return Credential._internal(
       url: hostedUrl,
-      token: json['token'] is String ? json['token'] as String : null,
       unknownFields: unknownFields,
+      token: _optional('token'),
```
sigurdm, line 68: What happens if there is both a token and an env key in the input? I guess currently we check for the env-var first. Should we instead consider that invalid?

lib/src/authentication/credential.dart
themisir: This might throw an error if field is in another type (int, map, list). I think, I should throw FormatException instead to make sure field type is String or null.

sigurdm: Yes - of course - I looked at this backwards...

lib/src/authentication/credential.dart
themisir: There's `Credential.isValid` method for validity check used to filter out invalid values when looking for token. I'm not checking validity in fromJson otherwise invalid values will be removed from pub-tokens.json when running token add or remove. That might cause pub-tokens.json generated by future version of SDK where contains entries with both token and env values that are used for some purpose to be removed.

themisir: Also currently I'm using `env` value if provided otherwise `token` value [here](https://github.com/dart-lang/pub/pull/3115/files#diff-9d07b56c1956f7361cc0b0c4b37054c0035634e87f45dffc4cd3d5f2817bc00eR112-R123).

sigurdm: So maybe `isValid` should also check that both are not set?

themisir: > So maybe `isValid` should also check that both are not set?

Sure.

test/token/add_token_test.dart
themisir: I'll work on tests with mocked server on another PR.

lib/src/command/token_add.dart
```diff
@@ -61,4 +63,37 @@ class TokenAddCommand extends PubCommand {
...
+  Future<void> _addTokenFromStdin(Uri hostedUrl) async {
+    final token = await stdinPrompt('Enter secret token:', echoMode: false)
+        .timeout(const Duration(minutes: 15));
+    if (token.isEmpty) {
+      usageException('Token is not provided.');
+    }
+
+    tokenStore.addCredential(Credential.token(hostedUrl, token));
+    log.message(
+      'Requests to $hostedUrl will now be authenticated using the secret '
```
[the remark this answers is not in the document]
jonasfj: Quotes are important, it makes it easier to discover typos :D

lib/src/command/token_add.dart
jonasfj: I think we should warn... but I can see the argument against it given how github actions work (ie. you only have to pass the env var with the secret to the command that does the operation).

For example, the following would produce a warning:
```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/dart...@v...
      - run: dart pub token add pub.dev --env=AUTH_TOKEN
      - run: dart pub publish
        env:
          AUTH_TOKEN: ${{ secrets.PUB_TOKEN }}
```

The warning probably wouldn't cause much alarm if the following command succeeds :D

on the other hand, this wouldn't:
```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/dart...@v...
      - run: dart pub token add pub.dev --env=AUTH_TOKEN
        env:
          AUTH_TOKEN: ${{ secrets.PUB_TOKEN }}
      - run: dart pub publish
        env:
          AUTH_TOKEN: ${{ secrets.PUB_TOKEN }}
```

Doing option two here is more work, but it also wouldn't be less secure.

-----------

I think that to help users do the right thing we need to do either:
 * (A) restrict allowed env var names to `[a-zA-Z_][a-zA-Z0-9_]*`, and/or,
 * (B) warn if the environment variable isn't defined when `dart token add <hosted-url> --env-var=<NAME>` is called.

IMO, we should lean towards keeping things simple and do both (A) and (B).
But I could also be sold on only doing one of these two things.

IMO; we can always remove the warning in (B), or re-introduce it later.

We can easily remove a restriction like (A), but we can't add it later.

Hence, I suggest we do (A), and (B). I know sigurdm@ leans a little towards doing neither, but I'm happy letting him pick at-least one of them that we should do.

Note. I think it's important that we do either (A) or (B) to help users when they are making typos.

jonasfj (commented):
Mostly just nits, overall this looks good to me :D

lib/src/command/token_add.dart
```diff
@@ -61,4 +63,37 @@ class TokenAddCommand extends PubCommand {
...
+      'token.',
+    );
+  }
+
+  Future<void> _addEnvVarToken(Uri hostedUrl) async {
+    if (envVar.isEmpty) {
+      throw DataException('Cannot use the empty string as --env-var');
+    }
```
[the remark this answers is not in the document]
jonasfj: At a minimum we should restrict to the legal environment variable names:
   * forbidding `=` on Linux, Mac OS
   * forbidding `=` on Windows

Sources:
 * https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html
 * https://docs.microsoft.com/en-us/windows/win32/procthread/environment-variables

I would suggest referencing the source in a comment explaining why we forbid `=` in the environment variable name.

------

We could further remove footguns from the user by saying that we require env vars to be:
> uppercase letters, digits, and the <underscore> ( '_' ) from the characters defined in Portable Character Set and do not begin with a digit.

Which is what the POSIX.1-2017 enforces for its variables.. But I think @sigurdm suggested that we don't artificially restrict the environment variable name. If we want to help the user we could detect if the env-var name we're given violates this rule and then warns about it.

Example:

If I explicitly pass `$MY_TOKEN` instead of `MY_TOKEN`, then maybe instead of saying:
```
dart pub token add https://some-pub.com --env-var="\$MY_TOKEN"

Requests to "https://some-pub.com" will now be authenticated using the secret 
token stored in the environment variable `$MY_TOKEN`.
Environment variable `$MY_TOKEN` is not defined.
```
We could say:

```
dart pub token add https://some-pub.com --env-var="\$MY_TOKEN"

Requests to "https://some-pub.com" will now be authenticated using the secret 
token stored in the environment variable `$MY_TOKEN`.

Environment variable `$MY_TOKEN` is not defined.

WARNING: Environment variable name given `$MY_TOKEN` starts with a dollar sign,
this is unusually for environment variable names. Are you sure you didn't want to use
the environment variable name: `MY_TOKEN`
```

We could also do:

```
dart pub token add https://some-pub.com --env-var="MY_TOKEN "

Requests to "https://some-pub.com" will now be authenticated using the secret 
token stored in the environment variable `MY_TOKEN `.

Environment variable `MY_TOKEN ` is not defined.

WARNING: Environment variable name given `MY_TOKEN ` starts with a dollar sign,
this is unusually for environment variable names. Are you sure you didn't want to use
the environment variable name: `MY_TOKEN `
```
and maybe even:
```
export MY_TOKEN="bestpassword1234"
dart pub token add https://some-pub.com --env-var="$MY_TOKEN"  # <-- I accidentally gave the value of the env var instead of the name

Requests to "https://some-pub.com" will now be authenticated using the secret 
token stored in the environment variable `bestpassword1234`.

Environment variable `bestpassword1234` is not defined.

WARNING: Environment variable name given `bestpassword1234` does not use
upper-case characters, 0-9, and underscore. If you didn't want to use this environment variable name
you can overwrite it with:
  dart pub token add https://some-pub.com --env-var=VARIABLE_NAME
Or remove it using:
  dart pub token remove https://some-pub.com
```

Warnings like this wouldn't be perfect, but they would certainly help cases where people accidentally pass the value of an environment variable rather than the actual env var name. Or pass the name, but add extract spaces or prefix it with a dollar sign by accident. These things are easy to accidentally do, and unless you read carefully, they can be hard to debug.

jonasfj (commented):
This generally looks good to me, we should fix the tests:
```
  Expected stderr:
  | Environment variable TOKEN is not defined.
  Got:
  X Environment variable `TOKEN` is not defined.
```
And also restrict the strings we accept as legal environment variable names in `--env-var`.

sigurdm:
I am ok with any solution. In my world it is enough to report an error at the usage site, and therefore I don't see much need for the other warnings or restrictions (maybe a special-case warning for vars starting with $ or % could be useful) but I don't see big harm in doing them if we (the pub-collective we) think it can help someone.

jonasfj (approved):

PR #3116, opened 2021-09-15
Do not allow: dart pub token remove --no-all

jonasfj:
Just fixing another nit. So that instead of:
```
$ dart run pub token remove --help
The [hosted-url] for a package repository must be specified.

Usage: pub token remove
-h, --help        Print this usage information.
    --[no-]all    Remove all secret tokens.

```

We have:
```
$dart run pub token remove --help
Remove secret token for package repository.

Usage: pub token remove
-h, --help    Print this usage information.
    --all     Remove all secret tokens.

```

Maybe we should make some tests using `golden_file.dart` ?

/cc @themisir

### DIFF
--- a/lib/src/command/token_remove.dart
+++ b/lib/src/command/token_remove.dart
@@ -23,7 +23,11 @@ class TokenRemoveCommand extends PubCommand {
   bool get isAll => argResults['all'];
 
   TokenRemoveCommand() {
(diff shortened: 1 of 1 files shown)


sigurdm (approved):

sigurdm:
> Maybe we should make some tests using golden_file.dart ?

I think we should have that for all our commands: https://github.com/dart-lang/pub/issues/3122

PR #3118, opened 2021-09-15
Don't resolve to rectracted package versions

szakarias:


lib/src/solver/package_lister.dart
```diff
@@ -36,6 +36,13 @@ class PackageLister {
   /// solve isn't a `pub get`.
   final PackageId _locked;
 
+  // The version of this package that, if retracted, is still allowed in the
+  // current version solve.
+  //
+  // We don't allow a retracted version during solving unless it is already
+  // present in pubspec.lock or pinned in dependency_overrides.
```
jonasfj, line 43:
```suggestion
  // present in `pubspec.lock` or pinned in `dependency_overrides`.
```

jonasfj: Should this just be a `Version`, or not a `PackageId`? maybe it doesn't matter...

jonasfj: Please document that this is `null` if no exceptions are to be made.

lib/src/solver/version_solver.dart
```diff
@@ -431,7 +431,9 @@ class VersionSolver {
       List<PackageId> packages) async {
```
jonasfj, line 431: Should we specify that `_getAvailableVersions` never includes retracted versions, instead? Where is the result used?

lib/src/solver/version_solver.dart
```diff
@@ -500,6 +508,17 @@ class VersionSolver {
     return _lockFile.packages[package];
   }
 
+  /// Gets the version of [package] which can be allowed during version solving
+  /// even if that version is marked as retracted.
+  ///
+  /// We only allow resolving to a retracted version if it is already in the
+  /// pubspec.lock or pinned in dependency_overrides.
```
jonasfj, line 515:
```suggestion
  /// `pubspec.lock` or pinned in `dependency_overrides`.
```

lib/src/source.dart
```diff
@@ -161,15 +161,20 @@ abstract class BoundSource {
   /// Sources should not override this. Instead, they implement [doGetVersions].
   ///
   /// If [maxAge] is given answers can be taken from cache - up to that age old.
-  Future<List<PackageId>> getVersions(PackageRef ref, {Duration maxAge}) {
+  ///
+  /// The [allowedRetractedVersion], if given, is the only version which can be
+  /// selected even if it is marked as retracted. Otherwise, all the returned
```
jonasfj, line 166: Maybe improve the sentence a little...

test/get/hosted/resolve_with_retracted_package_versions_test.dart
```diff
@@ -0,0 +1,120 @@
...
+            '''Because every version of foo depends on bar ^1.0.0 which doesn't match any versions, foo is forbidden. 
+            So, because myapp depends on foo 1.0.0, version solving failed.''');
+  });
+
+  /// Currently retraction does not affect prioritization. I.e., if
+  /// pubspec.lock already contains a retracted version, which is the newest
+  /// satisfying the dependency contstraint we will not choose to downgrade.
+  /// In this case we expect a newer version to be published at some point which
+  /// will then cause pub upgrade to choose that one.
+  test('Allow retracted version when it was already in pubspec.lock', () async {
```
jonasfj, line 51: nit: `///` for dartdoc documentation comments, maybe just use `//`

jonasfj (approved):

lib/src/source.dart
```diff
@@ -161,15 +161,20 @@ abstract class BoundSource {
...
+      {Duration maxAge, Version allowedRetractedVersion}) {
     if (ref.isRoot) {
       throw ArgumentError('Cannot get versions for the root package.');
     }
     if (ref.source != source) {
       throw ArgumentError('Package $ref does not use source ${source.name}.');
     }
 
-    return doGetVersions(ref, maxAge);
+    return doGetVersions(ref, maxAge, allowedRetractedVersion);
```
sigurdm, line 177: If possible we should do the filtering here instead of in each concrete source.

lib/src/source.dart
```diff
@@ -161,15 +161,20 @@ abstract class BoundSource {
   /// Sources should not override this. Instead, they implement [doGetVersions].
   ///
   /// If [maxAge] is given answers can be taken from cache - up to that age old.
-  Future<List<PackageId>> getVersions(PackageRef ref, {Duration maxAge}) {
+  ///
+  /// The [allowedRetractedVersion], if given, is the only version which can be
+  /// selected even if it is marked as retracted. Otherwise, all the returned
+  /// IDs correspond to non-retracted versions.
+  Future<List<PackageId>> getVersions(PackageRef ref,
+      {Duration maxAge, Version allowedRetractedVersion}) {
```
sigurdm, line 169: Maybe this should be 
```suggestion
      {Duration maxAge, bool allowedRetractedVersions = false}) {
```
To not let the solver logic affect the version listing

lib/src/solver/package_lister.dart
```diff
@@ -36,6 +36,15 @@ class PackageLister {
...
   final PackageId _locked;
 
+  // The version of this package that, if retracted, is still allowed in the
+  // current version solve.
+  //
+  // We don't allow a retracted version during solving unless it is already
+  // present in `pubspec.lock` or pinned in `dependency_overrides`.
+  //
+  // This is `null` if there is no retracted version that can be allowed.
+  final Version allowedRetractedVersion;
```
sigurdm, line 46: I have a feeling we might want `_locked` and `allowedRetractedVersion` should be the same type.

lib/src/solver/version_solver.dart
```diff
@@ -431,7 +431,9 @@ class VersionSolver {
       List<PackageId> packages) async {
     var availableVersions = <String, List<Version>>{};
     for (var package in packages) {
-      var cached = _packageListers[package.toRef()]?.cachedVersions;
+      var packageLister = _packageListers[package.toRef()];
```
sigurdm, line 434: I get a feeling we should never ask the packageLister here.

The sources will cache the available versions themselves.

This might be for another PR...

lib/src/solver/package_lister.dart
sigurdm, line 46: Can this be private?

lib/src/source.dart
szakarias: As we discussed, we will not do this, but keep the `Version allowedRetractedVersion` argument.

lib/src/solver/package_lister.dart
szakarias: Done

szakarias: I have the same feeling, but I think it might be `_locked` that needs to change. I'll add to the cleanup I want to do in a separate PR.

lib/src/solver/version_solver.dart
szakarias: Noted for another PR

lib/src/solver/version_solver.dart
szakarias: I agree, done.

lib/src/source.dart
```diff
@@ -161,15 +161,32 @@ abstract class BoundSource {
...
     }
     if (ref.source != source) {
       throw ArgumentError('Package $ref does not use source ${source.name}.');
     }
 
-    return doGetVersions(ref, maxAge);
+    var versions = await doGetVersions(ref, maxAge);
+
+    versions = (await Future.wait(versions.map((id) async {
+      final packageStatus = await status(id, Duration(days: 3));
```
sigurdm, line 180: Use `maxAge` here instead of 3 days.

lib/src/source/hosted.dart
```diff
@@ -454,6 +454,7 @@ class BoundHostedSource extends CachedSource {
       versionListing ??= await _cachedVersionListingResponse(ref, maxAge);
     }
     versionListing ??= await _scheduler.schedule(ref);
+
```
sigurdm, line 457: Perhaps revert this newline...

test/get/hosted/resolve_with_retracted_package_versions_test.dart
```diff
@@ -0,0 +1,120 @@
...
+      ..serve('foo', '1.0.0', deps: {'bar': '^1.0.0'})
+      ..serve('bar', '1.0.0'));
+    await d.appDir({'foo': '1.0.0'}).create();
+
+    globalPackageServer
+        .add((builder) => builder..retractPackageVersion('bar', '1.0.0'));
+    await pubGet(
+        error:
+            '''Because every version of foo depends on bar ^1.0.0 which doesn't match any versions, foo is forbidden. 
+            So, because myapp depends on foo 1.0.0, version solving failed.''');
```
sigurdm, line 43: Nit: the tripple-quote string ignores leading white-space, so you can write the (IMO) slightly nicer:
```suggestion
            '''
Because every version of foo depends on bar ^1.0.0 which doesn't match any versions, foo is forbidden. 
So, because myapp depends on foo 1.0.0, version solving failed.''');
```

sigurdm (approved):
LGTM with a few comments

lib/src/source.dart
szakarias: Of course! Done.

lib/src/source/hosted.dart
szakarias: Done

sigurdm (approved):
LGTM

PR #3120, opened 2021-09-15
Improved handling for authentication errors

### LINKED ISSUE #3117: Invalid www-authenticate produces poor CLI output
<details> 
<summary> Setup a dummy test server using: https://beeceptor.com/console/other-pub</summary> 

9C3ggW9Lukzxj8r
</details>

Steps to reproduce:
```
dart create _dummy_pkg
cd _dummy_pkg

echo -e "\npublish_to: https://other-pub.free.beeceptor.com/\n" >> pubspec.yaml
echo 'Whatever' > LICENSE

$ dart ../bin/pub.dart publish
Publishing _dummy_pkg 0.0.1 to https://other-pub.free.beeceptor.com:
|-- CHANGELOG.md
|-- LICENSE
|-- README.md
|-- analysis_options.yaml
|-- bin
|   '-- hack.dart
'-- pubspec.yaml
Package validation found the following potential issues:
* It's strongly recommended to include a "homepage" or "repository" field in your pubspec.yaml
* ./CHANGELOG.md doesn't mention current version (0.0.1).
  Consider updating it with notes on this version prior to publication.
* Packages with an SDK constraint on a pre-release of the Dart SDK should themselves be published as a pre-release version. If this package needs Dart version 2.15.0-51.0.dev, consider publishing the package as a pre-release instead.
  See https://dart.dev/tools/pub/publishing#publishing-prereleases For more information on pre-releases.

Package has 3 warnings.. Do you want to publish _dummy_pkg 0.0.1 (y/N)? y
Uploading... 
Failed to parse header val
```
This happens if the header value is:
```
www-authenticate:  Bearer realm=pub message="Bad token visit other-pub.com/tokens to get a token!"
```
Note the missing comma after `realm=pub`.

We should probably make sure we have test cases covering this. In this case, we have to just ignore the `www-authenticate` header, maybe print a warning that we were unable to parse it, but that's it. We can't assume the user is also the server administrator, so we have to produce output that is reasonably useful even when the server is giving us a bad header.

/cc @themisir

themisir:
Closes #3117

### DIFF
--- a/lib/src/authentication/client.dart
+++ b/lib/src/authentication/client.dart
@@ -48,15 +48,17 @@ class _AuthenticatedClient extends http.BaseClient {
 
     try {
       final response = await _inner.send(request);
-      if (response.statusCode == 401) _handleError(response);
+      if (response.statusCode == 401) {
+        _throwAuthException(response);
+      }
       return response;
-    } on PubHttpException catch (error) {
-      if (error.response?.statusCode == 403) _handleError(error.response);
+    } on PubHttpException catch (e) {
+      if (e.response?.statusCode == 403) _throwAuthException(e.response);
       rethrow;
     }
   }
 
-  void _handleError(http.BaseResponse response) {
+  void _throwAuthException(http.BaseResponse response) {
     String serverMessage;
     try {
       final wwwAuthenticateHeaderValue =
--- a/lib/src/authentication/client.dart
+++ b/lib/src/authentication/client.dart
@@ -55,7 +55,9 @@ class _AuthenticatedClient extends http.BaseClient {
       }
       return response;
     } on PubHttpException catch (e) {
-      if (e.response?.statusCode == 403) _throwAuthException(e.response);
+      if (e.response?.statusCode == 403) {
+        _throwAuthException(e.response);
+      }
       rethrow;
     }
   }
--- a/lib/src/authentication/client.dart
+++ b/lib/src/authentication/client.dart
@@ -75,7 +75,7 @@ class _AuthenticatedClient extends http.BaseClient {
         final challenge = AuthenticationChallenge.parseHeader(header)
             .firstWhereOrNull((challenge) =>
                 challenge.scheme == 'bearer' &&
-                challenge.parameters['realm'] == 'pub'
+                challenge.parameters['realm'] == 'pub' &&
                 challenge.parameters['message'] != null);
         serverMessage = challenge?.parameters['message'];
       } on FormatException {
--- a/lib/src/authentication/client.dart
+++ b/lib/src/authentication/client.dart
@@ -6,7 +6,9 @@
 
 import 'dart:io';
 
+import 'package:collection/collection.dart';
 import 'package:http/http.dart' as http;
+import 'package:http_parser/http_parser.dart';
 
 import '../exceptions.dart';
 import '../http.dart';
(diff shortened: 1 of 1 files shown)


lib/src/authentication/client.dart
```diff
@@ -34,17 +40,68 @@ class _AuthenticatedClient extends http.BaseClient {
...
+    String serverMessage;
+    try {
+      final wwwAuthenticateHeaderValue =
+          response.headers[HttpHeaders.wwwAuthenticateHeader];
+      if (wwwAuthenticateHeaderValue != null) {
+        final parsedValue = HeaderValue.parse(wwwAuthenticateHeaderValue,
+            parameterSeparator: ',');
+        if (parsedValue.parameters['realm'] == 'pub') {
+          serverMessage = parsedValue.parameters['message'];
+        }
```
jonasfj, line 69: I doubt this parsing complies with [RFC 7235 section 4.1](https://datatracker.ietf.org/doc/html/rfc7235#section-4.1).

Also make sure to reference `https://datatracker.ietf.org/doc/html/rfc7235#section-4.1` in code comments.
So that we clearly state that we intend to implement this specification correctly.

Problems:
1.  We never check the `auth-scheme` from the `challenge` (we only support `Bearer`).
2.  We do not support multiple `challenge` entries in `www-authenticate` header.

Relevant grammar from RFC 7235:
```
WWW-Authenticate = 1#challenge

challenge   = auth-scheme [ 1*SP ( token68 / #auth-param ) ]

auth-scheme    = token

auth-param     = token BWS "=" BWS ( token / quoted-string )

token68        = 1*( ALPHA / DIGIT /
                        "-" / "." / "_" / "~" / "+" / "/" ) *"="
```

(Notice that [section 1.2](https://datatracker.ietf.org/doc/html/rfc7235#section-1.2) specifies `#` operator as comma separated list, e.g. `#foo` is zero or more `foo` separated by commas.

**Example**
```
WWW-Authenticate: Bearer realm="apps", type=1, title="hello \"user\"", Basic realm="webdav", d2d2d2__, Bearer realm=pub, message="Get a token at \"https://other-pub.com\""
```

We might want to create a file with a function for parsing this, and write a handful of unit-tests for this.

Parsing this might not be entirely trivial as there can be commas inside the quoted strings. And quoted-string probably has a few corner cases around escaping.

In practice I doubt many servers will support multiple authentication challenges. But if we can parse these challenges correctly now, that would be fantastic. Because maybe in the future we'll see things like:

```
WWW-Authenticate: Bearer realm=pub, message="Get a token at \"https://other-pub.com\"",
                  Bearer realm=pub, serviceAccount="<service-account-name>@<project-id>.iam.gserviceaccount.com"
```

Hmm, we should probably make sure to iterate through all the challenges provided by the server! And we should probably ignore challenges that with `realm=pub` if they don't include a `message`.

lib/src/authentication/client.dart
```diff
@@ -34,17 +40,68 @@ class _AuthenticatedClient extends http.BaseClient {
...
+      }
+    } catch (_) {
```
jonasfj, line 71: Hopefully, the parsing can only throw `FormatException`, we should check this, but if so it would be preferable to do:
```suggestion
    } on FormatException {
```

lib/src/authentication/client.dart
```diff
@@ -34,17 +40,68 @@ class _AuthenticatedClient extends http.BaseClient {
...
+    if (credential != null &&
+        credential.canAuthenticate(request.url.toString())) {
       request.headers[HttpHeaders.authorizationHeader] =
           await credential.getAuthorizationHeaderValue();
     }
-    return _inner.send(request);
+
+    try {
+      final response = await _inner.send(request);
+      if (response.statusCode == 401) _handleError(response);
```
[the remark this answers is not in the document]
themisir: > In general, please always use if () { ... }

Sure, I've seen in some places (in pub source code) there's code with `if (cond) ...` style thought that's what you're using for short if statements.

lib/src/authentication/client.dart
```diff
@@ -34,17 +40,68 @@ class _AuthenticatedClient extends http.BaseClient {
...
+      return response;
+    } on PubHttpException catch (error) {
+      if (error.response?.statusCode == 403) _handleError(error.response);
```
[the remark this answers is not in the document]
themisir: Yes cause `_ThrowingClient` doesn't throws exception on 401 because it's handled by oauth client.

lib/src/authentication/client.dart
themisir: Current implementation throws `HttpException` afaik. I haven't dig deep into implementation code so I'm not sure if it might throw any other type of exception.

jonasfj (approved):
We really should do some golden tests for these errors... but from where I sit this looks like an improvement.

@sigurdm, any concerns?

themisir:
> We really should do some golden tests for these errors... but from where I sit this looks like an improvement.

Before that we have to do mock server tests

jonasfj:
Okay, this is still a strict improvement, I filed https://github.com/dart-lang/pub/issues/3144 for adding more tests.

PR #3121, opened 2021-09-15
Disable echo mode on secret token input

### LINKED ISSUE #3119: Token masking missing in `dart pub token add`
```
dart pub token add https://other-pub.com
Enter secret token: hello world
Requests to https://other-pub.com will now be authenticated using the secret token.
```

<details>
<summary>Expected result</summary>

```
dart pub token add https://other-pub.com
Enter secret token:
Requests to https://other-pub.com will now be authenticated using the secret token.
```
ie. no output at all, or

```
dart pub token add https://other-pub.com
Enter secret token: ***********
Requests to https://other-pub.com will now be authenticated using the secret token.
```
</details>

This might be complicated, require terminal and platform detection and possibly other hacks, see:
https://stackoverflow.com/questions/6856635/hide-password-input-on-terminal

/cc @themisir

themisir:
Closes #3119

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -575,13 +575,21 @@ Future<bool> confirm(String message) {
 }
 
 /// Writes [prompt] and reads a line from stdin.
-Future<String> stdinPrompt(String prompt) {
+Future<String> stdinPrompt(String prompt, {bool echoMode}) {
   if (runningFromTest) {
     log.message(prompt);
   } else {
     stdout.write('$prompt ');
   }
-  return _stdinLines.first;
+  if (echoMode != null && stdin.hasTerminal) {
+    final previousEchoMode = stdin.echoMode;
+    stdin.echoMode = echoMode;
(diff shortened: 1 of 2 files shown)


sigurdm (approved):
LGTM

sigurdm:
Thanks!

lib/src/io.dart
```diff
@@ -575,13 +575,21 @@ Future<bool> confirm(String message) {
 }
 
 /// Writes [prompt] and reads a line from stdin.
-Future<String> stdinPrompt(String prompt) {
+Future<String> stdinPrompt(String prompt, {bool echoMode}) {
```
jonasfj, line 578: Maybe we should move these utilities into:
https://github.com/dart-lang/cli_util

And make sure that it's possible to do more than one prompt in a process.

This is not important, just a possible cleanup we could do, which would also allow other Dart programs to benefit from the work we're doing.

Issue #3102, opened 2021-09-10
pub publish is ignoring parent .gitignore


ryanheise:
# Environment

- `pub version` or `flutter pub version`: 2.14.0
- OS version: Linux 5.12.14-arch1-1
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

This happens when publishing this directory:

https://github.com/ryanheise/audio_service/tree/56487d8e36f8db8bfab5271094955772bb362d7f/audio_service

If you have run `flutter run` in the `example` directory first, which generates a lot of temporary files (all covered in `../.gitignore`), publishing fails. But if you copy the `.gitignore` file from the parent into the current directory it succeeds.

```
$ flutter packages pub publish --dry-run
```

## Expected behavior

If the pub tool respects `.gitignore` by default, it should look for `.gitignore` in the same places that the git tool does. In federated plugins for Flutter, it is common practice to store multiple related packages in the same git repository, and share the `.gitignore` file at the root of the git repository. In fact the 1st party Flutter plugins by Google, the same approach is used with a single shared `.gitignore` at the root of the git repository.

## Actual behavior

The pub tool only finds `.gitignore` if it is in the current directory which is not how `.gitignore` is supposed to work.

## --trace output

```
$ cd example; flutter run; cd ..
$ flutter packages pub publish --dry-run
...
* Your package is 373.5 MB. Hosted packages must be smaller than 100 MB. Consider adding a .gitignore to avoid including temporary files.
$ git status -s
(empty list)
```

But it succeeds after doing either a `flutter clean` in both the plugin directory and its `example` subdirectory, or by copying the parent `.gitignore` file into the current directory.

sigurdm:
Hmm - yes this is not ideal - we should take any .gitignores into account until we meet a `.git` directory. 

...But there are even more git places gitignore rules might come from...
:(

jonasfj:
I think we should take the opinion that "global git-ignore configuration" is ignored. Looking up parent directories and using `.gitignore` from those folders, if we find a `.git/` folder in some parent folder is fine.

We could probably do that.

PR #3126, opened 2021-09-20
Use ignore files up to git root

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3102

We are still not taking all the same gitignores as git would into account (eg. https://stackoverflow.com/a/5724484) but this is hopefully good enough.

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -227,8 +227,9 @@ class Package {
     var root = dir;
     if (git.isInstalled) {
       try {
-        root = git
-            .runSync(['rev-parse', '--show-toplevel'], workingDir: dir).first;
+        root = p.normalize(
+          git.runSync(['rev-parse', '--show-toplevel'], workingDir: dir).first,
+        );
       } on git.GitException {
         // Not in a git folder.
       }
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -211,20 +211,37 @@ class Package {
   /// [recursive] is true, this will return all files beneath that path;
   /// otherwise, it will only return files one level beneath it.
   ///
-  /// This will take .pubignore and .gitignore files into account. For each
-  /// directory a .pubignore takes precedence over a .gitignore.
+  /// This will take .pubignore and .gitignore files into account.
+  ///
+  /// If [dir] is inside a git repository, all ignore files from the repo root
+  /// are considered.
+  ///
+  /// For each directory a .pubignore takes precedence over a .gitignore.
   ///
   /// Note that the returned paths won't always be beneath [dir]. To safely
   /// convert them to paths relative to the package root, use [relative].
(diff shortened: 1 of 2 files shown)


jonasfj (approved):

jakemac53:
This caused a global breakage for package:test - since I published from an SDK that included this commit https://github.com/dart-lang/test/issues/1594.

The issue is that we had a `.gitignore` for a file that was actually required for the package (and included in the git repository). It was presumably added to the git repository before the gitignore lines were added.

sigurdm:
We actually validate for files that are .gitignored but checked into git. But  we forgot to update the validator to follow along with this PR.
https://github.com/dart-lang/pub/pull/3169 should fix this.

sigurdm:
I would rather not revert this, as there are issues on the other side of the coin as well (such as #3102).

jakemac53:
That should be sufficient to help avoid this issue - although if we already know those files are checked into git I think we should probably just publish them with the repo instead of warning about them? Or do we have known use cases for that as well?

sigurdm:
>  if we already know those files are checked into git I think we should probably just publish them with the repo instead of warning about them? Or do we have known use cases for that as well

That was part of the reason for implementing .pubignore - sometimes a file should be ignored by git, but published by pub and vice versa.

See issues #2222, #2786, #2566

We implemented it such that a .pubignore in a folder takes precedence over a .gitignore in the same folder. That kind of override would not be possible if we just asked git to evaluate the .gitignores.

I start to think we over engineered this. But I am sure other approaches would have their downsides too.

jakemac53:
It does seem like changing this behavior at this point might be more likely to cause additional problems for people, so I am fine with leaving it as is. If I had gotten the validation warning the issue likely would not have happened, so assuming that goes in it should be good enough.

> That was part of the reason for implementing .pubignore - sometimes a file should be ignored by git, but published by pub and vice versa.

Ya, and we can (and probably will) resolve this in package:test by using `.pubignore`. But it feels like we have the wrong default here to me. I would expect by default my published package to contain all the files that are contained in my git repo - including those that _would be_ covered by the `.gitignore` but were explicitly added.

> See issues #2222, #2786, #2566

I would argue that what https://github.com/dart-lang/pub/issues/2222 is asking for should not be supported. All files published should be committed to the git repo. Otherwise you can't have a tag on git that matches up with what you have published, and there are a ton of foot guns around those files being out of date/missing/etc.

> We implemented it such that a .pubignore in a folder takes precedence over a .gitignore in the same folder. That kind of override would not be possible if we just asked git to evaluate the .gitignores.

This one does get interesting I agree - specifically for https://github.com/dart-lang/pub/issues/2566. I guess maybe I would have taken a layered approach - where you still always only include files committed by git and then the `.pubignore` is only used to exclude additional files. But that only works if you agree with my assertion that we shouldn't publish files that aren't committed to git :).

PR #3128, opened 2021-09-20
Migrate exit_codes.dart to null-safety

jonasfj:


sigurdm (approved):
Thus it begins....

LGTM

PR #3132, opened 2021-09-20
Migrate utils and dependencies to null-safety

jonasfj:
I couldn't stop myself... let's see if tests pass... We should probably move to `package:yaml_edit` and migrate `ignore.dart` next...
After that I think it's going to be hard to identify things we migrate alone.

### DIFF
--- a/lib/src/exceptions.dart
+++ b/lib/src/exceptions.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-// @dart=2.10
-
 import 'dart:io';
 import 'dart:isolate';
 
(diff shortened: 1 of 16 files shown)


lib/src/exceptions.dart
```diff
@@ -92,11 +90,14 @@ class ConfigException extends ApplicationException {
...
-  final Sdk missingSdk;
-
-  PackageNotFoundException(String message,
-      {innerError, StackTrace innerTrace, this.missingSdk})
-      : super(message, innerError, innerTrace);
+  final Sdk? missingSdk;
+
+  PackageNotFoundException(
+    String message, {
+    Object? innerError,
```
szakarias, line 97: It's as bad as before, but it would be nice if we could be more specific here.

szakarias (approved):

PR #3133, opened 2021-09-20
Shorthand syntax for hosted dependencies

### LINKED ISSUE #1806: Shorter syntax for hosted dependencies on a different pub server
_From @robbecker-wf on February 16, 2018 17:26_

# Overview

It's rather cumbersome and repetitive to specify hosted dependencies that are on a different server than the main public server. Let's take a look at how a dependency is declared currently

```
dependencies:
  my_package_name:
    hosted:
      name: my_package_name # duplicated information in this dependency
      url: https://pub.domain.com  # duplicated information across multiple dependencies
    version: ^2.1.0
  my_package_name2:
    hosted:
      name: my_package_name2  # duplicated information in this dependency
      url: https://pub.domain.com  # duplicated information across multiple dependencies
    version: ^1.2.3
```

As you can see, there is a lot of duplicated information, careful indentation and key names to remember or copy and paste to add a new hosted dependency. The minimal needed information is the dependency **name**, **pub server name**, and **version**. 

# Proposal
```
dependencies:
  browser: ^0.10.0 # Comes from the main public pub server. pub.dartlang.org
  rxdart@pub.dartlang.org: ^0.15.0 # Also comes from the main public pub server. pub.dartlang.org
  my_package_name@pub.domain.com: ^2.1.0 # Comes from a private hosted pub server
  my_package_name2@pub.domain.com: ^1.2.3
```
In this scenario, the fact that it is a hosted dependency and the https scheme can be inferred.
If the `@pub.domain.com` is left off, then the default public pub server would be defaulted.

_Copied from original issue: dart-lang/sdk#32201_

simolus3:
- Support two shorthand forms of non-default hosted dependencies:
  - by omitting its name: `foo: {hosted: {url: <url>}}`
  - by using a URL directly: `foo: {hosted: <url>}`
- Only allow the new syntax when adding a min SDK constraint of `2.15`

Closes #1806.

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -19,6 +19,7 @@ import '../authentication/client.dart';
 import '../exceptions.dart';
 import '../http.dart';
 import '../io.dart';
+import '../language_version.dart';
 import '../log.dart' as log;
 import '../package.dart';
 import '../package_name.dart';
@@ -182,8 +183,10 @@ class HostedSource extends Source {
   /// given name from the default host, while a map with keys "name" and "url"
   /// refers to a package with the given name from the host at the given URL.
   @override
-  PackageRef parseRef(String name, description, {String containingPath}) {
-    return PackageRef(name, this, _parseDescription(name, description));
+  PackageRef parseRef(String name, description,
+      {String containingPath, LanguageVersion languageVersion}) {
+    return PackageRef(
+        name, this, _parseDescription(name, description, languageVersion));
   }
 
   @override
@@ -208,13 +211,21 @@ class HostedSource extends Source {
   ///
   /// If the package parses correctly, this returns a (name, url) pair. If not,
   /// this throws a descriptive FormatException.
-  _HostedDescription _parseDescription(String packageName, description) {
+  _HostedDescription _parseDescription(
+      String packageName, description, LanguageVersion version) {
     if (description == null) {
       // Simple dependency without a `hosted` block, use the default server.
       return _HostedDescription(packageName, defaultUrl);
     }
 
+    final canUseShorthandSyntax = version >= _minVersionForShorterHostedSyntax;
+
     if (description is String) {
+      if (!canUseShorthandSyntax) {
+        throw FormatException('Using `hosted:` with a direct URL requires a '
+            'min Dart SDK constraint of $_minVersionForShorterHostedSyntax!');
+      }
+
       // We have a dependency like `foo: {hosted: '<url>'}`
       return _HostedDescription(
           packageName, validateAndNormalizeHostedUrl(description));
@@ -224,9 +235,12 @@ class HostedSource extends Source {
       throw FormatException('The description must be a package name or map.');
     }
 
-    var name = description['name'] ?? packageName;
+    var name = description['name'];
+    if (canUseShorthandSyntax) name ??= packageName;
+
     if (name is! String) {
-      throw FormatException("The 'name' key must have a string value.");
+      throw FormatException("The 'name' key must have a string value without "
+          'a min Dart SDK constraint of $_minVersionForShorterHostedSyntax.');
     }
 
     var url = defaultUrl;
@@ -240,6 +254,9 @@ class HostedSource extends Source {
 
     return _HostedDescription(name, url);
   }
+
+  static const LanguageVersion _minVersionForShorterHostedSyntax =
+      LanguageVersion(2, 15);
 }
 
 /// Information about a package version retrieved from /api/packages/$package
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -173,15 +173,14 @@ class HostedSource extends Source {
   /// Ensures that [description] is a valid hosted package description.
   ///
   /// Simple hosted dependencies only consist of a plain string, which is
-  /// resolved against the default host.
+  /// resolved against the default host. In this case, [description] will be
+  /// null.
   ///
   /// Hosted dependencies may also specify a custom host from which the package
-  /// is fetched.
-  /// TODO
+  /// is fetched. There are two syntactic forms of those dependencies:
   ///
-  /// There are two valid formats. A plain string refers to a package with the
-  /// given name from the default host, while a map with keys "name" and "url"
-  /// refers to a package with the given name from the host at the given URL.
+  ///  1. With an url and an optional name in a map: `hosted: {url: <url>}`
+  ///  2. With a direct url: `hosted: <url>`
   @override
   PackageRef parseRef(String name, description,
       {String containingPath, LanguageVersion languageVersion}) {
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -128,7 +128,7 @@ class HostedSource extends Source {
   /// should be downloaded. [url] most be normalized and validated using
   /// [validateAndNormalizeHostedUrl].
   PackageRef refFor(String name, {Uri url}) =>
-      PackageRef(name, this, _descriptionFor(name, url));
+      PackageRef(name, this, _HostedDescription(name, url));
 
   /// Returns an ID for a hosted package named [name] at [version].
   ///
@@ -136,7 +136,7 @@ class HostedSource extends Source {
   /// should be downloaded. [url] most be normalized and validated using
   /// [validateAndNormalizeHostedUrl].
   PackageId idFor(String name, Version version, {Uri url}) =>
-      PackageId(name, this, version, _descriptionFor(name, url));
+      PackageId(name, this, version, _HostedDescription(name, url));
 
   /// Returns the description for a hosted package named [name] with the
   /// given package server [url].
@@ -154,7 +154,8 @@ class HostedSource extends Source {
 
   @override
   dynamic serializeDescription(String containingPath, description) {
-    return _asDescription(description).originalNode;
+    final desc = _asDescription(description);
+    return _descriptionFor(desc.packageName, desc.uri);
   }
 
   @override
@@ -182,14 +183,23 @@ class HostedSource extends Source {
   /// refers to a package with the given name from the host at the given URL.
   @override
   PackageRef parseRef(String name, description, {String containingPath}) {
-    _parseDescription(name, description);
-    return PackageRef(name, this, description);
+    return PackageRef(name, this, _parseDescription(name, description));
   }
 
   @override
   PackageId parseId(String name, Version version, description,
       {String containingPath}) {
-    return PackageId(name, this, version, _asDescription(description));
+    final serializedDescription = (description as Map).cast<String, String>();
+
+    return PackageId(
+      name,
+      this,
+      version,
+      _HostedDescription(
+        serializedDescription['name'],
+        Uri.parse(serializedDescription['url']),
+      ),
+    );
   }
 
   _HostedDescription _asDescription(desc) => desc;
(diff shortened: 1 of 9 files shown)


lib/src/source/hosted.dart
```diff
@@ -225,6 +268,26 @@ class _VersionInfo {
   _VersionInfo(this.pubspec, this.archiveUrl, this.status);
 }
 
+class _HostedDescription {
```
jonasfj, line 271: Can we give this a documentation comment? Maybe say it's the value of `PackageRef.description` for `HostedSource`.

I guess this is replacing the `Pair<String, Uri>` we had before. That's probably good :D

lib/src/source/hosted.dart
```diff
@@ -212,8 +252,11 @@ class HostedSource extends Source {
       url = validateAndNormalizeHostedUrl(u);
     }
 
-    return Pair<String, Uri>(name, url);
+    return _HostedDescription(name, url);
   }
+
+  static const LanguageVersion _minVersionForShorterHostedSyntax =
```
jonasfj, line 258: Adding a few comments here might help future generations maintain this code :D 

```suggestion
  /// Minimum language version at which short hosted syntax is supported.
  ///
  /// This allows `hosted` dependencies to be expressed as:
  /// ```yaml
  /// dependencies:
  ///   foo:
  ///     hosted: https://some-pub.com/path
  ///     version: ^1.0.0
  /// ```
  static const LanguageVersion _minVersionForShortHostedSyntax =
```

lib/src/source/hosted.dart
```diff
@@ -152,55 +153,94 @@ class HostedSource extends Source {
...
+    }
+
+    final canUseShorthandSyntax =
+        version == null || version >= _minVersionForShorterHostedSyntax;
+
     if (description is String) {
-      return Pair<String, Uri>(description, defaultUrl);
+      if (!canUseShorthandSyntax) {
+        throw FormatException('Using `hosted:` with a direct URL requires a '
+            'min Dart SDK constraint of $_minVersionForShorterHostedSyntax!');
```
jonasfj, line 226:
```suggestion
        throw FormatException('The syntax `hosted: <hosted-url>` requires a '
            'minimum Dart SDK constraint of $_minVersionForShorterHostedSyntax!');
```

jonasfj (commented):
hmm, we should probably fix the test cases.. but otherwise this looks promising :D

Also we should add some integration style tests.. like the ones we have in `will_normalize_hosted_url_test.dart`

simolus3:
I've added documentation comments and a basic integration test.

lib/src/source/hosted.dart
```diff
@@ -227,14 +227,16 @@ class HostedSource extends Source {
...
-            'minimum Dart SDK constraint of $_minVersionForShorterHostedSyntax!');
+      // Old versions of pub (pre Dart 2.15) interpret `hosted: foo` as
+      // `hosted: {name: foo, url: <default>}`.
+      // For later versions, we treat it as `hosted: {name: <inferred>,
+      // url: foo}` if a user opts in by raising their min SDK environment.
+      if (canUseShorthandSyntax) {
+        return HostedDescription(
+            packageName, validateAndNormalizeHostedUrl(description));
+      } else {
+        return HostedDescription(description, defaultUrl);
```
jonasfj, line 238: Hmm, I know I said I thought this was smart... but thinking about it now, it has a clear downside:
If you write `hosted: <url>` with the wrong SDK constraint, then you'll get a weird error about the name containing illegal characters or worse some package can't be found :see_no_evil: 
You won't be told that the format `hosted: <url>` is only supported with lower bound SDK constraint of 2.15. Hence, users will try all sorts of weird things, instead of making the right choice and increasing the lower bound SDK constraint.

---------------------

This makes a lot of sense when interpreting the `pubspec.yaml` of a package that we depend on, though, because here we have to be compatible.

But if we see `hosted: foo` in the root `pubspec.yaml` we could choose to complain that lower-bound SDK constraint isn't 2.15. This is a breaking change, but it's only breaking the root package `pubspec.yaml` -- this is usually a file the user can change, it's not breaking any packages that have been published.

---------------

Another option is that we could choose to throw a `FormatException` if:
 * Lower bound SDK constraint is less than 2.15 (short syntax not supported), and,
 * We see `hosted: <value>`, and,
 * `<value>` is not a valid package name (`^(?:[a-zA-Z_]+)(\.?:[a-zA-Z_]+)*$`, see https://github.com/dart-lang/pub/blob/master/lib/src/pubspec.dart#L26-L32)

This way we get to point out the fact that `hosted: <url>` doesn't work pre-2.15 SDK constraint, and `hosted: <value>` is not a valid package name.

It's not perfect because `hosted: pub.example.com` would still cause some obscure error about `package pub.example.com not found on pub.dev`...

--------

Any ideas?

test/pubspec_test.dart
```diff
@@ -420,19 +442,6 @@ dependencies:
...
-          expectPubspecException(
-              '''
-name: pkg
-dependencies:
-  foo:
-    hosted: https://example.org/pub/
-''',
-              (pubspec) => pubspec.dependencies,
-              'The syntax `hosted: <hosted-url> requires a` minimum Dart SDK '
-                  'constraint of 2.15!');
```
jonasfj: hmm, yeah, it's probably best that find a way to produce an error like this... Rather than interpreting the URL as a package name.

We could also print a warning.. But maybe it's enough to be able to detect the difference between a package name and a URL.

lib/src/source/hosted.dart
simolus3: Distinguishing between the root pubspec and pubspecs of dependencies will further complicate parsing (we'd have to add this option to `Pubspec.fromMap` and pass it down to the sources). I think I like the option of throwing a `FormatException` for values that aren't valid package names for Dart <2.15.

> It's not perfect because hosted: `pub.example.com` would still cause some obscure error about `package pub.example.com not found on pub.dev` ...

We require hosted urls to have a scheme (regardless of the syntax used), so at least there's no ambiguity for valid hosts...

From the comment on the regex, it looks like package names with dots in them are a Google-internal thing not realistically used by anyone else. So maybe we could just consider package names with a single segment (`^[a-zA-Z_]+$`) when attempting to check whether a url was meant. Then this would only be a problem if some pubspec in Google-internal code used `hosted: foo.bar.baz` which I imagine is somewhat easier (for you :D) to find and migrate?

Alternatively, we could also emit a warning only instead of refusing to accept the pubspec. In that case users would get something like "Warning: hosted: pub.example.com will be interpreted as a package name. Try raising the minimum SDK version to 2.15 for pub to interpret it as a host" followed by `package pub.example.com not found on pub.dev`.

jonasfj: Okay, let's throw a `FormatException` if:
 * Lower bound SDK constraint is less than 2.15 (short syntax not supported), and,
 * We see `hosted: <value>`, and,
 * `<value>` is not a valid package name without dots, ie. not matching `^[a-zA-Z_]+[a-zA-Z0-9_]*$`.

This will cause problems for packages containing `hosted: foo.bar.baz`, but such names can't be published to `pub.dev`, and if you are using a custom pub.dev server, then: You have to specify `hosted: {url: ..., name: ...}`, so you won't be able to use the `hosted: <name>`, unless you're overwriting `PUB_HOSTED_URL`.

So this would break users who:
 * custom pub servers, AND,
 * using package names containing dots (`.`), AND,
 * using the obscure syntax `hosted: <name>`.

All of those things happening at the same time seems unlikely. So this seems like a reasonable corner case to break.

lib/src/source/hosted.dart
```diff
@@ -152,55 +154,117 @@ class HostedSource extends Source {
...
+    return PackageId(
+      name,
+      this,
+      version,
+      HostedDescription(serializedDescription['name'],
+          Uri.parse(serializedDescription['url'])),
+    );
   }
 
+  HostedDescription _asDescription(desc) => desc;
```
jonasfj, line 213: What do we need this for? I'm guessing it's a left-over that we can now remove :D

lib/src/source/hosted.dart
```diff
@@ -152,55 +154,117 @@ class HostedSource extends Source {
...
+        if (_looksLikePackageName.hasMatch(description)) {
+          // Valid use of `hosted: package` dependency with an old SDK
+          // environment.
+          return HostedDescription(description, defaultUrl);
+        } else {
+          throw FormatException(
+            'Using `hosted: <url>` is only supported with a minimum SDK '
+            'constraint of $_minVersionForShorterHostedSyntax.\n'
+            'If `$description` was meant as a package name, please use '
+            '`hosted: {name: "$description"}` instead.',
```
jonasfj, line 252: I would suggest omitting this.

test/hosted/short_syntax_test.dart
```diff
@@ -0,0 +1,87 @@
...
+      );
+      final sources = SourceRegistry();
+      final lock =
+          LockFile.load(p.join(d.sandbox, appPath, 'pubspec.lock'), sources);
+
+      expect(
+          lock.packages['foo'].description,
+          isA<HostedDescription>()
+              .having((e) => e.packageName, 'packageName', 'foo')
+              .having((e) => e.uri, 'uri', Uri.parse(globalPackageServer.url)));
```
jonasfj, line 45: Can we not just YAML parse the `pubspec.lock` and check the values? So that we don't have to use `LockFile.load` in tests -- afaik we only use it a few places -- in general we seem to avoid mporting pub logic in the tests, and instead do end-to-end tests.

This would also save us from having to make `HostedDescription` visible testing.

test/pubspec_test.dart
```diff
@@ -294,6 +297,172 @@ dependencies:
...
+    hosted:
+      url: https://example.org/pub/
+      name: bar
+''',
+          sources,
+        );
+
+        var foo = pubspec.dependencies['foo'];
+        expect(foo.name, equals('foo'));
+        expect(foo.source, isA<HostedSource>());
```
jonasfj, line 316: Let's skip this check, and leave `HostedSource` private, testing `foo.source.serializeDescription` should be fine :D

jonasfj (approved):
I only have nits.

If @szakarias or @sigurdm has a chance to take a look that would be good.

--------------

We could do a follow up to add matching support in `dart pub add` when SDK constraint allows using the shorter syntax.

lib/src/source/hosted.dart
simolus3: The description we get from package names or refs is always `dynamic`, so this is essentially a cast. I'll make the code more obvious by spelling out `desc as HostedDescription`. We could also inline the cast into all usages.

lib/src/source/hosted.dart
```diff
@@ -152,55 +153,115 @@ class HostedSource extends Source {
...
 
+  _HostedDescription _asDescription(desc) => desc as _HostedDescription;
+
   /// Parses the description for a package.
   ///
   /// If the package parses correctly, this returns a (name, url) pair. If not,
   /// this throws a descriptive FormatException.
-  Pair<String, Uri> _parseDescription(description) {
+  _HostedDescription _parseDescription(
+      String packageName, description, LanguageVersion version) {
```
sigurdm, line 219: nit: can we rename this local - it can be confused with the version of the package.
```suggestion
      String packageName, description, LanguageVersion languageVersion) {
```

lib/src/source/hosted.dart
```diff
@@ -152,55 +153,115 @@ class HostedSource extends Source {
...
+            packageName, validateAndNormalizeHostedUrl(description));
+      } else {
+        if (_looksLikePackageName.hasMatch(description)) {
+          // Valid use of `hosted: package` dependency with an old SDK
+          // environment.
+          return _HostedDescription(description, defaultUrl);
+        } else {
+          throw FormatException(
+            'Using `hosted: <url>` is only supported with a minimum SDK '
+            'constraint of $_minVersionForShorterHostedSyntax.',
```
sigurdm, line 249: '<=2.15' is not a valid sdk-constraint. You'll need to say '<=2.15.0'. Either this should be rephrased in terms of language version, or we need the final '.0'.

lib/src/source/hosted.dart
```diff
@@ -152,55 +153,115 @@ class HostedSource extends Source {
...
-    if (!description.containsKey('name')) {
-      throw FormatException("The description map must contain a 'name' key.");
-    }
-
     var name = description['name'];
+    if (canUseShorthandSyntax) name ??= packageName;
+
     if (name is! String) {
-      throw FormatException("The 'name' key must have a string value.");
+      throw FormatException("The 'name' key must have a string value without "
```
sigurdm, line 263: Nit: this message will read a bit weird if you have something like:
```
environment:
  sdk: ^2.15.0

dependencies:
  foo:
    hosted:
      name: 12
```

PR #3135, opened 2021-09-21
Report available non-retracted version

szakarias:


test/get/hosted/report_retracted_package_test.dart
```diff
@@ -38,6 +38,54 @@ Changed 2 dependencies!
...
+    // Delete the cache to trigger the report.
+    final barVersionsCache =
+        p.join(globalPackageServer.cachingPath, '.cache', 'bar-versions.json');
+    expect(fileExists(barVersionsCache), isTrue);
+    deleteEntry(barVersionsCache);
+    await pubGet(output: '''
+ Resolving dependencies...
+   bar 1.0.0 (retracted, 2.0.0 available)
+ Got dependencies!
+''');
```
jonasfj, line 65: We could consider making it:
```suggestion
    await pubGet(output: contains('bar 1.0.0 (retracted, 2.0.0 available)'));
```

That way we are less likely to break tests when we change how other parts of the output look.

jonasfj (approved):

test/get/hosted/report_retracted_package_test.dart
szakarias: Done.

PR #3138, opened 2021-09-22
Delete lib/src/yaml_edit and use yaml_edit package instead

szakarias:


jonasfj (approved):
Let's make sure we're ready to roll into Dart SDK once we merge this..

PR #3146, opened 2021-09-24
Warn against likely bad env-var name

jonasfj:
Just figured I would write this while I had a good idea. We should definitely add tests too.

cc @themisir


themisir (approved):
LGTM. But the message changes might affect test cases

sigurdm (approved):
LGTM!

PR #3147, opened 2021-09-25
Tests related to token authentication

themisir:
This PR contains tests for the following cases:

- Successful authentication with a pre-existing opaque token
- Successful authentication with a pre-existing environment token
- Removes token on 401 response
- Persists token on 403 response
- Error message handling for 401 and 403 responses


test/token/error_message_test.dart
```diff
@@ -0,0 +1,84 @@
...
+    // that pub cli supposed to sanitize.
+    //
+    // [1] https://github.com/dart-lang/sdk/blob/main/sdk/lib/_http/http_headers.dart#L653-L662
+  });
+
+  test('trims and prints dirty www-authenticate message', () async {
+    var message = List.generate(2048, (_) => 'a').join();
+
+    respondWithWwwAuthenticate('bearer realm="pub", message="$message"');
+    await expectPubErrorMessage(contains(message.substring(0, 1024)));
```
jonasfj, line 60: This is a useless test... you need to test that it does not contain `message`... if the 2048 chars long message is printed, then it'll still contain a 1024 substring of the message.

test/token/error_message_test.dart
```diff
@@ -0,0 +1,84 @@
...
+
+  test('prints message for bearer challenge for pub realm only', () async {
+    respondWithWwwAuthenticate(
+      'basic realm="pub", message="enter username and password", '
+      'newAuth message="use web portal to login", '
+      'bearer realm="api", message="contact IT dept to enroll", '
+      'bearer realm="pub", '
+      'bearer realm="pub", message="pub realm message"',
+    );
+    await expectPubErrorMessage(contains('pub realm message'));
```
jonasfj, line 82: Love this!

jonasfj (approved):

test/token/error_message_test.dart
themisir: ah ok

PR #3159, opened 2021-09-29
Use source.hasMultipleVersions to decide when to show version constraint.

isoos:
- reduces cross-linking a slight bit
- future-proofing new sources

### DIFF
--- a/lib/src/package_name.dart
+++ b/lib/src/package_name.dart
@@ -9,9 +9,7 @@ import 'package:pub_semver/pub_semver.dart';
 
 import 'package.dart';
 import 'source.dart';
-import 'source/git.dart';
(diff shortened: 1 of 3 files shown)


szakarias (approved):

isoos:
Apparently some output does use ` on foo any from sdk` and the PR changes it to `on foo from sdk`.
What is the expected one? /cc @jonasfj @sigurdm

jonasfj:
> Apparently some output does use `on foo any from sdk` and the PR changes it to `on foo from sdk`.

I think that change is fine.. my guess is that error messages were designed before `SdkSource` was added...

sigurdm (approved):
LGTM

PR #3161, opened 2021-09-30
Introducing command-line interface tests

jonasfj:
This adds a folder where we can write test cases command-line
interface behavior. Starting with tests for `--help` on all commands.


test/cli/testdata/help/pub-downgrade.txt
```diff
@@ -0,0 +1,15 @@
+$ pub downgrade --help
+Downgrade the current package's dependencies to oldest versions.
+
+
+
+Usage: pub downgrade [dependencies...]
+-h, --help               Print this usage information.
+    --[no-]offline       Use cached packages instead of accessing the network.
```
jonasfj, line 8: This just adds what we currently have... but I think we should do some follow-up PRs to improve these messages.
Like why do w need `--no-offline`? If we want to keep it for compatibility can we at-least hide the `--no-offline` variant.

I'm happy to keep `--offline`, and to keep it visible, it just seems the inverse is weird.

sigurdm: I also don't think we need `--no-offline`.

test/cli/help_test.dart
```diff
@@ -0,0 +1,54 @@
...
+  List<String> parents,
+  Iterable<Command> cmds,
+) sync* {
+  if (parents.isNotEmpty) {
+    yield parents;
+  }
+  // Track that we don't add more than once, we don't want to test aliases
+  final names = <String>{};
+  yield* cmds
+      .where((sub) => !sub.hidden && names.add(sub.name))
```
sigurdm, line 33: Maybe we should even do hidden commands - their help might be important to those who knows about it?

jonasfj:
> Perhaps we should instead be testing the interface of the embeddableCommand?

I agree, filed https://github.com/dart-lang/pub/issues/3182 -- as I think we need more test utilities, then we can start to migrating our tests from one to the other.

Issue #3152, opened 2021-09-27
Incorrect errors when trying to publish


stuartmorgan:
The flutter/plugins CI, which uses Flutter `master`, ground to a half sometime in the last couple of days due to bogus errors from `pub publish --dry-run` in our CI

Example (from https://cirrus-ci.com/task/5222632699199488):
```
ERROR: Package validation found the following errors:
* The pubspec is hidden, probably by .gitignore or pubignore.
* You must have a LICENSE file in the root directory.
  An open-source license helps ensure people can legally use your code.
Package validation found the following potential issues:
* Please add a README.md file that describes your package.
* Please add a`CHANGELOG.md` to your package. See https://dart.dev/tools/pub/publishing#important-files.
Sorry, your package is missing some requirements and can't be published yet.
For more information, see: https://dart.dev/tools/pub/cmd/pub-lish.
ERROR: pub finished with exit code 65
Unable to publish url_launcher
```

None of that is correct; the package has all of those things:
https://github.com/flutter/plugins/tree/master/packages/url_launcher/url_launcher

This seems to be happening with all of our packages, and reproduces locally with Flutter `master` but not `stable`.

jonasfj:
@sigurdm might this be related to the `.gitignore` changes we've made?

gspencergoog:
I am seeing this too, in the `flutter/assets-for-api-docs` repo, which also has a "packages" directory with packages in it.

Maybe we can include some of Flutter's repos in the CI test suite for `dart pub` (at least one of `flutter/packages` or `flutter/plugins`) as a "real world" test?

PR #3162, opened 2021-09-30
Don't ignore 'packages/' over the root-dir

sigurdm:
Fixes:  https://github.com/dart-lang/pub/issues/3152

The issue was that we have a legacy ignore rule (always ignore folders named `packages`. After we started interpreting gitignores from the git-root we would no longer venture into `packages/url_launcher` at all.

We should consider if we want to keep the legacy rule. Dart has not created packages/ for a loong time - and it is a pretty generic name bound to come up somewhere else.


szakarias (approved):
LGMT

PR #3163, opened 2021-09-30
Add in-memory cache to store version listings

szakarias:


lib/src/source/hosted.dart
```diff
@@ -351,26 +360,35 @@ class BoundHostedSource extends CachedSource {
...
+      }
+    }
     final cachePath = _versionListingCachePath(ref);
     final stat = io.File(cachePath).statSync();
     final now = DateTime.now();
     if (stat.type == io.FileSystemEntityType.file) {
       if (maxAge == null || now.difference(stat.modified) < maxAge) {
         try {
-          final cachedDoc = jsonDecode(await readTextFileAsync(cachePath));
+          final cachedDoc = jsonDecode(readTextFile(cachePath));
```
sigurdm, line 376: This might bring us back into the file-descriptor situation

lib/src/source/hosted.dart
```diff
@@ -402,6 +420,9 @@ class BoundHostedSource extends CachedSource {
           },
         ),
       );
+      // Delete the entry in the in-memory cache to maintain the invariant that
+      // cached information in memory is the same as that on the disk.
+      _responseCache.remove(ref);
```
sigurdm, line 425: Instead you should insert the new value in the cache - I think.

sigurdm (approved):
LGTM

PR #3168, opened 2021-10-04
migrate 3 leaf files to null-safety

szakarias:


lib/src/package_config.dart
```diff
@@ -23,30 +19,30 @@ class PackageConfig {
   /// Date-time the `.dart_tool/package_config.json` file was generated.
   ///
   /// This property is **optional** and may be `null` if not given.
```
sigurdm, line 21: Perhaps simplify to
```suggestion
  /// `null` if not given.
```

lib/src/package_config.dart
```diff
@@ -23,30 +19,30 @@ class PackageConfig {
...
   /// For `pub` this is the Dart SDK version from which `pub get` was called.
   ///
   /// This property is **optional** and may be `null` if not given.
-  Version generatorVersion;
+  Version? generatorVersion;
 
   /// Additional properties not in the specification for the
   /// `.dart_tool/package_config.json` file.
-  Map<String, dynamic> additionalProperties;
+  Map<String, dynamic>? additionalProperties;
```
sigurdm, line 41: Could this be non-nullable?

sigurdm (approved):
LGTM with comments

PR #3198, opened 2021-10-29
Rename timing category

sigurdm:
It seems the dash prevents the timings from showing up in the UI.


sigurdm:
That seemed not to be true...

Issue #2161, opened 2019-06-28
Can we accommodate more use-cases for dependency_overrides


jonasfj:
Use-cases for `dependency_overrides`:

**A) App development**, when you know better than the version constraint solver and you wish to override the dependencies of your application it makes sense to add `dependency_overrides` in `pubspec.yaml`.

**B) mono repository setup**, when developing in a mono repository it makes sense to override dependencies for the packages in the same repository, such that you are testing with and against these. These overrides also makes sense in CI, in fact they make sense in any context where `dev_dependencies` is installed.

**C) poly repository setup**, when working on package A that is used by package B in two different repositories, it makes sense to temporary override the dependency on A in B and run tests for B to ensure that these work, before submitting a patch for A. This is often the case when fixing a bug in package A. Overrides are not useful in CI, and overrides are generally specific to the local development setup.

------

Are we missing any use-cases? @sigurdm, @lemz90, @StorminGorman,

jonasfj:
Options currently proposed:

 1) Add an `pubspec.override.yaml` file (#2150)
 2) Add an `--ignore-overrides` option (#2151)
 3) Add an `--overrides=my-overrides.yaml` option (https://github.com/dart-lang/pub/pull/2151#issuecomment-506667067)
 4) Add a `dev_dependency_overrides` to `pubspec.yaml` which is only used when installing locally.

----
It seems to me that `dependency_overrides` as we already have solves (A) fairly well.
And perhaps using `dependency_overrides` in the `pubspec.yaml` to solve (B) and (C) is a misappropriation.

This suggests that maybe (2) isn't a good option. Similarly, (4) is not great for (C).

The me the downside with (1) is that introduces a new configuration file that we will have to support going forward (which feels a bit magic). Removing it later would probably be controversial, even if there is just a few people using it -- moreover it would be hard to know if it has few or many users.

----

While (3) isn't easy to use, it could be used by other tools.. it could also be a `--pubspec=my-pubspec.yaml` option. An option to simply specify a different file, instead of `pubspec.yaml`.

In this scenario, the `mono_repo pub get` tool could read the `pubspec.yaml` file, add overrides matching the other packages in the repository, and write the file to `/tmp/<random>-pubspec.yaml` and run `pub get --pubspec=/tmp/<random>-pubspec.yaml`.

[...]

PR #3199, opened 2021-10-30
Allow third-party tools to resolve dependencies

blaugold:
Tools like [melos](https://github.com/invertase/melos) need to be able to resolve dependencies, to link up local packages for development. 

With this change, up-to-date checks of `package_config.json`, `.packages` and `pubspec.lock` are skipped if `package_config.json` has been generated by a tool other than `pub`. This behavior is limited to `pub run` and `dart run`. `pub get` does the normal checks and overrides third-party modifications.

This change also eliminates an [unnecessary call](https://github.com/dart-lang/pub/blob/0035a40f25d027130c0314571da53ffafc6d973b/lib/src/executable.dart#L97) of `Entrypoint.assertUpToDate`. Just a few lines above, `Entrypoint.resolveExecutable` is called, which calls `assertUpToDate` in the process.


jonasfj:
Hmm, should we consider solving this problem by adding first class support for local dependency overrides with something like a `pubspec-overrides.yaml` sitting next to `pubspec.yaml`.

Maybe, this isn't a problem in reality, but if we encourage third-party tools to generate `package-config.json`, might we not risk that adding new features, build steps, etc. to `dart pub get` will be harder in the future.

-------

On the flip side: we could also simplify our logic and only check the generator when timestamps don't align.

jonasfj:
For the record, I'm not principally opposed to this change, just considering/brainstorming possible future implications :)

blaugold:
I think your right @jonasfj, that a cleaner, more universal approach is to support something like `pubspec-overrides.yaml`.
In issue #2161, a few possible solutions have been discussed, and it seems to me that a `pubspec-overrides.yaml` file provides the best trade-off between usability and flexibility. Users can write them manually, or have a tool generate them and still see what the current configuration is. It also works well with other tools running `pub get` for the user, such as flutter and IDE extensions.

Currently, melos is working around `pub` not having something like `pubspec-overrides.yaml`. It recreates the workspace with all of its packages in a temporary directory, but with only the files required to run `pub get`. It then calculates the required `dependency-overrides`, modifies the `pubspec.yaml` files and runs `pub get` in the temporary directory. The last step is to copy some of the files `pub` generated (e.g. `package_config.json`), back to the actual packages.

blaugold:
Following up on my last comment, I'm closing this PR since it's an attempt to enable a solution which is already a workaround.

sigurdm:
In any case - thanks for the contribution. I think it highlights an area where we need to find a better solution!

Salakar:
I would note there are places (at least in Flutter tooling) where 'third-party tool detection and doing something different' is already a thing, e.g. https://github.com/flutter/flutter/blob/master/packages/flutter_tools/test/general.shard/dart/pub_get_test.dart#L49

I'm not against having a `pubspec-overrides.yaml` file but my initial thoughts are that this sounds like a user editable file that is VCS tracked - having tools edit/generate into it could be problematic - especially where the file already exists in a project. I'd personally rather avoid having Melos modify user editable & CCS tracked files 😅

PR #3201, opened 2021-11-01
migrate test/cache and test/downgrade/ to null-safety

szakarias:


test/embedding/embedding_test.dart
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
-
 // @dart=2.10
```
sigurdm, line 4: Was this intended?

sigurdm (approved):
lgtm except for `test/embedding/embedding_test.dart`

test/embedding/embedding_test.dart
szakarias: No, thanks.

PR #3204, opened 2021-11-01
migrate test/lish/, test/run/, test/upgrade/ and golden_file.dart to null-safety

szakarias:


test/lish/many_files_test.dart
```diff
@@ -35,7 +33,7 @@ void main() {
       'the command line', () async {
     await d.validPackage.create();
 
-    int argMax;
+    num argMax;
     if (Platform.isWindows) {
       // On Windows, the maximum argument list length is 8^5 bytes.
       argMax = math.pow(8, 5);
```
sigurdm, line 39: Instead of using `num`, maybe it's better to do 
```suggestion
      argMax = math.pow(8, 5).toInt();
```
Or even just:
```suggestion
      argMax = 32768; // 8^5
```

sigurdm (approved):
LGTM with a single suggestion

test/lish/many_files_test.dart
szakarias: Done

PR #3206, opened 2021-11-01
migrate test/oath2/ test/outdated/ test/token test/validator/ to null-safety

szakarias:


test/oauth2/logout_test.dart
```diff
@@ -30,7 +28,7 @@ void main() {
     await servePackages();
     await d
         .credentialsFile(
-          globalPackageServer,
+          globalPackageServer!,
```
sigurdm, line 31: I start thinking we should have made this one `late`. Most of the changes revolve around this...

sigurdm (approved):
LGTM

test/oauth2/logout_test.dart
szakarias: I have also had the same thought, but not entirely sure it make sense. I'll give it a shot in a separate PR

PR #3212, opened 2021-11-05
Refine what a relative uri means in a git path

sigurdm:


lib/src/source/git.dart
```diff
@@ -147,9 +147,15 @@ class GitSource extends Source {
     return {'relative': relative, 'url': url};
   }
 
-  /// Returns [path] normalized.
+  /// Normalizes [path].
   ///
-  /// Throws a [FormatException] if [path] isn't a relative url or null.
+  /// Throws a [FormatException] if [path] isn't a String parsing as arelative
```
lrhn, line 152: String -> [String]
arelative -> a relative

lib/src/source/git.dart
```diff
@@ -147,9 +147,15 @@ class GitSource extends Source {
...
+  /// url or null.
```
lrhn, line 153: Ob-pedantry: 

url -> URL
null -> `null`

lib/src/source/git.dart
```diff
@@ -159,9 +165,13 @@ class GitSource extends Source {
...
     final parsed = Uri.parse(path);
-    if (parsed.isAbsolute) {
+    if (parsed.hasAbsolutePath ||
+        parsed.hasScheme ||
+        parsed.hasAuthority ||
+        parsed.hasFragment ||
+        parsed.hasQuery) {
       throw FormatException(
-          "The 'path' field of the description must be relative.");
+          "The 'path' field of the description must be a relative path url.");
```
lrhn, line 174: url -> URL

lrhn (approved):

szakarias (approved):

lib/src/source/git.dart
sigurdm: Done

lib/src/source/git.dart
sigurdm: Done

lib/src/source/git.dart
sigurdm: Done

PR #3215, opened 2021-11-07
Add support for pubspec overrides file

### LINKED ISSUE #2161: Can we accommodate more use-cases for dependency_overrides

blaugold:
This change adds support for a new file `pubspec_overrides.yaml` to override parts of `pubspec.yaml`. Overrides are only active for the `get` and `upgrade` commands.

An overrides file has the same structure as `pubspec.yaml`, but only supports a subset of its fields. Currently, only the following top-level field override `pubspec`:

- `dependency_overrides`

All other fields in the overrides file causes an error.

A dependency_overrides, in the overrides file, completely replace a dependency_overrides in `pubspec.yaml`. The two files are merged in a way that preserves source references to provide correct error messages.

When overrides are active, a warning is logged.

Fixes #2161


blaugold:
I rebased onto master.

blaugold:
Sounds good.

blaugold:
@jonasfj
Looking at all the solutions considered in the Google doc, it seems to me that a simple overrides file next to `pubspec.yaml` would support all use cases, while keeping complexity low for users, tooling, implementation and maintenance.

Let me know if there is anything I can do to support making a decision on how to go forward.

yupengyang:
I like this solution! It can save a lot of times if i need to rebase my code from remote after I changed my pubspec.yaml.

If we have the `pubspce.override.yaml`, we can add it to .gitignore and modify our dependencies without any git changes. Then we can directly `git pull` from remote without stashing pubspec.yaml and deleting pubspec.lock

lib/src/command.dart
```diff
@@ -288,6 +290,34 @@ and include the logs in an issue on https://github.com/dart-lang/pub/issues/new
...
+  ///
+  /// This adds the `--pubspec-overrides` option to the command and enables
+  /// the pubspec override feature for the package for which the command
+  /// is being executed.
+  void allowPubspecOverrides() {
+    assert(!_allowPubspecOverrides);
+    _allowPubspecOverrides = true;
+
+    argParser.addOption(
+      'pubspec-overrides',
```
jonasfj, line 304: Let's not have a command line option

jonasfj (changes requested):

lib/src/command.dart
jonasfj: I'll find some time to review more later..

blaugold: Done

lib/src/entrypoint.dart
```diff
@@ -139,6 +139,16 @@ class Entrypoint {
...
 
+  /// Whether the entrypoint package's pubspec contains overrides from an
+  /// overrides files.
+  ///
+  /// If this is `true`, [pubspecOverridesPath] contains the path to the
+  /// overrides file.
+  bool get hasPubspecOverrides => pubspecOverridesPath != null;
+
+  /// The path to the entrypoint package's pubspec overrides file.
+  String? get pubspecOverridesPath => root.pubspec.overridesLocation?.path;
```
jonasfj, line 150: The path is the same regardless of whether or not it exists... Better to have a boolean saying if it exists, and always have the path be the same.

lib/src/pubspec.dart
```diff
@@ -635,3 +690,63 @@ class Pubspec extends PubspecBase {
...
+/// Merges [overrides] from an overrides file into a [pubspec] file.
+///
+/// The merging is performed at the stage where the pubspec and overrides files
+/// have both been parsed into [YamlMap]s. Source references of [YamlNode]s are
+/// preserved, so that error messages will point to the correct source location.
+///
+/// Only the the following top-level fields of a pubspec are merged:
+///
+/// - `dependencies`
+/// - `dev_dependencies`
```
jonasfj, line 703: What is the value in overriding `dependencies` and `dev_dependencies`, when will that work?

lib/src/utils.dart
```diff
@@ -499,6 +503,113 @@ String yamlToString(data) {
...
+///
+/// The map contains entries for all of the keys of [original]. If both
+/// [original] and [overrides] contain an entry for a given key, [mergeEntry]
+/// is called to merge the values. Otherwise, the entry from [original] is taken
+/// over as is.
+///
+/// Entries that only exist in [overrides] are taken over as is.
+///
+/// The [YamlNode]s of keys of merged entries always come from [original].
+class OverrideYamlMap with MapMixin, UnmodifiableMapMixin implements YamlMap {
```
jonasfj, line 520: Can't we just mutate the original `YamlMap` or create a new `YamlMap` with the new values?

lib/src/utils.dart
```diff
@@ -499,6 +503,113 @@ String yamlToString(data) {
...
+  OverrideYamlMap(this.original, this.overrides, this.mergeEntry);
+
+  /// The original map which is being overridden.
+  final YamlMap original;
+
+  /// The map containing the overrides.
+  final YamlMap overrides;
```
jonasfj, line 527: If we did this, why would these need to be public?

lib/src/utils.dart
sigurdm, line 520: I think it is better to just build the merged map explicitly...

lib/src/pubspec.dart
```diff
@@ -381,8 +407,37 @@ class Pubspec extends PubspecBase {
...
+        throw PubspecException(error.message, error.span);
+      }
+
+      YamlMap overridesMap;
+      if (overridesNode is YamlScalar && overridesNode.value == null) {
+        overridesMap = YamlMap(sourceUrl: overridesLocation);
+      } else if (overridesNode is YamlMap) {
+        overridesMap = overridesNode;
+      } else {
+        throw PubspecException('The pubspec_overrides must be a YAML mapping.',
```
jonasfj, line 430: What is `pubspec_overrides` there is `pubspec_overrides.yaml`

lib/src/pubspec.dart
```diff
@@ -635,3 +690,63 @@ class Pubspec extends PubspecBase {
...
+        // Only dependencies can be overridden, currently.
+        return pubspecField;
+      }
+
+      return _mergeDependencyOverrides(pubspecField, overridesField);
+    });
+
+/// Merges a dependency section form an overrides file into an existing section
+/// of a pubspec file.
+YamlNode _mergeDependencyOverrides(
```
sigurdm, line 725: I think we should not merge overrides.
If there is a `pubspec_overrides.yaml` we should only use those overrides, ignoring the ones from pubspec.yaml.

I think that leads to a much simpler mental model

lib/src/pubspec.dart
jonasfj: If you override `dependencies` to use a path dependency, then anything with a transitive dependency on the path-dependency will have a problem.. because the source of a dependency is part of its identity.

So if you override `dependencies` then you most likely end up using `dependency_overrides`.

lib/src/pubspec.dart
```diff
@@ -69,6 +70,10 @@ class Pubspec extends PubspecBase {
   // initialization can throw a [PubspecException], that error should also be
   // exposed through [allErrors].
 
+  /// The location of the overrides file, if this pubspec file is incorporating
+  /// overrides from an overrides file.
+  final Uri? overridesLocation;
```
sigurdm, line 75: I think the location of the overrides file should only belong in entrypoint.dart

lib/src/pubspec.dart
```diff
@@ -293,8 +298,18 @@ class Pubspec extends PubspecBase {
           pubspecPath);
     }
 
+    String? overridesContents;
+    final overridesPath = path.join(packageDir, 'pubspec_overrides.yaml');
```
jonasfj, line 302: Can we not make `Pubspec.load` take `pubspecOverridesFile` as an optional argument?

lib/src/entrypoint.dart
```diff
@@ -246,6 +256,11 @@ class Entrypoint {
     required PubAnalytics? analytics,
     bool onlyReportSuccessOrFailure = false,
   }) async {
+    if (!onlyReportSuccessOrFailure && hasPubspecOverrides) {
+      log.warning(
+          'Warning: pubspec.yaml has overrides from pubspec_overrides.yaml');
```
sigurdm, line 261: Perhaps use `${pubspec_overrides_path}` such that the relative path comes out right.

lib/src/pubspec.dart
sigurdm: Also if we merge them we would need for a way to remove overrides

lib/src/pubspec.dart
```diff
@@ -635,3 +690,63 @@ class Pubspec extends PubspecBase {
...
+    OverrideYamlMap(pubspec, overrides, (key, pubspecField, overridesField) {
+      const dependencyKeys = [
+        'dependencies',
+        'dev_dependencies',
+        'dependency_overrides'
+      ];
+
+      if (!dependencyKeys.contains(key)) {
+        // Only dependencies can be overridden, currently.
+        return pubspecField;
```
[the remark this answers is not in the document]
blaugold: I think we should actually even throw an exception.

lib/src/pubspec.dart
blaugold: You're right, we should only support `dependency_overrides` in `pubspec_overrides.yaml`.

lib/src/pubspec.dart
blaugold: You're right, we should only support `dependency_overrides` in `pubspec_overrides.yaml`.

lib/src/utils.dart
blaugold: `YamlMap.nodes` is unmodifiable and `YamlMap` has no public API (`YamlMap.internal` is internal) to create a new one from`YamlNode`s. We need to be able to build the new map from `YamlNodes` from two different `YamlMap`s if we want to keep the source references of the `YamlNode`s intact, for useful error messages.

lib/src/utils.dart
blaugold: No need, they should be private.

lib/src/entrypoint.dart
blaugold: I'm not sure that's useful since `pubspec_overrides.yaml` is always going to be right next to `pubspec.yaml`.

sigurdm: Yes, but `pubspec.yaml` is not always in the current working directory (if pub is invoked with `--directory`)

lib/src/pubspec.dart
```diff
@@ -293,8 +298,18 @@ class Pubspec extends PubspecBase {
           pubspecPath);
     }
 
+    String? overridesContents;
```
sigurdm, line 301: Actually information about the overrides probably doesn't belong in the pubspec at all.

Only a root node can have pubspec_overrides, so we should probably have it in the entrypoint along with pubspec and lockfile.

The version solver should then take an explicit map of overrides that can be given from Entrypoint.acquireDependencies

lib/src/utils.dart
blaugold: Please see https://github.com/dart-lang/pub/pull/3215#discussion_r779486656

lib/src/pubspec.dart
blaugold: I'm following the example a few lines above:
```The pubspec must be a YAML mapping.```

Should I change both?

File: lib/src/pubspec.dart (hunk 8 above)
Comment (blaugold) [reply lib/src/pubspec.dart]: Done

File: lib/src/pubspec.dart (hunk 9 above)
Comment (blaugold) [reply lib/src/pubspec.dart]: Done

File: lib/src/pubspec.dart (hunk 11 above)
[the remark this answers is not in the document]
Comment (blaugold) [reply lib/src/pubspec.dart]: Done

File: lib/src/entrypoint.dart (hunk 2 above)
Comment (blaugold) [reply lib/src/entrypoint.dart]: Done

File: lib/src/pubspec.dart (hunk 3 above)
Comment (blaugold) [reply lib/src/pubspec.dart]: Done

File: lib/src/entrypoint.dart (hunk 10 above)
Comment (blaugold) [reply lib/src/entrypoint.dart]: Done

File: lib/src/utils.dart (hunk 5 above)
Comment (blaugold) [reply lib/src/utils.dart]: Done

File: lib/src/pubspec.dart (hunk 7 above)
Comment (blaugold) [reply lib/src/pubspec.dart]: Misunderstood that comment first, but get it now. `dependency_overrides` from `pubspec_overrides.yaml` now fully replace existing `dependency_overrides` in `pubspec.yaml`.

File: lib/src/pubspec.dart (hunk 12 above)
Comment (blaugold) [reply lib/src/pubspec.dart:300]: I added a `withPubspecOverrides` option to `Package.load` and `Pubspec.load`, which is only set to `true` by `Entrypoint(String rootDir, this.cache)`.

All the code to parse dependency specifications is in `Pubspec`, which we would need to parse `pubspec_overrides.yaml`. Factoring that code out to parse `pubspec_overrides.yaml` separately seems a bit overkill to me, but we can do that, too.

Comment (blaugold) [reply lib/src/pubspec.dart]: I have reworked the PR to parse `pubspec_overrides.yaml` into a separate `PubspecOverrides` file, which gets rid of the hacky `OverrideYamlMap`.

Seems to me that `Package` is a more natural fit to store the loaded `PubspecOverrides`. Quite a few places use `Package.dependencyOverrides` and this where we can apply the `dependencyOverrides` from `PubspecOverrides`.
I had to update the `PackageLister` slightly, so that is uses `Package.dependencyOverrides`.

WDYT?

**Comment by blaugold** [comment]
@jonasfj @sigurdm 
I have reworked the PR, as suggested by @sigurdm's [comment](https://github.com/dart-lang/pub/pull/3215#discussion_r779507884).

`PubspecOverrides` is now its own class, into which `pubspec_overrides.yaml` is parsed.

It would be great if you could take another look.

**Comment by spydon** [comment]
Any update on how this PR is coming along?
(Sorry for the ping, just really excited to see this merged :sweat_smile:)

File: lib/src/solver/package_lister.dart [hunk 13]
```diff
@@ -52,6 +52,11 @@ class PackageLister {
   /// The set of package names that were overridden by the root package.
   final Set<String> _overriddenPackages;
```
sigurdm, line 53: Do we still need this? Won't it contain all the keys of `_overriddenPackageRanges`?

lib/src/package.dart
```diff
@@ -62,15 +62,22 @@ class Package {
   /// The parsed pubspec associated with this package.
   final Pubspec pubspec;
 
+  /// The parsed pubspec overrides associated with this package.
+  ///
+  /// If this package has been loaded without overrides or
+  /// `pubspec_overrides.yaml` does not exist, this will be `null`.
+  late final PubspecOverrides? pubspecOverrides;
```
sigurdm, line 69: I think this can be private

lib/src/pubspec.dart
```diff
@@ -635,3 +666,121 @@ class Pubspec extends PubspecBase {
...
+    Iterable<PackageRange>? dependencyOverrides,
+    Map? fields,
+    SourceRegistry? sources,
+  })  : _packageName = packageName,
+        _languageVersion = languageVersion,
+        _dependencyOverrides = dependencyOverrides == null
+            ? null
+            : Map.fromIterable(dependencyOverrides, key: (range) => range.name),
+        _fields = fields == null ? YamlMap() : YamlMap.wrap(fields),
+        _sources = sources;
```
sigurdm, line 744: Is this used anywhere?

lib/src/pubspec.dart
```diff
@@ -409,6 +420,25 @@ class Pubspec extends PubspecBase {
     _collectError(_ensureEnvironment);
     return errors;
   }
+}
+
+/// A mixin that contains the shared logic for parsing of [Pubspec] and
+/// [PubspecOverrides].
+mixin _PubspecParsingMixin {
```
sigurdm, line 427: Would it be possible to just use top-level functions for doing the dependency-map parsing? I feel that a mixin is overly complicating things here.

Something like:

```
Map<String, PackageRange> parseDependencyMap(String field, YamlNode? node, LanguageVersion laguageVersion, SourceRegistry sources) {...}

File: test/descriptor.dart [hunk 17]
```diff
@@ -107,6 +107,16 @@ Descriptor libPubspec(String name, String version,
   return pubspec(map);
 }
```
sigurdm, line 108: I think we need some integration tests of this, showing eg. that the solver picks up pubspec_overrides.yaml and that they take precedence over dependency_overrides. They probably belong in test/version_solver_test.dart.

sigurdm:
I think this starts looking really good.

lib/src/pubspec.dart
blaugold: `_parseDependencies` uses a few smaller utilitly methods. Some of those use the fileds of `_PubspecParsingMixin` (or before `Pubspec`). They would have to become top-level functions, too, and they would need a few more arguments.

`Pubspec._parseEnvironment` and `Pubspec._computeFeatures` use those utilitly methods aswell.

How about use composition by pulling the parsing logic out into a `_PubspecParser` to avoid the mixin?

lib/src/pubspec.dart
blaugold: I think I added it for testing but didn't end up using it, so no. I'll remove it.

test/descriptor.dart
blaugold: Done

lib/src/pubspec.dart
blaugold: Done

lib/src/package.dart
blaugold: Done

lib/src/solver/package_lister.dart
blaugold: Done

nik-v-hax:
@jonasfj  do it jonas

blaugold:
@jonasfj I fixed the uninitialized late variable that made CI fail. Locally, all test pass now.

sigurdm:
I did a refactor of PubspecOverrides into Pubspec. Now the handling of override-filess is transparent from the Package level (except for the constructor).

blaugold:
@sigurdm Thanks for refactoring that. Looks good.

I fixed that last test that failed on Windows.

lib/src/solver/package_lister.dart
```diff
@@ -49,8 +49,8 @@ class PackageLister {
   /// The type of the dependency from the root package onto [_ref].
   final DependencyType _dependencyType;
 
-  /// The set of package names that were overridden by the root package.
-  final Set<String> _overriddenPackages;
+  /// The set of packages that were overridden by the root package.
+  final Set<PackageRange> _overriddenPackages;
```
jonasfj, line 53: Wouldn't it make sense to keep this as a `Set<String>`?

jonasfj (approved):

test/pubspec_test.dart
```diff
@@ -817,5 +819,83 @@ features:
...
+
+        final foo = pubspec.dependencyOverrides['foo']!;
+        expect(foo.name, equals('foo'));
+        expect(foo.source, equals(sources['hosted']));
+        expect(foo.constraint, VersionConstraint.parse('1.0.0'));
+      });
+
+      test('throws exception with correct source references', () {
+        expectPubspecOverridesException('''
+dependency_overrides:
```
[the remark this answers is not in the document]
sigurdm: I made it fail.

lib/src/solver/package_lister.dart

blaugold:
Thanks @sigurdm and @jonasfj for the code reviews and getting the PR to where it can be merged!

iqfareez:
Sorry to interrupt, may I know when this feature is expected to land in stable. Thank you.

blaugold:
@iqfareez It's available in stable since Dart 2.16.0.

PR #3221, opened 2021-11-09
Refactor notices/hints for solver conflicts.

jonasfj:
Untangled a lot of type assertions and digging into details of the
incompatibility when printing the incompatibility. Notably, we introduce
a concept of notices and hints.

Notices are messages that an `IncompatibilityCause` can have, they are
human readable messages printed before the incompatibility explanation.
This allows `SdkCause` to point out the current version of an SDK that
is causing a conflict (if the SDK is available).

Hints are messages that an `IncompatibilityCause` can have, they are
human readable message with actionable hints to be printed after the
incompatibility explanation. This allows `PackageNotFoundCause` to hint
that trying without `--offline` might work.

Hints are printed after the incompatibility explanation because they
actionable, they don't serve to explain what went wrong. They are
intended to be actionable, hence, it's preferable the user sees these as
one of the first things.

------------------------

This is an initial step to being able to improve the errors that results from things like missing credentials, etc..

### DIFF
--- a/lib/src/solver/failure.dart
+++ b/lib/src/solver/failure.dart
@@ -7,7 +7,6 @@ import 'package:collection/collection.dart';
 import '../exceptions.dart';
 import '../log.dart' as log;
 import '../package_name.dart';
-import '../sdk.dart';
 import '../utils.dart';
 import 'incompatibility.dart';
 import 'incompatibility_cause.dart';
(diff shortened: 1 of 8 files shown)


lib/src/solver/failure.dart
```diff
@@ -93,39 +92,22 @@ class _Writer {
...
-        sdkConstraintCauses.add(cause.sdk);
-      }
-    }
-
     // If the failure was caused in part by unsatisfied SDK constraints,
     // indicate the actual versions so we don't have to list them (possibly
     // multiple times) in the main body of the error message.
     //
     // Iterate through [sdks] to ensure that SDKs versions are printed in a
     // consistent order
```
sigurdm, line 100: This comment needs to be updated

lib/src/solver/failure.dart
```diff
@@ -93,39 +92,22 @@ class _Writer {
...
-    var wroteLine = false;
-    for (var sdk in sdks.values) {
-      if (!sdkConstraintCauses.contains(sdk)) continue;
-      if (!sdk.isAvailable) continue;
-      wroteLine = true;
-      buffer.writeln('The current ${sdk.name} SDK version is ${sdk.version}.');
+    final notices = _root.externalIncompatibilities
+        .where((c) => c.cause.notice != null)
+        .map((c) => c.cause.notice)
+        .whereNotNull()
```
sigurdm, line 104: Am I misunderstanding, or you do double null-element elimination here?

```suggestion
        .map((c) => c.cause.notice)
        .whereNotNull()
```

lib/src/solver/failure.dart
```diff
@@ -159,15 +141,22 @@ class _Writer {
...
-      if (sdk.isAvailable) continue;
-      if (sdk.installMessage == null) continue;
+    // Iterate through all hints, these are intended to be actionable, such as:
+    //  * How to install an SDK, and,
+    //  * How to provide authentication.
+    // Hence, it makes sense to show these at the end of the explanation, as the
+    // user will ideally see these before reading the actual conflict and
+    // understand how to fix the issue.
+    _root.externalIncompatibilities
+        .where((c) => c.cause.hint != null)
```
sigurdm, line 151: Same here.

sigurdm (approved):
LGTM with a few comments

I had imagined some kind of "footnote" style binding between the incompatibility and the note/hint - it might be overkill, but something like:

```
Because myapp depends on foo any which doesn't exist (could not find
package foo in cache [1]), version solving failed.
              
[1] Try again without --offline!
```

This might not be needed for a single hint, but a complicated trace could have several different hints relating to different places in the failure trace.

lib/src/solver/failure.dart
jonasfj: good catch

jonasfj:
Closing in favor of https://github.com/dart-lang/pub/pull/3223 where this is merged in...

PR #3222, opened 2021-11-09
Add env var for writing golden files

jonasfj:


test/golden_file.dart
```diff
@@ -69,6 +69,12 @@ class GoldenTestContext {
   }
 
   void _readGoldenFile() {
+    if (RegExp(r'^1|(?:true)$', caseSensitive: false)
+        .hasMatch(Platform.environment['_PUB_TEST_WRITE_GOLDEN'] ?? '')) {
+      _goldenFileExists = false;
```
sigurdm, line 74: Perhaps rename:
```suggestion
      _shouldRegenerateGoldenFile = true;
```
It is a bit hard to follow the logic flow of the field goldenFileExists between methods that have to be called in the right order.

Perhaps raise this logic to a `late final` initializer.

```dart
  late final _shouldRegenerateGoldenFile = trueRegexp.hasMatch(Platform.environment['_PUB_TEST_WRITE_GOLDEN'] ?? '') || !_goldenFile.existsSync();

```

sigurdm (approved):
Nice LGTM

PR #3223, opened 2021-11-09
Better error messages round 2

jonasfj:
Based on https://github.com/dart-lang/pub/pull/3221, so maybe review that first, then skip the first commit.


jonasfj:
@sigurdm, there is a chance this will work (last commit is a bit of a necessary hack)... but we need to completely refactor `http.dart` if we really want to cleanup this mess...

Things like:
https://github.com/dart-lang/pub/blob/c888b018d588d73cced4220c7ac32c93d477fd27/lib/src/http.dart#L210-L216

Causes hacks like:

https://github.com/dart-lang/pub/blob/c888b018d588d73cced4220c7ac32c93d477fd27/lib/src/authentication/client.dart#L49-L60

I'm not sure how that got past review :see_no_evil: 
It doesn't matter, but it's clearly about time that we completely rip-out `http.dart` and refactor it.

----------------------

Maybe, we want to keep using `package:http` abstractions, but then we certainly shouldn't violate them by throwing `PubHttpException` and certainly not embed custom logic it in, like matching URLs and hosts :see_no_evil: 

I have mixed feelings around wrapping `Client` from `package:http` it seems we easily end up with something that doesn't behave like a `Client` should  :rofl:

Perhaps, we need a simple function that can do HTTP requests, retry them and where it's clearly documented what exceptions it can throw.

lib/src/authentication/client.dart
```diff
@@ -129,26 +129,29 @@ Future<T> withAuthenticatedClient<T>(
...
-      message = '$hostedUrl package repository requested authentication! '
+      hint = '$hostedUrl package repository requested authentication!\n'
           'You can provide credential using:\n'
           '    pub token add $hostedUrl';
     }
     if (error.statusCode == 403) {
-      message = 'Insufficient permissions to the resource in $hostedUrl '
-          'package repository. You can modify credential using:\n'
+      hint = 'Insufficient permissions to the resource in $hostedUrl '
+          'package repository.\nYou can modify credential using:\n'
```
sigurdm:
```suggestion
          'package repository.\nYou can modify credentials using:\n'
```

lib/src/authentication/client.dart
```diff
@@ -129,26 +129,29 @@ Future<T> withAuthenticatedClient<T>(
...
       }
-      message = '$hostedUrl package repository requested authentication! '
+      hint = '$hostedUrl package repository requested authentication!\n'
           'You can provide credential using:\n'
           '    pub token add $hostedUrl';
     }
     if (error.statusCode == 403) {
-      message = 'Insufficient permissions to the resource in $hostedUrl '
-          'package repository. You can modify credential using:\n'
+      hint = 'Insufficient permissions to the resource in $hostedUrl '
```
sigurdm:
```suggestion
      hint = 'Insufficient permissions to the resource at the $hostedUrl '
```

lib/src/authentication/client.dart
```diff
@@ -129,26 +129,29 @@ Future<T> withAuthenticatedClient<T>(
...
+    var message = 'authentication failed';
 
+    assert(error.statusCode == 401 || error.statusCode == 403);
     if (error.statusCode == 401) {
       if (systemCache.tokenStore.removeCredential(hostedUrl)) {
         log.warning('Invalid token for $hostedUrl deleted.');
       }
-      message = '$hostedUrl package repository requested authentication! '
+      hint = '$hostedUrl package repository requested authentication!\n'
           'You can provide credential using:\n'
```
sigurdm:
```suggestion
          'You can provide credentials using:\n'
```

See https://english.stackexchange.com/a/252075

sigurdm (approved):
LGTM with a few comments.

This is really nice!

lib/src/command.dart
```diff
@@ -212,15 +212,20 @@ and include the logs in an issue on https://github.com/dart-lang/pub/issues/new
...
       if (packageNotFound != null) exception = packageNotFound;
     }
     while (exception is WrappedException && exception.innerError is Exception) {
-      exception = exception.innerError;
+      exception = exception.innerError!;
     }
 
+    /* if (exception is PackageNotFoundException && exception.hint != null) {
+      // If there was a hint, then there is some actionable information for
+      // the user, and then it must be a problem with DATA (or CONFIG).
```
sigurdm, line 226: remove

Issue #3236, opened 2021-11-24
Error message gives incorrect instructions for tracing pub


Hixie:
When pub fails with a crash like in https://github.com/dart-lang/sdk/issues/47759, it says to run `dart pub --trace pub deps`, but that's not valid. It's even worse when run with `flutter`.

For example, set up a `pubspec.yaml` that crashes `pub deps` as in https://github.com/dart-lang/sdk/issues/47759, then run `flutter pub deps`. It tells you to run `dart pub --trace deps`. If you run that, it fails to trace and crashes telling you to run `dart pub --trace pub --trace deps`. If you run that, you get told `Cannot specify arguments before a command.`.

sigurdm:
Yeah this is broken. `--trace` doesn't work. :)

I suggest we instead dump all log-info into a `.dart-tool/pub/pub-log.txt` - file on errors (limited to a finite size) and mention that in the error message.

PR #3240, opened 2021-11-26
Write log trace to $PUB_CACHE/log/pub_log.txt when crashing

sigurdm:
And suggest using that in case of a unexpected error.

Also write it when --verbose is given.

The log trace also contains other useful [diagnostics](https://github.com/dart-lang/pub/pull/3240/files#diff-87eef0fe2290c93a3497987f30069e3faab3af194eb0aff72f2f43d1c94add5eR338).

Fixes: https://github.com/dart-lang/pub/issues/3236


test/testdata/goldens/embedding/embedding_test/unexpected exception writes the right message.txt
```diff
@@ -0,0 +1,371 @@
...
+[E] and attach the transcript in an issue on https://github.com/dart-lang/pub/issues/new
+
+-------------------------------- END OF OUTPUT ---------------------------------
+
+File containing information about the latest pub run.
+Before making this file public, make sure to remove any sensitive information!
+
+Pub version: 0.1.2+3
+FLUTTER_ROOT: <not set>
+Command: /usr/local/google/home/sigurdm/projects/dart-sdk/sdk/out/ReleaseX64/dart-sdk/bin/dart
```
jonasfj, line 30: I'm guessing this golden test isn't very portable? :D

lib/src/log.dart
```diff
@@ -313,25 +315,72 @@ void exception(exception, [StackTrace? trace]) {
...
+/// Prints the previously recorded log transcript to stderr.
+void dumpTranscriptToFile(String path, String command, Entrypoint? entrypoint) {
+  final buffer = StringBuffer();
+  buffer.writeln('File containing information about the latest pub run.');
+  buffer.writeln(
+      'Before making this file public, make sure to remove any sensitive information!');
+  buffer.writeln();
+  buffer.writeln('Pub version: ${sdk.version}');
+  buffer.writeln(
+      'FLUTTER_ROOT: ${Platform.environment['FLUTTER_ROOT'] ?? '<not set>'}');
```
jonasfj, line 346: Let's also dump `PUB_HOSTED_URL` that will save us a lot of pain...
And maybe `PUB_CACHE`?

lib/src/log.dart
```diff
@@ -313,25 +315,72 @@ void exception(exception, [StackTrace? trace]) {
...
+}
+
+/// Prints the previously recorded log transcript to stderr.
+void dumpTranscriptToFile(String path, String command, Entrypoint? entrypoint) {
+  final buffer = StringBuffer();
+  buffer.writeln('File containing information about the latest pub run.');
+  buffer.writeln(
+      'Before making this file public, make sure to remove any sensitive information!');
+  buffer.writeln();
+  buffer.writeln('Pub version: ${sdk.version}');
```
jonasfj, line 344: Let's also add a date-time here... we'll need it... it'll help users not upload logs that they generated months ago..

test/testdata/goldens/embedding/embedding_test/unexpected exception writes the right message.txt
sigurdm: OOps :)

lib/src/log.dart
sigurdm: Great ideas!

lib/src/log.dart
sigurdm: ++

szakarias (approved):

test/embedding/embedding_test.dart
```diff
@@ -175,5 +206,102 @@ main() {
...
+String _filter(String input) {
+  return input
+      .replaceAll(p.toUri(d.sandbox).toString(), r'file://$SANDBOX')
+      .replaceAll(d.sandbox, r'$SANDBOX')
+      .replaceAll(Platform.pathSeparator, '/')
+      .replaceAll(Platform.operatingSystem, r'$OS')
+      .replaceAll(globalServer.port.toString(), r'$PORT')
+      .replaceAll(
+        RegExp(r'^Created:(.*)$', multiLine: true),
+        r'Created: $TIME',
```
jonasfj, line 223: This is a lot of new hacks, just to a goldens test...
and that goldens test will break whenever we refactor code a bit, because it embeds a stacktrace...

Are we sure we want something this fragile?

sigurdm: I was hoping that having a test of our stacktrace is a good thing. Hopefully it's easy to update the golden file.

I added a TODO, that we might reconsider this.

sigurdm: Also I'm removing line-numbers from the stack-trace for making this a bit more stable over time :)

PR #3244, opened 2021-11-30
Support publishing to a hosted-url with a /path

jonasfj:
We had a slash to much when publishing to custom hosted pub servers with a `/path` in the _hosted-url_.

I'm pretty sure that if the URL doesn't have any path segments then it doesn't matter if we do:
 * `server.resolve('/api/packages/versions/new');` (wrong),
 * `server.resolve('api/packages/versions/new');`

So this change should only affect cases where `PUB_HOSTED_URL` or `publish_to` contains path-segments.

------------------------------

This is not the prettiest of tests, but that might be subject for cleanup another day.

### DIFF
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -91,7 +91,7 @@ class LishCommand extends PubCommand {
 
     try {
       await log.progress('Uploading', () async {
-        var newUri = server.resolve('/api/packages/versions/new');
+        var newUri = server.resolve('api/packages/versions/new');
         var response = await client.get(newUri, headers: pubApiHeaders);
         var parameters = parseJsonResponse(response);
 
--- a/test/lish/archives_and_uploads_a_package_test.dart
+++ b/test/lish/archives_and_uploads_a_package_test.dart
@@ -37,6 +37,36 @@ void main() {
     await pub.shouldExit(exit_codes.SUCCESS);
   });
 
+  test('publishes to hosted-url with path', () async {
+    await servePackages();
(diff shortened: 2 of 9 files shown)


lib/src/command/lish.dart
```diff
@@ -91,7 +91,7 @@ class LishCommand extends PubCommand {
 
     try {
       await log.progress('Uploading', () async {
-        var newUri = server.resolve('/api/packages/versions/new');
+        var newUri = server.resolve('api/packages/versions/new');
```
athomas, line 94: `Uri.parse('https://pub.dev/path').resolve('api/packages/versions/new')` -> `https://pub.dev/api/packages/versions/new`

Would that be expected to be `https://pub.dev/path/api/packages/versions/new`?

`Uri.parse('https://pub.dev/path/').resolve('api/packages/versions/new')` -> `https://pub.dev/path/api/packages/versions/new`

athomas (changes requested):

lib/src/command/lish.dart
jonasfj: We normalize URLs to end in slash if path is non-empty...

athomas (approved):

MrTam:
Hey Guys, can confirm this fix has done the trick for us. Thanks so much for looking @jonasfj!

jonasfj:
Merged in https://github.com/dart-lang/pub/commit/dcb6abac2d7d43258c03b348be42bf4aab9529b1

PR #3249, opened 2021-12-06
Fix repository specification: isRetracted -> retracted

jonasfj:
Fun story, our spec differs from the client implementation and our tests... This fixes the spec :rofl:


isoos (approved):

szakarias (approved):
Thanks

PR #3267, opened 2022-01-03
folder/* should not ignore the folder itself, only the contents

sigurdm:


jonasfj (approved):

lib/src/ignore.dart
```diff
@@ -448,7 +448,8 @@ _IgnoreParseResult _parseIgnorePattern(String pattern, bool ignoreCase) {
     if (nextChar == null) break;
     current++;
     if (nextChar == '*') {
-      if (peekChar() == '*') {
+      final peek = peekChar();
```
jonasfj, line 451: I think calling `peekChar` multiple times is fine...

sigurdm: ++

Issue #1178, opened 2015-06-05
Parallel instances of `pub get` interact poorly


DartBot:
_Originally opened as dart-lang/sdk#21408_

_This issue was originally filed by daniel.robinson...&#064;gmail.com_

---

**What steps will reproduce the problem?**
1. I manually clicked to `run pub get` on my project pubspec.yaml file whilst a `pub get` was already in process on that pubspec. I was then in the state where my download of unittest 0.11.0+5 package was corrupted as it didn't have its own pubspec file inside.
2. Then running `pub cache repair` failed also because unittests pubspec didnt exist

**What is the expected output? What do you see instead?**
I think if a `pub get` is in progress it should probably queue up further `pub get` calls so as to not corrupt package downloads (assuming that was the cause of the corruption) If you run `pub cache repair` and it comes across a corrupt package that doesn't have a pubspec.yaml file in it then it should delete the package completely and redownload it, rather than just getting stuck.

**What version of the product are you using?**

Dart Editor version 1.8.0.dev_00_00 (DEV)
Dart SDK version 1.8.0-dev.0.0

**On what operating system?**
windows 7 64bit

DartBot:
*This comment was originally written by daniel.robinson...&#064;gmail.com*
_____
bug filed from here http://stackoverflow.com/questions/26545278/could-not-find-pubspec-yaml-in-unittest-0-11-05

DartBot:
<a href="https://github.com/sethladd"><img src="https://avatars.githubusercontent.com/u/5479?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [sethladd](https://github.com/sethladd)**

----

*Added [Area-Pub](../labels/Area-Pub), [Triaged](../labels/Triaged) labels.*

DartBot:
<a href="https://github.com/nex3"><img src="https://avatars.githubusercontent.com/u/188?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [nex3](https://github.com/nex3)**

----

There are two separate issues here: one about multiple instances of pub stepping on each others' toes when they run in parallel and one about "pub cache repair" failing when a pubspec is missing. I'm going to retarget this issue to the former, and I've filed issue dart-lang/sdk#21418 to cover the latter.
______
 cc @munificent.
*Changed the title to: "Parallel instances of `pub get` interact poorly".*

raphaeladam-wf:
[...]

PR #3277, opened 2022-01-07
Avoid race condition for hosted source on package extraction

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/1178

The basic idea is that if two pub processes are in a race to install a package-version, we let the first one win, and just ignore the failure to install the second one.

### DIFF
--- a/lib/src/command/cache_repair.dart
+++ b/lib/src/command/cache_repair.dart
@@ -23,6 +23,8 @@ class CacheRepairCommand extends PubCommand {
 
   @override
   Future<void> runProtected() async {
+    // Delete any eventual temp-files left in the cache.
+    cache.deleteTempDir();
     // Repair every cached source.
     final repairResults = (await Future.wait(
             cache.sources.all.map(cache.source).map((source) async {
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -361,51 +361,52 @@ bool dirExists(String dir) => Directory(dir).existsSync();
 /// when we try to delete or move something while it's being scanned. To
 /// mitigate that, on Windows, this will retry the operation a few times if it
 /// fails.
-void _attempt(String description, void Function() operation) {
+///
+/// For some operations it makes sense to handle ERROR_DIR_NOT_EMPTY
+/// differently. They can passe [ignoreEmptyDir] = `true`.
+void _attempt(String description, void Function() operation,
+    {bool ignoreEmptyDir = false}) {
   if (!Platform.isWindows) {
     operation();
     return;
   }
 
   String? getErrorReason(FileSystemException error) {
+    // ERROR_ACCESS_DENIED
     if (error.osError?.errorCode == 5) {
       return 'access was denied';
     }
 
+    // ERROR_SHARING_VIOLATION
     if (error.osError?.errorCode == 32) {
       return 'it was in use by another process';
     }
 
-    if (error.osError?.errorCode == 145) {
+    // ERROR_DIR_NOT_EMPTY
+    if (!ignoreEmptyDir && error.osError?.errorCode == 145) {
       return 'of dart-lang/sdk#25353';
     }
 
     return null;
   }
 
-  for (var i = 0; i < 2; i++) {
+  for (var i = 0; i < 3; i++) {
     try {
       operation();
-      return;
     } on FileSystemException catch (error) {
       var reason = getErrorReason(error);
       if (reason == null) rethrow;
 
-      log.io('Pub failed to $description because $reason. '
-          'Retrying in 50ms.');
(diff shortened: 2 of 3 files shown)


lib/src/io.dart
```diff
@@ -451,14 +452,19 @@ void cleanDir(String dir) {
...
+
+bool isDirectoryNotEmptyException(FileSystemException e) {
+  final errorCode = e.osError?.errorCode;
+  return
+      // https://man7.org/linux/man-pages/man3/errno.3.html
+      (Platform.isLinux && errorCode == 39) ||
+          // https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-
+          (Platform.isWindows && errorCode == 145) ||
+          // Seems undocumented, found by observation.
+          (Platform.isMacOS && errorCode == 66);
```
jonasfj, line 467:
```suggestion
      // On Linux rename will fail with ENOTEMPTY if directory exists:
      // https://man7.org/linux/man-pages/man2/rename.2.html
      // #define	ENOTEMPTY	39	/* Directory not empty */
      // https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/asm-generic/errno.h#n20
      (Platform.isLinux && errorCode == 39) ||
          // On Windows this may fail with ERROR_DIR_NOT_EMPTY
          // https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-
          (Platform.isWindows && errorCode == 145) ||
          // On MacOS rename will fail with ENOTEMPTY if directory exists.
          // #define ENOTEMPTY       66              /* Directory not empty */
          // https://github.com/apple-oss-distributions/xnu/blob/bb611c8fecc755a0d8e56e2fa51513527c5b7a0e/bsd/sys/errno.h#L190
          (Platform.isMacOS && errorCode == 66);
```

Took a while to find these, but I like that our magic numbers are documented.

While Linux `man errno` says that the number may change depending on architecture:
https://man7.org/linux/man-pages/man0/errno.h.0p.html
it doesn't seem this is the case for `ENOTEMPTY`.

lib/src/io.dart
```diff
@@ -361,52 +361,53 @@ bool dirExists(String dir) => Directory(dir).existsSync();
...
     }
 
+    // ERROR_SHARING_VIOLATION
     if (error.osError?.errorCode == 32) {
       return 'it was in use by another process';
     }
 
-    if (error.osError?.errorCode == 145) {
+    // ERROR_DIR_NOT_EMPTY
+    if (!ignoreEmptyDir && error.osError?.errorCode == 145) {
```
jonasfj, line 386: Why not reuse `isDirectoryNotEmptyException` ?
```suggestion
    if (!ignoreEmptyDir && isDirectoryNotEmptyException(error)) {
```

lib/src/source/hosted.dart
```diff
@@ -812,14 +812,23 @@ class BoundHostedSource extends CachedSource {
...
+      try {
+        renameDir(tempDir, destPath);
+      } on io.FileSystemException catch (e) {
+        tryDeleteEntry(tempDir);
+        if (!isDirectoryNotEmptyException(e)) {
+          rethrow;
+        }
+        log.fine('''
+Destination directory $destPath already existed.
+Assuming a concurrent pub invocation installed it.''');
```
jonasfj, line 830: Shouldn't we just move `tryDeleteEntry(tempDir);` into a `finally` section?
```suggestion
        try {
          if (!isDirectoryNotEmptyException(e)) {
            rethrow;
          }
          log.fine('''
Destination directory $destPath already existed.
Assuming a concurrent pub invocation installed it.''');
        } finally {
          tryDeleteEntry(tempDir);
        }
```

jonasfj (approved):

lib/src/command/cache_repair.dart
```diff
@@ -23,6 +23,8 @@ class CacheRepairCommand extends PubCommand {
 
   @override
   Future<void> runProtected() async {
+    // Delete any eventual temp-files left in the cache.
+    cache.deleteTempDir();
```
jonasfj, line 27: Nice, now we do actually use it somewhere :D

lib/src/source/hosted.dart
sigurdm: Good idea

sigurdm: Or on second thought that would look a bit funny in the verbose log. (successful rename of a to b, followed by deleting a)
I'll keep as-is

sigurdm:
cc @jakemac53 - you might be interested.

jakemac53 (approved):

jonasfj (approved):

Issue #3165, opened 2021-10-01
Running `pub global run` concurrently


sigurdm:
If the snapshot is missing we do `deleteBinstubs` before recreating them, this opens a window for a race condition.

dcharkes:
Thanks!

cc @cskau-g 

The PR that started using this: https://github.com/flutter/flutter/pull/87231

gspencergoog:
I just tried a test: I ran a short-running command: `dart pub global run snippets --help`, 10000 times concurrently on a 72-core machine. It didn't have an incident, and completed all 10000 runs.

However, if I did this instead:

1) remove ~/.pub-cache
2) `dart pub global activate snippets 0.2.3`
3) invoked the same 10000 runs of `dart pub global run snippets --help` concurrently

Then it fails, all of the executions crash, probably because on the first run after install, 72 cores are busy trying to build the package executable, and something isn't quite locked right.

(I used the example app in https://github.com/google/process_runner/tree/master/example to do the concurrent runs, which makes a task queue that keeps 72 running at the same time)

gspencergoog:
Here's some of the stderr output:

```
Dart_LoadScriptFromKernel: The binary program does not contain 'main'.
Cannot delete file, path = '/usr/local/google/home/gspencer/.pub-cache/bin/snippets' (OS Error: No such file or directory, errno = 2)
Could not make "/usr/local/google/home/gspencer/.pub-cache/bin/snippets" executable (exit code 1):
chmod: cannot access '/usr/local/google/home/gspencer/.pub-cache/bin/snippets': No such file or directory
Can't load Kernel binary: Invalid kernel binary: Indicated size is invalid.
Dart_LoadScriptFromKernel: The binary program does not contain 'main'.
```

(I eliminated duplicate lines)

dcharkes:
That is exactly what we were seeing on the flutter analyze step on the HHH bot. Thanks @gspencergoog!

cskau-g:
> > `chmod: cannot access '/usr/local/google/home/gspencer/.pub-cache/bin/snippets': No such file or directory`
> 
> That is exactly what we were seeing on the flutter analyze step on the HHH bot. Thanks @gspencergoog!

In fact all of those errors you list, Greg, is what we've been seeing. Which is great because it ties it all together and points to the same underlying issue.


[...]

Issue #1111, opened 2015-06-05
Remove special handling of the old global lockfiles in pub


DartBot:
<a href="https://github.com/nex3"><img src="https://avatars.githubusercontent.com/u/188?v=3" align="left" width="96" height="96"hspace="10"></img></a> **Issue by [nex3](https://github.com/nex3)**
_Originally opened as dart-lang/sdk#20703_

---

Once Dart 1.6 is sufficiently old, we should remove pub's handling of the lockfiles it generated for &quot;pub global install&quot;.

PR #3285, opened 2022-01-13
Avoid raceconditions in `global activate`, `run` and `global run`

sigurdm:
Attempts to fix: https://github.com/dart-lang/pub/issues/3165

This is done by:

## `global activate`:
* Activating a global package into a temp-dir and renaming the dir into the right place.
  Two concurrent activations of the same package `foo` will not be able to both contribute to the resulting files, an observer will always see the result of one or the other or none in `$PUB_CACHE/global_packages/foo`.

* Creating binstubs in a temp-dir, making them executable and then moving them into the right place.
   It won't be possible to observe half a binstub, or for chmod to have the binstub removed before it can operate on it.

## `run`/`global run`:
* Copying the incremental compilation data before compilation, and renaming the result to the right location.
   We already did a copy to preserve the incremental data, the change is to do the copy **before** compilation not after.

All of this builds on renaming being an atomic operation, another process will either see the old file or the new, never anything in-between.

There is still a possibility of confusing results if you concurrently `dart pub global activate` two different packages with overlapping binstub-names, then you can end up with any possible mixture of these binstubs - don't do that.

The pr contains quite some refactoring to allow global packages to exist in a temporary directory.

Also I took out support for pre dart-1.6 style lockfiles.
This fixes: https://github.com/dart-lang/pub/issues/1111


sigurdm:
@jakemac53 you might be interested in the changes to the precompilation (in `dart.dart`).

jakemac53 (approved):
These changes LGTM

jonasfj (approved):

Issue #3286, opened 2022-01-13
Allow token authentication for insecure hosted repositories


dtretyakov:
Hi, in the following PR https://github.com/dart-lang/pub/pull/3007 was introduced support of hosted repositories, but as a result, `dart` client throws an exception when you're trying to add a token for repository with HTTP protocol:
```
url must be https://, insecure repositories cannot use authentication.
```

Reference: https://github.com/dart-lang/pub/blob/efd24e64cbfefac5133f59623ec3dad21f2b79a1/lib/src/command/token_add.dart#L47

We're working on [support of Dart packages hosting](https://youtrack.jetbrains.com/issue/SPACE-8398) as part of [our product](https://www.jetbrains.com/help/space/packages.html) and would like to run integration tests by using the current `dart` client against local service (HTTP) in CI environment. Currently, it's impossible to opt-out from that security verification check.

What we could do with that:
* Introduce the ability to disable security check as it is done in other tools, e.g. [`skopeo --tls-verify=false`](https://github.com/containers/skopeo/blob/main/docs/skopeo-login.1.md).
* Fork the client and make our own builds without HTTPS enforcement.

The first option is preferred for us, but what is the Dart team opinion about that?

ricardoboss:
@sigurdm I also ran into this issue and created a PR following your suggestion. Can you take a look?

PR #3288, opened 2022-01-17
Allow storing tokens for http repositories in tests

satamas:
Associated issue: https://github.com/dart-lang/pub/issues/3286


sigurdm:
Thanks for taking time to contribute.

We decided we don't want this - at least not in this form - see the comment in https://github.com/dart-lang/pub/issues/3286 .

PR #3289, opened 2022-01-17
Add LFS support

### LINKED ISSUE #1433: Support git-lfs files in the lib directory

westito:
Fixes #1433 - Support git-lfs files in the lib directory 

LFS files need to be fetched when mirror git repository

### DIFF
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -489,6 +489,7 @@ class BoundGitSource extends CachedSource {
     assert(!_updatedRepos.contains(path));
     try {
       await _clone(ref.description['url'], path, mirror: true);
+      await _fetchLfs(path);
     } catch (_) {
       await _deleteGitRepoIfInvalid(path);
       rethrow;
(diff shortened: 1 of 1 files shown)


sigurdm:
Thanks for taking time to contribute.

We already have a pr doing this: https://github.com/dart-lang/pub/pull/2782 . Closing this in favor of that.

Issue #3257, opened 2021-12-14
pub upgrade --major-versions <pkg> should only do as little upgrading as possible to other packages.


sigurdm:
Probably it is enough to change `UpgradeType.upgrade` to `UpgradeType.get` when we only have a single package.

Also we can improve the help message (and the documentation page is outdated too):

Something like:
```
Upgrade the current package's dependencies to latest versions.

Example:
* To upgrade all dependencies to latest version compatible with
  current constraints in `pubspec.yaml` run:
      dart pub upgrade
* To upgrade all dependency constraints in `pubspec.yaml` use:
      dart pub upgrade --major-versions

Usage: dart pub upgrade [dependencies...]
-h, --help               Print this usage information.
    --[no-]offline       Use cached packages instead of accessing the network.
-n, --dry-run            Report what dependencies would change but don't change
                         any.
    --[no-]precompile    Precompile executables in immediate dependencies.
    --null-safety        Upgrade constraints in pubspec.yaml to null-safety
                         versions
    --major-versions     Upgrades packages to their latest resolvable versions,
                         and updates pubspec.yaml.
-C, --directory=<dir>    Run this in the directory<dir>.

Run "dart help" to see global options.
See https://dart.dev/tools/pub/cmd/pub-upgrade for detailed documentation.
```

jonasfj:
In short, we should do a `get` after patching `pubspec.yaml` with the new constraint, that doing a `get` here:
https://github.com/dart-lang/pub/blob/c888b018d588d73cced4220c7ac32c93d477fd27/lib/src/command/upgrade.dart#L244

And:
https://github.com/dart-lang/pub/blob/c888b018d588d73cced4220c7ac32c93d477fd27/lib/src/command/upgrade.dart#L232

PR #3295, opened 2022-01-31
Upgrade other versions conservatively with --major-versions

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3257


jonasfj (approved):

lib/src/command/upgrade.dart
```diff
@@ -389,6 +391,28 @@ You may have to:
     _showOfflineWarning();
   }
 
+  Pubspec _applyChanges(
```
jonasfj, line 394: Please give this a documentation comment...

We could also consider tweak `_updatePubspec` to the YAML string instead writing the updated `pubspec.yaml`.

Then we could use `_updatePubspec` and parse YAML again, this has the benefit that:
 * (A) We don't have two methods doing the same thing in different ways.
 * (B) We abort earlier if we fail to parse YAML after modification (or if we fail to modify the YAML).

(B) can happen in certain cases, for example if using references / anchors in YAML in the path being modified.

jonasfj (approved):

lib/src/command/upgrade.dart
sigurdm: Good idea

Done

jonasfj (approved):

PR #3297, opened 2022-02-01
fix: relative to the current directory rules

2ZeroSix:
fixes handling of  statements that should be handled against directory of "ignore" file

quote from https://git-scm.com/docs/gitignore#_pattern_format
> - The slash / is used as the directory separator. Separators may occur at the beginning, middle or end of the .gitignore search pattern.
> 
> - If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

### DIFF
--- a/lib/src/ignore.dart
+++ b/lib/src/ignore.dart
@@ -285,8 +285,8 @@ class Ignore {
       }
       if (currentIsDir) {
         final ignore = ignoreForDir(normalizedCurrent);
-        ignoreStack
-            .add(ignore == null ? null : _IgnorePrefixPair(ignore, current));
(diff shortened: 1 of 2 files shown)


jonasfj:
CC, @sigurdm do you concur?

sigurdm (approved):
Thanks for the fix!

LGTM

2ZeroSix:
fixed format issues

2ZeroSix:
@jonasfj it seems to be ready for merge

jonasfj (approved):

PR #3304, opened 2022-02-04
dependency_services

sigurdm:
This will provide support for dependabot resolution.

See https://github.com/dependabot/dependabot-core/pull/4510 for usage.

This will not be exposed in the sdk.


bin/dependency_services.dart
```diff
@@ -0,0 +1,105 @@
...
+      'solver',
+      'all'
+    ], allowedHelp: {
+      'error': 'Show only errors.',
+      'warning': 'Show only errors and warnings.',
+      'normal': 'Show errors, warnings, and user messages.',
+      'io': 'Also show IO operations.',
+      'solver': 'Show steps during version resolution.',
+      'all': 'Show all output including internal tracing messages.'
+    });
```
jonasfj, line 67: Should we maybe just stick with `--verbose`, do we really need this?

bin/dependency_services.dart
```diff
@@ -0,0 +1,105 @@
...
+
+import 'package:args/args.dart';
+import 'package:args/command_runner.dart';
+import 'package:pub/src/command.dart';
+import 'package:pub/src/command/dependency_services.dart';
+import 'package:pub/src/exit_codes.dart' as exit_codes;
+import 'package:pub/src/io.dart';
+import 'package:pub/src/log.dart' as log;
+
+class DependencyServicesCommandRunner extends CommandRunner<int>
```
jonasfj, line 18: Does this libary need to export anything? (besides `main`)

```suggestion
class _DependencyServicesCommandRunner extends CommandRunner<int>
```

lib/src/io.dart
```diff
@@ -1080,3 +1080,13 @@ final String? dartConfigDir = () {
     return null;
   }
 }();
+
+/// Escape [x] for users to copy-paste in bash.
+///
+/// If x is alphanumeric we leave it as is.
+///
+/// Otherwise, wrap with single quotation, and use '\'' to insert single quote.
+String protectArgument(String x) =>
```
jonasfj, line 1089:
```suggestion
String escapeShellArgument(String x) =>
```

lib/src/solver/report.dart
```diff
@@ -288,7 +288,7 @@ class SolveReport {
           // Only show newer prereleases for versions where a prerelease is
           // already chosen.
           newId.version.isPreRelease && newerUnstable) {
-        message = '(${maxAll(versions)} available)';
+        message = '(${maxAll(versions, Comparable.compare)} available)';
```
jonasfj, line 291: Why do we need this?

lib/src/system_cache.dart
```diff
@@ -146,4 +148,44 @@ class SystemCache {
     log.fine('Clean up system cache temp directory $tempDir.');
     if (dirExists(tempDir)) deleteEntry(tempDir);
   }
+
+  /// Get the latest version of [package].
+  ///
+  /// Will include prereleases in the comparison if [allowPrereleases].
+  ///
+  /// If [package] is a [PackageId] with a prerelease version and there are no
+  /// later stable version we return a prerelease version if it exists.
```
jonasfj, line 157:
```suggestion
  /// Will consider _prereleases_ if:
  ///  * [allowPrereleases] is true, or,
  ///  * [package] is a [PackageId] with a prerelease version, and no later prerelease exists.
```

test/dependency_services/dependency_services_test.dart
```diff
@@ -0,0 +1,212 @@
...
+        'dependencies': {
+          'foo': '^1.0.0',
+        },
+      })
+    ]).create();
+    await pubGet();
+    await listReportApply(context, [
+      _PackageVersion('foo', Version.parse('2.2.3')),
+      _PackageVersion('transitive', null)
+    ]);
```
jonasfj, line 124: As much as I love golden tests, maybe we should add some assertions here?

There is a non-zero risk that we won't notice minor changes in the golden files.

If we could do a simple assertion like:
```dart
expect(report['dependencies'].any((depReport) => depReport['name'] == 'foo' && depReport['singleBreaking'].any(
  (d) => d['name'] == 'transitive' && d['version'] == null),
), isTrue);
```

Or something like that... just a few, we don't have to check every possible property :D

test/dependency_services/dependency_services_test.dart
```diff
@@ -0,0 +1,212 @@
...
+          'foo': '^1.0.0',
+          'bar': '^1.0.0',
+        },
+      })
+    ]).create();
+    await pubGet();
+    server.serve('foo', '1.2.4');
+    await listReportApply(context, [
+      _PackageVersion('foo', Version.parse('1.2.3')),
+      _PackageVersion('transitive', null)
```
jonasfj, line 146: Why this? copy/pasta :D

jonasfj: Maybe we should have some sanity check in `apply` that makes this fail?

test/testdata/goldens/dependency_services/dependency_services_test/multibreaking.txt
```diff
@@ -0,0 +1,156 @@
...
+        }
+      ]
+    }
+  ]
+}
+
+-------------------------------- END OF OUTPUT ---------------------------------
+
+## Section apply
+$ echo '{"dependencyChanges":[{"name":"foo","version":"3.0.1","constraint":"^3.0.0"},{"name":"bar","version":"2.0.0"}]}' | dependency_services apply
```
jonasfj, line 130: Shouldn't there also be a `constraint` for `bar`? Just asking... I'm not actually sure we require that... probably because dependabot loses this information...

test/testdata/goldens/directory_option_test/commands taking a --directory~-C parameter work.txt
```diff
@@ -17,7 +17,7 @@ Changed 1 dependency in myapp!
 -------------------------------- END OF OUTPUT ---------------------------------
 
 ## Section 2
-$ pub -C myapp/example get --directory=myapp bar
+$ pub -C 'myapp/example' get --directory=myapp bar
```
jonasfj, line 20: Should we do a tweak to escaping mechanism, so that we don't need to escape when there is a `/` or `.` in the string... those are generally safe in bash, right? (and quite common in our tests)

bin/dependency_services.dart
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Support for automated upgrades.
+///
+/// For now this is not a finalized interface. Don't rely on this.
+import 'dart:async';
```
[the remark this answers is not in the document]
sigurdm: Right! I somehow remembered that https://github.com/dart-lang/dartdoc/issues/1082 was fixed...

bin/dependency_services.dart
sigurdm: No - but being in bin/ I would also not expect anyone to import this...

I can make everything private if it makes us sleep better.

test/testdata/goldens/dependency_services/dependency_services_test/multibreaking.txt
sigurdm: I don't think so. This is a single-breaking change. Only foo should have its constraint updated.

Also depandabot would lose this information

bin/dependency_services.dart
sigurdm: I also removed --trace and --version.

test/testdata/goldens/dependency_services/dependency_services_test/multibreaking.txt
jonasfj: I think this is the multibreaking example.... but yes dependabot will probably loose this information.

sigurdm: Ah you're right.

jonasfj (approved):

PR #3305, opened 2022-02-04
Refactor descriptors

sigurdm:
* Use strongly typed package-descriptions internally.

* Manifests the difference between a resolved description and an unresolved. This is only relevant for git-dependencies.

* Removes the notion of a bound source. Now there are only sources - for methods that require a cache they now take the cache.

* Makes an explicit root-source.

* Removes the `PackageName` abstraction. Now you always have to be clear whether you are working with a PackageId or a PackageRef. They don't share code through inheritance.

 * When global packages installed from git are referenced with `p.prettyUrl` when formatted in the ui, (uses a relative path when possible, and full url otherwise). They are still stored in the lock-file with the absolute url.

* Removes the never-launched implementation of "features".


lib/src/command/list_package_dirs.dart
```diff
@@ -45,8 +45,7 @@ class ListPackageDirsCommand extends PubCommand {
     // Include the local paths to all locked packages.
     var packages = mapMap(entrypoint.lockFile.packages,
         value: (String name, PackageId package) {
-      var source = entrypoint.cache.source(package.source);
-      var packageDir = source.getDirectory(package);
+      var packageDir = package.source.getDirectory(package, cache);
```
sigurdm, line 48: We could have:
```
  cache.getDirectory(package)

File: lib/src/source.dart [hunk 2]
```diff
@@ -276,16 +139,41 @@ abstract class BoundSource {
...
+
+abstract class Description<T extends Description<T>> {
+  Source<T> get source;
+  Object? serializeForPubspec(
+      {required String? containingDir,
+      required LanguageVersion languageVersion});
+
+  /// Converts `this` into a human-friendly form to show the user.
+  ///
+  /// Paths are always rrelative to current dir.
```
sigurdm, line 159:
```suggestion
  /// Paths are always relative to current dir.
```

lib/src/source.dart
```diff
@@ -276,16 +139,41 @@ abstract class BoundSource {
...
+  String format();
+}
 
-  /// Stores [pubspec] so it's returned when [describe] is called with [id].
+abstract class ResolvedDescription<T extends Description<T>> {
```
jonasfj, line 163: Can we give this a documentation comment?

lib/src/source/hosted.dart
```diff
@@ -946,80 +923,140 @@ class BoundHostedSource extends CachedSource {
...
-/// This uses the system cache to get the list of available packages and does
-/// no network access.
-class _OfflineHostedSource extends BoundHostedSource {
-  _OfflineHostedSource(HostedSource source, SystemCache systemCache)
-      : super(source, systemCache);
+/// The [PackageName.description] for a [HostedSource], storing the package name
+/// and resolved URI of the package server.
+class HostedDescription extends Description<HostedDescription> {
+  final String packageName;
+  final String url;
```
jonasfj, line 930: This is the URL to the package repository, right?
```suggestion
  /// Url for package repository containing [packageName] described in the this [Description].
  final String hostedUrl;
```

That way, we don't think this is the URL to package archive or something else :D

lib/src/source/hosted.dart
```diff
@@ -946,80 +923,140 @@ class BoundHostedSource extends CachedSource {
...
   }
 
   @override
-  Future<PackageStatus> status(PackageId id, {Duration? maxAge}) async {
-    // Do we have a cached version response on disk?
-    final versionListing = await _cachedVersionListingResponse(id.toRef());
+  HostedSource get source => HostedSource.instance;
+}
+
+class ResolvedHostedDescription extends ResolvedDescription<HostedDescription> {
```
jonasfj, line 962: Just curious, but this looks an awful lot like `HostedDescription`, is there any reasons we need `ResolvedDescription` and `Description`?

I'm not suggesting we refactor more now. But if it doesn't make sense to have both of them, could we consider leaving a TODO for future refactoring?

-------------------

Hmm, I see that in `GitResolvedDescription` we have `resolved-ref`, so presumably we need both because this actually the place we could store the `content-hash` is that, right? (just curious).

lib/src/system_cache.dart
```diff
@@ -146,4 +141,72 @@ class SystemCache {
...
+
+  /// Loads the (possibly remote) pubspec for the package version identified by
+  /// [id].
+  ///
+  /// This may be called for packages that have not yet been downloaded during
+  /// the version resolution process. Its results are automatically memoized.
+  ///
+  /// Throws a [DataException] if the pubspec's version doesn't match [id]'s
+  /// version.
+  Future<Pubspec> describe(PackageId id) async {
```
jonasfj, line 156: It seems weird to me that these things live on the `SystemCache`.
In my mind it makes more sense that a `Source` instance would be passed `SystemCache` when created, and then registered with `SourceRegistry`. And then `SourceRegistry` is responsible for calling methods on `Source` and ideally nobody else would ever interact with concrete source implementations.

Everything goes through `SourceRegistry`, and the rest of the code can call methods here... and when it returns a `PackageRef` / `PackageRange` / `PackageId` then the `Description` / `ResolvedDescription` are private members on  `PackageRef` / `PackageRange` / `PackageId`. But `PackageRef` / `PackageRange` / `PackageId` allows for calling generic methods available on `Description` (like serialize for `pubspec.yaml` or `pubspec.lock`, etc).

But I'm guessing we might have leaked the abstractions too many places to clean this up now, right?

I'm just asking questions here, not even suggesting we write a TODO, hehe :D

jonasfj (approved):

lib/src/system_cache.dart
jonasfj: I'm not 100% sure we couldn't avoid leaking the abstractions here. But if it's even remotely possible it's probably best to leave it for a follow-up PR.

lib/src/command/add.dart
```diff
@@ -176,6 +174,7 @@ class AddCommand extends PubCommand {
       /// ensure that the modification timestamp on `pubspec.lock` and
       /// `.dart_tool/package_config.json` is newer than `pubspec.yaml`,
       /// ensuring that [entrypoint.assertUptoDate] will pass.
+
```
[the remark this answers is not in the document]
sigurdm: Removed

jonasfj (approved):

PR #3313, opened 2022-02-18
Upgrade `package:tar` to version `0.5.4`.

simolus3:
This upgrades the vendored `tar` package from https://github.com/simolus3/tar/commit/b5c5a11d8969f458ccdeb8cf01615f692fed3e97 to https://github.com/simolus3/tar/commit/ce8aae4825ac4cbed80829cd0d6e181b924b3db2.

Noteworthy changes in `package:tar`:

- Decode tar archives with malformed UTF-8 data in non-important PAX headers.
- Forward pause/resume requests when writing tar streams.
- Use a faster, tar-specifc implementation to split the input stream into suitably-sized blocks.
- Support writing tar entries synchronously.
- Faster checksum implementation.
- Avoid copying when reading old GNU sparse maps.

Overall, I expect that these changes will improve the performance of extracting or archiving packages. Quite a bit has changed between the two versions though, so maybe we should once again check published packages and make sure they behave identically? I'm happy to write a script for that if needed.

### DIFF
--- a/lib/src/third_party/tar/src/utils.dart
+++ b/lib/src/third_party/tar/src/utils.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert';
 import 'dart:math';
 import 'dart:typed_data';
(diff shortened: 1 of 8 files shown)


jonasfj:
> so maybe we should once again check published packages and make sure they behave identically? I'm happy to write a script for that if needed.

This is always worthwhile, untarring issues can be very painful :see_no_evil: 

Hmm, I actually think we already have a script for it, maybe just try to run:
https://github.com/dart-lang/pub/blob/master/tool/extract_all_pub_dev.dart

simolus3:
I've pushed a branch with both tar versions and changed the script to compare the results between the two: https://github.com/simolus3/pub/commit/a13a3043dcc38b026219e5a95b12bb11b0a5fa40#diff-f806e63054146be86bbe1bc4362e918d350bc033e0c38302b4d49c983000536e

Unfortunately I've only gotten to packages starting with B with my mediocre internet connection, but at least that worked without any difference. Can you checkout that version and do the full comparison run?

sigurdm:
> Can you checkout that version and do the full comparison run?

I'll do that! Thanks!

lib/src/third_party/tar/src/utils.dart
```diff
@@ -229,3 +236,325 @@ Stream<List<int>> zeroes(int length) async* {
     yield Uint8List(remainingBytes);
   }
 }
+
+/// An optimized reader reading 512-byte blocks from an [input] stream.
+class BlockReader {
+  final Stream<List<int>> input;
```
jonasfj, line 242: nit: does this need to be public?

jonasfj (approved):

PR #3330, opened 2022-03-04
Allow use of token for talking to pub.dev

jonasfj:
This refactors testing of publishing a bit... and allows use of `dart pub token add` when talking to `pub.dartlang.org`.

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -117,8 +117,15 @@ class HostedSource extends Source {
     // Clearly, a bit of investigation is necessary before we update this to
     // pub.dev, it might be attractive to do next time we change the server API.
     try {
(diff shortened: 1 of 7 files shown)


sigurdm (approved):
LGTM

lib/src/source/hosted.dart
```diff
@@ -117,8 +117,15 @@ class HostedSource extends Source {
     // Clearly, a bit of investigation is necessary before we update this to
     // pub.dev, it might be attractive to do next time we change the server API.
     try {
+      var defaultHostedUrl = 'https://pub.dartlang.org';
```
kevmoo, line 120: This seems the type of thing that should be referenced from a shared const 🤷

sigurdm: The `HostedSource.defaultUrl` while not const is the shared location for this information.

Issue #3328, opened 2022-03-04
Messages from validator/directory.dart should link to dart.dev documentation


jonasfj:
We should link to `https://dart.dev/tools/pub/package-layout` from:
https://github.com/dart-lang/pub/blob/b93bf88f3195003c3411a8047df240598de377ff/lib/src/validator/directory.dart#L35-L47

We could possibly do [dart.dev/go/package-layout](https://github.com/dart-lang/site-www/blob/main/firebase.json#L137-L156) link, so we don't have to have a long URL in the CLI output.

Issue #3327, opened 2022-03-04
Respect .pubignore in validators


jonasfj:
All (I think all) validators should ignore files that are ignored with `.pubignore` as they won't be part of the package:
https://github.com/dart-lang/pub/blob/b93bf88f3195003c3411a8047df240598de377ff/lib/src/validator/directory.dart#L28

See also @sigurdm comment in https://github.com/dart-lang/pub/issues/3326#issuecomment-1058980922

PR #3331, opened 2022-03-04
Don't analyze ignored directories in directory-validator

sigurdm:
Also write a reference to the package-layout documentation.

Fixes: https://github.com/dart-lang/pub/issues/3328
Fixes: https://github.com/dart-lang/pub/issues/3327

### DIFF
--- a/lib/src/validator/directory.dart
+++ b/lib/src/validator/directory.dart
@@ -30,7 +30,7 @@ class DirectoryValidator extends Validator {
     for (final file in entrypoint.root.listFiles()) {
       // Find the topmost directory name of [file].
       final dir = path.join(entrypoint.root.dir,
-          path.split(path.relative(file, from: entrypoint.root.dir))[0]);
+          path.split(path.relative(file, from: entrypoint.root.dir)).first);
       if (!visited.add(dir)) continue;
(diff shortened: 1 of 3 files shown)


lib/src/validator/directory.dart
```diff
@@ -22,30 +22,35 @@ class DirectoryValidator extends Validator {
...
-          warnings.add('Rename the top-level "$dir" directory to "example".\n'
-              'This allows Pub to find your examples and create "packages" '
-              'directories for them.\n');
-        }
+  Future<void> validate() async {
+    final visited = <String>{};
+    for (final file in entrypoint.root.listFiles()) {
+      // Find the topmost directory name of [file].
+      final dir = path.join(entrypoint.root.dir,
+          path.split(path.relative(file, from: entrypoint.root.dir))[0]);
```
jonasfj, line 33:
```suggestion
          path.split(path.relative(file, from: entrypoint.root.dir)).first);
```

jonasfj (approved):

lib/src/validator/directory.dart
sigurdm: Done

Issue #3275, opened 2021-12-28
Unable to update my flutter package


jnknewaj:
I am trying to update my published flutter package. But when I a run this command `flutter pub publish --dry-run` following error shows: 

```
Git error. Command: `git ls-files --cached --exclude-standard --recurse-submodules`
stdout: 
stderr: error: unknown option `recurse-submodules'
usage: git ls-files [<options>] [<file>...]

    -z                    paths are separated with NUL character
    -t                    identify the file status with tags
    -v                    use lowercase letters for 'assume unchanged' files
    -c, --cached          show cached files in the output (default)
    -d, --deleted         show deleted files in the output
    -m, --modified        show modified files in the output
    -o, --others          show other files in the output
    -i, --ignored         show ignored files in the output
    -s, --stage           show staged contents' object name in the output
    -k, --killed          show files on the filesystem that need to be removed
    --directory           show 'other' directories' names only
    --empty-directory     don't show empty directories
    -u, --unmerged        show unmerged files in the output
    --resolve-undo        show resolve-undo information
    -x, --exclude <pattern>
                          skip files matching pattern
    -X, --exclude-from <file>
                          exclude patterns are read from <file>
    --exclude-per-directory <file>
                          read additional per-directory exclude patterns in <file>
    --exclude-standard    add the standard git exclusions
    --full-name           make the output relative to the project top directory
    --error-unmatch       if any <file> is not in the index, treat this as an error
    --with-tree <tree-ish>
                          pretend that paths removed since <tree-ish> are still present
    --abbrev[=<n>]        use <n> digits to display SHA-1s
    --debug               show debugging data

exit code: 129
pub finished with exit code 69
```

jonasfj:
What version of the Dart SDK are you using, do you have a link to your repository? What version of `git` do you have installed?

jnknewaj:
> @jnknewaj, What version of the Dart SDK are you using, do you have a link to your repository? What version of `git` do you have installed?

Dart SDK version :  2.15.1
Git Version : 2.7.4
Repository link : https://github.com/jnknewaj/Super-UI

sigurdm:
I think '--recourse-submodules' was only introduced in Git 2.14 (https://git-scm.com/book/en/v2/Git-Tools-Submodules).

Upgrading git would most likely help.

Maybe pub should check the git-version and fail if it is not high enough?

modulovalue:
I can confirm that updating git resolves this issue.

[...]

PR #3332, opened 2022-03-07
Warn if git version is not high enough for supporting all features

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3275

### DIFF
--- a/lib/src/git.dart
+++ b/lib/src/git.dart
@@ -8,6 +8,7 @@ import 'dart:async';
 import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 
+import 'command_runner.dart';
 import 'exceptions.dart';
 import 'io.dart';
 import 'log.dart' as log;
@@ -145,8 +146,8 @@ bool _tryGitCommand(String command) {
       // We just warn here, as some features might work with older versions of
       // git.
       log.warning('''
-Pub needs at least git version 2.14 for all features to work.
-You seem to have ${output.substring('git '.length)}.
+You have a very old version of git (version ${output.substring('git version '.length)}),
+for $topLevelProgram it is recommended to use git version 2.14 or newer.
 ''');
     }
     return true;
--- a/lib/src/git.dart
+++ b/lib/src/git.dart
@@ -6,6 +6,7 @@
 import 'dart:async';
 
 import 'package:path/path.dart' as p;
+import 'package:pub_semver/pub_semver.dart';
 
 import 'exceptions.dart';
 import 'io.dart';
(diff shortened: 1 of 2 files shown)


lib/src/git.dart
```diff
@@ -121,13 +122,32 @@ String? repoRoot(String dir) {
...
+    if (match == null) return false;
+    // Git seems to use many parts in the version number. We just check the
+    // first two.
+    final major = int.parse(match[1]!);
+    final minor = int.parse(match[2]!);
+    if (Version(major, minor, 0) < _minSupportedGitVersion) {
+      // We just warn here, as some features might work with older versions of
+      // git.
+      log.warning('''
+Pub need at least git version 2.14 for all features to work.
```
szakarias, line 146: need -> needs

szakarias (approved):

lib/src/git.dart
sigurdm: Thanks!

lib/src/git.dart
```diff
@@ -121,13 +122,32 @@ String? repoRoot(String dir) {
   return null;
 }
 
+final _minSupportedGitVersion = Version(2, 14, 0);
```
jonasfj, line 125: This needs a documentation string that explains what features were adding in Git 2.14 that causes it to the be the minimum supported version.

jonasfj: We only used that in:
```sh
$ git grep recurse-submodules
lib/src/validator/gitignore.dart:        '--recurse-submodules'
```
So I'm guessing this is actually a problem we've caused, one could even call it a regression.

lib/src/git.dart
```diff
@@ -121,13 +122,34 @@ String? repoRoot(String dir) {
...
+    // Git seems to use many parts in the version number. We just check the
+    // first two.
+    final major = int.parse(match[1]!);
+    final minor = int.parse(match[2]!);
+    if (Version(major, minor, 0) < _minSupportedGitVersion) {
+      // We just warn here, as some features might work with older versions of
+      // git.
+      log.warning('''
+Pub needs at least git version 2.14 for all features to work.
+You seem to have ${output.substring('git '.length)}.
```
jonasfj, line 149: We have a variable for topLevelCommand... 
```suggestion
You have a very old version of git (version ${output.substring('git '.length)}),
for $topLevelCommand it is recommanded to use git version 2.14 or newer.
```

lib/src/git.dart
sigurdm: This was used in Package.listFiles before that. So the regression is not introduced at that point.

jonasfj (approved):

PR #3335, opened 2022-03-08
Remove `uploader` command

sigurdm:
We now have a web-interface for managing the uploaders. That should be the preferred way.

Leaving a (hidden) stub telling the user to use the web interface instead.

------------

See also breaking change request: https://github.com/dart-lang/sdk/issues/48526


lib/src/command/uploader.dart
```diff
@@ -17,7 +17,7 @@ class UploaderCommand extends PubCommand {
   String get name => 'uploader';
   @override
   String get description =>
-      'Manage uploaders for a package on pub.dartlang.org.';
+      'Manage uploaders for a package on pub.dartlang.org. (Deprecated)';
```
jonasfj, line 20: Instead of writing this in the description, I think we should make the command `hidden: true`:
https://pub.dev/documentation/args/latest/command_runner/Command/hidden.html

Then we should add it to the `CHANGELOG.md` in the SDK

lib/src/command/uploader.dart
```diff
@@ -74,6 +74,10 @@ the \$PUB_HOSTED_URL environment variable.''',
     }
 
     final package = argResults['package'] ?? entrypoint.root.name;
+    log.message(
```
jonasfj, line 77:
```suggestion
    log.warning(
```

jonasfj (approved):

lib/src/command/uploader.dart
sigurdm: Done

lib/src/command/uploader.dart
sigurdm: I now `fail()` instead

jonasfj (approved):

jonasfj (approved):

PR #3346, opened 2022-03-18
Only call Package.listFiles once per publish.

sigurdm:
Each call to package.listFiles comes with at least one call to git, it is better to minimize this.

Refactors the validators slightly.


lib/src/validator.dart
```diff
@@ -190,4 +202,27 @@ abstract class Validator {
       }
     });
   }
+
+  /// Returns the [files] that are inside [dir] (relative to the package
+  /// entrypoint).
+  List<String> filesBeneath(String dir, {required bool recursive}) {
```
szakarias, line 208: Maybe we should place this in a more central place?

szakarias (approved):

lib/src/validator.dart
sigurdm: Let's do that later.
I added a todo.

PR #3350, opened 2022-03-21
Fix 3342, avoid downloading packages in dependency-services

### LINKED ISSUE #3342: `dependency_services apply` should not download package archives.
And not update `.dart_tool/package_config.json`

It should only update `pubspec.yaml` and `pubspec.lock`.

jonasfj:
Fixes #3342, @sigurdm this isn't the prettiest solution. I'm open to other options..

But it's also possible we just clean up some other day :see_no_evil:

### DIFF
--- a/lib/src/command/dependency_services.dart
+++ b/lib/src/command/dependency_services.dart
@@ -368,11 +368,16 @@ class DependencyServicesApplyCommand extends PubCommand {
     if (lockFileEditor != null && lockFileEditor.edits.isNotEmpty) {
       writeTextFile(entrypoint.lockFilePath, lockFileEditor.toString());
     }
+    // TODO: Handle HTTP exceptions gracefully!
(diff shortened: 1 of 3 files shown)


lib/src/command/dependency_services.dart
```diff
@@ -368,11 +368,16 @@ class DependencyServicesApplyCommand extends PubCommand {
     if (lockFileEditor != null && lockFileEditor.edits.isNotEmpty) {
       writeTextFile(entrypoint.lockFilePath, lockFileEditor.toString());
     }
+    // TODO: Handle HTTP exceptions gracefully!
```
jonasfj, line 371: I just discovered that this was unhandled, we should probably check how HTTP errors are handled in `dependency-services`.

I suspect things could happen in `acquireDependencies`.

jonasfj:
Closed in favor of https://github.com/dart-lang/pub/pull/3352

PR #3352, opened 2022-03-22
dependency_services: Don't download archives on apply

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3342

### DIFF
--- a/lib/src/command/dependency_services.dart
+++ b/lib/src/command/dependency_services.dart
@@ -16,6 +16,7 @@ import '../command.dart';
 import '../entrypoint.dart';
 import '../exceptions.dart';
 import '../io.dart';
+import '../lock_file.dart';
 import '../log.dart' as log;
 import '../package.dart';
 import '../package_name.dart';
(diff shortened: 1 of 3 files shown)


lib/src/command/dependency_services.dart
```diff
@@ -362,17 +362,31 @@ class DependencyServicesApplyCommand extends PubCommand {
...
 
-    if (pubspecEditor.edits.isNotEmpty) {
-      writeTextFile(entrypoint.pubspecPath, pubspecEditor.toString());
-    }
-    if (lockFileEditor != null && lockFileEditor.edits.isNotEmpty) {
-      writeTextFile(entrypoint.lockFilePath, lockFileEditor.toString());
-    }
+    final updatedLockfile = lockFileEditor == null
+        ? null
+        : LockFile.parse(lockFileEditor.toString(), cache.sources);
```
jonasfj, line 367: Since JSON is a strict subset of YAML, you don't have to use `YamlEditor` for modifying the `pubspec.lock` anymore...

jonasfj (approved):

lib/src/command/dependency_services.dart
sigurdm: Yeah - it doesn't really hurt either. Leaving this for now

Issue #3353, opened 2022-03-22
dart pub token add https://pub.dev or https://pub.dartlang.org should be the same


jonasfj:
The client still uses `https://pub.dartlang.org` interally.

But:
 * `dart pub token add https://pub.dev`
 * `dart pub token add https://pub.dartlang.org`
Should result in the same..

Probably we should only store one token under `https://pub.dartlang.org` and in the future we can migrate it.

PR #3358, opened 2022-03-24
Use pub.dartlang.org instead of pub.dev

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3353

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -48,15 +48,6 @@ Uri validateAndNormalizeHostedUrl(String hostedUrl) {
   Uri u;
   try {
     u = Uri.parse(hostedUrl);
-
-    // pub.dev and pub.dartlang.org are identical.
-    //
-    // We rewrite here to avoid caching both, and to avoid having different
-    // credentials for these two.
-    if (u == Uri.parse('https://pub.dev')) {
-      log.message('Using https://pub.dartlang.org instead of https://pub.dev.');
-      u = Uri.parse('https://pub.dartlang.org');
-    }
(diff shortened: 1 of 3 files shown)


lib/src/source/hosted.dart
```diff
@@ -48,6 +48,15 @@ Uri validateAndNormalizeHostedUrl(String hostedUrl) {
...
     u = Uri.parse(hostedUrl);
+
+    // pub.dev and pub.dartlang.org are identical.
+    //
+    // We rewrite here to avoid caching both, and to avoid having different
+    // credentials for these two.
+    if (u == Uri.parse('https://pub.dev')) {
+      log.message('Using https://pub.dartlang.org instead of https://pub.dev.');
+      u = Uri.parse('https://pub.dartlang.org');
+    }
```
jonasfj, line 59: You might want to do this after we've done:
```dart
  u = u.normalizePath();
  // If we have a path of only `/`
  if (u.path == '/') {
    u = u.replace(path: '');
  }
  // If there is a path, and it doesn't end in a slash we normalize to slash
  if (u.path.isNotEmpty && !u.path.endsWith('/')) {
    u = u.replace(path: u.path + '/');
  }
```

So right before return.. Otherwise, we might have`https://pub.dev/` getting normalized to `https://pub.dev`, but `https://pub.dev` gets normalized to `https://pub.dartlang.org` :rofl:

jonasfj (changes requested):

lib/src/source/hosted.dart
sigurdm: Good catch, done!

Issue #3382, opened 2022-04-05
pub to support git dependency resolution


exvuma:
## Request

Pub is able to resolve dependencies that use a git hash to specify a dependency

## Context

Our motivation on the Dependabot team is to have `pub`  be as well (if not better 😄 ) supported as other ecosystems. For example, npm, yarn, gomod, etc.. can use git URLs to upgrade dependencies. Note in  [documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem) pub is the ecosystem that does not support git deps

Not having this on pub causes problems for multiple use cases: 

* when private git dependency is included in the pubspec, currently pub is not able to resolve the dependency tree and therefore cannot update any dependencies.
* unable to reach the private dependency. the user will need to configure a private git registry in order for this to be possible though.

Thanks for working with us on this project. Let me know if you need more context or how we can help 

CC/ Dependabot engineers @jurre @mctofu

jonasfj:
Current support for git-dependencies in `pubspec.yaml` is documented in:
https://dart.dev/tools/pub/dependencies#git-packages

In short, you write:
```yaml
dependencies:
  # This specifies dependency on package kittens from a git repository
  kittens:
    git:
      # Required, URL to the repository, this can also be https://...
      url: git@github.com:munificent/kittens.git
      
      # Optional, this can be a revision / branch / tag
      ref: main
      
      # Optional, path to package
      path: packages/kittens/

  # It is also possible to write a version constraint, but nobody does this.
  # And it doesn't really make sense, since we only ever try one commit
  # from the git-repository.
  #version: ^1.2.3
```

The hard part here is that `dart pub get/upgrade` will only ever try one revision from the git-repository. So version constraints and version numbers don't really matter for a git-dependency.

### When `ref` is a branch or omitted
When `ref` is a branch, or `ref`is not specified which is the same as specifying the default branch (`ref: main`).

Then `dart pub get` will use the revision from `pubspec.lock`, and if the lock file doesn't contain a revision, then it'll resolve the `ref` to a revision and use that.
Running `dart pub upgrade` will resolve `ref` and ignore the value in `pubspec.lock`.

This makes sense, and this kind of upgrade should be reasonably for `dependency_services` to support.

### When `ref` is a tag or commit-hash
Then resolving the `ref` always yields the same commit. So doing `dart pub get` or `dart pub upgrade` makes no difference.

[...]

PR #3383, opened 2022-04-07
dependency_services: use revisions as versions for git deps.

sigurdm:
This is part of solving: https://github.com/dart-lang/pub/issues/3382 (this will most likely require changes in dependabot).

Instead of the version of a git dependency we write the commit hash.
We special case applying these to the lockfile.

### DIFF
--- a/lib/src/command/dependency_services.dart
+++ b/lib/src/command/dependency_services.dart
@@ -23,6 +23,7 @@ import '../package_name.dart';
 import '../pubspec.dart';
 import '../pubspec_utils.dart';
 import '../solver.dart';
+import '../source/git.dart';
 import '../system_cache.dart';
 import '../utils.dart';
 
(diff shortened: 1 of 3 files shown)


lib/src/git.dart
```diff
@@ -62,6 +62,7 @@ Future<List<String>> run(List<String> args,
         workingDir: workingDir,
         environment: {...?environment, 'LANG': 'en_GB'});
     if (!result.success) {
+      print(result.stderr);
```
jonasfj, line 65: Who isn't guilty of printf debugging :rofl:

lib/src/command/dependency_services.dart
```diff
@@ -407,8 +470,27 @@ class DependencyServicesApplyCommand extends PubCommand {
...
+    return null;
+  }
+}
+
+String? tryParseHash(String v) {
+  if (RegExp(r'^[a-fA-F0-9]+$').hasMatch(v)) {
+    return v;
+  }
+  return null;
 }
```
jonasfj, line 494: Do these need to be exported? can't they just be private?

jonasfj (approved):

lib/src/command/dependency_services.dart
sigurdm: They might as well be private.

PR #3392, opened 2022-04-14
Fix crash when checking code with bad import

### LINKED ISSUE #1731: pub publish crashes when using invalid syntax for an import
The dependency validator causes pub to crash if there is a line like:

```dart
import 'package:$bad';
```

We end up trying to read the length of null somewhere in `_findPackages`.

I think the correct behavior should be to warn, but allow publishing.

natebosch:
Closes #1731

If an import attempts to use string interpolation it will have a null
`stringValue` in the analyzer. Fix the resulting crash trying to parse
the null string.

Refactor to `Uri.tryParse` instead of catching the `FormatException`.

### DIFF
--- a/lib/src/validator/strict_dependencies.dart
+++ b/lib/src/validator/strict_dependencies.dart
@@ -42,11 +42,9 @@ class StrictDependenciesValidator extends Validator {
 
       for (var directive in directives) {
         Uri? url;
-        try {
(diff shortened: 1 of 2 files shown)


jonasfj (approved):

jonasfj:
Nice catch!

PR #3419, opened 2022-05-09
Ignore `pubspec_overrides.yaml` for `publish` command

### LINKED ISSUE #3393: `pub publish` false positive when using `pubspec_overrides.yaml`
# Problem

When using the new `pubspec_overrides.yaml` to override dependencies during dev:

```yaml
# pubspec_overrides.yaml
dependency_override:
  my_package:
    path: path/to/my_package
```

then trying to publish the package, pub complains with:

```
Package validation found the following error:
* Your pubspec.yaml must not override non-dev dependencies.
  This ensures you test your package against the same versions of its dependencies
  that users will have when they use it.
Sorry, your package is missing a requirement and can't be published yet.
For more information, see: https://dart.dev/tools/pub/cmd/pub-lish.
```

## Expected behavior

My understanding is that pubspec_overrides.yaml applies only during development (and not to published packages). As such, I'd expect this to be valid and publish to work

blaugold:
Fixes #3393

### DIFF
--- a/lib/src/command.dart
+++ b/lib/src/command.dart
@@ -73,7 +73,12 @@ abstract class PubCommand extends Command<int> {
   ///
   /// This will load the pubspec and fail with an error if the current directory
   /// is not a package.
-  late final Entrypoint entrypoint = Entrypoint(directory, cache);
+  late final Entrypoint entrypoint =
+      Entrypoint(directory, cache, withPubspecOverrides: withPubspecOverrides);
+
+  /// Whether `pubspec_overrides.yaml` is taken into account, when creating
+  /// [entrypoint].
+  bool get withPubspecOverrides => true;
 
   /// The URL for web documentation for this command.
   String? get docUrl => null;
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -32,6 +32,8 @@ class LishCommand extends PubCommand {
   String get docUrl => 'https://dart.dev/tools/pub/cmd/pub-lish';
   @override
   bool get takesArguments => false;
+  @override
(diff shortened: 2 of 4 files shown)


sigurdm (approved):

sigurdm:
This has just been rolled into the SDK https://dart-review.googlesource.com/c/sdk/+/247962 - Sorry - we got a bit behind on that.

This should be in the next dev release of the dart SDK, and should also be in the next stable (2.18)

jakubjodelka:
Great, thanks for the information!

Yongle-Fu:
still get this issue, for why 😂
```
Validating package... (4.2s)
Package validation found the following hint:
* Non-dev dependencies are overridden in pubspec_overrides.yaml.
  
  This indicates you are not testing your package against the same versions of its
  dependencies that users will have when they use it.
  
  This might be necessary for packages with cyclic dependencies.
  
  Please be extra careful when publishing.

Package has 0 warnings and 1 hint.. Do you want to publish liboxyz 1.24.0-dev.3 to https://dart-pub.brainco.cn (y/N)? y
Uploading... 
Successfully uploaded package.
➜  liboxyz git:(clear) melos --version
6.1.0
➜  liboxyz git:(clear) dart --version
Dart SDK version: 3.3.4 (stable) (Tue Apr 16 19:56:12 2024 +0000) on "macos_x64"
```

sigurdm:
> still get this issue, for why 😂

This is now a warning instead of an error, and thus still allow publishing. I think it is fair to get that warning no matter how the package is overridden.

That changed in: https://github.com/dart-lang/pub/pull/3565 where we started doing a full `pub get` before publishing.

PR #3429, opened 2022-05-17
Verify signatures from hosted dependencies

simolus3:
I'm opening this to start a discussion and to play around with it. This addresses the verification part of https://github.com/dart-lang/pub-dev/issues/5750. The actual implementation for signing may have to change a lot, but this is in a state where it can be experimented with. It would be interesting to know whether this approach is acceptable in general.

For commands downloading packages, adds an `--verify` option to verify signatures. The allowed options are:

1. `ignore`: Never attempt to verify signatures.
2. `softIfPresent`: If a package has a signature, attempt to verify it. Print a warning message if that fails. I've picked this as a default, but `ignore` may be more reasonable.
3. `soft`: Like `softIfPresent`, but also print a warning message for packages without a signature.
4. `strict`: Like `soft`, but fail and abort instead of printing a warning message. This effectively enforces all packages to have a verified signature.

I've pushed a signed package to my pub server for testing:

```yaml
dependencies:
  picocss:
    # signed with https://simonbinder.eu/email.asc
    hosted: https://simonbinder.eu
    version: ^1.5.0
```

If the general approach seems reasonable, there are remaining todos with the implementation:

- Obviously this needs extensive testing.
- This currently uses GPGME as a library to verify packages. That library is LGPL licensed. But since this is an optional feature, I think the library wouldn't have to be part of the Dart SDK. Pub could just load the library if the user happens to have installed. It might be safer to find a project with a more permissive license to to just use the `gpg` binary in a subprocess for verification.


simolus3:
Closing this as the overall story around signing will need way more planning. This PR helped start a discussion about signing so it has served its purpose :D

Issue #3434, opened 2022-05-23
`dart pub publish --dry-run` prints truncated output


jamesderlin:
When I run `dart pub publish --dry-run` in my package, I see:

```
Publishing dartbag 0.3.0 to https://pub.dartlang.org:
|-- CHANGELOG.md
|-- LICENSE
|-- README.md
|-- analysis_options.yaml
|-- lib
|   |-- analysis_options.yaml
|   |-- byte_data.dart
|   |-- debug.dart
|   | (5 more...)
|   |-- src
|   |   |-- mem_equals32.dart
|   |   '-- mem_equals64.dart
|   |-- timer.dart
|   '-- tty.dart
|-- pubspec.yaml
...
```

Some files are omitted from the listing and are replaced with "(5 more...)".  While that's helpful for readability, it defeats the point of me using `--dry-run` to see and verify exactly what will be uploaded.  The "(5 more...)" substitution occurs even when running `dart pub publish --dry-run --verbose`.  The full contents can be determined from the verbose messages detailing the generated `.tar.gz` file, albeit without the pretty-printed directory tree.

1. `--verbose` should print the full directory tree.
2. It would be nice if there were an option to print the full directory tree without also having to wade through the additional spew from `--verbose`.

sigurdm:
Yeah - I don't think it makes a lot of sense to leave out files.

I think we should always print all the files.

PR #3440, opened 2022-05-31
List all files in pub publish

sigurdm:
Even though there can be many files, this is really the only sanity check we have before publishing.

Fixes https://github.com/dart-lang/pub/issues/3434


jonasfj:
I don't mind the potentially very long list..
But then I think we should change the message, so that important things like:
`Publishing <package> <version> to <server>`

isn't at the first line, which the user not read...

jonasfj (approved):

PR #3446, opened 2022-06-02
Use non-nullable parameters for operator ==

natebosch:
The null value will never be passed to the equals operator
implementation. Widening the type causes any other class which
implements this interface to also widen the type.

### DIFF
--- a/lib/src/third_party/tar/src/format.dart
+++ b/lib/src/third_party/tar/src/format.dart
@@ -28,7 +28,7 @@ class TarFormat {
   int get hashCode => _value;
 
   @override
-  bool operator ==(Object? other) {
+  bool operator ==(Object other) {
     if (other is! TarFormat) return false;
 
     return _value == other._value;
--- a/lib/src/third_party/tar/src/sparse.dart
+++ b/lib/src/third_party/tar/src/sparse.dart
@@ -21,7 +21,7 @@ class SparseEntry {
   String toString() => 'offset: $offset, length $length';
 
   @override
-  bool operator ==(Object? other) {
+  bool operator ==(Object other) {
     if (other is! SparseEntry) return false;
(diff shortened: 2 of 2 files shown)


sigurdm:
This is vendored code - the fix should probably land in[ package:tar](https://pub.dev/packages/tar) first.

cc @simolus3

simolus3:
> Widening the type causes any other class which implements this interface to also widen the type.

To be fair, `TarFormat` is `@sealed` and `SparseEntry` is not exported.

Thanks for the pointers though, I'll fix this in `packag:tar` and open another PR for a roll into pub.

sigurdm:
> Thanks for the pointers though, I'll fix this in packag:tar and open another PR for a roll into pub.

If you want you can be dog-fooding jonas's new vendor-package: https://pub.dev/packages/vendor for importing the tar package!

simolus3:
I've opened #3447 which includes these changes.

> If you want you can be dog-fooding jonas's new vendor-package: https://pub.dev/packages/vendor for importing the tar package!

I've tried it, but it keeps the `lib` folder in the path whereas we just put contents of `lib` directly into `third_party/tar` (https://github.com/google/dart-neats/issues/183). That second `lib/` in the path triggers some lints, so it's probably easier to keep the directory structure as it is now.

natebosch:
> To be fair, `TarFormat` is `@sealed` and `SparseEntry` is not exported.

Yeah, the impact is pretty limited. I wanted to clean up what I found though, because I'd like to push for this to be added as a lint.

https://github.com/dart-lang/linter/issues/3441

natebosch:
> and open another PR for a roll into pub.

Thanks!

Superseded by https://github.com/dart-lang/pub/pull/3447

PR #3447, opened 2022-06-03
Roll tar to 0.5.5+1

simolus3:
The only non-trivial change is that the reader now handles pause events happening after the last chunk of a tar entry but before the stream is fully closed.
I don't believe these changes to have an impact on pub, but if you want me to I can once again setup a comparison script ensuring all packages extract without changes.

This also contains the changes from #3446.


sigurdm (approved):
LGTM

Issue #3334, opened 2022-03-07
pub fails to rename directory


timsneath:
# Environment

- `pub version` or `flutter pub version`: Dart SDK version: 2.17.0-179.0.dev (dev) (Sat Mar 5 02:12:06 2022 -0800) on "windows_x64"
- OS version: Windows 10.0.19042.1566
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

On a Google Windows workstation, `pub` intermittently fails to download dependencies:

```
[C:\src\win32] dart pub get
Resolving dependencies...
+ _fe_analyzer_shared 36.0.0
+ analyzer 3.3.1
...
+ winmd 1.0.26
+ yaml 3.1.0
Downloading winmd 1.0.26...
Downloading pub_semver 2.1.1...
Pub failed to rename directory because access was denied.
This may be caused by a virus scanner or having a file
in the directory open in another application.
```

I suspect this is because of a race condition with virus scanning or other malware protection. 

I see other examples of this from customers outside Google. 

I suspect it may be worth adding a wait/retry to this step to give a virus scanner time to execute and unlock the file; alternatively, there may be a non-blocking way to do this (e.g. call `WaitForSingleObject` to ensure that the file handle isn't locked before renaming).

mit-mit:
https://github.com/dart-lang/pub/issues/1798 looks related

mit-mit:
cc @jonasfj, @timsneath might have a repro of this, how do we proceed?

sigurdm:
> I suspect it may be worth adding a wait/retry to this step to give a virus scanner time to execute and unlock the file;

As a matter of fact when you see that error that means the rename has been retried with delays:
https://github.com/dart-lang/pub/blob/master/lib/src/io.dart#L367

> e.g. call WaitForSingleObject to ensure that the file handle isn't locked before renaming).

That might be worth looking into.

The pub team has very little windows experience, and for multiple reasons we would far prefer staying within dart:io. That said, if we could solve these issues we should at least have a go at it.

jonasfj:
We don't know that this is anti-virus, it could be anything that's messing with our files.
That said, my understanding is that these are generally temporary folders that we create, add files to and then are unable to atomically move.
So if it's not anti-virus (or a virus, hehe) that is messing with the system, then it's hard to know what is going on. But maybe there is a way to lock temporary directories in Windows, or create them in ways that prevents others from using them until we're done.

mit-mit:
[...]

PR #3451, opened 2022-06-07
Extend retries for file-ops on windows

sigurdm:
Attempt to mitigate/fix https://github.com/dart-lang/pub/issues/3334.

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -390,7 +390,7 @@ void _attempt(String description, void Function() operation,
     return null;
   }
 
-  const maxRetries = 10;
+  const maxRetries = 50;
   for (var i = 0; i < maxRetries; i++) {
     try {
       operation();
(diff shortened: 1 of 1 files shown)


jonasfj (approved):

lib/src/io.dart
```diff
@@ -390,18 +390,20 @@ void _attempt(String description, void Function() operation,
...
     } on FileSystemException catch (error) {
       var reason = getErrorReason(error);
       if (reason == null) rethrow;
 
-      if (i < 2) {
+      if (i < maxRetries - 1) {
+        final delayMilliseconds = 50 * (i + 1);
         log.io('Pub failed to $description because $reason. '
-            'Retrying in 50ms.');
-        sleep(Duration(milliseconds: 50));
```
jonasfj, line 404: this is low cost, we can just do it every 50ms and give it more retries...

jonasfj (approved):

sigurdm:
@timsneath do you have a chance to try this - and see if it indeed helps with: https://github.com/dart-lang/pub/issues/3334 ?

lib/src/io.dart
sigurdm: Done

sigurdm:
The failures are from changes in the sdk used for testing and unrelated to this PR.

@alexmarkov could the changes in stack-trace be related to https://github.com/dart-lang/sdk/commit/af4da780be986e9aad22ab18395f1548ac1d1937 ? 

https://github.com/dart-lang/pub/runs/6813277082?check_suite_focus=true

PR #3460, opened 2022-06-10
Fix stack-trace expectation

sigurdm:
Dart 2.18 will have better stack-traces.

Hardcoded filter of the stacktrace to be compatible with both worlds.

### DIFF
--- a/test/embedding/embedding_test.dart
+++ b/test/embedding/embedding_test.dart
@@ -231,7 +231,7 @@ main() {
 }
 
 String _filter(String input) {
-  return input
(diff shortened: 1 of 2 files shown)


jonasfj (approved):

jonasfj (approved):
We could also simplify stack trace in golden tests to just say STACKTRACE

Because we should not care how those look, should we?

sigurdm:
> Because we should not care how those look, should we?

Yeah - I guess we shouldn't... but somehow I like that we know what we spew at people.

Issue #3455, opened 2022-06-09
A change in PUB_CACHE does not result in dependencies being considered outdated by `entrypoint.assertUpToDate()`


sigurdm:
Consider this scenario for a checkout of the pub client:

```
> dart pub get
Resolving dependencies... (1.0s)
Got dependencies!
> PUB_CACHE=$HOME/.pub_cache2 dart pub
Could not find a file named "pubspec.yaml" in "/usr/local/google/home/sigurdm/.pub_cache2/hosted/pub.dartlang.org/_fe_analyzer_shared-40.0.0".
#0      new Pubspec.load (package:pub/src/pubspec.dart:307:7)
#1      new Package.load (package:pub/src/package.dart:128:29)
#2      SystemCache.load (package:pub/src/system_cache.dart:105:20)
#3      Entrypoint._createPackageGraph (package:pub/src/entrypoint.dart:119:63)
#4      Entrypoint.packageGraph (package:pub/src/entrypoint.dart:114:54)
#5      getExecutableForCommand (package:pub/src/executable.dart:338:19)
#6      RunCommand.run (package:dartdev/src/commands/run.dart:250:32)
#7      CommandRunner.runCommand (package:args/command_runner.dart:209:27)
#8      DartdevRunner.runCommand (package:dartdev/dartdev.dart:231:30)
#9      CommandRunner.run.<anonymous closure> (package:args/command_runner.dart:119:25)
#10     new Future.sync (dart:async/future.dart:301:31)
#11     CommandRunner.run (package:args/command_runner.dart:119:14)
#12     runDartdev (package:dartdev/dartdev.dart:66:29)
#13     main (file:///b/s/w/ir/cache/builder/sdk/pkg/dartdev/bin/dartdev.dart:11:9)
#14     _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:293:32)
#15     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:192:12)
```

jonasfj:
Why does this happen?

sigurdm:
Another view is that the resolution still is up-to-date (with respect to the old PUB_CACHE) - we should not resolve the executable location with respect to the PUB_CACHE, rather with respect to package_config.json.

sigurdm:
I'm not able to reproduce this with dart 3.9.0.

Probably something changed somewhere.

PR #3461, opened 2022-06-10
Create packageGraph from package_config.json

sigurdm:
This implicates that `dart run` will rely on an existing package_config.json even if it was build with another PUB_CACHE set.

This fixes (at least part of the problem) in https://github.com/dart-lang/pub/issues/3455


sigurdm:
The deleted test was maintaining support with pub caches from before dart 1.1.2 I think it is fair to drop that support now.

https://codereview.chromium.org//1292003005

sigurdm:
This was superseded by: https://github.com/dart-lang/pub/pull/3533

Issue #2890, opened 2021-02-23
dart pub get --pristine/--locked


jonasfj:
It would be nice to have a mode where `dart pub get` fails if `pubspec.lock` is missing or changes to `pubspec.lock` is required to satisfy `pubspec.yaml`.

Could be called `dart pub get --locked` or `dart pub get --pristine`.

**Motivation**: When doing a production build (or deployment) you explicitly want to avoid any deviation from `pubspec.lock` as this could result in shipping dependencies you have not tested.

adarsh-technocrat:
Hello, @jonasfj 👋 I want to go ahead and work on this issue would you please give me the initial steps to get started with :)

jonasfj:
I would suggest figuring out how to run tests locally, and then experiment, it probably just means:
 * Adding an option to [`dart pub get`](https://github.com/dart-lang/pub/blob/master/lib/src/command/get.dart)
 * Passing a new `pristine` argument to [`acquireDependencies`](https://github.com/dart-lang/pub/blob/master/lib/src/entrypoint.dart#L219)
 * Figuring out all the adverse effects this could have, elsewhere in the code, if any :D
 * Writing a lot of tests to show that `--pristine` works, both positive and negative tests.

adarsh-technocrat:
> @adarsh-technocrat, I would suggest figuring out how to run tests locally, and then experiment, it probably just means:
>  * Adding an option to [`dart pub get`](https://github.com/dart-lang/pub/blob/master/lib/src/command/get.dart)
>  * Passing a new `pristine` argument to [`acquireDependencies`](https://github.com/dart-lang/pub/blob/master/lib/src/entrypoint.dart#L219)
>  * Figuring out all the adverse effects this could have, elsewhere in the code, if any :D
>  * Writing a lot of tests to show that `--pristine` works, both positive and negative tests.

Thanks, @jonasfj for guiding I'll look into it..:p

yjbanov:
I don't think adding an extra option to `pub get` will work in practice. People type `pub get` (and `flutter packages get`) by hand on the command-line, and they won't use this option. I think a better approach is to add this option to `pubspec.yaml` and have `pub get` pick it up from there. Then we don't have to rely on humans remembering to use this option.

jonasfj:
> I don't think adding an extra option to `pub get` will work in practice. People type `pub get` (and `flutter packages get`) by hand on the command-line, and they won't use this option.

Just to be clear:
[...]

Issue #2905, opened 2021-03-08
`locked` option in `pubspec.yaml`


yjbanov:
The default behavior of `pub` is non-hermetic. Running `pub get` from the same codebase frequently leads to different sets of dependencies, causing flaky builds. `pubspec.lock` exists and can be checked into the repo, but `pub get` silently overwrites it, and therefore it does not solve the hermeticity issue. Furthermore, the silent overwriting is annoying. You end up with unintended local changes from unrelated actions, such as opening a project in VSCode (which runs `pub get`/`flutter packages get` behind-the-scenes).

A better mode would be to have `pub get` do the following two things:

1. Validate that `pubspec.lock` is in sync with `pubspec.yaml`, and fail with a non-zero exit code if it's not (missing `pubspec.lock` is considered out-of-sync). Nothing is written to disk as part of this step.
2. Fetch dependencies exactly as specified in `pubspec.lock`. The only file written to is `.packages`.

To generate or update `pubspec.lock` a user would run `pub get --overwrite` (or `-w` for short).

To allow package maintainers opt into this mode a new top-level `pubspec.yaml` property can be introduced, e.g.:

```yaml
locked: true
```

/cc @kevmoo

kevmoo:
@yjbanov – moved this to the pub repo

yjbanov:
> This really shouldn't happen. Unless you've changed the contents of `pubspec.yaml` since `pubspec.lock` was written (or upgraded the Flutter SDK, which often forces new versions of some packages).

Unfortunately it does happen. To reproduce:

- `git clone https://github.com/felixblaschke/simple_animations.git`
- `cd simple_animations`
- `flutter packages get`

I think what you mean is that it shouldn't happen as long as `pubspec.yaml` and `pubspec.lock` are in sync with each other. That may be true, but the problem is that it's too easy for the two to end up out of sync. The problem is less technical, and more social. I think the https://github.com/dart-lang/pub/issues/2890 proposal solves the problem technically, but not socially, because people will simply not type `--pristine` or `--locked`.

jonasfj:
> Unfortunately it does happen. To reproduce:
> * `git clone https://github.com/felixblaschke/simple_animations.git`
> * `cd simple_animations`
> * `flutter packages get`

Great example :D 


> I think what you mean is that it shouldn't happen as long as `pubspec.yaml` and `pubspec.lock` are in sync with each other. That may be true, but the problem is that it's too easy for the two to end up out of sync.

[...]

Issue #2462, opened 2020-04-26
Per-package sha256 checksum field in pubspec.lock


paulyoung:
I would like to be able to verify the integrity of individually downloaded packages by comparing sha256 checksums stored in `pubspec.lock`.

My motivation for doing so is to improve the experience of using [`pub2nix`](https://github.com/paulyoung/pub2nix), my solution for building pub packages with nix.

I’m currently working around this by parsing the lock file and generating nix expressions containing the hash ahead of time. This is necessary because nix doesn’t allow network access at runtime unless a hash of the expected result is provided.

If the entry for each package in `pubspec.lock` also contained a `sha256` field then I could eliminate the manual code generation step altogether and have a seamless experience.

In the meantime I plan to update `pub2nix` to behave as if this field was present, and update the codegen step to modify the lock file and add the hashes.

sigurdm:
This seems like a good idea.
I cannot give a timeline currently.

Is sha256 a strict requirement for nix? Or does it just have to be some hash?

paulyoung:
One feature also supports sha1 and sha512 but sha256 is used extensively throughout nix tooling.

I’m not aware of any other hashing algorithms that are supported out of the box.

FWIW I’m close to being done on refactoring pub2nix in the way I mentioned and patching the lock file is working well in the meantime.

jonasfj:
@paulyoung, I'm curious, how do you install the dependencies? `pub get` can't use tarballs. Of course you can extract the tarballs a head of running `pub get` and do a similar for `git` dependencies. But afaik `pub` directory naming conventions inside `PUB_CACHE` are not obvious (especially for git and http-only sources).

If you only want to support `pub.dev` you could make a proxy on localhost that downloads, hashes and cache the response requested from `pub.dev`. Just set the `PUB_HOSTED_URL=http://localhost:3000` (or something like that), and make the proxy handle patterns from: https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md

paulyoung:
What I did in https://github.com/paulyoung/pub2nix/blob/6726c11766a0883ac7c103594930db5831bce6a2/install.nix has been working for me so far.

Like I said I’ve been doing some refactoring and haven’t quite gotten to reviewing what happens there again. It’s been a while since I first wrote it.

Based on #2359 I’ve switched to producing `.dart_tool/package_config.json` instead of `.packages` so I anticipate there may be some changes relating to that.

paulyoung:
[...]

PR #3482, opened 2022-07-05
Content hashing of archives

sigurdm:
Fix of https://github.com/dart-lang/pub/issues/2462

* Compute a hash of each downloaded archive and store it in: `$PUB_CACHE/hosted/<hosted-url>/.hashes/<package>-<version>.sha256` (details here still subject to change)

* New optional field in the package listing api for the server to provide the content-hash. If that is provided - it is verified against the downloaded archive.

* When writing a pubspec.lock file, the sha256 is included in the description of each hosted package.

* On `pub get` If the description of a package from pubspec.lock doesn't match the one in the cache, the archive is redownloaded - if the hash still doesn't match, the resolution fails with an error.

* **Has been moved to a follow-up PR** Introduce a new option `dart pub get --enforce-lockfile` A mode that will NOT modify pubspec.lock. That means:
  - won't add hashes if missing,
  - will refuse to resolve if pubspec.yaml isn't satisfied,
  - will refuse to resolve if hashes don't match cached hashes.
  - will refuse to resolve if pubspec.lock is missing
  - will verify that the extracted package content matches the contents of the original archive.
 
  This is useful when deploying to production.
  Fixes: https://github.com/dart-lang/pub/issues/2890 and https://github.com/dart-lang/pub/issues/2905

An unfortunate side-effect of this change is that all already downloaded packages will be re-downloaded (because we don't store the archives, only the extracted files) to compute their hashes.


sigurdm:
The basic functionality is here.

We could use some work on the error-messages. Maybe other details need to be fixed.

sigurdm:
From the test failures and a bit of debugging it seems that `gzip.encode` from dart:io is not stable across platforms.
```dart
import 'dart:io';

main() {
  print(gzip.encode([
    112,
    117,
    98,
    115,
    112,
    101,
    99,
    46,
    121,
    97,
    109,
    108,
    0,
    0,
  ]));
}

```
Prints:
```
[31, 139, 8, 0, 0, 0, 0, 0, 0, 3, 43, 40, 77, 42, 46, 72, 77, 214, 171, 76, 204, 205, 97, 96, 0, 0, 244, 120, 3, 30, 14, 0, 0, 0] # on my linux workstation
[31, 139, 8, 0, 0, 0, 0, 0, 0, 19, 43, 40, 77, 42, 46, 72, 77, 214, 171, 76, 204, 205, 97, 96, 0, 0, 244, 120, 3, 30, 14, 0, 0, 0] # on my mac
```
This makes testing harder. We'll have to consider how best to test this. Maybe we can find a gzip-encoder written in dart to use for testing...

sigurdm:
Reading https://www.rfc-editor.org/rfc/rfc1952.html it seems there is a OS field in gzip. We might be able to just filter that away. Also there is an mtime stamp, that luckily it seems dart:io doesn't fill in.

jonasfj:
> On pub get If the description of a package from pubspec.lock doesn't match the one in the cache, the archive is redownloaded - if the hash still doesn't match, the resolution fails with an error.

I haven't verified if the code actually does this, but as recall we wanted it to update `pubspec.lock` (and print a warning) if the server version listing returns a different sha256.

jonasfj (commented):
Just some initial comments, I'm not all done. Let's walk through it when you get back.

doc/repository-spec-v2.md
```diff
@@ -256,6 +258,15 @@ parameters. This allows for the server to return signed-URLs for S3, GCS or
 other blob storage service. If temporary URLs are returned it is wise to not set
 expiration to less than 25 minutes (to allow for retries and clock drift).
 
+The `archive_sha256` should be the hex-encoded sha256 checksum of the file are
```
sigurdm, line 261:
```suggestion
The `archive_sha256` should be the hex-encoded sha256 checksum of the file
```

lib/src/command/get.dart
```diff
@@ -52,12 +56,17 @@ class GetCommand extends PubCommand {
       log.warning(log.yellow(
           'The --packages-dir flag is no longer used and does nothing.'));
     }
+
+    if (argResults['dry-run'] && argResults['enforce-lockfile']) {
+      fail('Cannot do a dry-run with enforce-lockfile.');
+    }
```
jonasfj, line 62: `--dry-run` with `--enforce-lockfile` is fine. In this case it should do a resolution and check that nothing in the lock file changed.
We won't need to download archives, but we could opt to check that content-hashes match what is one the server.

lib/src/command/get.dart
```diff
@@ -28,6 +29,9 @@ class GetCommand extends PubCommand {
         negatable: false,
         help: "Report what dependencies would change but don't change any.");
 
+    argParser.addFlag('enforce-lockfile',
+        negatable: false, help: 'Only use resolution from existing pubspec.lock.');
```
jonasfj, line 33:
```suggestion
        negatable: false, help: 'Fail resolution if pubspec.lock does not satisfy pubspec.yaml.');
```

or 

```suggestion
        negatable: false, help: 'Enforce pubspec.lock, and fail resolution if it does not satisfy pubspec.yaml.');
```

test/package_server.dart
```diff
@@ -220,13 +233,30 @@ class PackageServer {
...
+        return replaceOs(Stream.fromIterable(entries)
             .transform(tarWriterWith(format: OutputFormat.gnuLongName))
-            .transform(gzip.encoder);
+            .transform(gzip.encoder));
       },
     );
   }
 
+  /// Replaces the 9th entry in [stream] with a 0. This replaces the os entry
+  /// of a gzip stream, giving us the same stream on all platforms.
```
sigurdm, line 244: Point to spec.

test/package_server.dart
```diff
@@ -220,13 +233,30 @@ class PackageServer {
...
+  Stream<List<int>> replaceOs(Stream<List<int>> stream) async* {
```
jonasfj, line 245: https://www.rfc-editor.org/rfc/rfc1952.html

And maybe explain a little... perhaps make this simpler by converting everyting to a single Uint8List and then modify it... and return a `Stream.single()`

jonasfj: Maybe make it private.

test/package_server.dart
```diff
@@ -243,6 +273,16 @@ class PackageServer {
...
+
+  /// Useful for testing handling of a wrong hash.
+  void setSha256(String name, String version, String sha256) {
+    _packages[name]!.versions[version]!.sha256 = sha256;
+  }
+
+  Future<String> getSha256(String name, String version) async {
+    final v = _packages[name]!.versions[version]!;
+    return v.sha256 ?? hexEncode((await sha256.bind(v.contents()).first).bytes);
+  }
```
jonasfj, line 285: Unused, let's delete it

test/package_server.dart
```diff
@@ -243,6 +273,16 @@ class PackageServer {
   void retractPackageVersion(String name, String version) {
     _packages[name]!.versions[version]!.isRetracted = true;
   }
+
+  /// Useful for testing handling of a wrong hash.
+  void setSha256(String name, String version, String sha256) {
+    _packages[name]!.versions[version]!.sha256 = sha256;
+  }
```
jonasfj, line 280:
```suggestion
  void overrideArchiveSha256(String name, String version, String sha256) {
    _packages[name]!.versions[version]!.sha256 = sha256;
  }
```

lib/src/source/hosted.dart
```diff
@@ -689,6 +762,31 @@ class HostedSource extends CachedSource {
...
+  /// stored hash of all downloaded packages.
+  String hashPath(PackageId id, SystemCache cache) {
+    final description = id.description.description;
+    if (description is! HostedDescription) {
+      throw ArgumentError('Wrong source');
+    }
+    final rootDir = cache.rootDirForSource(this);
+
+    var dir = _urlToDirectory(description.url);
+    return p.join(rootDir, dir, '.hashes', '${id.name}-${id.version}.sha256');
```
jonasfj, line 778: Maybe, do `$PUB_CACHE/hosted-hashes/<server>/<package>-<version>.sha256`

test/content_hash_test.dart
```diff
@@ -0,0 +1,193 @@
...
+
+  test('archive_sha256 is checked on download', () async {
+    final server = await servePackages();
+    server.serve('foo', '1.0.0');
+    server.setSha256('foo', '1.0.0',
+        'e7a7a0f6d9873e4c40cf68cc3cc9ca5b6c8cef6a2220241bdada4b9cb0083279');
+    await appDir({'foo': 'any'}).create();
+    await pubGet(
+      error:
+          contains('Downloaded archive for foo-1.0.0 had wrong content-hash.'),
```
jonasfj, line 62: TODO: Tweak message

test/content_hash_test.dart
```diff
@@ -0,0 +1,193 @@
...
+    await pubGet();
+    server.serve('foo', '1.0.0',
+        contents: [file('new_file.txt', 'This file could be malicious.')]);
+    // Pub get will not revisit the file-listing if everything resolves, and only compare with a cached value.
+    await pubGet();
+    // Deleting the version-listing cache will cause it to be refetched, and the error will happen.
+    File(p.join(globalServer.cachingPath, '.cache', 'foo-versions.json'))
+        .deleteSync();
+
+    await pubGet(
```
jonasfj, line 81: Maybe, do `pub upgrade` instead

test/content_hash_test.dart
```diff
@@ -0,0 +1,193 @@
...
+      error: allOf(
+        contains('Cached version of foo-1.0.0 has wrong hash - redownloading.'),
+        contains(
+            'Cache entry for foo-1.0.0 does not have content-hash matching lockfile.'),
```
jonasfj, line 85: TODO: Tweak messages

test/get/enforce_lockfile_test.dart
```diff
@@ -0,0 +1,109 @@
...
+      packageConfigEntry(name: 'foo', version: '1.0.0'),
+    ]).validate();
+  });
+
+  test('Refuses to get if no lockfile exists', () async {
+    await appDir({}).create();
+    await pubGet(
+        args: ['--enforce-lockfile'],
+        error:
+            'Retrieving dependencies failed. Cannot do `--enforce-lockfile` without an existing `pubspec.lock`.');
```
jonasfj, line 38: TODO: Tweak message

test/get/enforce_lockfile_test.dart
```diff
@@ -0,0 +1,109 @@
...
+    await appDir({}).create();
+    await pubGet();
+    await appDir({'foo': 'any'}).create();
+
+    await pubGet(
+        args: ['--enforce-lockfile'],
+        error: 'Dependency foo is not already locked in `pubspec.lock`.');
+  });
+
+  test('Refuses to get if package is locked to version not matching constraint',
```
jonasfj, line 53: Add edge case where `1.0.0` was removed from the server, but version `1.1.0` is still available, and `1.0.0` is in pubspec.lock

test/get/enforce_lockfile_test.dart
```diff
@@ -0,0 +1,109 @@
...
+    final server = await servePackages();
+    server.serve('foo', '1.0.0');
+    server.serve('foo', '2.0.0');
+    await appDir({'foo': '^1.0.0'}).create();
+    await pubGet();
+    await appDir({'foo': '^2.0.0'}).create();
+    await pubGet(
+        args: ['--enforce-lockfile'],
+        error:
+            'Dependency foo is locked to foo 1.0.0 in `pubspec.lock` but resolves to foo 2.0.0.');
```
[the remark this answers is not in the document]
jonasfj: We'll need to consider the edge case where `1.0.0` was removed from the server, but version `1.1.0` is still available, and `1.0.0` is in pubspec.lock

lib/src/system_cache.dart
```diff
@@ -208,10 +214,17 @@ class SystemCache {
     return id.source.doGetDirectory(id, this, relativeFrom: relativeFrom);
   }
 
-  Future<void> downloadPackage(PackageId id) async {
+  Future<void> downloadPackage(
```
jonasfj, line 217: Can we have a documentation comment for what `allowOutdatedHashChecks` means?

lib/src/solver/result.dart
```diff
@@ -105,6 +94,35 @@ class SolveResult {
...
+      final newDescription = package.description;
+
+      if (previousDescription is ResolvedHostedDescription) {
+        final newHash = (newDescription as ResolvedHostedDescription).sha256;
+
+        if (!bytesEquals(previousDescription.sha256, newHash)) {
+          throw ApplicationException(
+              'Dependency ${package.name} is locked to a different content-hash than what was resolved.');
+        }
+      }
```
jonasfj, line 122: Enforce lock file should not allow packages to be removed.
We should also have a test case for this.

lib/src/solver/result.dart
```diff
@@ -105,6 +94,35 @@ class SolveResult {
...
 
+  /// Checks that the SolveResult is compatible with [_previousLockfile]
+  ///
+  /// Throws if pubspec.yaml isn't satisfied.
+  Future<void> enforceLockfile() async {
+    for (final package in packages) {
+      if (package.isRoot) continue;
+      final previousPackage = _previousLockFile.packages[package.name];
+      if (previousPackage == null) {
+        throw ApplicationException(
```
jonasfj, line 105: Consider using the summary as error message.
We can make the summary as a list of changes only and omit the `(x.y.z available)` notices (might need to tweak code for this).

Then we can make the error message something like:
```
Resolving...
+ tar 0.5.5+1
- foo 1.2.3
> bar 1.2.3 (was 1.0.0)
Resolution of pubspec.yaml does not satisfy pubspec.lock, which is required by --enforce-lockfile.
```

lib/src/lock_file.dart
```diff
@@ -308,25 +314,80 @@ Generated by pub on ${DateTime.now()}.''';
...
+        if (cachedHash == null) {
+          // This can happen if we resolve from a server not providing hashes,
+          // but we are not downloading the archives. (eg. for a
+          // dependency_services run).
+          log.fine(
+              'No hash of downloaded archive for ${id.name}-${id.version} found');
+        } else if (description.sha256 == null) {
+          // We have resolved from a server without archive_sha256 in the
+          // version listing. Use the sha256 from the archive instead.
+          description = description.withSha256(cachedHash);
```
jonasfj, line 334: What does this do? Don't we need to do something with this?

lib/src/lock_file.dart
```diff
@@ -308,25 +314,80 @@ Generated by pub on ${DateTime.now()}.''';
...
+        } else {
+          // TODO(Should we really not fail here???).
```
jonasfj, line 336: Let's make a test case or two that demonstrates this.

I'm thinking that this is possibly a reason to:
 * Change the lockfile
 * Re-download the file

Or maybe it's an internal error, or maybe I missed something obvious.

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1275,17 @@ class _RefAndCache {
...
 }
+
+/// A sink that can only have `add` called once, and that can retrieve the
+/// value.
+class _SingleValueSink<T> implements Sink<T> {
+  late final T value;
+
+  @override
+  void add(T data) {
+    value = data;
```
jonasfj, line 1286: throw `StateError` if this is called more than once.

test/package_server.dart
sigurdm: I'm now using it.

test/package_server.dart
sigurdm: Done

lib/src/solver/result.dart
```diff
@@ -67,27 +93,56 @@ class SolveResult {
...
+              if (originalHash != null) {
+                if (!bytesEquals(cachedHash, originalHash)) {
+                  log.warning('''
+Content of ${id.name}-${id.version} has changed compared to what was locked your pubspec.lock.
+
+This might indicate that:
+* The content has changed on the server since you created the pubspec.lock.
+* The pubspec.lock has been corrupted.
+
+Your pubspec.lock has been updated according to what is on your server.
```
sigurdm, line 123:
```suggestion
Your pubspec.lock has been updated according to what is on the server.
```

lib/src/solver/result.dart
```diff
@@ -67,27 +93,56 @@ class SolveResult {
...
+          }
+          final originalEntry = _previousLockFile.packages[id.name];
+          if (originalEntry != null && originalEntry.version == id.version) {
+            final originalDescription = originalEntry.description;
+            if (originalDescription is ResolvedHostedDescription) {
+              final Uint8List? originalHash = originalDescription.sha256;
+              if (originalHash != null) {
+                if (!bytesEquals(cachedHash, originalHash)) {
+                  log.warning('''
+Content of ${id.name}-${id.version} has changed compared to what was locked your pubspec.lock.
```
sigurdm, line 117:
```suggestion
Content of ${id.name}-${id.version} has changed compared to what was locked in your pubspec.lock.
```

doc/repository-spec-v2.md
```diff
@@ -256,6 +258,15 @@ parameters. This allows for the server to return signed-URLs for S3, GCS or
 other blob storage service. If temporary URLs are returned it is wise to not set
 expiration to less than 25 minutes (to allow for retries and clock drift).
 
+The `archive_sha256` should be the hex-encoded sha256 checksum of the file at
+archive_url. It is an optional field that allows the pub client to verify the
+integrity of downloaded archive.
```
szakarias, line 263: of downloaded -> of the downloaded ?

lib/src/solver/result.dart
szakarias: Or perhaps: 

The content of $ {id.name}-${id.version} on the server doesn't match the content of what was locked in your pubspec.lock.

Either way is okay, I think.

lib/src/source/hosted.dart
```diff
@@ -319,48 +313,53 @@ class HostedSource extends CachedSource {
...
-    throw FormatException('versions must be a list');
+    return versions.map((map) {
+      final pubspecData = map['pubspec'];
+      if (pubspecData is! Map) {
+        throw FormatException('pubspec must be a map');
+      }
+      var pubspec = Pubspec.fromMap(pubspecData, cache.sources,
+          expectedName: ref.name, location: location);
+      final archiveSha256 = map['archive_sha256'];
+      if (archiveSha256 != null && archiveSha256 is! String) {
```
jonasfj, line 340: Won't this be enough?
```suggestion
      if (archiveSha256 is! String) {
```

test/content_hash_test.dart
```diff
@@ -0,0 +1,213 @@
...
+          'e7a7a0f6d9873e4c40cf68cc3cc9ca5b6c8cef6a2220241bdada4b9cb0083279'),
+    ]).create();
+
+    await pubGet(
+        warning: 'Cached version of foo-1.0.0 has wrong hash - redownloading.');
+    await hostedHashesCache([
+      file('foo-1.0.0.sha256', originalHash),
+    ]).validate();
+  });
+  test(
```
jonasfj, line 164:
```suggestion
  });

  test(
```

lib/src/command/dependency_services.dart
```diff
@@ -457,8 +463,44 @@ class DependencyServicesApplyCommand extends PubCommand {
...
+                description.sha256 == null) {
+              // We removed the hash above before resolution - as we get the
+              // locked id back we need to find the content-hash from the
+              // version listing.
+              //
+              // `pub get` gets this version-listing from the downloaded archive
+              // but we don't want to download all archives - so we copy it from
+              // the version listing.
+              var listedId = (await cache.getVersions(package.toRef()))
+                  .firstWhere((id) => id == package);
```
jonasfj, line 481: Not sure if this is wise, it should never be necessary.
```suggestion
              var listedId = (await cache.getVersions(package.toRef()))
                  .firstWhere((id) => id == package, orElse: () => package);
```
I'm just thinking the error we get from trying to download it, might be better :D

lib/src/source/hosted.dart
```diff
@@ -868,23 +991,63 @@ class HostedSource extends CachedSource {
...
+        if (expectedHash != null && output.value != expectedHash) {
+          log.fine(
+              'Expected content-hash for ${id.name}-${id.version} $expectedHash actual: ${output.value}.');
+          throw FormatException('''
+Downloaded archive for ${id.name}-${id.version} had wrong content-hash.
+
+This indicates a problem on the package repository: `${description.url}`.
+
+See $contentHashesDocumentationUrl.
+''');
```
jonasfj, line 1034: This is the only thing I'm unsure of...

Options:
 * (A) Retry the request, this makes sense if the response didn't have a crc32 checksum.
 * (B) Don't retry the request, this makes sense if the response had a crc32 checksum.

But I don't like making logic that depends on whether we have crc32 checksum or not.

(C) If response carries a header `*-pub-content-sha256: <hex sha256>` then we abort fetching if the header doesn't match the `archive_sha256` from the version-listing API. If the downloaded content doesn't match the value of `*-pub-content-sha256` header, then we retry downloading.
If there is no `*-pub-content-sha256` and `archive_sha256` mismatches downloaded content, then we retry.

This is a bit of a hack, because the header will have to be anything that ends with `-pub-content-sha256`. Because on GCS we can attach custom headers, they are just prefixed `x-goog-meta-`, so our custom headers would be `x-goog-meta-pub-content-sha256` and if hosted on S3 a package repository would have `x-amx-meta-pub-content-sha256` or something like that.

See:
 * https://cloud.google.com/storage/docs/xml-api/reference-headers#xgoogmeta
 * https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html

--------------

But option (C) might also be overkill. Certainly it's something we should do in a follow up PR.

------------------

Back on topic, it's tempting to retry, because:
 * Most likely it's a network issues -- rarely should there be disagreement between version-listing API and the archive stored. So if there is disagreement, it's very possibly a network issue.
 * In cases where there is disagreement, its possible that it would resolve from retrying -- it's likely a temporary inconsistency in cloud APIs -- like updating a GCS object cached in a CDN might take some time to propagate.
 * If the disagreement persist, then it's a bug in the package repository, if the user have to wait for retries to take place before being told that the content-hash of the downloaded archive is wrong, then maybe that's not the end of the world. It's not like the user is waiting for a positive outcome -- it's not a common scenario.

So when in doubt, maybe it's better to retry :D

jonasfj (approved):

jonasfj (approved):

Issue #3398, opened 2022-04-23
Github Downloaded Project Class 'String' has no instance method 'containsKey' Error


algorr:
Hi everyone, 
I have an issue about project that is downloaded from my own github. After a while I developed the project, I download the all project from github desktop application and try to open in VsCode or Android Studio. After when I start to compile or debug I take an issue like that above:

Resolving dependencies...
Class 'String' has no instance method 'containsKey'.
Receiver: "dependency"
Tried calling: containsKey("version")
dart:core                                                        Object.noSuchMethod
package:pub/src/lock_file.dart 130:24                            LockFile._parse.<fn>
dart:collection                                                  MapMixin.forEach
package:pub/src/lock_file.dart 128:22                            LockFile._parse
package:pub/src/lock_file.dart 80:21                             new LockFile.load
package:pub/src/entrypoint.dart 103:35                           Entrypoint._loadLockFile
package:pub/src/entrypoint.dart 97:42                            Entrypoint.lockFile
package:pub/src/entrypoint.dart 286:21                           Entrypoint.acquireDependencies.<fn>
package:pub/src/entrypoint.dart 280:68                           Entrypoint.acquireDependencies.<fn>
package:pub/src/log.dart 428:18                                  progress
package:pub/src/entrypoint.dart 280:26                           Entrypoint.acquireDependencies
package:pub/src/command/get.dart 52:22                           GetCommand.runProtected
package:pub/src/command.dart 183:45                              PubCommand.run.<fn>
package:pub/src/command.dart 183:33                              PubCommand.run.<fn>
dart:async                                                       new Future.sync
package:pub/src/utils.dart 109:12                                captureErrors.wrappedCallback
dart:async                                                       runZonedGuarded
package:pub/src/utils.dart 126:5                                 captureErrors
package:pub/src/command.dart 183:13                              PubCommand.run
package:args/command_runner.dart 209:27                          CommandRunner.runCommand
package:dartdev/dartdev.dart 231:30                              DartdevRunner.runCommand
package:args/command_runner.dart 119:25                          CommandRunner.run.<fn>
dart:async                                                       new Future.sync
package:args/command_runner.dart 119:14                          CommandRunner.run
package:dartdev/dartdev.dart 66:29                               runDartdev
C:\b\s\w\ir\cache\builder\sdk\pkg\dartdev\bin\dartdev.dart 11:9  main
This is an unexpected error. The full log and other details are collected in:

[...]

PR #3487, opened 2022-07-08
Cleaner pubspec.lock parsing

sigurdm:
There were some corner cases where a malformed pubspec.lock would crash pub.

This PR tries to be a bit more systematic.

Ideally we should probably use something like json_serializable - but that would add another dependency to the sdk...

Fixes https://github.com/dart-lang/pub/issues/3398


test/list_package_dirs/lockfile_error_test.dart
```diff
@@ -19,7 +19,7 @@ void main() {
       'list-package-dirs',
       '--format=json'
     ], outputJson: {
-      'error': contains('The lockfile must be a YAML mapping.'),
+      'error': contains('Expected a YAML mapping.'),
```
jonasfj, line 22: Should we not say "lockfile" in the error message anymore? or is it u necessary because we're using source_span?

jonasfj (approved):

lib/src/lock_file.dart
```diff
@@ -340,6 +405,11 @@ ${yamlToString(data)}
...
   }
 
+  static const directMain = 'direct main';
+  static const directDev = 'direct dev';
+  static const directOverridden = 'direct overridden';
+  static const transitive = 'transitive';
+
   /// Returns the dependency classification for [package].
   String _dependencyType(String package) {
     if (_mainDependencies.contains(package)) return 'direct main';
```
sigurdm, line 415: Use the const also here

lib/src/lock_file.dart
```diff
@@ -340,6 +405,11 @@ ${yamlToString(data)}
 ''';
   }
 
+  static const directMain = 'direct main';
```
sigurdm, line 408: Make private

sigurdm: Done

test/list_package_dirs/lockfile_error_test.dart
sigurdm: I don't think it is needed when we have the file name from the source span

lib/src/exceptions.dart
```diff
@@ -118,3 +119,34 @@ bool isUserFacingException(error) {
       error is YamlException ||
       error is UsageException;
 }
+
+/// An exception thrown when parsing a pubspec.yaml or a pubspec.lock.
```
jonasfj, line 123:
```suggestion
/// An exception thrown when parsing a `pubspec.yaml` or a `pubspec.lock`.
```

lib/src/exceptions.dart
```diff
@@ -118,3 +119,34 @@ bool isUserFacingException(error) {
...
+///
+/// These exceptions are often thrown lazily while accessing pubspec properties.
```
jonasfj, line 125: Not really, we explicitly call `allErrors` :D

But yes, it makes sense that they are lazy, because we have to read the SDK constraint before we care if there are other errors.

lib/src/exceptions.dart
```diff
@@ -118,3 +119,34 @@ bool isUserFacingException(error) {
...
+///
+/// By being an [ApplicationException] this will not trigger a stack-trace on
+/// normal operations.
+///
+/// Works as a SourceSpanFormatException, but can contain more context:
```
jonasfj, line 130:
```suggestion
/// Works as a [SourceSpanFormatException], but can contain more context:
```

lib/src/entrypoint.dart
```diff
@@ -101,7 +102,17 @@ class Entrypoint {
...
+      try {
+        return _lockFile = LockFile.load(lockFilePath, cache.sources);
+      } on SourceSpanException catch (e) {
+        throw SourceSpanApplicationException(
+          e.message,
+          e.span,
+          explanation: 'Failed parsing lock file:',
+          hint:
+              'Consider deleting the file and running `$topLevelProgram pub get` to recreate it.',
+        );
```
jonasfj, line 114: Lovely, before we just crashed :D

jonasfj (approved):

jonasfj:
@sigurdm should we remember to merge this and get it shipped?

lib/src/exceptions.dart
sigurdm: I think we only call that during publication.

Issue #3498, opened 2022-07-19
Ignore pubspec_overrides.yaml by default when publishing


jonasfj:
There is no reason to include `pubspec_overrides.yaml` when publishing, indeed one could even say that doing so is surprising behavior.

I'm guessing we just have to add it to the default patterns used in the gitignore/pubignore logic.

SunlightBro:
I could create a PR for this, if wanted.

- Adding pubspec_overrides.yaml to `_basicIgnoreRules` in [lib/src/package.dart#L186](https://github.com/dart-lang/pub/blob/master/lib/src/package.dart#L186)
```diff
    'pubspec.lock',
    '!pubspec.lock/', // We allow a directory called pubspec lock.
+   'pubspec_overrides.yaml',
  ];

```
- creating a `test/lish/does_not_include_pubspec_overrides_file.dart` very similar to [does_not_include_dot_file.dart](https://github.com/dart-lang/pub/blob/master/test/lish/does_not_include_dot_file.dart)

PR #3503, opened 2022-07-27
Add pubspec_overrides.yaml to basicIgnoreRules

SunlightBro:
Fixes: #3498 , related to dart-lang/pana#1089


lib/src/package.dart
```diff
@@ -191,6 +191,7 @@ class Package {
     'packages/',
     'pubspec.lock',
     '!pubspec.lock/', // We allow a directory called pubspec lock.
+    'pubspec_overrides.yaml',
```
sigurdm, line 194: I think we should do
```suggestion
    '/pubspec_overrides.yaml',
```

To only ignore the file at the root (not deeper in the hierarchy).

sigurdm:
Thanks for the contribution! Just one small comment, and I think this is good.

SunlightBro:
Correct me if I'm wrong, but I don't think that [embedding_test](https://github.com/dart-lang/pub/blob/master/test/embedding/embedding_test.dart#L120) failed because of my changes.

On first glance it looks like there has been a change in the Headers
```diff
+ x-powered-by: Dart with package:shelf\n
- server: dart:io with Shelf\n
```
and this change is not yet represented in:
test/testdata/goldens/embedding/embedding_test/[logfile is written with --verbose and on unexpected exceptions.txt](https://github.com/dart-lang/pub/blob/master/test/testdata/goldens/embedding/embedding_test/logfile%20is%20written%20with%20--verbose%20and%20on%20unexpected%20exceptions.txt)

There is also a [comment](https://github.com/dart-lang/pub/blob/master/test/embedding/embedding_test.dart#L127) (from you actually), saying that this test is fragile.

_If you want I can try fixing this test, but I would probably do it in a separat PR_

sigurdm (approved):

PR #3511, opened 2022-08-03
Fix typo "exitcode"

nilsreichardt:
I think it should be called "exit code" instead "exitcode". Correct me, if I'm wrong.


sigurdm (approved):
LGTM

Issue #3520, opened 2022-08-09
dart pub global activate -s git <repo> throws unexpected error => Illegal scheme character (at character 4)


jsroest:
# Environment

Dart version
- 2.17.6 (stable)
- 2.18.0-271.4.beta (beta)
- 2.19.0-59.0.dev (dev)

OS
- macOS 12.5

# Problem
When trying to activate a dart package in global, the installation fails with this message:
```
johannesroest@ws-z14 repos % dart pub global activate -sgit git@bitbucket.org:dalosy/0000000-create_flutter_mono_repo-dalosy.git
Illegal scheme character (at character 4)
git@bitbucket.org:dalosy/0000000-create_flutter_mono_repo-dalosy.git
   ^
dart:core                                    Uri.parse
package:path/src/context.dart 1077:33        _parseUri
package:path/src/context.dart 1052:22        Context.prettyUri
package:path/path.dart 461:34                prettyUri
package:pub/src/global_packages.dart 295:28  GlobalPackages._describeActive
package:pub/src/global_packages.dart 194:34  GlobalPackages._installInCache
package:pub/src/global_packages.dart 114:11  GlobalPackages.activateGit
This is an unexpected error. The full log and other details are collected in:

    /Users/johannesroest/.pub-cache/log/pub_log.txt

Consider creating an issue on https://github.com/dart-lang/pub/issues/new
and attaching the relevant parts of that log file.
```
#pub_log.txt
```
Information about the latest pub run.

If you believe something is not working right, you can go to 
https://github.com/dart-lang/pub/issues/new to post a new issue and attach this file.

Before making this file public, make sure to remove any sensitive information!

Pub version: 2.17.6
Created: 2022-08-09T15:06:12.415552
FLUTTER_ROOT: <not set>
PUB_HOSTED_URL: <not set>
PUB_CACHE: "<not set>"
Command: dart pub global activate -sgit 'git@bitbucket.org:dalosy/0000000-create_flutter_mono_repo-dalosy.git'
Platform: macos

---- Log transcript ----
FINE: Pub 2.17.6
IO  : Created temp directory /var/folders/kl/pzt95yzs61v1z101j60z6myc0000gn/T/pub_1z03xF
IO  : Spawning "git --version" in /Users/johannesroest/repos/.
IO  : Finished git. Exit code 0.
    | stdout:
    | | git version 2.23.0
    | Nothing output on stderr.
FINE: Determined git command git.
[...]

Issue #3497, opened 2022-07-19
Illegal scheme character when depending on git dependency


EstebanCardenas:
Illegal scheme character error thrown when including a git dependency in the `pubspec.yaml` file.

Pubspec.yaml file:

```
name: app
description: app.
publish_to: 'none'
version: 1.0.0+1

environment:
  sdk: ">=2.17.6 <3.0.0"

dependencies:
  authentication:
    git:
      url: git@gitlab.com:app/ifp/modules/authentication.git
      ref: main
  dashboard: 
    git:
      url: git@gitlab.com:app/ifp/modules/dashboard.git
      ref: main
```

Error:

```
flutter pub get
Running "flutter pub get" in app...                         
Illegal scheme character (at character 4)
git@gitlab.com:app/ifp/modules/dashboard.git
```

Stack trace:

```
dart:core                                           Uri.parse
package:path/src/context.dart 1077:33               _parseUri
package:path/src/context.dart 1052:22               Context.prettyUri
package:path/path.dart 461:34                       prettyUri
package:pub/src/source/git.dart 276:16              GitSource._describeUncached
===== asynchronous gap ===========================
package:pub/src/source/git.dart 224:11              GitSource.doGetVersions.<fn>
===== asynchronous gap ===========================
package:pool/pool.dart 127:14                       Pool.withResource
===== asynchronous gap ===========================
package:pub/src/source/git.dart 218:12              GitSource.doGetVersions
===== asynchronous gap ===========================
package:pub/src/system_cache.dart 186:20            SystemCache.getVersions
===== asynchronous gap ===========================
package:pub/src/solver/package_lister.dart 82:31    PackageLister._versions.<fn>
===== asynchronous gap ===========================
package:pub/src/solver/package_lister.dart 126:15   PackageLister.countVersions
===== asynchronous gap ===========================
package:pub/src/solver/version_solver.dart 336:14   VersionSolver._choosePackageVersion.<fn>
===== asynchronous gap ===========================
package:pub/src/utils.dart 281:26                   minByAsync
===== asynchronous gap ===========================
package:pub/src/solver/version_solver.dart 335:19   VersionSolver._choosePackageVersion
===== asynchronous gap ===========================
package:pub/src/solver/version_solver.dart 97:18    VersionSolver.solve.<fn>
===== asynchronous gap ===========================
package:pub/src/source/hosted.dart 954:14           HostedSource.withPrefetching.<fn>
===== asynchronous gap ===========================
[...]

Issue #3377, opened 2022-04-04
pub upgrade fails if url is specified in git@github.com format for authentication by ssh


koji-1009:
# Environment

- `pub version` or `flutter pub version`: `Pub 2.17.0-265.0.dev`
- OS version: macOS Monterey
- Are you using the Chinese community mirror or a corporate firewall?: No

# Problem

If you specify a git repository for dependency in the form `git@github.com:rrousselGit/provider.git`, then Succeeds in flutter's stable channel, but fails in flutter's master channel.

The operation can be reproduced by describing the dependency as follows.
(The project that noticed this problem refers to a private GitHub repository that only internal members have access to.)

```yaml
  provider:
    git:
      url: git@github.com:rrousselGit/provider.git
      ref: master
```

https://github.com/dart-lang/pub/commit/610ce7f280189f39ec411eb0a8592a191940d8d2#diff-1639c4669c428c26e68cfebd5039a33f87ba568795f2c058c303ca8528f62b77R720

In the above commit, `Uri.parse` is now called internally.
Dart's `Uri.parse` throws FormatException for formats like `git@github.com`.

[dartpad](https://dartpad.dev/?id=d8c86cd308caf7926ee82a8a44362ce0)

## Expected behavior

pub upgrade succeeds

## Actual behavior

pub upgrade fails

## --trace output

```
$ dart pub upgrade
Resolving dependencies... (4.0s)
Illegal scheme character (at character 4)
git@github.com:rrousselGit/provider.git
   ^
dart:core                                   Uri.parse
package:path/src/context.dart 1077:33       _parseUri
package:path/src/context.dart 1052:22       Context.prettyUri
package:path/path.dart 461:34               prettyUri
package:pub/src/source/git.dart 730:23      GitResolvedDescription.format
package:pub/src/solver/report.dart 326:40   SolveReport._writeId
package:pub/src/solver/report.dart 302:5    SolveReport._reportPackage
===== asynchronous gap ===========================
package:pub/src/solver/report.dart 114:7    SolveReport._reportChanges
===== asynchronous gap ===========================
package:pub/src/solver/report.dart 47:5     SolveReport.show
===== asynchronous gap ===========================
package:pub/src/solver/result.dart 112:5    SolveResult.showReport
===== asynchronous gap ===========================
package:pub/src/entrypoint.dart 321:7       Entrypoint.acquireDependencies
===== asynchronous gap ===========================
[...]

PR #3524, opened 2022-08-11
Fix remaining usages of path.prettyUri on git uris

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/3520
Fixes https://github.com/dart-lang/pub/issues/3497

There were remaining uses of `path.prettyUri` on git-URIs. These fails as `Uri.parse` does not handle `git@github.com/a/b.git` style urls. (As discussed in: https://github.com/dart-lang/pub/issues/3377#issuecomment-1091430936)

@mit-mit 
We should consider hotfixing this into 2.18 if there is time.

### DIFF
--- a/lib/src/global_packages.dart
+++ b/lib/src/global_packages.dart
@@ -295,7 +295,7 @@ To recompile executables, first run `$topLevelProgram pub global deactivate $nam
 
     if (description is GitDescription) {
       log.message('Package ${log.bold(name)} is currently active from Git '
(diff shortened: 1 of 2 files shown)


szakarias (approved):
LGTM

mit-mit:
> We should consider hotfixing this into 2.18 if there is time.

Not sure if it's too late, but please open a cherry pick bug, then it can follow in a hotfix release if needed.

Issue #3519, opened 2022-08-08
Switch external language from pub.dartlang.org to pub.dev


timsneath:
It's been a long time since we switched the external-facing site to pub.dev now, and most Flutter and Dart users never knowingly used the old pub.dartlang.org URL. While it remains a technical implementation detail, it's confusing that we reference it in our `pub publish` process, not least because we mix and match the two URLs in our logging messages.

I'd like to propose that we now change the two references to pub.dartlang.org below to pub.dev, matching the other references:

```
[C:\src\win32_registry] dart pub publish
Publishing win32_registry 1.0.2 to https://pub.dartlang.org:
|-- CHANGELOG.md
|-- CODE_OF_CONDUCT.md
|-- LICENSE
|-- README.md
|-- analysis_options.yaml
|-- coverage.cmd
|-- example
|   |-- create_values.dart
|   |-- example.dart
|   '-- read_values.dart
|-- lib
|   |-- src
|   |   |-- access_rights.dart
|   |   |-- registry.dart
|   |   |-- registry_hive.dart
|   |   |-- registry_key.dart
|   |   |-- registry_key_info.dart
|   |   |-- registry_value.dart
|   |   |-- registry_value_type.dart
|   |   '-- utils.dart
|   '-- win32_registry.dart
|-- pubspec.yaml
'-- test
    '-- win32_registry_test.dart

Publishing is forever; packages cannot be unpublished.
Policy details are available at https://pub.dev/policy

Do you want to publish win32_registry 1.0.2 to https://pub.dartlang.org (y/N)? y
Uploading...
Successfully uploaded https://pub.dev/packages/win32_registry "1.0.2".
```

@mit-mit for visibility / approval.

timsneath:
As an aside, it's weird that the version number is quoted in the last line; perhaps we can fix that at the same time?

mit-mit:
> I'd like to propose that we now change the two references to pub.dartlang.org below to pub.dev, matching the other references:

Yes, I think we can change these. We probably can't change the actual API endpoint, but these are UI messages, and they are about what you're publishing to which *is* pub.dev.

> As an aside, it's weird that the version number is "quoted" in the last line; perhaps we can fix that at the same time?

Yeah, that looks weird indeed, fixing in  https://github.com/dart-lang/pub-dev/pull/5941

PR #3528, opened 2022-08-16
Use pub.dev in writing

sigurdm:
Change all mentions of the pub.dartlang.org website to pub.dev in the ui.

Fixes: https://github.com/dart-lang/pub/issues/3519

### DIFF
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -154,7 +154,7 @@ class LishCommand extends PubCommand {
   Future<void> _publish(List<int> packageBytes) async {
     try {
       final officialPubServers = {
-        'https://dev',
+        'https://pub.dev',
         // [validateAndNormalizeHostedUrl] normalizes https://pub.dartlang.org
         // to https://pub.dev, so we don't need to do allow that here.
 
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -25,7 +25,7 @@ class LishCommand extends PubCommand {
   @override
   String get name => 'publish';
   @override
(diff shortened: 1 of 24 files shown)


lib/src/command/lish.dart
```diff
@@ -154,9 +154,9 @@ class LishCommand extends PubCommand {
   Future<void> _publish(List<int> packageBytes) async {
     try {
       final officialPubServers = {
-        'https://pub.dartlang.org',
-        // [validateAndNormalizeHostedUrl] normalizes https://pub.dev to
-        // https://pub.dartlang.org, so we don't need to do allow that here.
+        'https://dev',
```
timsneath, line 157: Should this be pub.dev?

sigurdm: Thanks for catching this!

timsneath:
Is this ready to be merged, @sigurdm?

sigurdm:
Unfortunately I think we are going to backpedal a bit here.

When trying it out some unpleasing behaviour became apparent when using an older pub with a pubspec.lock created after this patch: Pub (and dart run) would look for cached entries based on the server in the pubspec.lock (instead of the package_config.json as one could argue it should), and crash when it could not find them (fixed in https://github.com/dart-lang/pub/pull/3533)

I am working on a reduced version that only updates the writing, but still uses pub.dartlang.org for all actual operations.

timsneath:
Oh, for sure. So long as we can fix the UI text itself, it's fine if we still use pub.dartlang.org behind the scenes. Thanks!

jonasfj (approved):

PR #3533, opened 2022-08-19
Avoid packageGraph (and therefore getExecutableForCommand) depending on the lock-file

sigurdm:
The source of truth for the package environment should be the `.dart_tool/packageConfig.json` file.

This leads to us no longer being able to do a global run if the .dart_tool/packageConfig.json file is missing, but that was supporting compatibility with dart <1.12, and not clear that this compatibility would work in any way-shape-or-form as so many things have changed since.


lib/src/entrypoint.dart
```diff
@@ -107,6 +107,30 @@ class Entrypoint {
 
   LockFile? _lockFile;
 
+  late PackageConfig packageConfig = () {
```
jonasfj, line 110: Maybe, give a documentation comment that warns that this throws `DataException` if loading the package config fails.
And maybe, indicate that this should be used when this exception is handled, or desirable, for example `dart run`.

jonasfj (approved):

lib/src/entrypoint.dart
sigurdm: Done

Issue #3456, opened 2022-06-09
dart pub upgrade --major-versions is worse than without the option


sortie:
# Environment

- `pub version` or `flutter pub version`:

I don't have a `pub version` command but `dart --version` is `Dart SDK version: 2.17.3 (stable) (Wed Jun 1 11:06:41 2022 +0200) on "linux_x64`.

- OS version: Linux 5.17.6
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

`dart pub upgrade --major-versions` is failing to find solutions that `dart pub upgrade` finds.

```yaml
name: mypkg
version: 0.0.1
environment:
  sdk: '>=2.10.0 <3.0.0'
dependencies:
  collection: ^1.15.0-nullsafety.4
  crypto: ^3.0.1
  googleapis: ^8.1.0
  googleapis_auth: ^1.3.0
  markdown: ^4.0.1
  path: ^1.6.2
  protobuf: ^2.0.1
dev_dependencies:
  lints: ^1.0.1
  test: ^1.15.7
```

with with this pubspec.lock: https://gist.github.com/sortie/0e70bee328c285b783af01bdbee9e83c

## Expected behavior

`dart pub upgrade --major-version` just gives me all the good new stuff and finds at least the solutions the ordinary variant can find.

## Actual behavior

Running `dart pub upgrade --major-version` fails to resolve all the available versions:

```
Resolving dependencies... (1.4s)
  _discoveryapis_commons 1.0.2 (1.0.3 available)
  _fe_analyzer_shared 36.0.0 (40.0.0 available)
  analyzer 3.3.1 (4.1.0 available)
  args 2.3.0 (2.3.1 available)
  async 2.8.2 (2.9.0 available)
  collection 1.15.0 (1.16.0 available)
  convert 3.0.1 (3.0.2 available)
  coverage 1.1.0 (1.3.2 available)
  crypto 3.0.1 (3.0.2 available)
  fixnum 1.0.0 (1.0.1 available)
  frontend_server_client 2.1.2 (2.1.3 available)
> googleapis 9.1.0 (was 8.1.0)
  googleapis_auth 1.3.0 (1.3.1 available)
  http_parser 4.0.0 (4.0.1 available)
> lints 2.0.0 (was 1.0.1)
> markdown 5.0.0 (was 4.0.1)
  meta 1.7.0 (1.8.0 available)
  mime 1.0.1 (1.0.2 available)
  path 1.8.1 (1.8.2 available)
[...]

PR #3539, opened 2022-08-25
Don't be conservative in upgrade --major-versions without specific pa…

sigurdm:
…ckages

Fixes: https://github.com/dart-lang/pub/issues/3456

### DIFF
--- a/lib/src/command/upgrade.dart
+++ b/lib/src/command/upgrade.dart
@@ -225,7 +225,7 @@ be direct 'dependencies' or 'dev_dependencies', following packages are not:
     // When doing '--majorVersions' for specific packages we try to update other
     // packages as little as possible to make a focused change (SolveType.get).
     //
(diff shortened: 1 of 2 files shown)


lib/src/command/upgrade.dart
```diff
@@ -222,6 +222,14 @@ be direct 'dependencies' or 'dev_dependencies', following packages are not:
     }
     final newPubspecText = _updatePubspec(changes);
 
+    // When doing '--majorVersions' for specific packages we try to update other
+    // packages as little as possible to make a focused change (SolveType.get).
+    //
+    // But without a specific package we want to get as much non-major updates
```
szakarias, line 228: perhaps much -> many ?

szakarias (approved):
LGTM

lib/src/command/upgrade.dart
sigurdm: Done

Issue #2281, opened 2019-12-16
Validate CRC32c if `x-goog-hash` header is present


jonasfj:
The `x-goog-hash` header is present in responses from GCS, see: https://cloud.google.com/storage/docs/hashes-etags

Please note that if the object was [transcoded by GCS](https://cloud.google.com/storage/docs/transcoding) then the has will be invalid.
Also the validate must happen before gzip decompression, if the response is gzipped, and the request **must** be made with `Accept-Encoding: gzip`.

If we get this wrong, we won't be able to gzip data in GCS in the future, let's aim to avoid that.

jonasfj:
Probably we can use [package:crclib](https://pub.dev/documentation/crclib/latest/crclib/Crc32Iscsi/Crc32Iscsi.html).

jonasfj:
This is actually in:
https://pub.dev/documentation/crclib/latest/crclib/Crc32Iscsi-class.html

After viewing `http.dart` with @sigurdm it seems we'll have to do a minor refactor of all http handling before we can retry errors during the response phase of the requests.

jonasfj:
There is also a crc32 implementation here:
https://github.com/dart-lang/pana/blob/master/lib/src/license_detection/crc32.dart

Thought this isn't the variant used by GCS, but it could probably be tweaked.

PR #3546, opened 2022-09-01
Checksum validation of archives (#2281)

nehalvpatel:
Fixes: https://github.com/dart-lang/pub/issues/2281


lib/src/source/hosted.dart
```diff
@@ -860,16 +863,27 @@ class HostedSource extends CachedSource {
...
           p.join(tempDirForArchive, '$packageName-$version.tar.gz');
       var response = await withAuthenticatedClient(
-          cache,
-          Uri.parse(description.url),
-          (client) => client.send(http.Request('GET', url)));
+          cache, Uri.parse(description.url), (client) {
+        var req = http.Request('GET', url);
+        // TODO: How can we properly check if a request is GCS bound, so only
+        // those requests get this header? (for checksum validation)
+        req.headers['Cache-Control'] = 'no-transform';
```
nehalvpatel, line 870: How can we properly check if a request is GCS bound, so only those requests get this header?

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1114,95 @@ class _RefAndCache {
   @override
   bool operator ==(Object other) => other is _RefAndCache && other.ref == ref;
 }
+
+extension GCSChecksumValidation on http.StreamedResponse {
+  static const checksumsHeaderName = 'x-goog-hash';
+
+  bool get isFromGCS {
+    // TODO: Does this suffice as a heuristic?
+    return headers.containsKey(checksumsHeaderName);
```
nehalvpatel, line 1123: This is run after the request is already fired. It could be moved pre-request if we figure out a way to properly detect whether a request is GCS bound in the other TODO.

lib/src/source/hosted.dart
sigurdm: Hmm - could we just always send this header?
I think for the coming content-hashing: https://github.com/dart-lang/pub/pull/3482 any transformations to the archive (on any server) would be problematic.

Furthermore we can validate the crc32 checksum whenever it is provided, it doesn't strictly have to be from gcs...

We could even update the https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md to mention the header, and explain that it is checked.

jonasfj: I'd think it'd be safe to always send it.

But also unnecessary, I don't think we have any intermediate proxy, and if we did we shouldn't allow to do transformations. No need to compress a `tar.gz` again :see_no_evil:

pubspec.yaml
```diff
@@ -24,6 +24,7 @@ dependencies:
   shelf: ^1.1.1
   source_span: ^1.8.1
   stack_trace: ^1.10.0
+  stream_transform: ^2.0.0
```
jonasfj, line 27: I don't see it here: https://github.com/dart-lang/sdk/blob/main/DEPS

So we would need to add a dependency on it, the question becomes if it isn't less work to do:
```dart
transformStream(Stream<List<int>> stream) async* {
  await for (final chunk in stream) {
    crc.update(chunk);
    yield chunk;
  }
}
```

There might be dragons here I haven't thought of, but I suspect it's not super dangerous to do this.

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1114,95 @@ class _RefAndCache {
   @override
   bool operator ==(Object other) => other is _RefAndCache && other.ref == ref;
 }
+
+extension GCSChecksumValidation on http.StreamedResponse {
```
jonasfj, line 1118: Do we need an extension? Can't we do with utility methods..

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1114,95 @@ class _RefAndCache {
...
+    final checksumComputer = Crc32c();
+
+    return stream.tap(checksumComputer.update, onDone: () {
+      var computedCrc32c = checksumComputer.finalize();
+
+      log.io(
+          'Computed checksum ($computedCrc32c) for package with GCS checksum of ($gcsCrc32c).');
+
+      if (gcsCrc32c != computedCrc32c) {
+        throw PackageIntegrityException(
```
jonasfj, line 1150: crc32c is to protect against transport issues... I think we should retry, if it fails.

lib/src/source/hosted.dart
nehalvpatel: So, I wrote this under the assumption that [adding this header would impair performance](https://stackoverflow.com/a/20197125) for requests that are bound for custom hosts. But it sounds like this header is desired regardless for the upcoming content hashing feature.

I can keep the header and add a comment/note?

Also, let me know if you'd like me to update the repository spec to include this header. 🫡

pubspec.yaml
nehalvpatel: Oh yea, that looks like it should be a good alternative. I'll work on implementing that.

lib/src/source/hosted.dart
nehalvpatel: I made it an extension primarily because of how it grouped all the checksum related logic, and I had assumed that it would make extensibility (checksum validation for other hosts) and testing easier.

I can break it up into utility methods, though. I also see now that checksum validation seems like it will be more of a first-class feature for all hosts in the future, so the extension makes less sense.

lib/src/source/hosted.dart
nehalvpatel: Gotcha, I'll add retries.

pubspec.yaml
nehalvpatel: I ended up using a [conceptually similar approach](https://github.com/dart-lang/pub/pull/3546/commits/7755d07bbf4c736aa12567c298e2fb8f6d42149c) which should be less dragon-y. 😃

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1113,97 @@ class _RefAndCache {
   @override
   bool operator ==(Object other) => other is _RefAndCache && other.ref == ref;
 }
+
+extension GCSChecksumValidation on http.StreamedResponse {
+  static const checksumsHeaderName = 'x-goog-hash';
+
+  bool get isFromGCS {
```
sigurdm, line 1120: I don't think we have to mention GCS, we always check the hash if the header is present.

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1113,97 @@ class _RefAndCache {
...
+    }
+
+    final checksumComputer = Crc32c();
+
+    return stream
+        .transform(onDataTransformer(checksumComputer.update))
+        .transform(onDoneTransformer(() {
+      var computedCrc32c = checksumComputer.finalize();
+
+      log.io(
```
sigurdm, line 1147: I would use `log.fine` here if logging this at all...

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1113,97 @@ class _RefAndCache {
...
+    // https://cloud.google.com/storage/docs/xml-api/reference-headers#xgooghash
+
+    // Additionally, when the Dart http package encounters multiple response
+    // headers with the same key, it concatenates their values with a comma
+    // before inserting a single item with that key and concatenated value into
+    // its response "headers" Map. There is a TODO to refactor the
+    // response class to use a HttpHeaders object instead of a Map, but the
+    // person it was assigned to, nweiz@, has long since left the Dart team.
+    // https://github.com/dart-lang/http/issues/24
+    // https://github.com/dart-lang/http/blob/06649afbb5847dbb0293816ba8348766b116e419/pkgs/http/lib/src/base_response.dart#L29
```
sigurdm, line 1176: I like that we have some extra details, but no reason to call out usernames,
```suggestion
    // Additionally, when the Dart http client encounters multiple response
    // headers with the same key, it concatenates their values with a comma
    // before inserting a single item with that key and concatenated value into
    // its response "headers" Map. 
    // See https://github.com/dart-lang/http/issues/24
    // https://github.com/dart-lang/http/blob/06649afbb5847dbb0293816ba8348766b116e419/pkgs/http/lib/src/base_response.dart#L29
```

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1113,97 @@ class _RefAndCache {
...
+
+    // Therefore, we need to use this roundabout method for parsing the
+    // CRC32C checksum.
```
sigurdm, line 1179: I think this follows from the above.
```suggestion
```

sigurdm (commented):
I think this needs some testing.

I think you can set the test package-server up so that it returns a checksum, and then make a test that the client validates, and somehow also check that the request is retried.

lib/src/source/hosted.dart
nehalvpatel: Yea, I wasn't sure on what the proper etiquette was so I erred towards more information. Noted for the future, though!

lib/src/source/hosted.dart
```diff
@@ -883,6 +886,12 @@ class HostedSource extends CachedSource {
     });
   }
 
+  http.Request _createArchiveRequest(Uri url) {
+    var request = http.Request('GET', url);
+    request.headers[io.HttpHeaders.cacheControlHeader] = 'no-transform';
```
sigurdm, line 891: Could you write a comment about what this does and why we need it (are we sure we need it - is this not mainly relevant to images, I would expect the checksum that gcs serves to be reliable)?

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1109,91 @@ class _RefAndCache {
   @override
   bool operator ==(Object other) => other is _RefAndCache && other.ref == ref;
 }
+
+const checksumHeaderName = 'x-goog-hash';
+
+bool _hasChecksumHeader(Map<String, String> headers) {
```
sigurdm, line 1115: Perhaps just inline this helper

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1109,91 @@ class _RefAndCache {
...
+/// Adds a checksum validation "tap" to the response stream and returns a
+/// wrapped Stream object, which should be used to consume the incoming data.
+///
+/// As chunks are received, a CRC32C checksum is updated.
+/// Once the download is completed, the final checksum is compared with
+/// the one present in the checksum response header.
+///
+/// Throws [PackageIntegrityException] if anything is wrong with the checksum
+/// or if there is a checksum mismatch.
+Stream<List<int>> _responseStreamWithChecksumValidationTap(
```
sigurdm, line 1128: Probably easier to read this as an async* method. Something along the lines of:
```dart
Stream<List<int>> _validateChecksum(Stream<List<int>> stream, int? expectedChecksum) async* {
  if (checksum == null) yield* stream;
  final crc32 = Crc32();
  await for(final chunk in stream) {
    crc32.update(chunk);
  }
  final actualChecksum = crc32.finalize();
  if (actualChecksum != expectedChecksum) {
    // throw here.
  }
}
```

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1109,91 @@ class _RefAndCache {
...
+
+  final parts = checksums.split(',');
+  for (final part in parts) {
+    if (part.startsWith('crc32c=')) {
+      final undecoded = part.substring('crc32c='.length);
+
+      try {
+        final bytes = base64.decode(undecoded);
+        return bytesToUint32(bytes);
+      } catch (e) {
```
sigurdm, line 1175: Please avoid catch-alls.
Probably we just need to catch `FormatException` here.

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1109,91 @@ class _RefAndCache {
...
+/// wrapped Stream object, which should be used to consume the incoming data.
+///
+/// As chunks are received, a CRC32C checksum is updated.
+/// Once the download is completed, the final checksum is compared with
+/// the one present in the checksum response header.
+///
+/// Throws [PackageIntegrityException] if anything is wrong with the checksum
+/// or if there is a checksum mismatch.
+Stream<List<int>> _responseStreamWithChecksumValidationTap(
+    http.StreamedResponse response, String fileName) {
```
sigurdm, line 1129: I don't think you are using the file name here.

sigurdm (commented):
A few suggestions

lib/src/source/hosted.dart
nehalvpatel: So, I added that due to some guidance in the Decompressive transcoding section on [Transcoding of gzip-compressed files](https://cloud.google.com/storage/docs/transcoding#decompressive_transcoding):

> Preventing decompressive transcoding is useful, for example, if you want to reduce egress cost or time or if you want to validate the downloaded objects have the expected crc32c/md5 checksums.

I'm not familiar enough with the storage infrastructure to know if it's strictly needed for our case.

In any case, I should comment that and I will shortly.

lib/src/source/hosted.dart
nehalvpatel: Oh, yea that's a lot more compact and clear. I'll give it a shot.

lib/src/source/hosted.dart
nehalvpatel: So, I'm using it as diagnostic information in the "log.fine" call a few lines below.

`log.fine('Computed CRC32C ($crc32c) for $fileName with hosted CRC32C of ($hostedCrc32c)');`

I was not 100% happy with the solution of passing `fileName`, but I couldn't come up with a better way in time so I left it like that.

lib/src/source/hosted.dart
nehalvpatel: Yes, I'll remove the catch-all. I think we'll also need to catch `RangeError`.

lib/src/source/hosted.dart
nehalvpatel: Do you mean like this?

```dart
bool _hasChecksumHeader(Map<String, String> headers) =>
    headers.containsKey(checksumHeaderName);
```

Or remove the function and replace the function call?

```dart
var responseStream = response.headers.containsKey(checksumHeaderName)
          ? _validateChecksum(response, fileName)
          : response.stream;
```

lib/src/source/hosted.dart
nehalvpatel: Nevermind, I see that Errors should be handled differently. https://stackoverflow.com/questions/57000995/throwing-exceptions-in-dart-for-invalid-parameters-e-g-sign-in/57004304#57004304

lib/src/source/hosted.dart
nehalvpatel: So, I tried this out and unfortunately, it ends up consuming the file stream entirely instead of tapping it.

lib/src/utils.dart
```diff
@@ -148,6 +149,17 @@ Future<List<T>> waitAndPrintErrors<T>(Iterable<Future<T>> futures) {
   });
 }
 
+/// Returns a [StreamTransformer] that will call [onData] for each data
+/// event of the stream.
+///
+/// The stream will be passed through unchanged.
+StreamTransformer<T, T> onDataTransformer<T>(void Function(T data) onData) {
```
jonasfj, line 156: Can't this be done with `await for`?

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1138,83 @@ class _RefAndCache {
...
+    if (part.startsWith('crc32c=')) {
+      final undecoded = part.substring('crc32c='.length);
+
+      try {
+        final bytes = base64.decode(undecoded);
+        return bytesToUint32(bytes);
+
+        // Suppress lint because 1) we are dealing with arbitrary runtime values
+        // 2) we do not want to duplicate the validation logic that is already
+        // throwing the `RangeError`.
```
jonasfj, line 1197: Consider making it throw a `FormatException`

lib/src/source/hosted.dart
nehalvpatel: Actually, I forgot to add a `yield chunk;` link in the `await for` when I tried this yesterday. After adding that in, this approach works!

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1138,83 @@ class _RefAndCache {
...
+      } on RangeError {
+        continue;
+      } on FormatException {
+        continue;
+      }
+    }
+  }
+
+  throw PackageIntegrityException(
+    'Package archive "$fileName" has a malformed CRC32C checksum in its response headers',
```
jonasfj, line 1208: We could also ignore the case where the header is there, but there is no crc32c.
It feels safer to ignore, but I don't know of anyone that would make a server which returns an x-goog-hash header without crc32c in the header. But it seems like someone else emulating GCS might do that?

lib/src/source/hosted.dart
jonasfj: Generally, in Dart you're not supposed to catch errors, only exceptions.
Instead you should test upfront if your input satisfies the required properties of the function. If that's not possible or feasible then maybe the function should be throwing an exception instead.

This is just nit :)

nehalvpatel: So, the code throwing the `FormatError` is actually within the Dart SDK. There are two function calls we use here that could throw that: 1) `base64.decode()` and 2) `ByteData.getUint32()`, which is called from within `bytesToUint32()`.

For 1), I could write an extension method that just calls the original method but catches `FormatError` and re-throws as `FormatException`.

For 2), I could update `bytesToUint32()` to catch `FormatError` and re-throw as `FormatException`.

lib/src/utils.dart
nehalvpatel: So, Sigurd [suggested that as well](https://github.com/dart-lang/pub/pull/3546#discussion_r974138761), but I made a mistake when first trying it (forgot to yield chunks).

After adding that in, it seems to work. I'll test and make the change.

lib/src/source/hosted.dart
```diff
@@ -857,19 +858,43 @@ class HostedSource extends CachedSource {
...
+          if (expectedChecksum == null) {
+            // Skip validation if the checksum response header is not present.
+            await createFileFromStream(response.stream, archivePath);
+          } else {
+            // The checksum will be validated once this finishes consuming the
+            // stream. [PackageIntegrityException] will be thrown on failure.
+            await createFileFromStream(
+                _validateStream(response.stream, expectedChecksum, fileName),
+                archivePath);
+          }
```
jonasfj, line 889:
```dart

Steam<List<int>> stream = response.stream;

if(expectedChecksum...) {
  stream = ...
}
await createFileStream...
```

lib/src/source/hosted.dart
nehalvpatel: Sure, I can make that change. I think that seems like straightforward reasoning.

lib/src/source/hosted.dart
```diff
@@ -857,21 +860,46 @@ class HostedSource extends CachedSource {
...
+          // There are ways around this, and we might revisit this later.
+          await createFileFromStream(stream, archivePath);
+        },
+        // Retry if the checksum response header was malformed or the actual
+        // checksum did not match the expected checksum.
+        retryIf: (e) => e is PackageIntegrityException,
+        onRetry: (e, retryCount) => log.io(
+            'Retry #${retryCount + 1} because of checksum error with GET $url...'),
+        maxAttempts: math.max(
+          1, // Having less than 1 retry is **always** wrong.
```
sigurdm, line 895: is this 1 retry (two attempts) or one attempt? I would think that not retrying, but attempting once is acceptable.... (I think that is what the code does, so maybe just write)
```suggestion
          1, // Having less than 1 attempt doesn't make sense.
```

lib/src/io.dart
```diff
@@ -172,7 +172,7 @@ List<int> readBinaryFile(String file) {
 }
 
 /// Reads the contents of the binary file [file] as a [Stream].
-Stream<List<int>> readBinaryFileAsSream(String file) {
+Stream<List<int>> readBinaryFileAsStream(String file) {
```
sigurdm, line 175: Thanks!

sigurdm:
This starts to look good!

sigurdm (approved):

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1128,82 @@ class _RefAndCache {
...
+
+  final actualChecksum = crc32c.finalize();
+
+  log.fine(
+      'Computed checksum $actualChecksum for "$fileName" with expected CRC32C '
+      'of $expectedChecksum.');
+
+  if (actualChecksum != expectedChecksum) {
+    throw PackageIntegrityException(
+      'Package archive "$fileName" has a CRC32C checksum mismatch',
```
sigurdm, line 1159: Can we change this to show the URL instead of the filename? The filename is an artifact of how we process the archive data.

lib/src/source/hosted.dart
```diff
@@ -1100,3 +1128,82 @@ class _RefAndCache {
   @override
   bool operator ==(Object other) => other is _RefAndCache && other.ref == ref;
 }
+
+const checksumHeaderName = 'x-goog-hash';
```
jonasfj, line 1132: Does this need to be exported?
```suggestion
const _checksumHeaderName = 'x-goog-hash';
```

test/package_server.dart
```diff
@@ -18,15 +21,24 @@ import 'descriptor.dart' as d;
...
+  final shelf_io.IOServer _inner;
 
   /// Handlers of requests. Last matching handler will be used.
   final List<_PatternAndHandler> _handlers = [];
 
   // A list of all the requests recieved up till now.
   final List<String> requestedPaths = <String>[];
 
+  // Whether the [IOServer] should compress the content, if possible.
+  // See [HttpServer.autoCompress] for details.
```
jonasfj, line 34: Can we specify in the comment whether it defaults to being enabled or not?

jonasfj (approved):

lib/src/source/hosted.dart
jonasfj: Can we write something like:
> Package archive downloaded from "$url" has "x-goog-hash: crc32c=$crc" which doesn't match the checksum of the archive downloaded.

lib/src/source/hosted.dart
nehalvpatel: So, that export is used in "test/package_server.dart":

https://github.com/dart-lang/pub/blob/ce3a0ec5fd8598d609b6c4498b20ceb37c6798ae/test/package_server.dart#L122-L128

https://github.com/dart-lang/pub/blob/ce3a0ec5fd8598d609b6c4498b20ceb37c6798ae/test/package_server.dart#L282-L283

sigurdm: If it needs to be exported only for testing you can mark it `@visibleForTesting`

nehalvpatel: Nice! Added that.

PR #3565, opened 2022-09-16
 Do resolution before publishing.

sigurdm:
Also change publishing with dependency_overrides to be a warning instead of an error.

Otherwise it would be impossible to publish breaking changes to packages with a cyclic dependency.

This caused a lot of tests to fail (they were not set up to resolve with the default test sdk version).


sigurdm:
cc @jakemac53

jakemac53:
LGTM, thanks! This will make the publishing process for package:test much less sketchy :). cc @natebosch @kevmoo

kevmoo:
Hold up: do you MUTATE dep overrides on publish to remove them?

If someone publishes w/ overrides intact, it'll blow up like...bad....on usage, right?

natebosch:
> If someone publishes w/ overrides intact, it'll blow up like...bad....on usage, right?

I think only if they ignored the warning from pub, haven't tested against the versions of dependencies that it will resolve against, and there is some incompatibility.

If the concern is having the overrides listed in the published pubspec.yaml - that shouldn't matter. They won't have an effect.

test/descriptor.dart
```diff
@@ -33,7 +33,7 @@ TarFileDescriptor tar(String name, [List<Descriptor>? contents]) =>
 
 /// Describes a package that passes all validation.
 DirectoryDescriptor get validPackage => dir(appPath, [
-      libPubspec('test_pkg', '1.0.0', sdk: '>=1.8.0 <=2.0.0'),
+      libPubspec('test_pkg', '1.0.0', sdk: '>=0.1.2 <=0.2.0'),
```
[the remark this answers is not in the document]
sigurdm: We have a lot of tests testing publishing a "valid" package. The default sdk when running a test is v0.1.2+3, that went fine when `pub lish` did no resolution, but now it would fail. So Either we had to update each publish-test with a custom sdk version or to make the valid package compatible with the test version.

Will put a line in the pr description.

sigurdm:
> If the concern is having the overrides listed in the published pubspec.yaml - that shouldn't matter. They won't have an effect.

Yeah only overrides in the root package take effect.

test/lish/archives_and_uploads_a_package_test.dart
```diff
@@ -15,10 +15,9 @@ import '../test_pub.dart';
 import 'utils.dart';
 
 void main() {
-  setUp(d.validPackage.create);
```
jonasfj, line 18: Just curious, why?

test/lish/dry_run_errors_if_private_test.dart
```diff
@@ -10,7 +10,7 @@ import '../descriptor.dart' as d;
 import '../test_pub.dart';
 
 void main() {
-  test('preview shows an error if the package is private', () async {
+  test('dyr-run shows an error if the package is private', () async {
```
jonasfj, line 13:
```suggestion
  test('dry-run shows an error if the package is private', () async {
```

test/lish/package_creation_provides_invalid_json_test.dart
```diff
@@ -10,12 +10,13 @@ import '../test_pub.dart';
...
-  setUp(d.validPackage.create);
-
   test('package creation provides invalid JSON', () async {
     await servePackages();
+    await d.validPackage.create();
     await d.credentialsFile(globalServer, 'access token').create();
-    var pub = await startPublish(globalServer);
+    var pub = await startPublish(
+      globalServer,
+    );
```
jonasfj, line 19: Do we need this change? Don't get me wrong I love trailing comma :D
```suggestion
    var pub = await startPublish(globalServer);
```

test/validator/dependency_test.dart
```diff
@@ -56,213 +65,194 @@ Future<void> setUpDependency(dep,
...
-
-        test(
-            "and should use the other source's unquoted version if "
-            'concrete', () async {
-          await setUpDependency(
-              {'git': 'git://github.com/dart-lang/foo', 'version': '0.2.3'});
-          await expectValidationWarning('  foo: 0.2.3');
-        });
-      });
-    });
```
jonasfj, line 172: I'm not sure I follow why these where valid before, and why we're removing them now.
Just curious?

jonasfj (approved):

test/validator/dependency_test.dart
sigurdm: Thanks - they should not have been removed.

Reinstated them.

We might consider how useful these checks are: https://github.com/dart-lang/pub/issues/3567

test/lish/archives_and_uploads_a_package_test.dart
sigurdm: I find self-contained tests much easier to read.

test/lish/package_creation_provides_invalid_json_test.dart
sigurdm: No we don't - for a while I thought I had to add arguments here (because I needed a custom _PUB_TEST_SDK_VERSION)

Issue #3391, opened 2022-04-14
Pub lish should run `[dart|flutter] analyze`


jakemac53:
It would be a nice extra sanity check if `pub lish` ran analysis on the project, and gave a warning if there were any diagnostics.

This would have caught a mistake I made today, where I published a version of `test_core` (after removing dependency overrides on the local version of `test_api`). I thought there were no changes in test_api that test_core depended on, and so I hadn't published a new version of that, but when I removed the override it did in fact cause some analysis errors. I didn't notice it in time, and pub didn't alert me, so I had to retract the version. Super glad I can retract versions now though!

sigurdm:
I think this is a good idea. Surprised we don't already do this.

PR #3568, opened 2022-09-19
Warn on `pub publish` if `dart analyze` has messages

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3391

### DIFF
--- a/lib/src/validator/analyze.dart
+++ b/lib/src/validator/analyze.dart
@@ -1,0 +1,28 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+
+import '../io.dart';
+
+import '../validator.dart';
+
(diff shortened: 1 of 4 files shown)


lib/src/validator/analyze.dart
```diff
@@ -0,0 +1,28 @@
...
+  @override
+  Future<void> validate() async {
+    final result = await runProcess(Platform.resolvedExecutable, [
+      'analyze',
+      '--fatal-infos',
+      if (!p.equals(entrypoint.root.dir, p.current)) entrypoint.root.dir,
+    ]);
+    if (result.exitCode != 0) {
+      warnings.add(
+          '`dart analyze` found the following issue(s):\n${result.stdout.join('\n')}');
```
jonasfj, line 25: Should we limit the amount of output?

jonasfj (approved):

lib/src/validator/analyze.dart
sigurdm: Yeah - that's a good idea!

sigurdm: Done

stuartmorgan:
This can give false-positives for packages with `example` packages currently; see https://github.com/dart-lang/pub/issues/3618

Issue #3564, opened 2022-09-14
Consider improved messaging when an SDK upgrade would unblock version solving


matanlurey:
`dart --version`: `Dart SDK version: 2.17.6 (stable) (Tue Jul 12 12:54:37 2022 +0200) on "macos_arm64"`

![](https://user-images.githubusercontent.com/168174/190037816-5f753b04-0aa6-4429-8748-28a6b97b9d90.png)

It appears like the analyzer version (v5.0.0) isn't compatible with the state of the package, or am I missing something?

**EDIT**: After blasting away my `pubspec.lock` and calling `dart pub get`:

```txt
Resolving dependencies... 
The current Dart SDK version is 2.17.6.

Because test >=1.16.1 <1.16.6 depends on analyzer >=0.39.5 <2.0.0 and test >=1.16.6 <1.17.10 depends on analyzer ^1.0.0, test >=1.16.1 <1.17.10 requires analyzer >=0.39.5 <2.0.0.
And because test >=1.17.10 <1.20.0 depends on analyzer >=1.0.0 <3.0.0 and test >=1.20.0 <1.21.0 depends on analyzer >=2.0.0 <4.0.0, test >=1.16.1 <1.21.0 requires analyzer >=0.39.5 <4.0.0.
And because test >=1.21.0 <1.21.5 depends on analyzer >=2.0.0 <5.0.0 and test >=1.21.5 requires SDK version >=2.18.0-146.0.dev <3.0.0, test >=1.16.1 requires analyzer >=0.39.5 <5.0.0.
So, because linter depends on both analyzer ^5.0.0 and test ^1.16.1, version solving failed.
```

I think @pq you might have forgot to rev the pubspec dependencies?

Should the [github action][gha] check that `pub get` works for `stable` as a reminder to rev the pubspec?

[gha]: https://github.com/dart-lang/linter/blob/main/.github/workflows/build.yaml#L36-L38

sigurdm:
The current solver-traces tell you why something doesn't work, but don't tell you what would work.

```
Resolving dependencies... 
The current Dart SDK version is 2.17.6.

Because test >=1.16.1 <1.16.6 depends on analyzer >=0.39.5 <2.0.0 and test >=1.16.6 <1.17.10 depends on analyzer ^1.0.0, test >=1.16.1 <1.17.10 requires analyzer >=0.39.5 <2.0.0.
[...]

PR #3569, opened 2022-09-20
Suggest resolution updates

sigurdm:
A supplement to the solve-failure trace.

If some relaxation to a package constraint or update of your sdk can make the resolution go through we suggest it.

```
dart pub get
Resolving dependencies... (1.2s)
Because test >=1.16.6 depends on yaml ^3.0.0 and pub depends on yaml ^1.1.0, test >=1.16.6 is forbidden.
So, because pub depends on test ^1.21.5, version solving failed.

You can try one of the following suggestion(s) to resolve the failure:
* Try upgrading your constraint on yaml: `dart pub add yaml:^3.1.1`
```

There is a timeout of max spending 3 seconds on finding suggestions to avoid degenerate cases hanging.

Fixes: https://github.com/dart-lang/pub/issues/3564

Should not land before we can actually run `dart add` with both dev and non-dev packages: https://github.com/dart-lang/pub/pull/3571

Also fixes the syntax used by `dart pub add m:'{"hosted":"<URL>"}'` to be the new syntax if the target pubspec allows it.


lib/src/entrypoint.dart
```diff
@@ -916,6 +931,113 @@ See https://dart.dev/go/sdk-constraint
...
+  Future<String> _suggestResolutionAlternatives(SolveType type,
+      Incompatibility incompatibility, Iterable<String> unlock) async {
+    final b = StringBuffer();
+    final visited = <String>{};
+    final stopwatch = Stopwatch()..start();
+    var first = true;
+    void suggest(String suggestion) {
+      if (first) {
+        b.writeln(
+            '\nYou can try one of the following suggestion(s) to resolve the failure:');
```
jonasfj, line 947: nit: Can't we make `(s)` be there or not...

lib/src/entrypoint.dart
```diff
@@ -916,6 +931,113 @@ See https://dart.dev/go/sdk-constraint
...
+              if (originalConstraint is VersionRange) {
+                final min = originalConstraint.min;
+                if (min != null) {
+                  updateDirection =
+                      resolvedVersion < min ? 'downgrading' : 'upgrading';
+                }
+              }
+
+              suggest(
+                  '* Try $updateDirection your constraint on $name: `$topLevelProgram pub add $name:$newConstraint`');
```
jonasfj, line 1030: Maybe we can make it easier to copy/paste
```suggestion
                  '* Try $updateDirection your constraint on $name: $topLevelProgram pub add $name:$newConstraint');
```

Could the message be: `dart pub upgrade --major-version $name`? I'm guessing it's not better, because it does say what version you'll be getter in the message / command, so `dart pub add` is probably better.

lib/src/entrypoint.dart
```diff
@@ -916,6 +931,113 @@ See https://dart.dev/go/sdk-constraint
...
+  /// Looks through the root-[incompability] of a solve-failure and tries to see
+  /// if any single change can make a resolution go through. Returns a formatted
+  /// list of suggestions, or the empty String if no suggestions were found.
+  Future<String> _suggestResolutionAlternatives(SolveType type,
+      Incompatibility incompatibility, Iterable<String> unlock) async {
+    final b = StringBuffer();
+    final visited = <String>{};
+    final stopwatch = Stopwatch()..start();
+    var first = true;
+    void suggest(String suggestion) {
```
jonasfj, line 944: Why not collect them in a list and show them at the end...
I'm thinking we want to prioritize them, and ideally never show more than 5 suggestions.

lib/src/entrypoint.dart
```diff
@@ -916,6 +931,113 @@ See https://dart.dev/go/sdk-constraint
...
+              });
+
+          if (runningFromFlutter) {
+            suggest(
+                '* Try using the Flutter SDK version: ${bestRelease.flutterVersion}. ');
+          } else {
+            // Here we assume that any Dart version included in a Flutter
+            // release can also be found as a released Dart SDK.
+            suggest(
+                '* Try using the Dart SDK version: ${bestRelease.dartVersion}. See https://dart.dev/get-dart.');
```
jonasfj, line 991: Can we say "consider" whenever we're suggesting a downgrade?

lib/src/pubspec_utils.dart
```diff
@@ -103,12 +103,12 @@ Future<Pubspec> constrainedToAtLeastNullSafetyPubspec(
 /// [stripOnly] will have their upper bounds removed. If [stripOnly] is
 /// not specified or empty, then all packages will have their upper bounds
 /// removed.
-Pubspec stripVersionUpperBounds(Pubspec original,
-    {Iterable<String>? stripOnly}) {
+Pubspec stripVersionBounds(Pubspec original,
```
jonasfj, line 106: Please update documentation

lib/src/entrypoint.dart
```diff
@@ -916,6 +931,113 @@ See https://dart.dev/go/sdk-constraint
...
+          final name = term.package.name;
+          if (!visited.add(name)) {
+            continue;
+          }
+          final originalConstraint =
+              (root.dependencies[name] ?? root.devDependencies[name])
+                  ?.constraint;
+          if (originalConstraint != null) {
+            final relaxedPubspec = stripVersionBounds(root.pubspec,
+                stripOnly: [name], stripLowerBound: true);
```
jonasfj, line 1008: You don't want to try to strip upper-bound before stripping lower-bound?

sigurdm: No, the heuristic is already to try the newest version first, so in general this will give you the highest version that makes the resolution go through.

lib/src/entrypoint.dart
sigurdm: Done

lib/src/entrypoint.dart
sigurdm: Yeah - that was lazy coding.

Updated.

lib/src/entrypoint.dart
sigurdm: Done

lib/src/pubspec_utils.dart
sigurdm: Done

lib/src/entrypoint.dart
```diff
@@ -936,3 +1065,9 @@ bool detectWindowsLineEndings(String text) {
   }
   return windowsNewlines > unixNewlines;
 }
+
+class _ResolutionSuggestion {
+  final String suggestion;
+  final int priority;
```
jonasfj, line 1071: * List of package names affected
 * Number of upgrades
 * Number of downgrades

lib/src/solver/failure.dart
```diff
@@ -35,14 +37,17 @@ class SolveFailure implements ApplicationException {
...
             incompatibility.terms.single.package.isRoot);
 
   /// Describes how [incompatibility] was derived, and thus why version solving
   /// failed.
   @override
-  String toString() => _Writer(incompatibility).write();
+  String toString() => [
+        _Writer(incompatibility).write(),
+        if (suggestions != null) suggestions
+      ].join();
```
jonasfj, line 50: Are you sure you don't want:
```suggestion
      ].join('\n');
```

lib/src/solver/version_solver.dart
```diff
@@ -75,11 +75,21 @@ class VersionSolver {
   /// The set of packages for which the lockfile should be ignored.
   final Set<String> _unlock;
 
+  /// If present these represents the version of an SDK to assume during
+  /// resolution.
+  final Map<String, Version>? _sdkOverrides;
```
jonasfj, line 80:
```suggestion
  final Map<String, Version> _sdkOverrides;
```

lib/src/solver/version_solver.dart
```diff
@@ -75,11 +75,21 @@ class VersionSolver {
...
-  VersionSolver(this._type, this._systemCache, this._root, this._lockFile,
-      Iterable<String> unlock)
-      : _dependencyOverrides = _root.dependencyOverrides,
+  VersionSolver(
+    this._type,
+    this._systemCache,
+    this._root,
+    this._lockFile,
+    Iterable<String> unlock, {
+    Map<String, Version>? sdkOverrides,
```
jonasfj, line 90:
```suggestion
    Map<String, Version> sdkOverrides = const {},
```

lib/src/solver/solve_suggestions.dart
```diff
@@ -0,0 +1,216 @@
...
+import 'incompatibility_cause.dart';
+
+/// Looks through the root-[incompability] of a solve-failure and tries to see
+/// * if an update of the current sdk can make a resolution go through.
+/// * if any single change to a package constraint can make a resolution go
+///   through.
+/// * If removing the bounds on all constraints can make a resolution go through
+///   updating less than 5 dependencies.
+/// * If running `pub upgrade --major versions` would make the resolution go
+///   through.
```
jonasfj, line 23:
```suggestion
/// Looks through the root-[incompability] of a solve-failure and tries to see if
/// the conflict could resolved by any of the following suggestions:
/// * An update of the current SDK.
/// * Any single change to a package constraint.
/// * Removing the bounds on all constraints, changing less than 5 dependencies.
/// * Running `pub upgrade --major versions`.
```

lib/src/entrypoint.dart

lib/src/solver/package_lister.dart
```diff
@@ -97,17 +99,20 @@ class PackageLister {
...
-      {bool downgrade = false})
-      : _isDowngrade = downgrade;
+    this._systemCache,
+    this._ref,
+    this._locked,
+    this._dependencyType,
+    this._overriddenPackages,
+    this._allowedRetractedVersion, {
+    bool downgrade = false,
+    required Map<String, Version>? sdkOverrides,
```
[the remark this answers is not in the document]

lib/src/flutter_releases.dart
```diff
@@ -0,0 +1,99 @@
...
+
+String get flutterReleasesUrl =>
+    Platform.environment['_PUB_TEST_FLUTTER_RELEASES_URL'] ??
+    'https://storage.googleapis.com/flutter_infra_release/releases/releases_linux.json';
+
+// Retrieves all released versions of Flutter.
+Future<List<FlutterRelease>> _flutterReleases = () async {
+  final response = await retryForHttp(
+    'fetching available Flutter releases',
+    () => globalHttpClient.get(Uri.parse(flutterReleasesUrl)),
```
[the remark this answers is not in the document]
sigurdm: Done

lib/src/solver/solve_suggestions.dart
```diff
@@ -0,0 +1,214 @@
...
+    return null;
+  }
+}
+
+String packageAddDescription(Entrypoint entrypoint, PackageId id) {
+  final name = id.name;
+  final isDev = entrypoint.root.pubspec.devDependencies.containsKey(name);
+  final constraint = VersionConstraint.compatibleWith(id.version);
+  final devPart = isDev ? 'dev:' : '';
+  return '$devPart$name:$constraint';
```
jonasfj, line 213: This won't work if it's not a hosted dependency from the default server.
Should we make some check and avoid proposing suggestions for other than that kind of packages?

lib/src/solver/solve_suggestions.dart
```diff
@@ -0,0 +1,214 @@
...
+            entrypoint.root.devDependencies[name])
+        ?.constraint;
+    if (originalConstraint != null) {
+      final relaxedPubspec = stripVersionBounds(entrypoint.root.pubspec,
+          stripOnly: [name], stripLowerBound: true);
+
+      final result = await tryResolve(
+          type, cache, Package.inMemory(relaxedPubspec),
+          lockFile: entrypoint.lockFile, unlock: unlock);
+      if (result != null) {
```
jonasfj, line 77:
```suggestion
      if (result == null) {
        return;
      }
```
Then we get less indentation...

lib/src/solver/solve_suggestions.dart
```diff
@@ -0,0 +1,214 @@
...
+  Future<_ResolutionSuggestion?> suggestUnlockingAll(
+      {required bool stripLowerBound}) async {
+    final originalPubspec = entrypoint.root.pubspec;
+    final relaxedPubspec =
+        stripVersionBounds(originalPubspec, stripLowerBound: stripLowerBound);
+
+    final result = await tryResolve(
+        type, cache, Package.inMemory(relaxedPubspec),
+        lockFile: entrypoint.lockFile, unlock: unlock);
+    if (result != null) {
```
jonasfj, line 117:
```suggestion
      if (result == null) {
        return;
      }
```
Then we get less indentation...

lib/src/solver/solve_suggestions.dart
```diff
@@ -0,0 +1,214 @@
...
+      }
+    }
+    return null;
+  }
+
+  Future<_ResolutionSuggestion?> suggestUnlockingAll(
+      {required bool stripLowerBound}) async {
+    final originalPubspec = entrypoint.root.pubspec;
+    final relaxedPubspec =
+        stripVersionBounds(originalPubspec, stripLowerBound: stripLowerBound);
```
jonasfj, line 112: Should we strip stuff for git-dependencies and hosted dependencies that have a URL?

lib/src/solver/solve_suggestions.dart
```diff
@@ -0,0 +1,214 @@
...
+/// Returns a formatted list of suggestions, or the empty String if no
+/// suggestions were found.
+Future<String?> suggestResolutionAlternatives(
+  Entrypoint entrypoint,
+  SolveType type,
+  Incompatibility incompatibility,
+  Iterable<String> unlock,
+  SystemCache cache,
+) async {
+  Future<_ResolutionSuggestion?> suggestSdkUpdate(SdkCause cause) async {
```
jonasfj, line 32: Should we not make these top-level functions, and document what they do?

lib/src/solver/solve_suggestions.dart
```diff
@@ -0,0 +1,214 @@
...
+        if (!visited.add(name)) {
+          continue;
+        }
+        addSuggestion(await suggestSinglePackageUpdate(name));
+      }
+    }
+  }
+  if (suggestions.isEmpty) {
+    addSuggestion(await suggestUnlockingAll(stripLowerBound: true) ??
+        await suggestUnlockingAll(stripLowerBound: false));
```
jonasfj, line 177: Shouldn't this be the other way around?

lib/src/solver/solve_suggestions.dart
jonasfj: We could create a:
```dart

class _FailedResolutionContext {
  final Entrypoint entrypoint;
  final SolveType type;
  final Incompatibility incompatibility;
  final Iterable<String> unlock;
  final SystemCache cache;
}

```

So we don't have to repeat all the arguments...

lib/src/solver/solve_suggestions.dart
```diff
@@ -0,0 +1,214 @@
...
+        final name = term.package.name;
+
+        if (!visited.add(name)) {
+          continue;
+        }
+        addSuggestion(await suggestSinglePackageUpdate(name));
+      }
+    }
+  }
+  if (suggestions.isEmpty) {
```
jonasfj, line 175: The reason we don't want to suggest `upgrade --major-versions` is because we have a smaller suggestion, right?

Like if we look at the suggestions, all suggestions are:
 * A list of package names for which the constraints are tweaked
 * For each package where the constraints are tweaked:
    * Did we upgrade / downgrade?
    * Did we jump a major version or not?
    * How many versions did we jump?

Could we describe the size of a suggestion somehow? Then make the logic on whether include a suggestion more generic.

I think the space of all suggestions can be considered a partially-ordered-set, but if we want to do that we might have to think a bit :rofl:  maybe it's overkill.

lib/src/solver/solve_suggestions.dart
sigurdm: I instead added code, so we can update packages from any server

Also I now skip non-hosted dependencies, as they will never resolve anyway.

lib/src/solver/solve_suggestions.dart
sigurdm: Done

lib/src/solver/solve_suggestions.dart
sigurdm: This will only touch their version constraints - this is not going to make any difference for non-hosted dependencies (they usually have no version constraint, and if they do, stripping them will not make any difference to the resolution, as the source will not return multiple versions anyways)

lib/src/solver/solve_suggestions.dart
sigurdm: We first attempt stripping only the upper bound, if that fails we relax further and strip the lower bound as well...

lib/src/solver/solve_suggestions.dart
sigurdm: Yeah... that might make sense... still I'd like not having to compute the --major versions if we found an upgrade to a single package, however that would be required to do the compare...

lib/src/solver/solve_suggestions.dart
sigurdm: Did that, moved the local functions to be methods on the context.

I keep the incompatibility out - I have plans to also make suggestions for `dart pub upgrade` for the case where things resolve, but you don't end up on the latest.

sigurdm:
PTAL

lib/src/solver/solve_suggestions.dart
sigurdm: Done

lib/src/command/add.dart
```diff
@@ -563,6 +564,9 @@ Specify multiple sdk packages with descriptors.''');
             {
               'dependencies': {
                 packageName: parsedDescriptor,
+              },
+              'environment': {
+                'sdk': sdk.version.toString(),
```
jonasfj, line 569: This is an entirely unrelated fix, right?

test/sdk_test.dart
```diff
@@ -127,10 +127,8 @@ void main() {
         await pubCommand(
           command,
           environment: {'FLUTTER_ROOT': p.join(d.sandbox, 'flutter')},
-          error: equalsIgnoringWhitespace("""
-              Because myapp depends on foo ^1.0.0 from sdk which doesn't match
-                any versions, version solving failed.
-            """),
+          error: contains('''
+Because myapp depends on foo ^1.0.0 from sdk which doesn't match any versions, version solving failed.'''),
```
jonasfj, line 131: Why these changes, what is wrong with `equalsIgnoringWhitespace` isn't that more robust?

jonasfj (approved):

lib/src/command/add.dart
sigurdm: Well - it was discovered because the suggestion of
```
dart pub add bar:'{"version":"^2.0.0","hosted":"http://localhost:44091"}
```
would otherwise fail because it uses a new-style hosted url...

test/sdk_test.dart
sigurdm: Well - now these output more, so we are only recognizing a substring.

We should really have a `containsIgnoringWhitespace`...

Let's create that another day.

Issue #3324, opened 2022-03-03
Allow `pub add` to install multiple packages


exaby73:
Taking inspiration from npm and yarn, this feature would allow the `pub add` AKA `dart pub add` command to  install multiple packages by taking multiple packages as arguments. Current to install multiple packages, we have to do this:

```bash
dart pub add package_a
dart pub add package_b
dart pub add package_c
```

Allowing multiple packages to be installed would shorten it to this:

```bash
dart pub add package_a package_b package_c
```

sigurdm:
We have just implemented this recently.

Closing as duplicate of https://github.com/dart-lang/pub/issues/3273.

Issue #3273, opened 2022-01-04
Support multiple packages in `dart pub add ...`


DanTup:
Noted at https://github.com/Dart-Code/Dart-Code/issues/3737#issuecomment-1000548925, it would be neat to be able to provide a set of packages to add in one step, since running `dart pub add` multiple times takes much longer.

It would be really useful for copy/pasting from articles, both into the CLI or into VS Code's **Dart: Add Dependency** input box.

sigurdm:
This was discussed in https://github.com/dart-lang/pub/issues/3213 we closed as wont-fix.

Maybe we should reconsider...

We could allow several names, but only allow them from one source (basically would only work for hosted)
```
dart pub add foo bar baz # works, comes from default host
dart pub add foo bar baz --hosted https://package-repo.com # works
dart pub add foo bar baz --git https://git.com # only one package per git repo (fails)
```

That would cover the common usecase.

If anyone is interested in contributing I think we would accept something like that.

DanTup:
Restricting it to one source is probably reasonable for me (it'd be nice to not have to, but I suspect > 99% of users from VS Code would just want a few names from Pub).

If someone were to contribute it (not promising, but in case I do find time 🙂), am I right in thinking it would start with removing this:

https://github.com/dart-lang/pub/blob/17c5addbbf93c4462dc0dfeefcd43b48d713b1c1/lib/src/command/add.dart#L94-L95

And then running this part for each package (passing `updatedPubSpec` back in for subsequent calls):

https://github.com/dart-lang/pub/blob/17c5addbbf93c4462dc0dfeefcd43b48d713b1c1/lib/src/command/add.dart#L101-L103

Before then running the normal solving code below, and then the checks that is satisfies the requirements below also for each package?

sigurdm:
Yes - I think that looks like the right strategy!

ArV29:
I would like to work towards adding this feature if no one else has already started

DanTup:
I haven't started (and probably wouldn't get to in the next few weeks), so that sounds good to me :-)

Please CC me if you do file a PR, as I may be interested in testing it for VS Code integration prior to it being out in a release to get a headstart :-)

ArV29:
Sure,
This is going to be my first real contribution btw so I would probably need a bit of guidance initially

ArV29:
> I haven't started (and probably wouldn't get to in the next few weeks), so that sounds good to me :-)
> 
[...]

PR #3571, opened 2022-09-23
Allow adding from multiple sources

sigurdm:
And having some packages be dev_dependencies while others are not.
By having:
* each package carry its own (yaml-formatted) descriptor as it would be written in pubspec.yaml. 
* `dev:` be a prefix to the package instead of a top-level argument.

This allows eg.:

```
dart pub add foo:^1.2.3  'dev:bar:{"git":"../bar.git"}
```

This is a more full resolution to https://github.com/dart-lang/pub/issues/3273 than https://github.com/dart-lang/pub/issues/3324 but implemented in a backwards compatible way (the old arguments still work as long as they are not combined with descriptors (other than bare constraints)).

If a constraint is not given explicitly, either as `foo:^1.2.3` or `foo:{<source>:<descriptor>,"version":"<constraint"}` the "best" constraint is inferred and inserted as before (instead of being interpreted as 'any' as it would in a pubspec.yaml).


sigurdm:
The argument parsing I do in this PR is not really for the faint of heart. I believe it works, but would love to do this in a nicer way if possible...

sigurdm:
Update: now the arguments are not separated by `--` but instead based on descriptors. This makes the argument parsing more conventional :) .

test/add/path/relative_path_test.dart
```diff
@@ -135,4 +135,39 @@ void main() {
...
+    await d.appDir({}).create();
+
+    await pubAdd(
+      args: [
+        '--directory',
+        appPath,
+        'foo',
+        '--path',
+        'foo',
+        '--',
```
jonasfj, line 154: Are we still doing this?

sigurdm:
No - sorry I forgot to push before updating the description.

lib/src/command/add.dart
```diff
@@ -39,31 +39,18 @@ class AddCommand extends PubCommand {
...
 
   AddCommand() {
+    argParser.addSeparator('Package options');
+
     argParser.addFlag('dev',
         abbr: 'd',
         negatable: false,
-        help: 'Adds to the development dependencies instead.');
+        help: 'Adds to the development dependencies instead.',
+        hide: true);
```
sigurdm, line 53: unhide

lib/src/command/add.dart
```diff
@@ -39,31 +39,18 @@ class AddCommand extends PubCommand {
...
-  String? get path => argResults['path'];
-  String? get sdk => argResults['sdk'];
-
-  bool get hasGitOptions => gitUrl != null || gitRef != null || gitPath != null;
-  bool get hasHostOptions => hostUrl != null;
-
-  bool get isHosted => !hasGitOptions && path == null && path == null;
 
   AddCommand() {
+    argParser.addSeparator('Package options');
```
sigurdm, line 47: Remove this

lib/src/command/add.dart
```diff
@@ -94,20 +76,56 @@ class AddCommand extends PubCommand {
...
+    argParser.addFlag(
+      'example',
+      help:
+          'Also update dependencies in `example/` after modifying pubspec.yaml in the root package (if it exists).',
+      hide: true,
+    );
   }
 
+  /// Parses the arguments that are allowed after the first "--".
+  late final followingPartsArgParser = () {
```
sigurdm, line 88: Remove this

lib/src/command/add.dart
```diff
@@ -285,52 +308,71 @@ class AddCommand extends PubCommand {
...
-      final conflictingFlag = conflictingFlagSets
-          .where((s) => !s.contains(flag))
-          .expand((s) => s)
-          .firstWhereOrNull(argResults.wasParsed);
-      if (conflictingFlag != null) {
-        usageException(
-            'Packages can only have one source, "pub add" flags "--$flag" and '
-            '"--$conflictingFlag" are conflicting.');
-      }
+  _ParseResult _parsePackage(
```
szakarias, line 311: Update documentation.

It would be great to have examples for both the old and new supported way.

lib/src/command/add.dart
```diff
@@ -285,52 +308,71 @@ class AddCommand extends PubCommand {
...
+    PackageRef? ref;
+
+    if (descriptor != null) {
+      try {
+        // An unquoted version constraint is not always valid yaml.
+        // But we want to allow it here anyways.
+        constraint = VersionConstraint.parse(descriptor);
+      } on FormatException {
+        final parsedDescriptor = loadYaml(descriptor);
+        // Use the pubspec parsing mechanism for persing the descriptor.
```
szakarias, line 339: persing -> parsing

lib/src/command/add.dart
```diff
@@ -285,52 +308,71 @@ class AddCommand extends PubCommand {
...
+        final Pubspec dummyPubspec;
+        try {
+          dummyPubspec = Pubspec.fromMap({
+            'dependencies': {
+              packageName: parsedDescriptor,
+            }
+          }, cache.sources,
+              // Resolve relative paths relative to current, not where the pubspec-yaml is.
```
szakarias, line 347: nit: pubspec-yaml -> pubspec.yaml

lib/src/command/add.dart
```diff
@@ -39,31 +39,18 @@ class AddCommand extends PubCommand {
   String get description => 'Add dependencies to pubspec.yaml.';
```
jonasfj, line 39:
```suggestion
  String get description => '''
Add dependencies to `pubspec.yaml`.

Invoking `dart pub add foo bar` will add `foo` and `bar` to `pubspec.yaml`
with a default constraint derived from latest compatible version.
''';
```

`Command.summary` will be derived from the first-line, otherwise if we want to customize the summary, we can override it.

But I think a command with a usage this complicated will benefit greatly from examples in `--help`. Especially, since most users will be interested in doing `dart pub add <package>` or `dart pub add <package>:<constraint>`

lib/src/command/add.dart
```diff
@@ -94,20 +76,56 @@ class AddCommand extends PubCommand {
...
+        help: 'Git branch or commit to be retrieved');
+    argParser.addOption('git-path', help: 'Path of git package in repository');
+    argParser.addOption('hosted-url', help: 'URL of package host server');
+    argParser.addOption('path', help: 'Add package from local path');
+    argParser.addOption(
+      'sdk',
+      help: 'add package from SDK source',
+      allowed: ['flutter'],
+      valueHelp: '[flutter]',
+    );
```
jonasfj, line 108: Should we make all of these options hidden and deprecated?

lib/src/command/add.dart
```diff
@@ -39,31 +39,18 @@ class AddCommand extends PubCommand {
   String get description => 'Add dependencies to pubspec.yaml.';
   @override
   String get argumentsDescription =>
-      '<package>[:<constraint>] [<package2>[:<constraint2>]...] [options]';
+      '[options] ([dev:]<package>[:descriptor])+';
```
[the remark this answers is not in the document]
sigurdm: Done

lib/src/command/add.dart
sigurdm: Done

lib/src/command/add.dart
sigurdm: Done

lib/src/command/add.dart
sigurdm: Done

lib/src/command/add.dart
sigurdm: Done

lib/src/command/add.dart
sigurdm: Done

lib/src/command/add.dart
sigurdm: Done

lib/src/command/add.dart
```diff
@@ -36,49 +37,77 @@ class AddCommand extends PubCommand {
...
-      help:
-          'Also update dependencies in `example/` after modifying pubspec.yaml in the root package (if it exists).',
+      'dev',
+      abbr: 'd',
+      negatable: false,
+      help: 'Adds to the development dependencies instead.',
+      hide: true,
+    );
+
+    argParser.addOption(
```
jonasfj, line 81: All of these options need a comment explaining why we're hid them.
```suggestion
    // Following options are hidden/deprecated in favor of the new syntax: [dev:]<package>[:descriptor] ...
    // To avoid breaking changes we keep supporting them, but hide them from --help to discourage
    // further use. Combining these with new syntax will fail.
    argParser.addOption(
```

lib/src/command/add.dart
```diff
@@ -259,24 +307,33 @@ class AddCommand extends PubCommand {
...
   /// 'retry:>2.0.0 <3.0.1'
   /// retry:any
+  /// 'retry:{"path":"../foo"}'
+  /// 'retry:{"git":{"url":"../foo","ref":"branchname"},"version":"^1.2.3"}'
+  /// 'retry:{"sdk":"flutter"}'
+  /// 'retry:{"hosted":"mypub.dev"}'
   /// ```
   ///
+  /// The --path --sdk and --git-<option> arguments cannot be combined with a
+  /// non-string descriptor.
```
jonasfj, line 335: What about splitting this into two methods?
 * One for parsing when any of the options are given `_parsePackageAndConstraint` (which is deprecated)
 * One for parsing package and descriptor `_parsePackageAndDescriptor` (for the new syntax)

That way you can keep the old logic in `_parsePackageAndConstraint` and only add new logic in `_parsePackageAndDescriptor`, and you just look for options to decide which one to use.

lib/src/command.dart
```diff
@@ -87,11 +88,15 @@ abstract class PubCommand extends Command<int> {
...
+
   // Lazily initialize the parser because the superclass constructor requires
   // it but we want to initialize it based on [allowTrailingOptions].
-  @override
-  late final ArgParser argParser = ArgParser(
-      allowTrailingOptions: allowTrailingOptions, usageLineLength: lineLength);
+  late final ArgParser _argParser = ArgParser(
+    allowTrailingOptions: allowTrailingOptions,
+    usageLineLength: lineLength,
+  );
```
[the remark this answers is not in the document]
sigurdm: This is not needed. Reverted.

lib/src/command.dart
```diff
@@ -57,9 +57,10 @@ abstract class PubCommand extends Command<int> {
 
   String get directory =>
       (argResults.options.contains('directory')
-          ? argResults['directory']
+          ? argResults['directory'] as String
           : null) ??
-      _pubTopLevel.directory;
+      _pubTopLevel.directory ??
+      p.current;
```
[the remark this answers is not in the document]
sigurdm: Moved this to separate PR.

lib/src/command/add.dart
```diff
@@ -94,20 +123,34 @@ class AddCommand extends PubCommand {
...
+
+    if (argResults.rest.length > 1 && argResults.gitUrl != null) {
+      usageException('''
+--git-url cannot be used with multiple packages.
+Specify multiple git packages with descriptors.''');
+    } else if (argResults.rest.length > 1 && argResults.path != null) {
+      usageException('''
+--path cannot be used with multiple packages.
+Specify multiple path packages with descriptors.''');
+    }
```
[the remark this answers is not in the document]
sigurdm: Added check for sdk.

lib/src/command/add.dart
```diff
@@ -259,12 +336,12 @@ class AddCommand extends PubCommand {
...
 
-  /// Parse [package] to return the corresponding [PackageRange], as well as its
-  /// representation in `pubspec.yaml`.
+  /// Parse [package] to return the corresponding [_ParseResult] using the
+  /// arguments given in [argResults] to configure the description.
   ///
   /// [package] must be written in the format
-  /// `<package-name>[:<version-constraint>]`, where quotations should be used
-  /// if necessary.
+  /// `[dev:]<package-name>[:<version-constraint>]`, where quotations should be
```
jonasfj, line 343:
```suggestion
  /// `<package-name>[:<version-constraint>]`, where quotations should be
```

You've already parse `dev:` when you get here.

lib/src/command/add.dart
```diff
@@ -349,37 +432,137 @@ class AddCommand extends PubCommand {
...
       );
     }
-    return _ParseResult(ref, constraint);
+    return _PartialParseResult(ref, constraint);
+  }
+
+  /// Parse [package] to return the corresponding [_ParseResult].
+  ///
+  /// [package] must be written in the format
+  /// `dev:<package-name>[:descriptor>]`, where quotations should be used if
```
jonasfj, line 459:
```suggestion
  /// `<package-name>[:descriptor>]`, where quotations should be used if
```

You've already handled this.

lib/src/command/add.dart
```diff
@@ -349,37 +432,137 @@ class AddCommand extends PubCommand {
...
+  ///
+  /// Packages must either be a git, hosted, sdk, or path package. Mixing of
+  /// options is not allowed and will cause a [PackageParseError] to be thrown.
+  ///
+  /// If any of the other git options are defined when `--git-url` is not
+  /// defined, an error will be thrown.
+  ///
+  /// Returns a `ref` of `null` if the descriptor did not specify a source.
+  /// Then the source will be determined by the old-style arguments.
+  _PartialParseResult _parsePackage(String package, ArgResults argResults) {
```
jonasfj, line 498:
```suggestion
  _PartialParseResult _parsePackage(String packageDescriptorArgument) {
```
`argResults` is not used, you handle `--dev` before this.

lib/src/command/add.dart
```diff
@@ -349,37 +432,137 @@ class AddCommand extends PubCommand {
...
+  ///
+  /// [package] must be written in the format
+  /// `dev:<package-name>[:descriptor>]`, where quotations should be used if
+  /// necessary.
+  ///
+  /// `descriptor` is what you would put in a pubspec.yaml in the dependencies
+  /// section.
+  ///
+  /// Assumes that none of '--git-url', '--git-ref', '--git-path', '--path' and
+  /// '--sdk' are present in [argResults], but will obey '--dev'.
```
jonasfj, line 466: Lies: ` but will obey '--dev'` this has already been parsed.
```suggestion
  /// '--sdk' are present in [argResults].
```

lib/src/command/add.dart
```diff
@@ -349,37 +432,137 @@ class AddCommand extends PubCommand {
...
+    final descriptor = match.namedGroup('descriptor')?.substring(1);
+
+    /// We want to allow for [constraint] to take on a `null` value here to
+    /// preserve the fact that the user did not specify a constraint.
+    VersionConstraint? constraint;
+
+    /// The package to be added.
+    PackageRef? ref;
+
+    if (descriptor != null) {
```
jonasfj, line 515: `descriptor` is optional, right, so we just throw away `packageName` if no `descriptor` is given?

We probably need a test case that covers this. Or maybe this is parsed as legacy, I just think it might be worth supporting here for completeness.

Or maybe `descriptor` is always empty string...? In which case we should assert it with `!`.

lib/src/command/add.dart
```diff
@@ -349,37 +432,137 @@ class AddCommand extends PubCommand {
...
+  /// options is not allowed and will cause a [PackageParseError] to be thrown.
+  ///
+  /// If any of the other git options are defined when `--git-url` is not
+  /// defined, an error will be thrown.
+  ///
+  /// Returns a `ref` of `null` if the descriptor did not specify a source.
+  /// Then the source will be determined by the old-style arguments.
+  _PartialParseResult _parsePackage(String package, ArgResults argResults) {
+    final match =
+        RegExp(r'^(?<name>[^:]*)(?<descriptor>:.*)?$').firstMatch(package);
```
jonasfj, line 500:
```suggestion
        RegExp(r'^(?<name>[^:]*)(?::(?<descriptor>.*))?$').firstMatch(package);
```
Then you won't have to do: `match.namedGroup('descriptor')?.substring(1);` just do `match.namedGroup('descriptor');`

lib/src/command/add.dart
```diff
@@ -349,37 +432,137 @@ class AddCommand extends PubCommand {
...
+        // Use the pubspec parsing mechanism for parsing the descriptor.
+        final Pubspec dummyPubspec;
+        try {
+          dummyPubspec = Pubspec.fromMap({
+            'dependencies': {
+              packageName: parsedDescriptor,
+            }
+          }, cache.sources,
+              // Resolve relative paths relative to current, not where the pubspec.yaml is.
+              location: p.toUri(p.join(p.current, 'descriptor')));
```
jonasfj, line 531: Really? I don' t know the answer, just asking :rofl:

lib/src/command/add.dart
```diff
@@ -349,37 +432,137 @@ class AddCommand extends PubCommand {
...
+        } on FormatException catch (e) {
+          usageException('Failed parsing package specification: ${e.message}');
+        }
+        final range = dummyPubspec.dependencies[packageName]!;
+        if (parsedDescriptor is String) {
+          // Ref will be constructed by `_parsePackageOldStyleArgs()`
```
jonasfj, line 537: Shouldn't this only be used in legacy mode?

lib/src/command/add.dart
```diff
@@ -94,20 +130,56 @@ class AddCommand extends PubCommand {
...
+      if (p.startsWith('dev:')) {
+        if (argResults.isDev) {
+          usageException("Cannot combine 'dev:' with --dev");
+        }
+        isDev = true;
+        p = p.substring('dev:'.length);
+      }
+      var partial = _parsePackage(p, argResults);
+      if (partial.ref == null) {
+        // No yaml-style descriptor was given, parse old-style arguments.
```
jonasfj, line 173: Doing `dart pub add foo` isn't not old style is it?

It's still recommended behavior, right?

I guess it's an implementation detail in here, it just seems like it would be wise to switch between `_parsePackage` and `_parsePackageOldStyleArgs` and not use both in combination.

Like, maybe we do:
```dart
if (argResults.hasOldStyleOptions) {
  partial = _parsePackageOldStyleArgs(p, argResults);
} else {
  partial = _parsePackage(p); // doesn't need argResults
}
```

szakarias (approved):
LGTM

lib/src/command/add.dart
sigurdm: Yeah - that's better.

jonasfj (approved):

lib/src/command/add.dart
sigurdm: Removed

lib/src/command/add.dart
sigurdm: I think this is right. We need a `file name`, because the pubspec parsing `location` parameter expects the path to a file name, not the containing directory.

lib/src/command/add.dart
sigurdm: It fell into the old style parsing, to allow for old-style arguments to be passed along.

Restructured the flow...

lib/src/command/add.dart
```diff
@@ -534,7 +550,7 @@ Specify multiple sdk packages with descriptors.''');
         }
         final range = dummyPubspec.dependencies[packageName]!;
         if (parsedDescriptor is String) {
-          // Ref will be parsed by `_parsePackageOldStyleArgs()`
+          // Ref will be constructed by `_parsePackageOldStyleArgs()`
```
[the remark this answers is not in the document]
sigurdm: Yeah - outdated comment.

lib/src/command/add.dart
```diff
@@ -31,54 +32,93 @@ import '../utils.dart';
...
+  * Add a path dependency:
+    `$topLevelProgram pub add 'foo{"path":"../foo"}'`
+  * Add a hosted dependency:
+    `$topLevelProgram pub add 'foo{"hosted":"my-pub.dev"}'`
+  * Add an sdk dependency:
+    `$topLevelProgram pub add 'foo{"sdk":"flutter"}'`
+  * Add a git dependency:
+    `$topLevelProgram pub add 'foo{"git":"https://github.com/foo/foo"}'`
+  * Add a git dependency with a path and ref specified:
+    `$topLevelProgram pub add 'foo{"git":{"url":"../foo.git","ref":"branch","path":"subdir"}}'`''';
```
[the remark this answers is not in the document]
sigurdm: Yeah - let's try that.

lib/src/command/add.dart
```diff
@@ -645,10 +670,11 @@ extension on ArgResults {
   String? get gitUrl => this['git-url'];
   String? get gitPath => this['git-path'];
   String? get gitRef => this['git-ref'];
-  String? get hostUrl => this['hosted-url'];
+  String? get hostedUrl => this['hosted-url'];
   String? get path => this['path'];
   String? get sdk => this['sdk'];
-  bool get hasOldStyleOptions => hasGitOptions || path != null || sdk != null;
+  bool get hasOldStyleOptions =>
+      hasGitOptions || path != null || sdk != null || hostedUrl != null;
```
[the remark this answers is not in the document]
sigurdm: Hmm, yeah, let's put it there.

lib/src/command/add.dart
```diff
@@ -349,37 +454,141 @@ class AddCommand extends PubCommand {
...
+  /// retry:2.0.0
+  /// dev:retry:^2.0.0
+  /// retry:'>=2.0.0'
+  /// retry:'>2.0.0 <3.0.1'
+  /// 'retry:>2.0.0 <3.0.1'
+  /// retry:any
+  /// 'retry:{"path":"../foo"}'
+  /// 'retry:{"git":{"url":"../foo","ref":"branchname"},"version":"^1.2.3"}'
+  /// 'retry:{"sdk":"flutter"}'
+  /// 'retry:{"hosted":"mypub.dev"}'
```
MaryaBelanger, line 503: @sigurdm Is the syntax `package:{descriptor}` or `package{descriptor}`?

These lines indicate it's `package:{descriptor}`, but the examples in the help text show `package{descriptor}` (no colon between package and descriptor)

Are both allowed?

parlough: I believe the `:` is required from my local testing and the help message mentions "by giving a descriptor after a colon".

I also noticed the help message includes single quotes (`'`) around the `package:{descriptor}` syntax which seems to not work for me. Double quotes do work (`"`) though.

parlough: I've opened an issue at https://github.com/dart-lang/pub/issues/3749

sigurdm: Yeah - you need the colon - thanks for spotting this!

Issue #3575, opened 2022-09-25
Global executable from path source runs twice on first run on Windows 


blaugold:
# Environment

- Dart version (run `dart --version`): 2.18.1
- OS kind and version (e.g. "Windows 10, version 1809" or "macOS 12.4"): Microsoft Windows Server 2022 
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

When an executable that has been installed with `dart pub global activate --verbose --source path .` is executed for the first time on Windows, it is executed twice.

This [repository](https://github.com/blaugold/pub_global_executable_issue) contains a reproduction with the following executable script and a [GitHub Action workflow](https://github.com/blaugold/pub_global_executable_issue/blob/main/.github/workflows/ci.yaml):

```dart
// bin/foo.dart

import 'dart:io';

void main(List<String> arguments) {
  print('Running foo');
  sleep(Duration(seconds: 5));
}
```

This is the output of running `foo --verbose` for the fist time:

```
$: foo --verbose
FINE: Pub 2.18.1
IO  : Created temp directory D:\a\pub_global_pwsh\pub_global_pwsh\.dart_tool/pub\incremental\foo\tmp340e[8](https://github.com/blaugold/pub_global_pwsh/actions/runs/3122206521/jobs/5063995252#step:5:9)672
MSG : Building package executable...
MSG : Built foo:foo.
IO  : Renaming "D:\a\pub_global_pwsh\pub_global_pwsh\.dart_tool/pub\incremental\foo\tmp340e8672\foo.dart.incremental.dill.incremental.dill" to "D:\a\pub_global_pwsh\pub_global_pwsh\.dart_tool/pub\bin\foo\foo.dart-2.18.1.snapshot".
IO  : Deleting directory D:\a\pub_global_pwsh\pub_global_pwsh\.dart_tool/pub\incremental\foo\tmp340e8672.
FINE: Building package executable finished (0.3[11](https://github.com/blaugold/pub_global_pwsh/actions/runs/3122206521/jobs/5063995252#step:5:12)s).
FINE: Replacing old binstub C:\Users\runneradmin\.pub-cache\bin\foo.bat
IO  : Deleting file C:\Users\runneradmin\.pub-cache\bin\foo.bat.
IO  : Created temp directory C:\Users\runneradmin\.pub-cache\_temp\dir1765ba3e
IO  : Writing 630 characters to text file C:\Users\runneradmin\.pub-cache\bin\foo.bat.
FINE: Contents:
[...]

PR #3586, opened 2022-10-05
Make binstubs offset invariant

sigurdm:
After realizing that batch-files pick up changes while they are running: https://stackoverflow.com/questions/906586/changing-a-batch-file-when-its-running it seems that it is prudent to avoid changing the offsets inside the file when it is regenerated.

This pr changes two things:
* Always generate the full logic in the binstub (it will check for snapshot presence anyways)
* For windows we generate some padding after the snapshot path, such that changing the path will not affect indices.

This hopefully fixes: https://github.com/dart-lang/pub/issues/3575


jonasfj (approved):
So we always pad to 200 bytes in size...

blaugold:
@sigurdm Thanks! It's looking good. With the fix, the executable is only executed once.

Issue #3130, opened 2021-09-20
Sanitize output from handleJsonSuccess / handleJsonError / invalidServerResponse


jonasfj:
In `http.dart` we have few function that dumps output from the package repository server to terminal:
https://github.com/dart-lang/pub/blob/570cb28cdc68db2f153160d3559ad9c8d00c5752/lib/src/http.dart#L304-L346

It would probably be wise to apply some sanitizing to this output, similar to what @themisir did for `message="..."` in `www-authenticate`.

IMO, we should break message sanitizing logic into a utility function and use when printing output from a server.

In particular I think it's unreasonable to allow servers to print ANSI escape codes, and such... maybe a few newlines, but not too many. And not too long messages.

I haven't check if there is anything weird you can do here, I'm just imagining there could be...

jonasfj:
Playing around with this a bit it seems ANSI escape codes can be encoded JSON:
```dart
import 'dart:convert';

void main() {
  print('# Print test in bold:');
  print('\u001b[1m test \u001b[22m');
  print('');

  print('# Encode as JSON:');
  final j = json.encode({'msg': '\u001b[1m test \u001b[22m'});
  print(j);
  print('');

  print('# Print decoded JSON:');
  print(json.decode(j));
  print('');

  print('# Encode as utf8 and print decoded utf-8:');
  final b = utf8.encode(j);
  print(utf8.decode(b));
  print('');

  print('# Encode as utf8 and print decoded utf-8 and JSON:');
  print(json.decode(utf8.decode(b)));
}
```

We probably shouldn't allow package repositories to send ANSI escape codes to the terminal.

jonasfj:
When looking at this we should also clearify when to use:
 * `www-authenticate` header, or,
 * [JSON response error body](https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md#rejecting-requests)...

Issue #3325, opened 2022-03-03
Refactor pub http retry behavior


sigurdm:
* timeouts on dns lookups (if possible)
* timeouts on http requests
=> timeout 30 seconds on request-> response headers

* [ ] timeouts on body download
* [ ] require a chunk every 30 seconds. (keep alive)

Length limits:
* [ ] Stop downloading after content-length is reached.
* [ ] Check-sum of google checksum headers when present https://github.com/dart-lang/pub/issues/2281
* [ ] Sanity length limit on version listings.

Improved error behavior:

* [ ] Retry after 429 status code.
* [ ] Don't crash on 403: https://github.com/dart-lang/pub/issues/3256.
* [ ] Retry failure during response-streaming -> we need a fetch function
* [ ] Improve handling of "secret" error codes

nehalvpatel:
I've captured the current state of the current retry logic so we can reference it while refactoring.

Caveat: the [`retry()` method with the CRC32C validation](https://github.com/dart-lang/pub/pull/3546) is not merged as of the writing of this comment.

![Retry Logic - Before](https://user-images.githubusercontent.com/159021/192849809-52e9007a-adb8-4488-8a96-96a387b95aeb.svg)

nehalvpatel:
#3590 improves retries and addresses some of these issues:
- Mitigated intermittent connection/low-level networking errors with retries. We no longer check for OS-specific error codes. 
  - https://github.com/dart-lang/pub/blob/a73598b5f36cf32c2d498391c6b9e7b015c02a4f/lib/src/http.dart#L259-L294 
- Fixed case where client could crash from a 403 response. Consolidated logic for validating the response status code.
  - https://github.com/dart-lang/pub/blob/a73598b5f36cf32c2d498391c6b9e7b015c02a4f/lib/src/http.dart#L306-L335
- Added retries to OAuth2 login and publishing process

PR #3590, opened 2022-10-06
Refactor HTTP retries (#3325)

nehalvpatel:
### UPDATE FROM 10/26/2022 - Plan:
- http.dart exposes:
  - retryHttpLogic / retryForHttp
    - Accepts a function argument
    - Retries with exponential backoff. Default max = 7
    - Does retry when:
      - HTTP 408, 429
      - HTTP 5xx errors
      - `HttpException`, `SocketException`, `TlsException`, `WebSocketException`, `PubHttpException` if PubHttpException.isIntermittent
        - `PackageIntegrityException` will subclass `PubHttpException` and be used for call-site specific errors. For example: crc32c validation and sha256 validation
    - If retries fail, throws `ApplicationException` (investigation of existing behavior needed)
      - 404 by solver (PubHttpException)
      - SocketException?
    - Logs exceptions at a high level (DEBUG)?
    - Logs retry log at a high level (INFO/DEBUG)?
  - extension on http.BaseResponse called `throwIfNotOk()`
  - globalHttpClient
  - attachRequestMetadata(http.Request request)
    - attaches the x-pub-os / command, etc.. headers from zone variables
    - Stuff that _PubHttpClient currently does
    - Used when making requests to list-versions and download archive
- in oauth2.dart, wrap http.Client in RetryClient and pass to package:oauth2 

---

The goal is to simplify and condense the retry logic. A future PR will address the other items in #3325.


lib/src/http.dart
```diff
@@ -343,58 +231,202 @@ Map parseJsonResponse(http.Response response) {
...
+  void throwIfError() {
+    // Retry if the response indicates a server error.
+    if ([500, 502, 503, 504].contains(statusCode)) {
+      throw PubHttpException(this as http.Response, couldRetry: true);
+    }
+
+    // 401 responses should be handled by the OAuth2 client. It's very
+    // unlikely that they'll be returned by non-OAuth2 requests. We also want
+    // to pass along 400 responses from the token endpoint.
+    var tokenRequest = request!.url == oauth2.tokenEndpoint;
```
jonasfj, line 411: Isn't this a hack no necessary anymore?

Because the oauth stuff won't call this?

lib/src/source/hosted.dart
```diff
@@ -885,41 +885,41 @@ class HostedSource extends CachedSource {
...
-          final response = await withAuthenticatedClient(cache,
-              Uri.parse(description.url), (client) => client.send(request));
-          final expectedChecksum = _parseCrc32c(response.headers, fileName);
-
-          Stream<List<int>> stream = response.stream;
-          if (expectedChecksum != null) {
-            stream = _validateStream(
-                response.stream, expectedChecksum, id, archiveUrl);
-          }
+        await client.sendWithRetries(
```
jonasfj, line 891: Does this give us much over just using the retry function directly?

lib/src/http.dart
```diff
@@ -343,58 +231,202 @@ Map parseJsonResponse(http.Response response) {
 Never invalidServerResponse(http.Response response) =>
     fail(log.red('Invalid server response:\n${response.body}'));
 
+/// Exception thrown when an HTTP operation fails.
```
jonasfj, line 234: You now have two exceptions... PubHttpException and this... Documentation comments don't explain why.

lib/src/http.dart
```diff
@@ -343,58 +231,202 @@ Map parseJsonResponse(http.Response response) {
...
+        // in some cloud environments. Furthermore, since these error codes are
+        // platform-specific (undocumented) and essentially cargo-culted along
+        // skipping retries may lead to intermittent issues that could be fixed
+        // with a retry. Failing to retry intermittent issues is likely to cause
+        // customers to wrap pub in a retry loop which will not improve the
+        // end-user experience.
+
+        // TODO: should these return PubConnectException('msg', couldRetry: false)?
+        if (osError.isDnsError) {
+          fail('Could not resolve URL "${request.url.origin}".');
```
jonasfj, line 365: Maybe retry on dns errors... In EC2 I found them to be intermittent... Have we done any research on this?

Like how long time will the OS cache a failed DNS result?

lib/src/http.dart
nehalvpatel: So, `PubConnectException` was added as a temporary hack because `PubHttpException` requires a non-null "response" as part of its construction. This became an issue when I was trying to wrap `SocketException`s. When these happen, it's early enough within the request lifecycle that there is no "response" object I could use to create a `PubHttpException`.

#### What I ended up doing:
- Create `PubConnectException`, which extends `WrappedException`, with the intention to wrap `IOException`, `SocketException`, and possibly `TimeoutException` and `AuthenticationException`.

#### Why wrap?
Initially, I did it because it enables us to have a centralized `couldRetry` property and to drive down the complexity of the `retryIf` function. This is through `mapException`, which I'm using to parse and wrap exceptions before `retryIf` is called.

Thinking further on it, I think it could also improve logs and make it easier to do metrics for HTTP failures.

#### Some options I've been thinking about:
1. Don't wrap exceptions and remove `PubConnectException`.
    - Simplest option and possibly the best one?
2. Make "response" optional in `PubHttpException`.
    - This means updating some logging code to have null checks.
    - Make `PubHttpException` conform to `WrappedException`?
      - Use `PubHttpException` to wrap `IOException`, `SocketException`, and possibly `TimeoutException` and `AuthenticationException`.
3. Rename `PubHttpException` to `PubHttpResponseException` and rename `PubConnectException` to `PubHttpConnect(ion?)Exception`.
    - Use `PubHttpConnect(ion?)Exception` to wrap `IOException`, `SocketException`, and possibly `TimeoutException` and `AuthenticationException`.

lib/src/source/hosted.dart
nehalvpatel: TL;DR: Not really, I can change it to `retryForHttp(client, ...)` or `withHttpRetries(client, ...)`.

---

This came about from me needing access to the "request" object when trying to handle these OS errors: https://github.com/dart-lang/pub/blob/a8979a13f0a63ba28a3bb06882b367fba035aa7d/lib/src/http.dart#L363-L369

The approach we originally discussed, a `performRequest()` function argument, doesn't give us access to the "request" object.

So, I initially added a "request" argument to the HTTP retries method which was passed back to the user through a new argument in `performRequest()` like so: `performRequest(request)`. This was already starting to feel weird and there was no guarantee that the user would not use a different "request" object in `performRequest(request)`. And it was then disqualified when I saw that the "request" object cannot be re-used in retries because it gets finalized when sent.

Then, I reverted and replaced the `performRequest()` function argument with these two: `composeRequest()` and `onResponse(response)`. This lets us create a new "request" object for each attempt by invoking `composeRequest()`.

I could have kept `performRequest(request)`, but it has the same guarantee issue as above and feels super heavy to have to pass a function that just does:
```dart
(request) => await client.send(request);
```

Now, I had the choice to either pass "client" as an argument or to define the HTTP retries method as an extension on `http.Client`. I chose the extension option because `client.sendWithRetries(...)` felt nicer, but I think either option is good.

lib/src/http.dart
nehalvpatel: This is something I haven't yet touched from the original code. I'll take a closer look at this soon.

lib/src/http.dart
nehalvpatel: I can look into this. For this first draft, I've left the behavior as close to the original's as possible.

lib/src/utils.dart
```diff
@@ -673,7 +674,8 @@ Future<T> retry<T>(
   double randomizationFactor = 0.25,
   Duration maxDelay = const Duration(seconds: 30),
   int maxAttempts = 8,
-  FutureOr<bool> Function(Exception)? retryIf,
+  Exception Function(Exception, StackTrace)? mapException,
+  FutureOr<bool> Function(Exception, StackTrace)? retryIf,
```
jonasfj, line 678: Is there ever a good reason make the decision on whether or not to retry based on `StackTrace`?

lib/src/http.dart
```diff
@@ -343,58 +231,202 @@ Map parseJsonResponse(http.Response response) {
...
+      if (retryCount != 3) return;
+      if (!_retriedHosts.add(request.url.host)) return;
+      log.message('It looks like ${request.url.host} is having some trouble.\n'
+          'Pub will wait for a while before trying to connect again.');
+    }, maxAttempts: maxAttempts);
+  }
+}
+
+extension on http.BaseResponse {
+  void throwIfError() {
```
jonasfj, line 402:
```suggestion
  void throwIfNotOK() {
```

lib/src/source/hosted.dart
jonasfj: hmm, I think we need to review the sanity of these errors.. I did some refactor to retry all of them in:
https://github.com/dart-lang/pub/pull/2254

But even then they were essentially cargo-culted :see_no_evil: 


----------------------------

But it's probably fantastic to have some logic for converting a generic `SocketException` into human readable messages like `Unable to validate SSL certificate`. That's helpful for users.

What about:
```dart
/// Program-wide limiter for concurrent network requests.
final _httpPool = Pool(16);

/// Retry [fn] when HTTP exceptions are thrown, printing [operation] in log
/// and error messages.
///
/// The [operation] should be a human readable description of the operation
/// being done. Such that messages like:
///  * "[Something went wrong] when [operation]"
///  * "Retry #2 [operation]"
Future<T> retryForHttp<T>(
  String operation,
  FutureOr<T> Function() fn,
) async {
  try {
    return await retry(
      () async => await _httpPool.withResource(() async => await fn()),
      retryIf: (e, _) =>
          e is SocketException ||
          e is TlsException ||
          e is HttpException ||
          e is PubHttpException && e.couldRetry,
      onRetry: (_, n) => log.io('Retry #${n + 1} $operation'),
      maxAttempts: math.max(
        1, // Having less than 1 attempt doesn't make sense.
        int.tryParse(Platform.environment['PUB_MAX_HTTP_RETRIES'] ?? '') ?? 7,
      ),
    );
  } on SocketException catch (e) {
    final osError = e.osError;
    if (osError == null) {
      throw PubConnectException('', couldRetry: true);
    }

    if (osError.isDnsError) {
      fail('Could not resolve URL when $operation.');
    } else if (osError.isSslError) {
      fail('Unable to validate SSL certificate when $operation.');
    }
    rethrow;
  }
}
```

Then using it becomes:
```dart
      await withAuthenticatedClient(cache, Uri.parse(description.url),
          (client) async {
        // Attempt to download archive and validate its checksum.
        await retryForHttp('downloading "$archiveUrl"', () async {
          final req = http.Request('GET', archiveUrl);
          final res = await client.send(req).timeout(Duration(seconds: 30));
          res.throwIfNotOK();

          Stream<List<int>> stream = res.stream;

          final expectedCrc32 = _parseCrc32c(res.headers, fileName);
          if (expectedCrc32 != null) {
            stream = _validateStream(stream, expectedCrc32, id, archiveUrl);
          }

          await createFileFromStream(stream, archivePath);
        });
      });
```

lib/src/http.dart
```diff
@@ -343,58 +231,202 @@ Map parseJsonResponse(http.Response response) {
...
+
+    // 401 responses should be handled by the OAuth2 client. It's very
+    // unlikely that they'll be returned by non-OAuth2 requests. We also want
+    // to pass along 400 responses from the token endpoint.
+    var tokenRequest = request!.url == oauth2.tokenEndpoint;
+    if (statusCode < 400 ||
+        statusCode == 401 ||
+        (statusCode == 400 && tokenRequest)) {
+      return;
+    }
```
jonasfj, line 416: Are you sure we need to keep this?

lib/src/http.dart
```diff
@@ -343,58 +231,202 @@ Map parseJsonResponse(http.Response response) {
...
+
+    if (statusCode == 406 &&
+        request!.headers['Accept'] == pubApiHeaders['Accept']) {
+      fail('Pub ${sdk.version} is incompatible with the current version of '
+          '${request!.url.host}.\n'
+          'Upgrade pub to the latest version and try again.');
+    }
```
jonasfj, line 423: I don't think these can be null, but this is an error message, if we get it -- it's better that we get null than an null-error :see_no_evil: 
```suggestion
    if (statusCode == 406 &&
        request?.headers['Accept'] == pubApiHeaders['Accept']) {
      fail('Pub ${sdk.version} is incompatible with the current version of '
          '${request?.url.host}.\n'
          'Upgrade pub to the latest version and try again.');
    }
```

lib/src/http.dart
```diff
@@ -343,58 +231,202 @@ Map parseJsonResponse(http.Response response) {
...
+
+    if (statusCode == 500 &&
+        (request!.url.host == 'pub.dartlang.org' ||
+            request!.url.host == 'storage.googleapis.com')) {
+      fail('HTTP error 500: Internal Server Error at ${request!.url}.\n'
+          'This is likely a transient error. Please try again later.');
+    }
```
jonasfj, line 430: Do we need this?

Can't we just print a pretty message when mapping exceptions? Also do we really to special case this for pub.dev and gcs?

jonasfj: Just saying, let's cleanup this and make it generic.

lib/src/http.dart
jonasfj: > Don't wrap exceptions and remove PubConnectException.
>  Simplest option and possibly the best one?

Yes, wrapping exceptions is always extremely complicated.

> Make "response" optional in PubHttpException.

Or maybe we have two:
```dart

class PubHttpException {
  final bool isIntermittent;
  final String message;
}

class PubHttpResponseException extends PubHttpException {
  final int status;
  final Map<String,String> headers;
}

```

lib/src/source/hosted.dart
nehalvpatel: I like the "operation" argument. In your version, would `_parseCrc32c` and `_validateStream` have to be changed to throw `PubHttpException` instead of `PackageIntegrityException`?

lib/src/http.dart
nehalvpatel: The example code you have feels like a good option. I'll work on adopting it and then see how it feels overall.

lib/src/utils.dart
nehalvpatel: Hah, probably not.

lib/src/http.dart
nehalvpatel: Sounds good.

lib/src/http.dart
```diff
@@ -343,58 +231,202 @@ Map parseJsonResponse(http.Response response) {
...
+    http.StreamedResponse? managedResponse;
+
+    return await retry(() async {
+      final resource = await _httpPool.request();
+      request = await composeRequest();
+
+      http.StreamedResponse directResponse;
+      try {
+        directResponse = await send(request);
+        directResponse.throwIfError();
```
nehalvpatel, line 314:
```suggestion
        directResponse.throwIfNotOK();
```

lib/src/http.dart
nehalvpatel: Looking into this soon. This has been lifted over from the current iteration and I planned on revisiting it when I started applying `retryForHttp` to the rest of the HTTP call sites.

lib/src/source/hosted.dart
nehalvpatel: One blocker for this approach is the usage of `withResource()`. I've tried to make that work in a previous iteration and found that it makes it very easy to create deadlocks. The `withResource()` function has many cases where the inner request is either canceled, finished, or errored but the resource doesn't get released.

This complicated piece of code ends up being crucial:
https://github.com/dart-lang/pub/blob/b0978b0e80a1a83bd422b5565d9756e07a0fc35c/lib/src/http.dart#L318-L345

lib/src/oauth2.dart
```diff
@@ -17,6 +18,11 @@ import 'log.dart' as log;
 import 'system_cache.dart';
 import 'utils.dart';
 
+/// The global HTTP client with basic retries. Used instead of retryForHttp for
+/// OAuth calls because the OAuth2 package requires a client to be passed. While
+/// the retry logic is more basic, this is fine for the publishing process.
+final _retryHttpClient = RetryClient(globalHttpClient);
```
nehalvpatel, line 24: Do we want to pass any config to `RetryClient`?

lib/src/source/hosted.dart
```diff
@@ -374,20 +425,22 @@ class HostedSource extends CachedSource {
...
+        });
+      });
       final decoded = jsonDecode(bodyText);
       if (decoded is! Map<String, dynamic>) {
         throw FormatException('version listing must be a mapping');
       }
       body = decoded;
       result = _versionInfoFromPackageListing(body, ref, url, cache);
     } on Exception catch (error, stackTrace) {
-      final packageName = _asDescription(ref.description).packageName;
```
nehalvpatel, line 390: Will removing `_asDescription` cause any side effects? I moved the `packageName` definition above since I use it in the try block. Then I noticed there's already a type check on line 415 which makes that call unnecessary:
https://github.com/dart-lang/pub/blob/e6ae9474bcfe23369911c2b36e8f0453fab872d9/lib/src/source/hosted.dart#L411-L421

lib/src/error_group.dart
```diff
@@ -191,7 +191,9 @@ class _ErrorGroupFuture<T> implements Future<T> {
 
     // Make sure _completer.future doesn't automatically send errors to the
     // top-level.
-    _completer.future.catchError((_) {});
+    _completer.future.catchError(
+        // ignore: body_might_complete_normally_catch_error
```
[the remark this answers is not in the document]
sigurdm: Perhaps we can do as this: https://github.com/dart-lang/pub/pull/3631

lib/src/http.dart
```diff
@@ -345,56 +189,85 @@ Never invalidServerResponse(http.Response response) =>
...
+        request?.headers['Accept'] == pubApiHeaders['Accept']) {
+      fail('Pub ${sdk.version} is incompatible with the current version of '
+          '${request?.url.host}.\n'
+          'Upgrade pub to the latest version and try again.');
+    } else if (statusCode >= 500 ||
+        statusCode == HttpStatus.requestTimeout ||
+        statusCode == HttpStatus.tooManyRequests) {
+      // Throw if the response indicates a server error or an intermittent
+      // client error, but mark it as intermittent so it can be retried.
+      throw PubHttpResponseException(this as http.Response,
```
jonasfj, line 266: If you have to cast, maybe it's better to do `extension Throwing on http.Response` or, make `PubHttpResponseException` accept an `BaseResponse`

lib/src/oauth2.dart
jonasfj: `when` and `whenError` might be worth defining, the defaults are a little lax:
https://github.com/dart-lang/http/blob/master/pkgs/http/lib/retry.dart#L151-L153

Like `whenError: (e, st) => e is IoException` might be fine.
`when: (BaseResponse response) => response.statusCode >= 500` might also be nice...

This is less important.. but something like that.

lib/src/exceptions.dart
```diff
@@ -106,12 +107,8 @@ class PackageNotFoundException extends WrappedException {
...
 
 /// A class for exceptions where a package's checksum could not be validated.
-class PackageIntegrityException extends WrappedException {
-  PackageIntegrityException(
-    String message, {
-    Object? innerError,
-    StackTrace? innerTrace,
-  }) : super(message, innerError, innerTrace);
+class PackageIntegrityException extends PubHttpException {
+  PackageIntegrityException(super.message, {super.isIntermittent});
```
jonasfj, line 111: nit: Aren't these always intermittent?

lib/src/command/login.dart
```diff
@@ -46,7 +46,7 @@ class LoginCommand extends PubCommand {
 
   Future<_UserInfo?> _retrieveUserInfo() async {
     return await oauth2.withClient(cache, (client) async {
-      final discovery = await httpClient.get(Uri.https(
+      final discovery = await globalHttpClient.get(Uri.https(
           'accounts.google.com', '/.well-known/openid-configuration'));
```
jonasfj, line 50: Let's use `retryForHttp` around fetching this information.

lib/src/source/hosted.dart
```diff
@@ -167,6 +168,57 @@ class HostedSource extends CachedSource {
...
+    headers['X-Pub-OS'] = Platform.operatingSystem;
+    headers['X-Pub-Command'] = PubCommand.command;
+    headers['X-Pub-Session-ID'] = sessionId;
+
+    var environment = Platform.environment['PUB_ENVIRONMENT'];
+    if (environment != null) {
+      headers['X-Pub-Environment'] = environment;
+    }
+
+    var type = Zone.current[#_dependencyType];
```
jonasfj, line 203: We can't use this symbol here, I at-least I don't think we can. Since it's a private symbol, it should be distinct from the symbol defined in `http.dart`:
https://github.com/dart-lang/pub/blob/86536fd07dc62705a2b1b3eb46e1208b3a361f3a/lib/src/http.dart#L138

At-least that's my guess.

------------------------------------------------------

May I suggest that we make:
```dart
extension AttachHeaders on http.Request {
  void attachMetadataHeaders() {
      ...
  }
}
```

in `http.dart`.

We'll have to refactor the places that use `Client.get` and `Client.read` into creating `Request` objects  and passing them to `Client.send`.
But I don't think that matters much.

lib/src/error_group.dart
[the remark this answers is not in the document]
nehalvpatel: Sounds good! I will rebase to get that change.

lib/src/http.dart
nehalvpatel: Extending `http.Response` breaks the extension for the _download method, which uses `http.StreamedResponse`. Vice-versa for if we extend `http.StreamedResponse`, since the versions method uses `http.Response`.

---

#### The change to make `PubHttpResponseException` accept `http.BaseResponse` works, but causes build errors in lish.dart:

Screen Shot 2022-11-03 at 5 10 14 PM

---

#### If we change `handleJsonError` to accept `http.BaseResponse`, we get some more errors:

Screen Shot 2022-11-03 at 5 11 17 PM

---

#### And then finally, if we change `parseJsonResponse` to accept `http.BaseResponse`:

Screen Shot 2022-11-03 at 5 12 24 PM

---

So, this was the only way I could come up with to make it all work. 😅

lib/src/source/hosted.dart
sigurdm: I think that can be removed. That looks much better!

lib/src/source/hosted.dart
sigurdm: Makes me wonder if we have any tests covering the meta-data sending?

It would probably require changing `isPubDevUrl` to something like:

```dart
  static bool isPubDevUrl(String url) {
    final origin = Uri.parse(url).origin;

    return origin == pubDevUrl ||
        origin == pubDartlangUrl ||
        (runningFromTest &&
            origin == io.Platform.environment['_PUB_TEST_DEFAULT_HOSTED_URL']);
  }
```

lib/src/command/lish.dart
```diff
@@ -94,7 +94,8 @@ class LishCommand extends PubCommand {
     try {
       await log.progress('Uploading', () async {
         var newUri = host.resolve('api/packages/versions/new');
-        var response = await client.get(newUri, headers: pubApiHeaders);
+        var response = await client.get(newUri,
```
sigurdm, line 97: Don't we need to add retries here and in line 124 ?

sigurdm (commented):
Looks really good!

We don't have to do the metadata tests in this PR - bet they surely would be nice to have

lib/src/exceptions.dart
nehalvpatel: True! I'll clean that up.

lib/src/command/lish.dart
nehalvpatel: My understanding from talking to @jonasfj was that retries might be unnecessary for the publishing process. But I do see how they'd be useful, especially as we start working towards automated publishes.

lib/src/source/hosted.dart
nehalvpatel: I've changed the headers method to an extension in http.dart. 👍

Re: the metadata test, I'd like to do that test but need some clarification on the desired behavior for the metadata URL-checking logic. I've sent you a message on Chat.

lib/src/http.dart
jonasfj: Fair point, we can clean it up later.

lib/src/source/hosted.dart
jonasfj: How about we do tests for metadata sending a follow up PR (it's not tests we currently have). And it's probably better to keep this PR small.

lib/src/http.dart
jonasfj: One option could be to create `PubHttpResponse(BaseResponse, List<int>? body);` so that it has an optional body.

But maybe this is better left for a follow up PR...

jonasfj: I think we actually have: https://github.com/dart-lang/pub/issues/3130

For refactoring `handleJsonError`, I think if we're going to make better output -- that's also a great opportunity to re-think the API interface and use something like `BaseResponse` instead.

---------------------------

Actually, it might be cleaner to cast from `BaseResponse` to `Response` in `handleJsonError`, because -- if it's not a `Response` then it's same as not being able to read the body.

jonasfj:
@nehalvpatel I think if we fix the tests and retry in publishing, then this is pretty much good to go.

We can revisit the types in `PubHttpResponse` in a follow up PR, maybe while try to fix https://github.com/dart-lang/pub/issues/3130. It's probably a lot easier to do a few smaller PRs.

lib/src/http.dart
nehalvpatel: The cast from `BaseResponse` to `Response` looks good!

I ended up having to implement that in this PR because the cast in `throwIfNotOk` was causing tests to fail.

https://github.com/dart-lang/pub/blob/7afe52cd1aa68c4546dcfe595a53d8890ed3b024/lib/src/http.dart#L192-L197

lib/src/http.dart
```diff
@@ -311,13 +311,36 @@ extension Throwing on http.BaseResponse {
...
+  /// is successful, reads the whole response body and returns it.
+  ///
+  /// The send method on [http.Client], which returns a [http.StreamedResponse],
+  /// is the only method that accepts a request object. This method can be used
   /// when you need to send a request object but want a regular response object.
-  Future<http.Response> sendSync(http.BaseRequest request) async {
+  ///
+  /// If false is passed for [throwIfNotOk], the response will not be validated.
+  /// See [http.BaseResponse.throwIfNotOk] extension for validation details.
+  Future<http.Response> fetch(http.BaseRequest request,
```
jonasfj, line 347: I don't mind, but I have mixed feelings -- mostly because it's easy to call `send()` by accident instead of using `fetch()`.

or rather it's not very obvious that `fetch` is intended to throw something in certain cases. But maybe, this is something we can refactor later.

jonasfj: We could do:
```dart

class PubHttpClient {
  Future<http.Response> fetch(http.BaseRequest request);
  Future<http.StreamingResponse> fetchStream(http.BaseRequest request);

  // For the new cases where you really need it.
  http.Client rawClient();
}
```
But maybe, we do this sort of thing later. Now that we're moving `package:oauth2` into `pub` -- we could consider refactoring all of that, and just have a `fetch` method and `fetchStream` method. But it's a lot more refactoring, let's cleanup later.

jonasfj (approved):

PR #3602, opened 2022-10-13
update docs for running 'pub token'

devoncarew:
- update docs for running 'pub token'

I noticed that the error messages here are outdated - they specified running `pub token` instead of `dart pub token`.

### DIFF
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -126,7 +126,7 @@ class LishCommand extends PubCommand {
       if (error.statusCode == 401) {
         msg += '$host package repository requested authentication!\n'
             'You can provide credentials using:\n'
-            '    dart pub token add $host\n';
+            '    $topLevelProgram pub token add $host\n';
       }
       if (error.statusCode == 403) {
         msg += 'Insufficient permissions to the resource at the $host '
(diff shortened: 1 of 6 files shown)


lib/src/command/lish.dart
```diff
@@ -126,12 +126,12 @@ class LishCommand extends PubCommand {
       if (error.statusCode == 401) {
         msg += '$host package repository requested authentication!\n'
             'You can provide credentials using:\n'
-            '    pub token add $host\n';
+            '    dart pub token add $host\n';
```
sigurdm, line 129: Could you do
```suggestion
            '    $topLevelProgram pub token add $host\n';
```
to support `flutter pub token add`

lib/src/command/lish.dart
```diff
@@ -126,12 +126,12 @@ class LishCommand extends PubCommand {
...
       }
       if (error.statusCode == 403) {
         msg += 'Insufficient permissions to the resource at the $host '
             'package repository.\nYou can modify credentials using:\n'
-            '    pub token add $host\n';
+            '    dart pub token add $host\n';
```
sigurdm, line 134: Same here
```suggestion
            '    $topLevelProgram pub token add $host\n';
```

sigurdm (approved):

Issue #3607, opened 2022-10-17
unable to pub publish from from 2.19.0-dev SDK


devoncarew:
## Environment

- Dart version (run `dart --version`):

On or around `2.19.0-312.0.dev`.

- OS kind and version (e.g. "Windows 10, version 1809" or "macOS 12.4"):

Reproed on both macos and linux.

## Problem

I'm unable to publish any package using the 2.19.0 dev Dart SDK. I've tried several options to reset my local environment, including deleting my pub cache as well as my local auth tokens (in $HOME/.config/dart as well as the mac location) and running `dart pub logout`.

The error I see is:

```
Do you want to publish watcher 1.0.2 to https://pub.dartlang.org (y/N)? Y
Uploading... 
https://pub.dartlang.org package repository requested authentication!
You can provide credentials using:
    pub token add https://pub.dartlang.org
Authentication failed!
```

I was able to work around the issue by downloading an older SDK (in this case `1.17.0`) and using the dart command in there to run pub publish.

I didn't try bisecting the publishing issue to a specific sdk commit, but that could be useful for figuring out the regression.

jonasfj:
@sigurdm I'm 91% sure this is a bug from https://github.com/dart-lang/pub/pull/3528

sigurdm:
I checked manually that the fix indeed lets us publish again.

PR #3610, opened 2022-10-17
Fix publishing

jonasfj:
@sigurdm I didn't try this out, but I think this is the fix for https://github.com/dart-lang/pub/issues/3607.

I guess we can only validate this by trying it?


sigurdm (approved):
Thanks for catching and fixing this.

PR #3611, opened 2022-10-18
Completely switch from pub.dartlang.org to pub.dev

sigurdm:
As content-hashes are introduced with https://github.com/dart-lang/pub/pull/3482 all packages in the cache has to be downloaded again (to validate the hash).

We use this chance to clean up move the client entirely to using pub.dev.

pub.dartlang.org will be rewritten internally, so these should continue working:

* `dart pub token add pub.dartlang.org` (will add a pub.dev token)
* `dependencies: {foo: {hosted: pub.dartlang.org}}`
* `publish_to: pub.dartlang.org`
* `dart pub publish --server=pub.dartlang.org`

This pr will incur a disk-space overhead in existing caches where the pub.dartlang.org directory will linger.
But any kind of reusing the old directory will likely be buggy and have poor backwards compatibility and as mentioned redownloads would happen anyway because of content hashing.

Continues the effort we started when fixing: https://github.com/dart-lang/pub/issues/3519 .

### DIFF
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -155,9 +155,9 @@ class LishCommand extends PubCommand {
   Future<void> _publish(List<int> packageBytes) async {
     try {
       final officialPubServers = {
-        'https://pub.dartlang.org',
-        // [validateAndNormalizeHostedUrl] normalizes https://pub.dev
-        // to https://pub.dartlang.org, so we don't need to do allow that here.
+        'https://pub.dev',
+        // [validateAndNormalizeHostedUrl] normalizes https://pub.dartlang.org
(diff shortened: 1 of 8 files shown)


jonasfj (approved):

sigurdm:
cc @mit-mit Just a heads up that this switch is happening.

parlough:
In [Troubleshooting pub](https://dart.dev/tools/pub/troubleshoot) there is a section on [Getting a socket error trying to find a package](https://dart.dev/tools/pub/troubleshoot#pub-get-socket-error) which discusses adding pub.dartlang.org as a "trusted web address". Should we switch/add pub.dev?

PR #3613, opened 2022-10-18
Support `--source` in kernel compilation commands

dcharkes:
https://dart-review.googlesource.com/c/sdk/+/231334 introduced passing additional sources to `gen_kernel`.

This PR enables passing additional sources to the various pub commands compiling kernel, so that these arguments can be passed from dartdev.

This PR depends on:

* https://github.com/dart-lang/webdev/pull/1760

### DIFF
--- a/lib/src/dart.dart
+++ b/lib/src/dart.dart
@@ -145,6 +145,10 @@ class AnalyzerErrorGroup implements Exception {
 /// for `package:` uri resolution.
 ///
 /// The [name] is used to describe the executable in logs and error messages.
+///
+/// The [additionalSources], if provided, instruct the compiler to include
+/// additional source files into compilation even if they are not referenced
+/// from the main library.
 Future<void> precompile({
   required String executablePath,
   required String incrementalDillPath,
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -465,6 +465,10 @@ class Entrypoint {
   }
 
   /// Precompiles executable .dart file at [path] to a snapshot.
+  ///
+  /// The [additionalSources], if provided, instruct the compiler to include
+  /// additional source files into compilation even if they are not referenced
+  /// from the main library.
   Future<void> precompileExecutable(
     Executable executable, {
     List<String> additionalSources = const [],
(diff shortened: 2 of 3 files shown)


dcharkes:
CI is red because of https://github.com/dart-lang/pub/issues/3617.

lib/src/dart.dart
```diff
@@ -151,6 +151,7 @@ Future<void> precompile({
   required String name,
   required String outputPath,
   required String packageConfigPath,
+  List<String> additionalSources = const [],
```
jonasfj, line 154: Can we update the documentation comment?

jonasfj (approved):

lib/src/dart.dart
dcharkes: done (and in other files)

dcharkes:
Merging. This will turn CI red, which is tracked in https://github.com/dart-lang/pub/issues/3617.

dcharkes:
@devoncarew [your previous DEPS CL](https://dart-review.googlesource.com/c/sdk/+/265002) looks suspiciously like an automatically generated CL. Is there a tool I can invoke to do the same? Or is it automatic? Might I suggest adding a link to the tool documentation in the commit messages of the CLs it creates?

Pub also needs a bump now in the sdk DEPS file to include this PR.

devoncarew:
Yes, it is from a tool; we generally run the script (manually) 1-2 times per week. It's low enough cost in aggregate (per package) that I'm not too worried about further automation.

Anyone can run it (and is welcome to). It's just `dart tools/rev_sdk_deps.dart` from the sdk repo. That will use git the gclient CLI tool to make edits to the DEPS file. You then create a CL as normal from there. We could add some additional tooling around it to automatically create a CL (as `tools/manage_deps.dart` does), but haven't explored that yet.

Pub is currently excluded from auto-bumping via this tool, as that team likes to update the SDK's changelog when they roll (the linter and dart_style are also excluded, for separate reasons).

> Might I suggest adding a link to the tool documentation in the commit messages of the CLs it creates?

Sounds great; I'd take a CL that does that (in the tool's current form, that message would just be printed to stdout, and it would be up to the developer to copy the commit messages and that text into the CL's description).

Issue #3608, opened 2022-10-16
Global activate doesn't support --git-ref


matthew-carroll:
Global activate says it supports checking out Git source with a `--git-ref`. However, when trying to use this capability, the Git repo is checked out with a `depth` of `1`, which only finds the `main` branch. Therefore, it currently seems to be impossible to activate a branch of a repo.

mraleph:
Transferring to dart-lang/pub where this belongs.

Giuspepe:
I fixed this in my PR https://github.com/dart-lang/pub/pull/3620 which is ready to be reviewed

sigurdm:
This is fixed

PR #3620, opened 2022-10-21
Support --git-ref argument in pub global activate (#3608)

Giuspepe:
This fixes https://github.com/dart-lang/pub/issues/3608.

A branch name, tag, or full length commit SHA can be passed to the `--git-ref` argument in `dart pub global activate --source git`.

While fixing this I noticed that for `--source git`, `pub` downloads the git repository quite often. E.g. it is downloaded in the very beginning just to get the package name from the repository and then it is deleted immediately, just to be downloaded again in a later step. So I'm sure we can optimize the download behavior in another PR.

### DIFF
--- a/lib/src/command/global_activate.dart
+++ b/lib/src/command/global_activate.dart
@@ -14,8 +14,10 @@ import '../utils.dart';
 class GlobalActivateCommand extends PubCommand {
   @override
   String get name => 'activate';
+
   @override
   String get description => "Make a package's executables globally available.";
+
   @override
   String get argumentsDescription => '<package> [version-constraint]';
 
@@ -29,7 +31,8 @@ class GlobalActivateCommand extends PubCommand {
     argParser.addOption('git-path', help: 'Path of git package in repository');
 
     argParser.addOption('git-ref',
-        help: 'Git branch or commit to be retrieved');
+        help: 'Git branch, tag, or commit SHA to be retrieved. '
+            'When using a commit SHA, must be the full SHA (40 characters).');
 
     argParser.addMultiOption('features',
         abbr: 'f', help: 'Feature(s) to enable.', hide: true);
--- a/lib/src/command/global_activate.dart
+++ b/lib/src/command/global_activate.dart
@@ -14,10 +14,8 @@ import '../utils.dart';
 class GlobalActivateCommand extends PubCommand {
   @override
   String get name => 'activate';
-
   @override
   String get description => "Make a package's executables globally available.";
-
   @override
   String get argumentsDescription => '<package> [version-constraint]';
 
--- a/lib/src/command/global_activate.dart
+++ b/lib/src/command/global_activate.dart
@@ -31,8 +31,7 @@ class GlobalActivateCommand extends PubCommand {
     argParser.addOption('git-path', help: 'Path of git package in repository');
 
     argParser.addOption('git-ref',
-        help: 'Git branch, tag, or commit SHA to be retrieved. '
-            'When using a commit SHA, must be the full SHA (40 characters).');
+        help: 'Git branch, tag, or commit SHA to be retrieved.');
 
     argParser.addMultiOption('features',
         abbr: 'f', help: 'Feature(s) to enable.', hide: true);
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -199,9 +199,12 @@ class GitSource extends CachedSource {
   Future<String> getPackageNameFromRepo(
       String repo, String? ref, String? path, SystemCache cache) {
     // Clone the repo to a temp directory.
+
+    // TODO: why is it cloned to a temp dir? it's downloaded again later?!
+    // TODO: let's not delete the temp dir but instead move it to the right destination in .pub-cache
+
     return withTempDir((tempDir) async {
-      await _clone(repo, tempDir, shallow: true);
-      if (ref != null) await _checkOut(tempDir, ref);
+      await _clone(repo, tempDir, shallow: true, ref: ref);
       var pubspec = Pubspec.load(p.join(tempDir, path), cache.sources);
       return pubspec.name;
     });
@@ -425,7 +428,7 @@ class GitSource extends CachedSource {
 
   /// Ensures that the canonical clone of the repository referred to by [ref]
   /// contains the given Git [revision].
-  Future _ensureRevision(
+  Future<void> _ensureRevision(
     PackageRef ref,
     String revision,
     SystemCache cache,
@@ -448,7 +451,7 @@ class GitSource extends CachedSource {
 
   /// Ensures that the canonical clone of the repository referred to by [ref]
   /// exists and is up-to-date.
-  Future _ensureRepoCache(PackageRef ref, SystemCache cache) async {
+  Future<void> _ensureRepoCache(PackageRef ref, SystemCache cache) async {
     var path = _repoCachePath(ref, cache);
     if (_updatedRepos.contains(path)) return;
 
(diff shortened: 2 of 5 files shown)


Giuspepe:
Hey @jonasfj would you mind reviewing this? 😊

test/global/activate/git_package_test.dart
```diff
@@ -9,9 +9,9 @@ import '../../descriptor.dart' as d;
 import '../../test_pub.dart';
 
 void main() {
-  test('activates a package from a Git repo', () async {
-    ensureGit();
+  setUp(ensureGit);
```
jonasfj, line 12: Is this something we can do in a separate PR that's super easy to land -- just to reduce noise in this PR?

lib/src/source/git.dart
```diff
@@ -571,33 +570,52 @@ class GitSource extends CachedSource {
...
     bool mirror = false,
     bool shallow = false,
-  }) {
-    return Future.sync(() {
-      // Git on Windows does not seem to automatically create the destination
-      // directory.
-      ensureDir(to);
-      var args = [
+    String? ref,
+  }) async {
```
jonasfj, line 579: I'm less sure about this.
Curious what @sigurdm thinks?

Maybe, we  should use [`git worktree`](https://git-scm.com/docs/git-worktree).

Also if we're going to change this logic, let's aim to make it compatible with git lfs.

lib/src/command/global_activate.dart
```diff
@@ -29,7 +31,8 @@ class GlobalActivateCommand extends PubCommand {
     argParser.addOption('git-path', help: 'Path of git package in repository');
 
     argParser.addOption('git-ref',
-        help: 'Git branch or commit to be retrieved');
+        help: 'Git branch, tag, or commit SHA to be retrieved. '
+            'When using a commit SHA, must be the full SHA (40 characters).');
```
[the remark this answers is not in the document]
Giuspepe: My initial fix was only able to handle full commit SHAs, but it turns out that partial commit SHAs now work as well with the current fix! So it's great that you looked over this because that made me test it once again with partial SHAs. I tested it locally with a remote git repository and added a test for this as well.

lib/src/source/git.dart
```diff
@@ -425,7 +424,7 @@ class GitSource extends CachedSource {
 
   /// Ensures that the canonical clone of the repository referred to by [ref]
   /// contains the given Git [revision].
-  Future _ensureRevision(
+  Future<void> _ensureRevision(
```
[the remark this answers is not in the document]
Giuspepe: Sure, I removed all the noise :D

test/global/activate/git_package_test.dart
Giuspepe: The noise is removed now :)

lib/src/source/git.dart
```diff
@@ -576,21 +575,42 @@ class GitSource extends CachedSource {
...
+        await git.run(['remote', 'add', 'origin', from], workingDir: to);
+        await git.run([
+          'fetch',
+          'origin',
+          '--tags',
+          if (shallow) ...['--depth', '1'],
+          if (ref != null) ref,
+        ], workingDir: to);
+        await git.run(['reset', '--hard', 'FETCH_HEAD'], workingDir: to);
+        await git.run(['fetch', '--all'], workingDir: to);
```
jonasfj, line 597: What are the various different approaches we could take to doing this?

Giuspepe: @jonasfj @sigurdm 
The goal is to download the repository with the given ref. Otherwise `git checkout <ref>` which is executed later on will fail. That's the reason why on `master` it currently isn't possible to use the `--git-ref` parameter, because that ref is never downloaded when only calling `git clone --depth 1` like in the current implementation. Sadly it isn't possible to just pass a ref parameter to `git clone` which is why we need a workaround for `--git-ref` to work.

**Implemented approach in this PR**
With the `git init`, `git remote add origin`, `git fetch origin ref`, `git reset --hard FETCH_HEAD` approach it is possible to download the repository at a given ref (no matter if the ref is a full/partial commit ID, a branch, or a tag name). 

**Alternative approach: clone the repository with the entire history instead of just `--shallow`**
This would make the `_clone` method much easier and would solve the problem. Just download the repository with the entire history and the ref will be contained in the history, so we can check it out later.  
The drawback of this of course is that it isn't as efficient as just downloading the repository at the given ref.

However, there are some inefficiencies in the current downloading of git repositories which we could improve so that it would be okay to clone the entire history.

Currently (on `master`), the given repository is downloaded multiple times anyways:
1. First, with `--shallow` _just_ to get the name of the package. (Btw, this could lead to errors when the package name changes between refs). So the shallowly cloned repository is downloaded into a temporary directory just to get the package name and deleted directly after that.
2. Then, the git repository is cloned again during version solving. This time it is cloned with `--mirror` and it is cloned into the `.pub-cache/git/cache` directory, e.g. `.pub-cache/git/cache/<package-name>-<git-url-hash>`.
3. Once the version solver is done, the repository is copied from the `.pub-cache/git/cache` to the `.pub-cache/git` directory, e.g. `pub-cache/git/<package-name>-<git-commit-SHA>`

Because the git repository is cloned anyways during version solving with `--mirror` (which means that we're already downloading the whole history anyways), we could just clone it one step earlier with `--mirror` so we can use the same repository with the whole history for getting the package name (step 1) as well as version solving (step 2). We're downloading the repository with `--mirror` sooner or later anyways.

Although I don't entirely understand why we are cloning the repository with `--mirror` in the first place. Why do we need the whole history of the repository for version solving if we already specified a specific git-ref? Can't we just download the repository shallowly at that specific git-ref to avoid downloading the whole history?

**Other approaches**
I'm sure there are other approaches which would work - as long as the tests which I added in `git_package_test.dart` work, it should be fine. These tests ensure that you can use the `--git-ref` parameter in all conceivable ways (branch-name, tag, HEAD~, full commit SHA, partial commit SHA, HEAD~2).

jonasfj: (1) feels like a hack that is special cased for `pub global activate`, perhaps it's not super smart :see_no_evil: 

------------

I think (2) is `pub` trying to be smart. This is how git-dependencies works in general.
`pub` creates a mirror of the repository and then we can create a checkout of a specific revision using the local mirror of the repository.

That way, if you use the same git-dependency in different projects, or if you upgrade to a new git-revision, you don't clone the repository again, because there is a local mirror created in step (2).

I guess it's smart, but perhaps too smart :see_no_evil: 

We have an issue where git LFS doesn't work by default because the mirror, and later clone from mirror in step (3). I'm not sure why the LFS doesn't work when doing this -- but it doesn't -- probably because when we do the clone in step (3) we don't supply the URL for the original repository, we only supply the URL for the local mirror created in (2).

I think if in (3) we did a clone of the original repository and passed `--reference path/local/mirror` then git LFS would work.

---------------------

All of this makes me wonder a bit:
 * Should we do a local mirror at all?
 * Should we clone git dependencies into `.dart_tool/...` in the users' project instead of cloning into `PUB_CACHE`?
   * Unless you run `dart pub cache clean` there really is no mechanism for garbage collecting the `PUB_CACHE`.
   * Argument would have to be that:
     * git dependencies are more likely to updated frequently, and,
     * git dependencies are less likely to be reused.
 * If we want to do the local mirror in (2), should we perhaps not do a clone in (3), but instead do a `git work-tree` checkout instead? (I think that might also solve the git LFS issue)

Before we change how git-dependencies are cloned, I think it's worth making sure we actually want to keep doing (2) and (3) the way we currently are.

sigurdm:
Closing in favor of:  https://github.com/dart-lang/pub/pull/3656

We are grateful for your proposed contribution, but I think the approach in the other PR is the right one.

Issue #3532, opened 2022-08-18
New command `dart pub cache preload package.tar.gz`


sigurdm:
Should install a package in the pub cache.
Should write the content-hash.
Should use publish_to as the server, and default to PUB_HOSTED_URI should default to pub.dev (that gets written in the cache to pub.dartlang.org).

The command should be hidden in the --help text.

sigurdm:
If that package-version exists we overwrite it.

PR #3636, opened 2022-11-07
New command `dart pub cache preload`

sigurdm:
Will load one or more tar.gz archives of packages as those served by pub.get and add them to the cache.

Will overwrite an existing entry for the same package-version.

No attempt is made to verify this against what is really being served -> this can be done offline.
Writes the sha256 hash of the archive as the content-hash in the cache -> later pub get invocations will verify the cached hash against what is in the cache.

Fixes: https://github.com/dart-lang/pub/issues/3532

### DIFF
--- a/lib/src/command/cache_preload.dart
+++ b/lib/src/command/cache_preload.dart
@@ -21,6 +21,8 @@ class CachePreloadCommand extends PubCommand {
   @override
   String get docUrl => 'https://dart.dev/tools/pub/cmd/pub-cache';
 
+  /// The `cache preload` command is hidden by default, because it's really only intended for
+  /// `flutter` to use when pre-loading `PUB_CACHE` after being installed from `zip` archive.
   @override
   bool get hidden => true;
 
--- a/lib/src/command/cache_preload.dart
+++ b/lib/src/command/cache_preload.dart
@@ -10,7 +10,7 @@ import '../log.dart' as log;
 import '../source/hosted.dart';
 import '../utils.dart';
 
-/// Handles the `cache add` pub command.
+/// Handles the `cache preload` pub command.
 class CachePreloadCommand extends PubCommand {
   @override
   String get name => 'preload';
--- a/lib/src/command/cache_preload.dart
+++ b/lib/src/command/cache_preload.dart
@@ -1,0 +1,47 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import '../command.dart';
+import '../io.dart';
+import '../log.dart' as log;
+import '../source/hosted.dart';
+import '../utils.dart';
+
+/// Handles the `cache add` pub command.
+class CachePreloadCommand extends PubCommand {
+  @override
+  String get name => 'preload';
+  @override
+  String get description => 'Install packages from a .tar.gz archive.';
+  @override
+  String get argumentsDescription => '<package1.tar.gz> ...';
+  @override
+  String get docUrl => 'https://dart.dev/tools/pub/cmd/pub-cache';
+
+  @override
+  bool get hidden => true;
+
+  @override
+  Future<void> runProtected() async {
+    // Make sure there is a package.
+    if (argResults.rest.isEmpty) {
+      usageException('No package to preload given.');
+    }
+
+    for (String packagePath in argResults.rest) {
+      if (!fileExists(packagePath)) {
+        fail('Could not find file $packagePath.');
+      }
+    }
+    for (String archivePath in argResults.rest) {
+      final id = await cache.hosted.preloadPackage(archivePath, cache);
+      final url = (id.description.description as HostedDescription).url;
+
+      final fromPart = HostedSource.isFromPubDev(id) ? '' : ' from $url';
+      log.message('Installed $archivePath in cache as $id$fromPart.');
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


lib/src/command/cache_preload.dart
```diff
@@ -0,0 +1,47 @@
...
+  String get name => 'preload';
+  @override
+  String get description => 'Install packages from a .tar.gz archive.';
+  @override
+  String get argumentsDescription => '<package1.tar.gz> ...';
+  @override
+  String get docUrl => 'https://dart.dev/tools/pub/cmd/pub-cache';
+
+  @override
+  bool get hidden => true;
```
jonasfj, line 25:
```suggestion
  /// The `cache preload` command is hidden by default, because it's really only intended for
  /// `flutter` to use when pre-loading `PUB_CACHE` after being installed from `zip` archive.
  @override
  bool get hidden => true;
```

lib/src/command/cache_preload.dart
```diff
@@ -0,0 +1,47 @@
...
+
+import 'dart:async';
+
+import '../command.dart';
+import '../io.dart';
+import '../log.dart' as log;
+import '../source/hosted.dart';
+import '../utils.dart';
+
+/// Handles the `cache add` pub command.
```
jonasfj, line 13:
```suggestion
/// Handles the `cache preload` pub command.
```

lib/src/source/hosted.dart
```diff
@@ -1086,9 +1086,14 @@ See $contentHashesDocumentationUrl.
...
-      await extractTarGz(readBinaryFileAsStream(archivePath), tempDir);
+      try {
+        await extractTarGz(readBinaryFileAsStream(archivePath), tempDir);
 
-      ensureDir(p.dirname(destPath));
+        ensureDir(p.dirname(destPath));
+      } catch (e) {
+        deleteEntry(tempDir);
+        rethrow;
+      }
```
jonasfj, line 1096: Nice, unrelated fix, right? :rofl:

test/cache/preload_test.dart
```diff
@@ -0,0 +1,165 @@
...
+
+    await appDir({'foo': '^2.0.0'}).create();
+    // Do a `pub get` here to create a lock file in order to validate we later can
+    // `pub get --offline` with packages installed by `preload`.
+    await pubGet();
+
+    await runPub(args: ['cache', 'clean', '-f']);
+
+    final archivePath = p.join(sandbox, 'archive');
+    final archivePath2 = p.join(sandbox, 'archive2');
```
jonasfj, line 30:
```suggestion
    final archivePath1 = p.join(sandbox, 'foo-1.0.0-archive.tar.gz');
    final archivePath2 = p.join(sandbox, 'foo-2.0.0-archive.tar.gz');
```

test/cache/preload_test.dart
```diff
@@ -0,0 +1,165 @@
...
+
+    File(archivePath).writeAsBytesSync(await readBytes(
+        Uri.parse(server.url).resolve('packages/foo/versions/1.0.0.tar.gz')));
+    File(archivePath2).writeAsBytesSync(await readBytes(
+        Uri.parse(server.url).resolve('packages/foo/versions/2.0.0.tar.gz')));
```
jonasfj, line 35: nit: but are you sure you don't want a trailing comma here :rofl: (do what you like, I can't really care :see_no_evil: )
```suggestion
    File(archivePath).writeAsBytesSync(
      await readBytes(Uri.parse(server.url).resolve('packages/foo/versions/1.0.0.tar.gz')),
    );
    File(archivePath2).writeAsBytesSync(
      await readBytes(Uri.parse(server.url).resolve('packages/foo/versions/2.0.0.tar.gz')),
    );
```

test/cache/preload_test.dart
```diff
@@ -0,0 +1,165 @@
...
+    final archivePath = p.join(sandbox, 'archive');
+
+    File(archivePath).writeAsBytesSync(await readBytes(
+        Uri.parse(server.url).resolve('packages/foo/versions/1.0.0.tar.gz')));
+    await runPub(
+      args: ['cache', 'preload', archivePath],
+      environment: {'_PUB_TEST_DEFAULT_HOSTED_URL': 'pub.dev'},
+      output: allOf([
+        contains(
+            'Installed $archivePath in cache as foo 1.0.0 from ${server.url}.')
```
jonasfj, line 73: I don't understand this test case?

jonasfj (approved):

lib/src/source/hosted.dart
sigurdm: Yes :)

test/cache/preload_test.dart
```diff
@@ -0,0 +1,165 @@
...
+    final server = await servePackages();
+    server.serve('foo', '1.0.0');
+
+    final archivePath = p.join(sandbox, 'archive');
+
+    File(archivePath).writeAsBytesSync(await readBytes(
+        Uri.parse(server.url).resolve('packages/foo/versions/1.0.0.tar.gz')));
+    await runPub(
+      args: ['cache', 'preload', archivePath],
+      environment: {'_PUB_TEST_DEFAULT_HOSTED_URL': 'pub.dev'},
```
jonasfj, line 70:
```suggestion
      // '${server.url}' is the default value for _PUB_TEST_DEFAULT_HOSTED_URL in tests
      // by overwriting it, we make the pub command think '${server.url}' is a custom pub server.
      environment: {'_PUB_TEST_DEFAULT_HOSTED_URL': 'pub.dev'},
```

jonasfj (approved):

test/cache/preload_test.dart
sigurdm: Added a comment along those lines

parlough:
Sounds like we shouldn't document this on dart.dev, is that correct?

sigurdm:
> Sounds like we shouldn't document this on dart.dev, is that correct?

Yeah for now we keep this unannounced mainly for Flutter use.

PR #3637, opened 2022-11-07
dart pub get --enforce-lockfile

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2890

### DIFF
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -58,6 +58,9 @@ class SolveReport {
 
   /// Displays a report of the results of the version resolution in
   /// [_newLockFile] relative to the [_previousLockFile] file.
+  ///
+  /// Returns `true` if there was any change of dependencies relative to the old
+  /// lockfile.
   Future<bool> show() async {
     final hasChanges = await _reportChanges();
     _checkContentHashesMatchOldLockfile();


lib/src/solver/report.dart
```diff
@@ -199,9 +207,13 @@ $contentHashesDocumentationUrl
       for (var name in ordered(removed)) {
         await _reportPackage(name, alwaysShow: true);
       }
+      hasChanges = true;
     }
 
     log.message(_output);
+    if (enforceLockfile && hasChanges) {
+      dataError('Could not enforce the lockfile.');
```
jonasfj, line 215: Would it not be better if these error messages mentioned:
 * Resolution aborted: Unable to satisfy `pubspec.yaml` using the existing `pubspec.lock`
 * Try without `--enforce-lockfile` to update `pubspec.lock`.
 * X number of dependencies changed in `pubspec.lock`.
 * `pubspec.lock` does not exist (if that is the case).

Just some ideas for information that should go into this message.

---------------------------------

Interesting side note:

It's unclear if we actually should allow resolution to go ahead. I think it's the best implementation for us.
But it leaves the user with weird behavior, where `--enforce-lockfile` can cause a dependency conflict.
Which should never be possible, because `--enforce-lockfile` implies that I don't want resolution to be taking place at all.

But I'm not sure we want the number of hacks necessary to work around this.

lib/src/entrypoint.dart
```diff
@@ -352,22 +368,29 @@ class Entrypoint {
...
+      dryRun: dryRun,
+      enforceLockfile: enforceLockfile,
+      quiet: onlyReportSuccessOrFailure,
+    );
+
+    final hasChanges = await report.show();
+    await report.summarize();
+    if (enforceLockfile && hasChanges) {
+      dataError(
+          'Could not enforce the lockfile$suffix.${onlyReportSuccessOrFailure ? forDetails() : ''}');
```
jonasfj, line 386: I'm confused, shouldn't this say:
> Cannot not enforce `pubspec.lock`, try `dart pub get` without `--enforce-lockfile` to resolve dependencies.

lib/src/entrypoint.dart
```diff
@@ -337,11 +356,8 @@ class Entrypoint {
       });
     } catch (e) {
       if (onlyReportSuccessOrFailure && (e is ApplicationException)) {
-        final directoryOption = root.isInMemory || root.dir == '.'
-            ? ''
-            : ' --directory ${root.dir}';
         throw ApplicationException(
-            'Resolving dependencies$suffix failed. For details run `$topLevelProgram pub ${type.toString()}$directoryOption`');
+            'Resolving dependencies$suffix failed.${forDetails()}');
```
jonasfj, line 360: I don't think forDetails can be reused out of this flow. So why not move it in here.

lib/src/solver/report.dart
```diff
@@ -48,14 +50,22 @@ class SolveReport {
...
+        _quiet = quiet,
+        _enforceLockfile = enforceLockfile;
 
   /// Displays a report of the results of the version resolution in
   /// [_newLockFile] relative to the [_previousLockFile] file.
-  Future<void> show() async {
-    await _reportChanges();
-    await _reportOverrides();
+  ///
+
```
jonasfj, line 62: ???

sigurdm: Removed

lib/src/entrypoint.dart
sigurdm: I'm using it also below when the resolution succeeded, but something changed from the pubspec.lock.

lib/src/entrypoint.dart
sigurdm: The discrepancies will have been output by report.show+summarize at this point.

But adding the suggestions to try without `--enforce-lockfile` is good.

lib/src/solver/report.dart
sigurdm: > X number of dependencies changed in pubspec.lock.

This is mentioned by the report.summary.

> pubspec.lock does not exist (if that is the case).

That case is handled separately before even attempting to resolve.

lib/src/command/get.dart
```diff
@@ -28,6 +28,13 @@ class GetCommand extends PubCommand {
         negatable: false,
         help: "Report what dependencies would change but don't change any.");
 
+    argParser.addFlag(
+      'enforce-lockfile',
+      negatable: false,
+      help:
+          'Enforce pubspec.lock. Fail resolution if pubspec.lock does not satisfy pubspec.yaml',
```
jonasfj, line 35:
```suggestion
          'Enforce pubspec.lock, fails resolution if pubspec.lock does not satisfy pubspec.yaml',
```
Don't know, maybe I'm just crazy -- English is not my first language.

lib/src/solver/report.dart
```diff
@@ -48,14 +50,19 @@ class SolveReport {
...
+  })  : _dryRun = dryRun,
+        _quiet = quiet,
+        _enforceLockfile = enforceLockfile;
 
   /// Displays a report of the results of the version resolution in
   /// [_newLockFile] relative to the [_previousLockFile] file.
-  Future<void> show() async {
-    await _reportChanges();
-    await _reportOverrides();
+  Future<bool> show() async {
```
jonasfj, line 61: Write in documentation what it returns.

lib/src/solver/report.dart
```diff
@@ -380,38 +415,55 @@ $contentHashesDocumentationUrl
...
+
+  void warning(String message) {
+    if (_quiet) {
+      log.fine(message);
+    } else {
+      log.warning(message);
+    }
+  }
+
+  void message(String message) {
```
jonasfj, line 462: Do this need to be public?

jonasfj (approved):

lib/src/solver/report.dart
sigurdm: Nope

lib/src/solver/report.dart
sigurdm: Done

lib/src/command/get.dart
sigurdm: I think both are valid. Landing as is - we can follow up with a fix if you disagree.

Issue #3531, opened 2022-08-18
Create a README.md in the root of the PUB_CACHE


jonasfj:
Let's create a `README.md` file in the root of the `PUB_CACHE` which explains:
 * What the `PUB_CACHE` is for.
 * How to specify a different `PUB_CACHE`, using the `PUB_CACHE` environment variable.
 * What the implications of deleting the `PUB_CACHE` is.
 * How to repair the `PUB_CACHE` in case you have accidentally modified anything (`dart pub cache repair`)
 * How to delete the `PUB_CACHE` using `dart pub cache clean`.

It's entirely possible that the `$PUB_CACHE/README.md` should say something like:
```markdown
# Pub Package Cache

This folder is used by `dart pub` to store packages used in Dart / Flutter projects.
The contents should only be modified using the `dart pub` command.

TODO: What happens if you delete / clean  the folder.
TODO: You may want to delete it if it has grown too big, inorder to purge unused packages.

 * `dart pub cache repair`
 * `dart pub cache clean`

For more information, see:
https://dart.dev/go/pub-cache
```
or maybe just:
```markdown
See: https://dart.dev/go/pub-cache
```

And then have `https://dart.dev/go/pub-cache` redirect to either:
 * https://dart.dev/tools/pub/glossary#system-cache
 * https://dart.dev/tools/pub/cmd/pub-get#the-system-package-cache
 * Or make a new page and merge these two...

-----

Also we need to create a `https://dart.dev/go/pub-cache` page..

-----

We should check if the `$PUB_CACHE/README.md` exists every time we write a package to the `PUB_CACHE`. That way most invocations of `dart pub get` won't need to check if the file is there, because they just assert that packages exist. Only when a new resolution is done and packages are downloaded do we check if the `README.md` file exists (and write it if not).

PR #3650, opened 2022-11-14
Add a README.md to the pub cache after command ends

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3531


szakarias (approved):
LGTM

lib/src/system_cache.dart
```diff
@@ -270,6 +270,26 @@ class SystemCache {
...
+      writeTextFile(readmePath, '''
+# Pub Package Cache
+
+This folder is used by Pub to store packages used by Dart / Flutter
+projects.
+
+The contents should only be modified using the `dart pub`/`flutter pub`
+commands.
+
+For more information, see: https://dart.dev/go/pub-cache
```
jonasfj, line 289:
```suggestion
`PUB_CACHE` - Pub Package Cache
===============================

This folder is used by Pub to store cached packages used
in Dart / Flutter projects.

**DO NOT MODIFY THIS FOLDER**.

The contents of this folder should only be modified using the
`dart pub` and `flutter pub` commands.

For details on how manage the `PUB_CACHE`, see:
https://dart.dev/go/pub-cache
```

lib/src/system_cache.dart
```diff
@@ -270,6 +270,26 @@ class SystemCache {
 
     return latest;
   }
+
+  /// Write a README.md file in the root of the cache directory to document the
+  /// contents of the folder.
+  ///
+  void ensureReadme() {
+    final readmePath = p.join(rootDir, 'README.md');
```
jonasfj, line 278:
```suggestion
  }

 var _ensuredReadme = false;

  /// Write a README.md file in the root of the cache directory to document the
  /// contents of the folder.
  ///
  /// This should only be called when we are doing another operation that is
  /// modifying the `PUB_CACHE`. This ensures that users won't experience
  /// permission errors because we writing a `README.md` file, in a flow that
  /// the user expected wouldn't have issues with a read-only `PUB_CACHE`.
  void ensureReadme() {
    // Only ensure README once per execution
    if (_ensuredReadme) {
      return;
    }
    _ensuredReadme = true;
    
    final readmePath = p.join(rootDir, 'README.md');
```

lib/src/system_cache.dart
sigurdm: I wrote something like this but with less CAPS.

sigurdm:
PTAL

Issue #3641, opened 2022-11-09
assert up-to-date should run 'dart pub get' if the SDK has changed from what generated package_config.json


jonasfj:
assert up-to-date should run 'dart pub get' if the SDK has changed from what generated package_config.json

jonasfj:
and we should probably run "assert up-to-date" for `dart test/run/doc/analyze`

jonasfj:
Related to https://github.com/dart-lang/sdk/issues/50422

sigurdm:
Fixed in https://github.com/dart-lang/pub/pull/3652

PR #3652, opened 2022-11-18
Have executableForCommand rerun pub get if sdk changed minor version

sigurdm:
https://github.com/dart-lang/pub/issues/3641

Compares with the version in `.dart_tool/package_config.json['generatorVersion']`

Also if either previous version or this version of the sdk is a prerelease.


lib/src/entrypoint.dart
```diff
@@ -840,6 +845,25 @@ class Entrypoint {
...
+  /// Throws [DataException], if `.dart_tool/package_config.json` the version
+  /// changed sufficiently.
+  void _checkPackageConfigSameSdk() {
+    final generatorVersion = packageConfig.generatorVersion;
+    if (generatorVersion == null ||
+        generatorVersion.major != sdk.version.major ||
+        generatorVersion.minor != sdk.version.minor ||
+        generatorVersion.isPreRelease ||
+        sdk.version.isPreRelease) {
+      dataError('The sdk was updated since last package resolution');
```
jonasfj, line 863: Should we change to ensureUpToDate...
And avoid any risk that these messages become user visible

jonasfj (approved):

lib/src/entrypoint.dart
jonasfj: Probably a future refactor...

sigurdm: Yeah - I think we can do that when we provide the functionality from the external API.

PR #3656, opened 2022-11-21
Fix global activate --git-ref

sigurdm:
Fix of: https://github.com/dart-lang/pub/issues/3608

The problem was that using --shallow when cloning for retrieving the name was broken, it would not fetch that ref. 

The tests did not reveal this problem as it used a relative url to clone the repo, even with --shallow all refs would be available.

This PR skips the "clone to temp-dir" step and clones directly to the system cache, meaning that the same clone will be reused in the following steps.


lib/src/source/git.dart
```diff
@@ -196,17 +196,67 @@ class GitSource extends CachedSource {
...
-    return withTempDir((tempDir) async {
-      await _clone(repo, tempDir, shallow: true);
-      if (ref != null) await _checkOut(tempDir, ref);
-      var pubspec = Pubspec.load(p.join(tempDir, path), cache.sources);
-      return pubspec.name;
+    String url,
+    String? ref,
+    String? path,
+    SystemCache cache, {
+    required String containingDir,
```
jonasfj, line 207:
```suggestion
    required String relativeTo,
```

Just asking?

lib/src/source/git.dart
```diff
@@ -196,17 +196,67 @@ class GitSource extends CachedSource {
...
     });
   }
 
+  /// Lists the file as it is represented at the revision of [description].
+  ///
+  /// Assumes that revision is present in the cache already (can be done with
+  /// [_ensureRevision]).
+  Future<String> _showFileAtRevision(
+    GitResolvedDescription resolvedDescription,
+    String pathInProject,
```
jonasfj, line 234: Do you want to use this argument? Or rename the method to `pubspecAtRevision`

jonasfj (approved):

lib/src/source/git.dart
sigurdm: Good catch - yeah, we should use it. copy paste error :)

lib/src/source/git.dart
sigurdm: Yeah - that's right

a-siva:
@sigurdm This change appears to break Dart rolls into Flutter (see https://github.com/flutter/flutter/issues/116107) any ideas what the issue could be ?

PR #3658, opened 2022-11-21
Add env var flag to override stdout.hasTerminal for output animations

nehalvpatel:
Making a change necessary to get seamless stdout animation when running "pub get" from flutter_tools.
- related: https://github.com/flutter/flutter/issues/74700

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -426,6 +426,26 @@ bool get canUseAnsiCodes {
 /// Gets an ANSI escape if those are supported by stdout (or nothing).
 String getAnsi(String ansiCode) => canUseAnsiCodes ? ansiCode : '';
 
+/// Whether an environment variable overriding the [stdout.hasTerminal] check
+/// was passed.
+bool get forceTerminalOutput =>
+    Platform.environment.containsKey('_PUB_FORCE_TERMINAL_OUTPUT');
+
+/// Whether it makes sense to do stdout animation.
+///
+/// Checks if pub is in JSON output mode or if stdout has no terminal attached.
+/// The flutter tool sets an environment variable when running "pub get" that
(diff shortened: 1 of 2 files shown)


lib/src/utils.dart
```diff
@@ -426,6 +426,26 @@ bool get canUseAnsiCodes {
 /// Gets an ANSI escape if those are supported by stdout (or nothing).
 String getAnsi(String ansiCode) => canUseAnsiCodes ? ansiCode : '';
 
+/// Whether an environment variable overriding the [stdout.hasTerminal] check
+/// was passed.
+bool get forceTerminalOutput =>
+    Platform.environment.containsKey('_PUB_FORCE_TERMINAL_OUTPUT');
```
nehalvpatel, line 432: Alternative env var names:
- `_PUB_OVERRIDE_HAS_TERMINAL_CHECK`
- `_PUB_OVERRIDE_TERMINAL_CHECK`
- `_PUB_OVERRIDE_STDOUT_HAS_TERMINAL`

---

Also, we could use one of these existing flags from lib/command_runner.dart instead of a new env var. I'm not sure if there would be side effects for commands other than `pub get`, though.
https://github.com/dart-lang/pub/blob/97f638d091e1afa66f422d0b83f71a0d65692b4e/lib/src/command_runner.dart#L39-L44

sigurdm: We could also consider reusing the mechanism we have for deciding whether to show colors or not - I guess showing an animation could fall under the same umbrella...

nehalvpatel: True, yea I can take a look.

sigurdm:
Landing this in the hope that it will be useful as is.
If we end up not using it, it can be removed later as it is just an internal implementation

PR #3660, opened 2022-11-22
Suggest using 2.12 as lower-bound SDK constraint

### LINKED ISSUE #3659: 'dart pub get' gives a version constraint suggestion that is out of date
# Actual behavior
Create an empty directory, put a `pubspec.yaml` file:
```
name: testing_app

dependencies:
  test: ^1.22.0
```
Then run command:
```
$ dart --version
Dart SDK version: 2.18.4 (stable) (Tue Nov 1 15:15:07 2022 +0000) on "macos_arm64"

$ dart pub get
Resolving dependencies...
pubspec.yaml has no lower-bound SDK constraint.
You should edit pubspec.yaml to contain an SDK constraint:

environment:
  sdk: '>=2.10.0 <3.0.0'

See https://dart.dev/go/sdk-constraint
```

# Expected behavior
The suggested `environment.sdk` value should be `'>=2.12.0 <3.0.0'` as suggested by article https://dart.dev/null-safety#enable-null-safety. Otherwise, first-time users who didn't know about the recommended null-safety feature would not get null-safety by default.

Especially, that article specifically mentioned "Dart 3 and later will only support sound null safety" so it'd be better if Dart starts preparing as many new users as possible for that in order to minimize surprises.

jonasfj:
Fixes #3659

I'm down for us making progress on ^-syntax, but let's do this in a follow up. I think it's important that we roll this into 2.19.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -894,12 +894,12 @@ class Entrypoint {
   void _checkSdkConstraint(Pubspec pubspec) {
     final dartSdkConstraint = pubspec.dartSdkConstraint.effectiveConstraint;
     if (dartSdkConstraint is! VersionRange || dartSdkConstraint.min == null) {
-      // Suggest version range '>=2.10.0 <3.0.0', we avoid using:
+      // Suggest version range '>=2.12.0 <3.0.0', we avoid using:
       // [CompatibleWithVersionRange] because some pub versions don't support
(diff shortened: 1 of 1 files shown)


sigurdm (approved):
Change looks ok, but how about we suggest "compatible with the current sdk minor version"? Isn't that a better default?

PR #3663, opened 2022-11-23
Fix escapeShellArgument

sigurdm:
Just noticed that we are doing a useless replace. Second string needs to be raw.


jonasfj (approved):
r+

PR #3665, opened 2022-11-23
Only check for sdk updates during `resolveExecutable`

sigurdm:
This partially undoes https://github.com/dart-lang/pub/pull/3652 .

We realized that all `pub global activate`d packages would now be broken on sdk-update.

Still the situation with globally activated packages have other issues. But we need a coherent plan to resolve those.


szakarias (approved):
Lgtm

Issue #3639, opened 2022-11-08
Publish errors with custom lib/analysis_options.yaml


tenhobi:
When a package has custom `analysis_options.yaml` file in `lib/` (so other private packages can include it), and the file contains `analyzer.exlude` for some files that developers do not want to lint (`*.g.dart`, `*.freezed.dart`, ...), but want to preserve for the `pub publish`, it fails with an error:

```
Bad state: Unable to find the context to <path>\my_class.freezed.dart
package:pub/src/dart.dart 119:5                           AnalysisContextManager._getExistingSession
package:pub/src/dart.dart 83:23                           AnalysisContextManager.parse
package:pub/src/dart.dart 102:16                          AnalysisContextManager.parseImportsAndExports
package:pub/src/validator/strict_dependencies.dart 35:45  StrictDependenciesValidator._findPackages
dart:async                                                _SyncStarIterator.moveNext
package:pub/src/validator/strict_dependencies.dart 80:23  StrictDependenciesValidator._validateLibBin
package:pub/src/validator/strict_dependencies.dart 70:5   StrictDependenciesValidator.validate
package:pub/src/validator.dart 166:23                     Validator.runAll.<fn>
dart:async                                                Future.wait
package:pub/src/validator.dart 164:25                     Validator.runAll
===== asynchronous gap ===========================
package:pub/src/command/lish.dart 263:5                   LishCommand._validate
===== asynchronous gap ===========================
package:pub/src/command/lish.dart 236:19                  LishCommand.runProtected
```

When `analyzer.exclude` is removed, it works just fine. (not optimal, those files might start showing lints infos/warnings/...)

Renaming from `lib/analysis_options.yaml` to, for example, `lib/lints.yaml` works, and the `pub publish` process is successful. Linting work fine with both names, and major linting packages like [`pedantic` (deprecated)](https://pub.dev/packages/pedantic), [`effective_dart` (deprecated)](https://pub.dev/packages/effective_dart), [`Very Good Analysis`](https://pub.dev/packages/very_good_analysis), ..., also use `lib/analysis_options.yaml` files, so I assume it is not a forbidden naming.

> Flutter 3.3.7 • channel stable • https://github.com/flutter/flutter.git
> Framework • revision e99c9c7cd9 (7 days ago) • 2022-11-01 16:59:00 -0700
> Engine • revision 857bd6b74c
> Tools • Dart 2.18.4 • DevTools 2.15.0

sigurdm:
[...]

PR #3666, opened 2022-11-24
Ignore analysis_options.yaml everywhere

sigurdm:
Fix of: https://github.com/dart-lang/pub/issues/3639

The initialization of the analyzer is inspired by how we do it in pana: https://github.com/dart-lang/pana/blob/master/lib/src/tag/tagger.dart#L130

### DIFF
--- a/lib/src/dart.dart
+++ b/lib/src/dart.dart
@@ -36,7 +36,7 @@ class AnalysisContextManager {
                 optionsFile:
                     // We don't want to take 'analysis_options.yaml' files into
                     // account. So we replace it with an empty file.
-                    Platform.isWindows ? 'NUL' : '/dev/null',
+                    Platform.isWindows ? r'C:\NUL' : '/dev/null',
               ).first,
             )
             .currentSession;


lib/src/dart.dart
```diff
@@ -23,53 +20,26 @@ import 'io.dart';
...
+
+  AnalysisContextManager._(this.packagePath)
+      : _session = ContextBuilder()
+            .createContext(
+              contextRoot: ContextLocator().locateRoots(
+                includedPaths: [p.absolute(p.normalize(packagePath))],
+                optionsFile:
+                    // We don't want to take 'analysis_options.yaml' files into
+                    // account. So we replace it with an empty file.
+                    Platform.isWindows ? r'C:\NUL' : '/dev/null',
```
jonasfj, line 39: that works everywhere?

jonasfj (approved):

lib/src/dart.dart
sigurdm: At least on our CI bots...

sigurdm: Reading from a file called NUL on windows gives an empty file: https://ss64.com/nt/nul.html I use `C:\NUL` because the analyzer insists on absolute paths (WHY?)

Reading from /dev/null on posix gives an empty file: https://linuxhint.com/what_is_dev_null/ (near the bottom).

PR #3667, opened 2022-11-24
Support `--native-assets` in kernel compilation commands

dcharkes:
https://dart-review.googlesource.com/c/sdk/+/264842 introduces passing `--native-assets` to `gen_kernel`.

This PR enables passing native asset mapping files to the various pub commands compiling kernel, so that these arguments can be passed from dartdev.

Depends on:

* https://github.com/dart-lang/webdev/pull/1797


sigurdm (approved):
LGTM when the new version of frontend_server_client is published

PR #3670, opened 2022-11-29
Fix reading pubspecs from git

sigurdm:
Fix of https://github.com/flutter/flutter/issues/116107

I slightly refactored how we read a pubspec from a git revision without checking it out. But unfortunately made a mistake in how we join the lines coming from the git process stdout.

Unfortunately all tests we have of git uses a json descriptor that is not sensitive to removal of newlines.

### DIFF
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -254,7 +254,7 @@ class GitSource extends CachedSource {
       fail('Could not find a file named "$pathInCache" in '
           '${GitDescription.prettyUri(description.url)} $revision.');
     }
-    return lines.join();
+    return lines.join('\n');
   }
 
   @override
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -248,6 +248,8 @@ class GitSource extends CachedSource {
 
     late List<String> lines;
     try {
+      // TODO(sigurdm): We should have a `git.run` alternative that gives back
+      // a stream of stdout instead of the lines.
       lines = await git
           .run(['show', '$revision:$pathInCache'], workingDir: repoPath);
     } on git.GitException catch (_) {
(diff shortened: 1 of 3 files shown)


jonasfj (approved):

jonasfj:
Ideally, we shouldn't be calling `git` in a manner where we get back lines. I don't think it'll cause issues, I'm just imagining that the interface we're using here is designed for getting lines that we want to proxy as output.

It would be better to call `git` in a way where it writes the output to a temporary file. Or where we buffer all of stdout into a single buffer. So that we don't change the kinds of new-lines in the file.

I don't see why / how this would cause issues, but maybe windows + `.gitconfig` could cause hard to predict behavior in some cases.

lib/src/source/git.dart
```diff
@@ -254,7 +254,7 @@ class GitSource extends CachedSource {
       fail('Could not find a file named "$pathInCache" in '
           '${GitDescription.prettyUri(description.url)} $revision.');
     }
-    return lines.join();
```
jonasfj, line 257: Would it be wiser to make a temporary file and do `--output=tempFile` ?

sigurdm:
> Ideally, we shouldn't be calling git in a manner where we get back lines. I don't think it'll cause issues, I'm just imagining that the interface we're using here is designed for getting lines that we want to proxy as output.

I agree. However it has been like this since forever, and I would prefer changing as little as possible in this PR.
I added a TODO, such that we can look into this later.

Actually it seems there is already a more general todo for this here: https://github.com/dart-lang/pub/blob/13ca9de11069a40e03fb6a78fa56aacf3e7fc8d8/lib/src/io.dart#L1061

lib/src/source/git.dart
sigurdm: I think we should just not split the output in the first place. But let's do it later.

PR #3673, opened 2022-11-29
Use caret syntax in environment recommendations

sigurdm:


jonasfj (approved):
I don't mind, but I think this is better as a lint.

This affects all users, why should only package authors find this out, and why only when publishing.

Also we've only allowed it just now -- maybe give people sometime migrate before start nagging them.

jonasfj (approved):
Sorry, I read it wrong. This is awesome!

sigurdm:
Use ^-syntax if applicable when recommending constraints

PR #3676, opened 2022-11-30
Create output folder if it doesn't exist in the precompiler

dcharkes:
The exception is caught and printed, but not the stack-trace. So I have no idea which call site is responsible for omitting to create the output folder. This should likely be fixed at the call site.


sigurdm (approved):

dcharkes:
Thanks!

However, should we investigate which call site doesn't create the folder?
I'm not sure that this the right solution. If there's multiple concurrent invocations that share a parent folder that does not exist, this could race.

sigurdm:
Right - I guess I was too fast here

PR #3683, opened 2022-12-05
Use resident compiler to run tests

sigurdm:
This can be seen as a proof of concept as it is still slightly hacky. But the reduced time overhead makes that we might want to adapt it anyway - it will only affect ourselves.

There are 3 parts to this change:

* Use a resident compiler, such that we can rely on hot restart incremental compilation.
* Compile the tool/test.dart if needed with a small shell wrapper
* Invoke the test runner main directly from tool/test.dart saving another process invocation.

This together takes a "hot" test run of a single test with a single file change in the pub code from ~5 seconds from of ~25 seconds on my machine!

Downsides:
* Invoking the test runner is using private details from package:test
* Using the resident compiler requires us to skip frontend_server_client and talk directly to the compiler. I assume it has less guarantees of not breaking, but I'm not sure.
* Having a shell script on top of the dart script is annoying, but I think needed for now.
  In the future  we might be able to just `dart run` and get precompilation of the script, but that currently doesn't work when running a script not in `bin/`.
* It seems I need to jit-compile, as the test-runner relies on `Isolate.packageConfig` which is not available when aot-compiled.
  * which leads to us having to circumvent: https://github.com/dart-lang/sdk/issues/50615

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,11 +44,12 @@ To run pub from the Git repository, run:
 
 Before any change is made to pub, all tests should pass. To run a pub test, run:
 
-    dart tool/test.dart test/path/to_test.dart
+    tool/test.sh test/path/to_test.dart
 
 To run all tests at once, run:
 
-    dart tool/test.dart
+    tool/test.sh
+
 
 Changes to pub should be accompanied by one or more tests that exercise the new
 functionality. When adding a test, the best strategy is to find a similar test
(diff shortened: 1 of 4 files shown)


sigurdm:
Let's come back to this later.

Maybe we need to build this into `dart compile --resident`

PR #3684, opened 2022-12-06
Don't warn about overrides

sigurdm:
Using overrides is a common and supported use-case.

It is good to inform about them, but warning is too much.

Fixes: https://github.com/dart-lang/sdk/issues/50592

This is a bit of a band-aid. In general I think it is better to show no output on stderr at all when doing the implicit resolution no terminal is attached unless the resolution fails. But that can be done separately in a follow-up.


lib/src/entrypoint.dart
```diff
@@ -343,8 +343,7 @@ Try running `$topLevelProgram pub get` to create `$lockFilePath`.''');
     }
 
     if (!summaryOnly && hasPubspecOverrides) {
-      log.warning(
-          'Warning: pubspec.yaml has overrides from $pubspecOverridesPath');
+      log.message('pubspec.yaml has overrides from $pubspecOverridesPath');
```
jonasfj, line 346:
```suggestion
      log.message('Notice: pubspec.yaml has overrides from $pubspecOverridesPath');
```

jonasfj (approved):

lib/src/entrypoint.dart
sigurdm: I was thinking the same, but something in the back of my head reminded me of: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-empty-prose

> In particular, avoid saying "Note:". It adds nothing.

Not sure that rule counts for terminal output - but somehow I think it does...

jonasfj: I think it's different from source code comments.

I'm mostly thinking we should try to make it standout, so that it doesn't look like it's a continuation of the text on the previous line.

We could also write: 
```
(pubspec.yaml has overrides from $pubspecOverridesPath)
```

or 
```
hint: pubspec.yaml has overrides from $pubspecOverridesPath
```

What does it look like in normal output?

sigurdm: Yeah - making it stand out is a good argument. Let's go with "Notice:".

jonasfj: What we currently have is:
```
$dart pub get
Warning: pubspec.yaml has overrides from pubspec_overrides.yaml
Resolving dependencies... (1.0s)
  analyzer 0.41.2 (5.3.1 available)
  ...
  shelf_web_socket 0.2.4+1 (1.0.3 available)
! test 1.16.0 (overridden) (1.22.1 available)
  test_api 0.2.19 (0.4.17 available)
  web_socket_channel 1.2.0 (2.2.0 available)
  webkit_inspection_protocol 0.7.5 (1.2.0 available)
Warning: You are using these overridden dependencies:
! test 1.16.0 (1.22.1 available)
Got dependencies!
```

jonasfj: if `pubspec_overrides.yaml` is empty:
```
$ dart pub get
Warning: pubspec.yaml has overrides from pubspec_overrides.yaml
Resolving dependencies... 
Got dependencies!
```

jonasfj: If I have one override:
``` 
$ dart pub get
Warning: pubspec.yaml has overrides from pubspec_overrides.yaml
Resolving dependencies... 
! retry 3.0.0 (overridden) (3.1.0 available)
Warning: You are using these overridden dependencies:
! retry 3.0.0 (3.1.0 available)
Got dependencies!
```

jonasfj: I upgrade to latest version:
```
$ dart pub get
Warning: pubspec.yaml has overrides from pubspec_overrides.yaml
Resolving dependencies... 
Warning: You are using these overridden dependencies:
! retry 3.1.0
Got dependencies!
```

jonasfj (approved):

lib/src/entrypoint.dart
sigurdm: I have removed the `Warning: pubspec.yaml has overrides from pubspec_overrides.yaml` line and the `Warning: You are using these overridden dependencies:` list, and now instead always include overridden dependencies in the main update-list. And also mention if they are overridden in the pubspec_overrides.yaml file.

before:
```
> dart pub get
Warning: pubspec.yaml has overrides from pubspec_overrides.yaml
Resolving dependencies... (1.4s)
! meta 1.7.0 (overridden) (1.8.0 available)
Warning: You are using these overridden dependencies:
! meta 1.7.0 (1.8.0 available)
Got dependencies!
```
after:
```                                                                                                                                        
dart pub get
Resolving dependencies... (1.6s)
! meta 1.7.0 (overridden in ./pubspec_overrides.yaml) (1.8.0 available)
Got dependencies!
```

A lot less noisy!

sigurdm:
PTAL

lib/src/solver/report.dart
```diff
@@ -434,8 +426,11 @@ $contentHashesDocumentationUrl
     }
 
     // Highlight overridden packages.
-    if (isOverridden && highlightOverride) {
-      output.write(" ${log.magenta('(overridden)')}");
+    if (isOverridden) {
+      final location = _root.pubspec.usesOverridesFile
+          ? ' in ${p.join(_root.dir, Pubspec.pubspecOverridesFilename)}'
+          : '';
```
jonasfj, line 432: Shouldn't we only write this for packages are overridden in the pubspec_overrides.yaml file?

test/pubspec_overrides_test.dart
```diff
@@ -22,12 +23,14 @@ void main() {
...
+      final overridesPath = p.join('.', 'pubspec_overrides.yaml');
       await pubCommand(
         command,
-        warning:
-            'Warning: pubspec.yaml has overrides from pubspec_overrides.yaml\n'
-            'Warning: You are using these overridden dependencies:\n'
-            '! lib 2.0.0',
+        output: contains(
+          '''
+! lib 2.0.0 (overridden in $overridesPath)
```
jonasfj, line 31: Let's add another test cases where:
 * `foo` and `bar` is overriden in `pubspec.yaml`
 * `foo` is overriden in `pubspec_override.yaml`

Then we should check for:
```
! foo 2.0.0 (overridden in $overridesPath)
! bar 2.0.0 (overridden)
```

jonasfj (approved):

test/pubspec_overrides_test.dart
sigurdm: Added a case (but the override of foo is shadowed)

jonasfj (approved):

lib/src/solver/report.dart
sigurdm: overrides can only all come from the pubspec_overrides.yaml file or none of them

Issue #3686, opened 2022-12-07
Resolutions messages can be confusing in Dart 3


mit-mit:
During the roll attempts of Dart 3.0 alpha to Flutter, error messages like this one were seen by Flutter team members:

```
Because hello depends on google_maps_flutter >=0.0.2 <2.0.0-nullsafety 
which requires SDK version >=2.0.0-dev.47.0 <3.0.0, version solving failed.
```

The general conclusion from there was that the upper constraint of `<3.0.0` was the issue. But the real issue here is the lower constraint `>=2.0.0-dev.47.0`, which is below 2.12 and thus means this is a package without null safety support.

I think we should be more explicit in these and say something like this:

```
Because hello depends on google_maps_flutter >=0.0.2 <2.0.0-nullsafety 
which does not support null safety, version solving failed.
```

mit-mit:
The in-progress PR for this issue is https://github.com/dart-lang/pub/pull/3688

mit-mit:
Roll in https://dart-review.googlesource.com/c/sdk/+/276780

mit-mit:
Fixed in https://github.com/dart-lang/pub/pull/3723

mit-mit:
And rolling to SDK in https://dart-review.googlesource.com/c/sdk/+/278348

mit-mit:
Here's the output we ended up with in #3723:

```
Because hello depends on google_maps_flutter >=0.0.2 <2.0.0-nullsafety 
which does not support null safety, version solving failed.

The lower bound of "sdk: '>=2.10.0 <3.0.0'" must be 2.12.0 or higher to enable null safety.
For details, see https://dart.dev/null-safety
```

PR #3688, opened 2022-12-08
Solve-hint when a sdk-constraint is not nullable

sigurdm:
Proposed fix of https://github.com/dart-lang/pub/issues/3686


sigurdm:
First commit is a refactor of the appDir descriptor.

Consider reviewing second commit alone.

lib/src/solver/incompatibility_cause.dart
```diff
@@ -96,6 +97,11 @@ class SdkCause extends IncompatibilityCause {
 
   @override
   String? get hint {
+    if (sdk.isDartSdk &&
+        !LanguageVersion.fromSdkConstraint(constraint).supportsNullSafety &&
+        sdk.version! >= Version(3, 0, 0).firstPreRelease) {
+      return 'The constraint $constraint does not support null-safety.';
```
mit-mit, line 103: I think this may be easier to understand if we only mention the lower constraint?

lib/src/solver/incompatibility_cause.dart
```diff
@@ -100,7 +100,7 @@ class SdkCause extends IncompatibilityCause {
     if (sdk.isDartSdk &&
         !LanguageVersion.fromSdkConstraint(constraint).supportsNullSafety &&
         sdk.version! >= Version(3, 0, 0).firstPreRelease) {
-      return 'The constraint $constraint does not support null-safety.';
+      return 'The lower bound of "$constraint" does not opt into null safety.';
```
mit-mit, line 103: Maybe s/opt into/enable/?

sigurdm: Done

szakarias (approved):
LGTM

lib/src/solver/incompatibility_cause.dart
```diff
@@ -96,6 +97,11 @@ class SdkCause extends IncompatibilityCause {
 
   @override
   String? get hint {
+    if (sdk.isDartSdk &&
+        !LanguageVersion.fromSdkConstraint(constraint).supportsNullSafety &&
+        sdk.version! >= Version(3, 0, 0).firstPreRelease) {
+      return 'The lower bound of "$constraint" does not enable null safety.';
```
jonasfj, line 103:
```suggestion
      return 'The lower-bound of "$constraint" does not enable null safety.';
```

jonasfj (approved):

lib/src/solver/incompatibility_cause.dart
sigurdm: Whoops - I missed this. Will do a follow-up.

sigurdm: Actually researching, I think this should be spelled in two words.

Examples:

https://en.wikipedia.org/wiki/Upper_and_lower_bounds
https://www.dictionary.com/browse/lower-bound
https://www.geeksforgeeks.org/lower_bound-in-cpp/

PR #3689, opened 2022-12-08
Only output errors during implicit resolution

sigurdm:
Can be seen as a follow-up to https://github.com/dart-lang/pub/pull/3684

All calls to log.error are followed by non-success exit - so this will make a successful pub run silent inside `*unlessTerminal`.

This PR also introduces a mechanism for forcing terminal output during tests.

The added test only tests that log.message is suppressed; after https://github.com/dart-lang/pub/pull/3684 the only warning that can occur during resolution is if a content-hash has changed, and that seems a bit obscure to invoke in this test.


szakarias (approved):
LGTM

PR #3707, opened 2022-12-19
Remove --no-sound-null-safety from (global) run

sigurdm:
This flag will no longer have effect in dart 3.


lib/src/command/global_run.dart
```diff
@@ -32,8 +32,6 @@ class GlobalRunCommand extends PubCommand {
         help: 'Runs the executable in a VM with the given experiments enabled. '
             '(Will disable snapshotting, resulting in slower startup).',
         valueHelp: 'experiment');
-    argParser.addFlag('sound-null-safety',
-        help: 'Override the default null safety execution mode.');
```
jonasfj, line 36: Should we make it hidden and print an error message instead?

jonasfj (approved):

PR #3713, opened 2022-12-22
Remove features for migrating to null safety

sigurdm:
Dart 3 will only work with already migrated code. 2.19 should be used for any last migrations.

Removing:
- `dart pub outdated --mode=null-safety`
- `dart pub upgrade --null-safety`
- The mixed-mode validator (warns if you are publishing a mixed-mode package)

We keep the validator that gives a hint if the package you are publishing is the first null safe package - I believe this still makes sense to notify about even in a Dart 3 world.

Part of https://github.com/dart-lang/sdk/issues/50707 .


szakarias (approved):
LGTM

Issue #3700, opened 2022-12-16
Support add/remove on dependency overrides


nex3:
Currently, it's not possible to programmatically manipulate dependency overrides using `dart pub`. `dart pub add` and `dart pub remove` only touch the main and dev dependencies.

Dependency overrides are very useful when working on multiple interconnected repositories, in which case it's also useful to update dependencies between them automatically on release with CI tooling. But if a package is using dependency overrides, this isn't possible.

PR #3716, opened 2022-12-23
Allow adding and removing dependency overrides

sigurdm:
https://github.com/dart-lang/pub/issues/3700


lib/src/command/add.dart
```diff
@@ -319,20 +328,32 @@ Specify multiple sdk packages with descriptors.''');
...
 
   /// Split [arg] on ':' and interpret it with the flags in [argResult] either as
   /// an old-style or a new-style descriptor to produce a PackageRef].
   _ParseResult _parsePackage(String arg, ArgResults argResults) {
     var isDev = argResults['dev'] as bool;
-    if (arg.startsWith('dev:')) {
-      if (argResults.isDev) {
-        usageException("Cannot combine 'dev:' with --dev");
+    var isOverride = false;
+    final firstColon = arg.indexOf(':');
```
jonasfj, line 340: Let's do some better parsing here... So we don't have `dev:^1.2.3` acting as something that is added as dev-dependency... it's just weird.

If we try to match: `(?:(dev|override):)?(${identifierRegExp.pattern})(?::(.*))?$` then we'll have:
 * group 1: dev or override
 * group 2: package name
 * group 3: constraint / JSON

If it doesn't match this regexp, we can try to match: `(?:([a-z][a-z0-9_-]):)?(${identifierRegExp.pattern})(?::(.*))?$`, and if it matches, then we can suggest that the modifier in group 1, should be either `dev` or `override`.

Or we can just stick to say what the form is, instead of trying to match a second regex with the only purpose being that we provide a nicer error message.

sigurdm: Done

test/testdata/goldens/help_test/pub add --help.txt
```diff
@@ -29,12 +31,14 @@ For example:
...
+  * Add a dependency override:
+    `dart pub add 'override:foo:1.0.0`
   * Add a git dependency with a path and ref specified:
     `dart pub add \
       'foo:{"git":{"url":"../foo.git","ref":"branch","path":"subdir"}}'`
 
-Usage: pub add [options] [dev:]<package>[:descriptor] [[dev:]<package>[:descriptor]
-       ...]
+Usage: pub add [options] [dev:|override:]<package>[:descriptor]
+       [dev:|override:]<package>[:descriptor] ...]
```
jonasfj, line 41: We could also consider
```
Usage: pub add [options] [dev|override:]<package>[:descriptor]
       [dev|override:]<package>[:descriptor] ...]
```
or
```
Usage: pub add [options] [<dev|override>:]<package>[:descriptor]
       [<dev|override>:]<package>[:descriptor] ...]
```

or 
```
Usage: pub add [options] [<section>:]<package>[:descriptor]
       [<section>:]<package>[:descriptor] ...]
```

Or maybe, we omit this from the usage line, and explain it further down.
We might just be confusing usings.

lib/src/command/remove.dart
```diff
@@ -20,9 +20,19 @@ class RemoveCommand extends PubCommand {
...
+`dependencies` or `dev_dependencies` in `pubspec.yaml`.
+
+To remove a dependency override of a package prefix the package name with
+'override:'.
+''';
+
   @override
-  String get argumentsDescription => '<package>';
+  String get argumentsDescription =>
+      '[override:]<package1> [[override:]<package2>...]';
```
jonasfj, line 35: Should we consider:
```suggestion
  String get argumentsDescription => '<package1> [<package2>...]';
```
To keep the usage line simple. I guess it's wrong, I'm just tempted to :D

lib/src/command/remove.dart
```diff
@@ -108,21 +118,28 @@ class RemoveCommand extends PubCommand {
   }
 
   Pubspec _removePackagesFromPubspec(Pubspec original, Set<String> packages) {
+    final overrides = packages
+        .where((p) => p.startsWith('override:'))
+        .map((e) => e.substring('override:'.length));
+    final otherPackages = packages.where((p) => !p.startsWith('override:'));
```
jonasfj, line 124: This is argument parsing, why don't we do this in `runProtected()` and then pass the result to `_removePackagesFromPubspec` or `_writeRemovalToPubspec`

Maybe, the result can be `packagesToRemove` and `overridesToRemove` as two list of strings.

jonasfj (approved):

lib/src/command/add.dart
```diff
@@ -62,20 +64,22 @@ For example:
...
+    `$topLevelProgram pub add 'foo:{"path":"../foo"}'`
   * Add a hosted dependency:
-    `$topLevelProgram pub add 'foo{"hosted":"my-pub.dev"}'`
+    `$topLevelProgram pub add 'foo:{"hosted":"my-pub.dev"}'`
   * Add an sdk dependency:
-    `$topLevelProgram pub add 'foo{"sdk":"flutter"}'`
+    `$topLevelProgram pub add 'foo:{"sdk":"flutter"}'`
   * Add a git dependency:
-    `$topLevelProgram pub add 'foo{"git":"https://github.com/foo/foo"}'`
+    `$topLevelProgram pub add 'foo:{"git":"https://github.com/foo/foo"}'`
```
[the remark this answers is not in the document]
sigurdm: yes

Issue #3697, opened 2022-12-14
[pub] Unexpected behavior for sdk constraint solving with sdk 3.0.0 update


annagrin:
The sdk constraint  `sdk: ">=2.18.0 <3.0.0"` resolves, but `sdk: ">=2.10.0 <3.0.0"` does not with the current version of SDK `3.0.0-21.0.dev`

Is it intentional? I would expect them  both to fail to resolve.

Related: https://github.com/dart-lang/webdev/pull/1824

**pubscpec**

```
name: _test
version: 1.0.0
description: >-
  A fake package used for testing weak null-safety.
publish_to: none

environment:
  sdk: ">=2.18.0 <3.0.0"

dependencies:
  intl: ^0.17.0
  path: ^1.6.1

dev_dependencies:
  build_runner: '>=1.6.2 <3.0.0'
  build_web_compilers: '>=2.12.0 <4.0.0'
```

**Result**

```
➜  _test git:(annagrin) dart pub upgrade
...
  watcher 1.0.2
  web_socket_channel 2.2.0
  yaml 3.1.1
No dependencies changed.
1 package has newer versions incompatible with dependency constraints.
Try `dart pub outdated` for more information.
```

**pubspec**

```
name: _test
version: 1.0.0
description: >-
  A fake package used for testing weak null-safety.
publish_to: none

environment:
  sdk: ">=2.10.0 <3.0.0"

dependencies:
  intl: ^0.17.0
  path: ^1.6.1

dev_dependencies:
  build_runner: '>=1.6.2 <3.0.0'
  build_web_compilers: '>=2.12.0 <4.0.0'
```

**Result**

```
➜  _test git:(annagrin) dart pub upgrade
Resolving dependencies...
The current Dart SDK version is 3.0.0-21.0.dev.

Because _test requires SDK version >=2.10.0 <3.0.0, version solving failed.
```

annagrin:
Thanks @lrhn and @parlough for clarifying! My only problem with this is that it took me some time to understand what is happening with my tests - one was working and another didn't compile, and of course the main difference was the SDK constraint. I think it would be easier if the SDK constraint just failed to resolve in both cases and I had to update it.

I am not sure how common this situation would be for other people, as I am compiling weak null safety code using build runner to test that it is still working with current dwds.

mit-mit:
[...]

PR #3723, opened 2023-01-02
Refine Dart 3 no-null output

mit-mit:
Refine Dart 3 compatibility output

Fixes https://github.com/dart-lang/pub/issues/3697

### Output before
```
$ pub get
Resolving dependencies...
The current Dart SDK version is 3.0.0-76.0.dev.

Because oldapp requires SDK version >=2.9.0 <3.0.0, version solving failed.

The lower bound of ">=2.9.0 <3.0.0" does not enable null safety.
```

### Output after
```
$ ~/dev/pub/bin/pub.exe get
Resolving dependencies...
The current Dart SDK version is 3.0.0-76.0.dev.

Because oldapp doesn't support null safety, version solving failed.

The lower bound of ">=2.9.0 <3.0.0" must be 2.12.0 or higher for null safety.
```

### DIFF
--- a/lib/src/solver/incompatibility_cause.dart
+++ b/lib/src/solver/incompatibility_cause.dart
@@ -81,6 +81,10 @@ class SdkCause extends IncompatibilityCause {
   /// The SDK with which the package was incompatible.
   final Sdk sdk;
 
+  bool get noNullSafetyCause => sdk.isDartSdk &&
+        !LanguageVersion.fromSdkConstraint(constraint).supportsNullSafety &&
+        sdk.version! >= Version(3, 0, 0).firstPreRelease;
+
   @override
   String? get notice {
     // If the SDK is not available, then we have an actionable [hint] printed
@@ -97,10 +101,9 @@ class SdkCause extends IncompatibilityCause {
 
   @override
   String? get hint {
-    if (sdk.isDartSdk &&
-        !LanguageVersion.fromSdkConstraint(constraint).supportsNullSafety &&
-        sdk.version! >= Version(3, 0, 0).firstPreRelease) {
-      return 'The lower bound of "$constraint" does not enable null safety.';
+    if (noNullSafetyCause) {
+      return 'The lower bound of "$constraint" must be 2.12.0'
+      ' or higher for null safety.';
     }
     // If the SDK is available, then installing it won't help
     if (sdk.isAvailable) {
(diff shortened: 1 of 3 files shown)


mit-mit:
cd @sigurdm @jonasfj

test/dart3_sdk_constraint_hack_test.dart
```diff
@@ -142,7 +142,7 @@ void main() {
     });
     await pubGet(
       error: contains(
-        'The lower bound of ">=2.10.0 <3.0.0" does not enable null safety.',
+        'The lower bound of ">=2.10.0 <3.0.0" must be 2.12.0 or higher for null safety.',
```
sigurdm, line 145: perhaps:
```suggestion
        'The lower bound of ">=2.10.0 <3.0.0" must be 2.12.0 or higher to enable null safety.',
```

sigurdm, line 145: Should we suggest a 2.19 sdk for migrating?
```suggestion
        '''
 The lower bound of ">=2.10.0 <3.0.0" must be 2.12.0 or higher for null safety.
 
 Consider using a Dart 2.19 SDK to migrate to null safety.''',
```

sigurdm (approved):
I like it - a few suggestions - but LGTM

test/dart3_sdk_constraint_hack_test.dart
mit-mit: Maybe we're better off adding something like `For details, see https://dart.dev/go/null-safety` ?

sigurdm: Yes - that is even better!

mit-mit: Done

mit-mit: Done

lib/src/solver/incompatibility_cause.dart
```diff
@@ -81,6 +81,11 @@ class SdkCause extends IncompatibilityCause {
   /// The SDK with which the package was incompatible.
   final Sdk sdk;
 
+  bool get noNullSafetyCause =>
+      sdk.isDartSdk &&
+      !LanguageVersion.fromSdkConstraint(constraint).supportsNullSafety &&
+      sdk.version! >= Version(3, 0, 0).firstPreRelease;
```
jonasfj, line 87: I wish we could make this private, or something, but I don't see a good way to do it....

So I suggest we do this :D

jonasfj (approved):
I think we can land this, once we green-up the broken tests... I think tests are a bit broken on master, right now too.

So we'll land this once that's cleared up..

PR #3731, opened 2023-01-13
Allow synonyms 'pkg' and 'packages' for the pub command

sigurdm:
Fixes: https://github.com/dart-lang/sdk/issues/50971


jonasfj:
Actually, are we sure we want this?

sigurdm:
Yeah - let's decide that before landing... :)

sigurdm:
Updated this to just suggest using the `pub` command instead if you type eg. `dart packages get`.

jonasfj (approved):

Issue #3725, opened 2023-01-05
`dart pub token add` should sanity check the token


jonasfj:
We should validate that tokens consistent of characters that can be encoded in a header value.

Maybe we should stick to what is allowed in oauth bearer tokens:
```
     b64token    = 1*( ALPHA / DIGIT /
                       "-" / "." / "_" / "~" / "+" / "/" ) *"="
     credentials = "Bearer" 1*SP b64token
```

So maybe it just has to match `[a-zA-Z0-9._~+/=-]*`, then we won't get any weird errors from http.

sigurdm:
We should probably add this to the documentation too.

PR #3732, opened 2023-01-13
Verify token characters when added and when used

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3725

### DIFF
--- a/doc/repository-spec-v2.md
+++ b/doc/repository-spec-v2.md
@@ -135,7 +135,7 @@ details on token management see: `dart pub token --help`.
 
 The tokens are inserted verbatim in the header, therefore they have to adhere to
  https://www.rfc-editor.org/rfc/rfc6750#section-2.1. This means they must match
- the regex: `^[a-zA-Z0-9._~+/=-]*$`.
+ the regex: `^[a-zA-Z0-9._~+/=-]+$`.
 
 
 ### Missing Authentication or Invalid Token
--- a/doc/repository-spec-v2.md
+++ b/doc/repository-spec-v2.md
@@ -133,6 +133,10 @@ that the `<token>` is accidentally stored in shell history. For security reasons
 authentication can only be used when `<hosted-url>` uses HTTPS. For further
 details on token management see: `dart pub token --help`.
 
+The tokens are inserted verbatim in the header, therefore they have to adhere to
+ https://www.rfc-editor.org/rfc/rfc6750#section-2.1. This means they must match
+ the regex: `^[a-zA-Z0-9._~+/=-]*$`.
+
 
 ### Missing Authentication or Invalid Token
 If the server requires authentication and the request does not carry an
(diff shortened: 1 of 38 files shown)


doc/repository-spec-v2.md
```diff
@@ -133,6 +133,10 @@ that the `<token>` is accidentally stored in shell history. For security reasons
 authentication can only be used when `<hosted-url>` uses HTTPS. For further
 details on token management see: `dart pub token --help`.
 
+The tokens are inserted verbatim in the header, therefore they have to adhere to
+ https://www.rfc-editor.org/rfc/rfc6750#section-2.1. This means they must match
+ the regex: `^[a-zA-Z0-9._~+/=-]*$`.
```
jonasfj, line 138: nit, but we should not allow empty tokens? Should we?
```suggestion
 the regex: `^[a-zA-Z0-9._~+/=-]+$`.
```

lib/src/authentication/credential.dart
```diff
@@ -116,19 +117,26 @@ class Credential {
...
         );
       }
-      return Future.value('Bearer $value');
+      tokenValue = value;
+    } else {
+      tokenValue = token!;
+    }
+    if (!isValidBearerToken(tokenValue)) {
+      dataError('Credential token for $url is not a valid Bearer token. '
+          'It should match `^[a-zA-Z0-9._~+/=-]*\$`');
```
jonasfj, line 136:
```suggestion
          'It should match `^[a-zA-Z0-9._~+/=-]+\$`');
```

lib/src/authentication/credential.dart
```diff
@@ -144,6 +152,14 @@ class Credential {
   // Either [token] or [env] should be defined to be valid.
   bool isValid() => (token == null) ^ (env == null);
 
+  /// Whether [candidate] can be used as a bearer token.
+  ///
+  /// We limit tokens to be valid bearer tokens according to
+  /// https://www.rfc-editor.org/rfc/rfc6750#section-2.1
+  static bool isValidBearerToken(String candidate) {
+    return RegExp(r'^[a-zA-Z0-9._~+/=-]*$').hasMatch(candidate);
```
jonasfj, line 160:
```suggestion
    return RegExp(r'^[a-zA-Z0-9._~+/=-]+$').hasMatch(candidate);
```

doc/repository-spec-v2.md
jonasfj: Yeah, we don't allow empty tokens, see:
```
    if (token.isEmpty) {
      usageException('Token is not provided.');
    }
```

jonasfj (approved):
This is a potentially breaking change, let's make sure it's noted in CHANGELOG.

sigurdm:
> This is a potentially breaking change,

Hmm, before you would get an http error about an invalid header... not sure how breaking it is to catch it before - but sure let's put it in the changelog.

PR #3736, opened 2023-01-17
If pinned to an exact version, keep the pin when bumping

deivid-rodriguez:
We got the following issue reported at dependabot-core: https://github.com/dependabot/dependabot-core/issues/6446.

I think the request is reasonable and I believe it's what we do in other ecosystems.

I researched a bit and I fixed it by special casing exact pinning in pub's dependency services.

### DIFF
--- a/lib/src/command/dependency_services.dart
+++ b/lib/src/command/dependency_services.dart
@@ -606,6 +606,9 @@ Map<String, PackageRange>? _dependencySetOfPackage(
           : null;
 }
 
+/// Return a constraint compatible with [newVersion].
+///
+/// By convention if the original constraint is pinned we return [newVersion]. Otherwise use [VersionConstraint.compatibleWith].
 VersionConstraint _bumpConstraint(
   VersionConstraint original,
   Version newVersion,
--- a/lib/src/command/dependency_services.dart
+++ b/lib/src/command/dependency_services.dart
@@ -627,6 +627,9 @@ VersionConstraint _bumpConstraint(
   );
 }
 
+/// Return a constraint compatible with [newVersion], but including [original] as well.
+///
+/// By convention if the original constraint is pinned, we don't widen the constraint but return [newVersion] instead.
 VersionConstraint _widenConstraint(
   VersionConstraint original,
   Version newVersion,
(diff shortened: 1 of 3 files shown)


lib/src/command/dependency_services.dart
```diff
@@ -606,6 +606,24 @@ Map<String, PackageRange>? _dependencySetOfPackage(
           : null;
 }
 
+VersionConstraint _bumpConstraint(
```
sigurdm, line 609: Could you add a small dartdoc explaining
```suggestion
/// Return a version compatible with [newVersion].
///
/// By convention if the dependency is pinned in [originalVersion] we bump it to a pinned version. Otherwise use [VersionConstraint.compatibleWith].
VersionConstraint _bumpConstraint(
```

sigurdm:
Thanks - I wanted to get around to this at some point!

sigurdm:
Seems another golden file needs to be updated: test/testdata/goldens/dependency_services/dependency_services_test/Can update a git package.txt

Goldens can be updated with
```
_PUB_TEST_WRITE_GOLDEN=true tool/test.dart test/dependency_services/dependency_services_test.dart
```

deivid-rodriguez:
Thanks, running that command cause a lot of SHA hashes to get updated. Not sure if I was supposed to commit those updates too, but I located the one update related to this patch that I wanted and updated just that.

On top of that, I added the following changes:

* I tried to clarify the documentation you added because I believe it implied that the function returned always a version but it returns a version constraint in the general case.
* I added another update to also make widening requirements work in the same way. Widening requirements is normally used for libraries, and if a library has chosen an exact pin for a dependency, that probably has a reaso and Dependabot should not override that. Hopefully that makes sense.

sigurdm:
I think this looks good. Could you sign the CLA?

deivid-rodriguez:
@sigurdm Sorry for the silence here. I only have one personal Google account which I use only for personal stuff, so I'd like to not sign using that. The other Google account I own is from Github, but I need some paperwork to get clearance to sign the CLA using that. I'm currently working on that, will sign it as soon as I can!

sigurdm:
@deivid-rodriguez no worries - not in a hurry here, just wanted to not forget about this :)

sigurdm:
Another friendly ping: @deivid-rodriguez .

sigurdm:
Another friendly ping: @deivid-rodriguez .

deivid-rodriguez:
So sorry for the delay @sigurdm, I should be signed now, hopefully it's ok.

sigurdm:
No worries @deivid-rodriguez !

Issue #3738, opened 2023-01-15
dart 2.19 beta.4 reports incorrect version range


bsutton:
I have the following pubspec.yaml
```
name: library
version: 0.0.1
description: A simple command-line application created by dcli
environment: 
  sdk: '>=2.19.0-beta.1 <3.0.0'
dependencies: 
  dcli: 1.36.0
dev_dependencies: 
  lint_hard: ^1.0.0

```
Note the max sdk is < 3.0.0

When I run dart pub get

```
dart pub get
Resolving dependencies... 
The current Dart SDK version is 2.19.0-444.4.beta.

Because library requires SDK version >=2.19.0-beta.1 <4.0.0, version solving failed.

```

Note the upper version is displayed as < 4.0.0 rather than the expected < 3.0.0

FYI:
```
 dart --version
Dart SDK version: 2.19.0-444.4.beta (beta) (Tue Jan 10 10:24:44 2023 +0000) on "linux_x64"
```

sigurdm:
Yeah - this is because we decided to extend the compatibility forwards, such that
dart sdk constraints of `>=2.12.0 <3.0.0` gets interpreted as `>=2.12 <4.0.0`.

We only do this for constraints that enable null-safety:
dart sdk constraints of `>=2.9.0 <3.0.0` gets interpreted as `>=2.9.0 <3.0.0`.

The logic was implemented here: https://github.com/dart-lang/pub/pull/3572

Original issue here: https://github.com/dart-lang/pub/issues/3554

The intent is that you will not see the rewritten constraint in error messages (after dart 3 such a constraint is always matched with the current sdk) but for dart 2.19 this can happen... I think I did not really consider that case.

I wonder if this is serious enough that we should try to hotfix it for 2.19? The hotfix would probably be to revert https://github.com/dart-lang/pub/pull/3572 such that it only takes effect in dart 3.

It will never cause a resolution to fail that would otherwise have succeeded, only it will have a confusing error message.

@mit-mit WDYT?

jonasfj:
@sigurdm, won't this also be a source of confusion in 3.x?


I think at one-point we considered doing:
```
Because library requires SDK version >=2.19.0-beta.1 <3.0.0 (interpreted as <4.0.0), version solving failed.
```

I don't recall why we abandoned it, probably it was messy to carry the information around.

Maybe, we could do something like:
```
The current Dart SDK version is 2.19.0-444.4.beta.
[...]

PR #3739, opened 2023-01-20
Only rewrite sdk constraints after dart 3

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3738

This should be hot-fixed onto 2.19 if we can make it (otherwise it'll have no effect)


szakarias (approved):
LGTM

Issue #3363, opened 2022-03-24
Version solving errors are hard to parse


gaaclarke:

vsmenon:
Actually, not sure I triaged this correctly.  I missed these are SDK packages.  Are those handled differently?

gaaclarke:
The pubspec.yaml is in the dart sdk itself: https://github.com/dart-lang/sdk/blob/main/pkg/kernel/pubspec.yaml

The big problem is that it hard to read because there is no delineation between package names and pubspec concepts and English.  For example, I didn't author the pubspec so the idea that `expect` is a package name wasn't even on my radar.  I still have a hard time understanding what it says.

PR #3741, opened 2023-01-20
Use bold to distinguish package names in solve-traces

sigurdm:
Attempt at fixing: https://github.com/dart-lang/pub/issues/3363 by using ansi colors to highlight package names in the solve-trace. This should make it a bit easier to read - especially if packages have keyword-like names.

An alternative would be to prefix package names with `package:`...

This has an issue with how we cannot nest colorization of ansi colors...

Draft to get feedback for the idea if it is worth handling implementing ansi-colorization.

image

### DIFF
--- a/lib/src/package_name.dart
+++ b/lib/src/package_name.dart
@@ -4,6 +4,7 @@
 
 import 'package:pub_semver/pub_semver.dart';
 
+import 'log.dart';
 import 'package.dart';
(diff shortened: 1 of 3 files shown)


sigurdm:
@jonasfj, @szakarias WDYT?

sigurdm:
I still have some hope for this.

Thanks for reminding me

PR #3745, opened 2023-01-24
Invoke _post_pub_get hook after running get command within a Flutter package

nehalvpatel:


lib/src/command/get.dart
```diff
@@ -60,6 +67,50 @@ class GetCommand extends PubCommand {
     );
   }
 
+  bool shouldRunPostGetHook() {
+    final flutterRoot = Platform.environment['FLUTTER_ROOT'];
+    if (flutterRoot == null) {
+      return false;
+    }
+    if (!fileExists(p.join(flutterRoot, 'version'))) {
```
sigurdm, line 75: Reuse the logic in https://github.com/dart-lang/pub/blob/9796ba9364a3168eac87362f7f222e3ccba310e9/lib/src/sdk/flutter.dart#L23 to determine if flutter is available.

lib/src/command/get.dart
```diff
@@ -78,6 +129,12 @@ class GetCommand extends PubCommand {
       enforceLockfile: argResults['enforce-lockfile'],
     );
 
+    if (!argResults['dry-run'] && shouldRunPostGetHook()) {
+      log.message('Running post get hook...');
+      final exitCode = await runPostGetHook();
```
sigurdm, line 134: I think this belongs at the end of Entrypoint.acquireDependencies. It has to run for everything that makes a resolution (get, upgrade, downgrade, add, remove)

lib/src/command/get.dart
```diff
@@ -60,6 +67,50 @@ class GetCommand extends PubCommand {
...
+
+    return hasFlutterDependency;
+  }
+
+  Future<int> runPostGetHook() async {
+    final String flutterRoot = Platform.environment['FLUTTER_ROOT']!;
+    final String flutterToolPath = p.join(flutterRoot, 'bin', 'flutter');
+
+    final StreamSubscription<ProcessSignal> subscription =
+        ProcessSignal.sigint.watch().listen((e) {});
```
sigurdm, line 94: Why are we doing this? We should at least have a comment explaining this.

lib/src/command/get.dart
sigurdm: Entrypoint.acquireDependencies also has logic for determining if there where changes - and only in that case should the hook run.

lib/src/command/get.dart
```diff
@@ -60,6 +67,50 @@ class GetCommand extends PubCommand {
...
+      [
+        'pub',
+        '_post_pub_get',
+        '-C',
+        directory,
+        '--update-version-and-package-config',
+        '--regenerate-platform-specific-tooling',
+        argResults['example'] ? '--example' : '',
+      ],
+      mode: ProcessStartMode.inheritStdio,
```
sigurdm, line 106: Is there ever a need for the hook to be interactive?
I would prefer the hook to be run in `ProcessStartMode.normal`.

lib/src/command/get.dart
```diff
@@ -60,6 +67,50 @@ class GetCommand extends PubCommand {
...
+    return hasFlutterDependency;
+  }
+
+  Future<int> runPostGetHook() async {
+    final String flutterRoot = Platform.environment['FLUTTER_ROOT']!;
+    final String flutterToolPath = p.join(flutterRoot, 'bin', 'flutter');
+
+    final StreamSubscription<ProcessSignal> subscription =
+        ProcessSignal.sigint.watch().listen((e) {});
+    final Process process = await Process.start(
```
sigurdm, line 95: I think if we can run this non-interactively you can use `Process.run` instead.

lib/src/command/get.dart
```diff
@@ -60,6 +67,50 @@ class GetCommand extends PubCommand {
...
+    final hasFlutterDependency =
+        entrypoint.root.dependencies.values.any((package) {
+      return package.name == 'flutter' &&
+          package.source.runtimeType == SdkSource;
+    });
+
+    return hasFlutterDependency;
+  }
+
+  Future<int> runPostGetHook() async {
```
sigurdm, line 89: This should probably not return the exit-code but handle it on its own. Eg. by throwing an exception if non-zero.

lib/src/command/get.dart
```diff
@@ -60,6 +67,50 @@ class GetCommand extends PubCommand {
...
+        ProcessSignal.sigint.watch().listen((e) {});
+    final Process process = await Process.start(
+      flutterToolPath,
+      [
+        'pub',
+        '_post_pub_get',
+        '-C',
+        directory,
+        '--update-version-and-package-config',
+        '--regenerate-platform-specific-tooling',
```
sigurdm, line 103: Can we leave the specific logic out of pub? The ideal should be that the hook is a black box from pub's perspective.
```suggestion
```

sigurdm (commented):
Some initial comments

lib/src/command/get.dart
nehalvpatel: Sounds good!

lib/src/command/get.dart
nehalvpatel: I've added an exception with some simple text.

lib/src/command/get.dart
nehalvpatel: Sure, I think I'll add output to the flutter tool that will say which steps it is performing.

lib/src/command/get.dart
nehalvpatel: Now that I think about it, I'm not sure why I had that either 😄.

lib/src/entrypoint.dart
```diff
@@ -1017,4 +1024,55 @@ See https://dart.dev/go/sdk-constraint
...
+
+  /// Checks if the Flutter SDK is available and whether this is a Flutter app
+  /// package.
+  bool shouldRunFlutterPackageHook() {
+    if (!FlutterSdk().isAvailable) {
+      return false;
+    }
+
+    // TODO: sky_engine check
+    final hasFlutterDependency = root.dependencies.values.any((package) {
```
sigurdm, line 1036: We should make sure this also works for transitive dependencies. (Maybe a test-case?)

lib/src/entrypoint.dart
```diff
@@ -435,6 +437,11 @@ Unable to satisfy `$pubspecPath` using `$lockFilePath`$suffix.${forDetails()}$su
         log.exception(error, stackTrace);
       }
     }
+
+    if (!dryRun && hasChanges && shouldRunFlutterPackageHook()) {
+      log.message('Running Flutter package hook in ${root.dir}...');
```
sigurdm, line 442: I think this should be `log.info`

lib/src/entrypoint.dart
nehalvpatel: Gotcha! looking into it

mosuem:
Closing this, as it is stale and the files are out of date. Feel free to reopen in case you still want to land it!

Issue #3552, opened 2022-09-06
Warnings in `dart pub publish` for Dart 3


jonasfj:
Consider:
 * Publish warning if not Dart 3 compatible and using >=2.12.0 <3.0.0 as SDK constraint (to either change code or use different SDK constraint)
 * Publish hint if Dart 3 compatible and using <3.0.0, suggest using <4.0.0 as SDK constraint

sigurdm:
The first bullet is done.

I doubt we want to do the third.

sigurdm:
Second bullet is fixed by https://github.com/dart-lang/pub/pull/3752

PR #3752, opened 2023-01-26
Info when publishing <3.0.0 constraint that is interpreted as <4.0.0

sigurdm:
Fix of https://github.com/dart-lang/pub/issues/3552

### DIFF
--- a/lib/src/validator/sdk_constraint.dart
+++ b/lib/src/validator/sdk_constraint.dart
@@ -53,9 +53,9 @@ class SdkConstraintValidator extends Validator {
               errors.isEmpty &&
               originalConstraint != effectiveConstraint) {
         hints.add('''
-The declared sdk constraint is '$originalConstraint', this is interpreted as '$effectiveConstraint'.
+The declared SDK constraint is '$originalConstraint', this is interpreted as '$effectiveConstraint'.
 
-Consider updating the sdk constraint to:
+Consider updating the SDK constraint to:
 
 environment:
   sdk: '$effectiveConstraint'
--- a/lib/src/validator/sdk_constraint.dart
+++ b/lib/src/validator/sdk_constraint.dart
@@ -46,9 +46,13 @@ class SdkConstraintValidator extends Validator {
             'See https://dart.dev/tools/pub/publishing#publishing-prereleases '
             'For more information on pre-releases.');
       }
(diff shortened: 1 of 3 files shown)


szakarias (approved):

lib/src/validator/sdk_constraint.dart
```diff
@@ -58,6 +46,21 @@ class SdkConstraintValidator extends Validator {
...
+      if (
+          // We only want to give this hint if there was no other problems with
+          // the sdk constraint.
+          warnings.isEmpty &&
+              errors.isEmpty &&
+              originalConstraint != effectiveConstraint) {
+        hints.add('''
+The declared sdk constraint is '$originalConstraint', this is interpreted as '$effectiveConstraint'.
+
+Consider updating the sdk constraint to:
```
jonasfj, line 58: Do we normally capitalize SDK..

```suggestion
The declared SDK constraint is '$originalConstraint', this is interpreted as '$effectiveConstraint'.

Consider updating the SDK constraint to:
```

lib/src/validator/sdk_constraint.dart
```diff
@@ -58,6 +46,21 @@ class SdkConstraintValidator extends Validator {
             'See https://dart.dev/tools/pub/publishing#publishing-prereleases '
             'For more information on pre-releases.');
       }
+      if (
+          // We only want to give this hint if there was no other problems with
+          // the sdk constraint.
+          warnings.isEmpty &&
+              errors.isEmpty &&
+              originalConstraint != effectiveConstraint) {
```
jonasfj, line 54: This will only happen in Dart 3.x, right? and this change won't make it into 2.x?

jonasfj (approved):

lib/src/validator/sdk_constraint.dart
jonasfj: Context: I don't think we want people to write `<4.0.0` before they are using a 3.0.0-... SDK.

sigurdm: Yes, unless we cherry-pick it to the 2.19-series. The sdk main branch is already in dart 3-land.

lib/src/validator/sdk_constraint.dart
sigurdm: I guess so.

PR #3754, opened 2023-01-27
Normalize pub.dartlang.org to pub.dev when reading lockfile

sigurdm:
Fix of https://github.com/flutter/flutter/issues/119091

This PR ensures that HostedDescription is always created with a validated and normalized url. That was missing when creating a description from a pubspec.lock file - causing pub to consider the locked package coming from an unrelated host, and thus unlocking it.

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -51,7 +51,7 @@ const contentHashesDocumentationUrl = 'https://dart.dev/go/content-hashes';
 /// backwards compatibility with `pubspec.lock`-files which contain
 /// `https://pub.dartlang.org`.
 ///
-/// Throws [FormatException] if there is anything wrong [hostedUrl].
+/// Throws [FormatException] if there is anything wrong with [hostedUrl].
 ///
 /// [1]: ../../../doc/repository-spec-v2.md
 Uri validateAndNormalizeHostedUrl(String hostedUrl) {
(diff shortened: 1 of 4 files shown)


sigurdm:
This will need a CP into the dart 2.19 series.

lib/src/source/hosted.dart
```diff
@@ -195,8 +195,8 @@ class HostedSource extends CachedSource {
   /// Returns a reference to a hosted package named [name].
   ///
   /// If [url] is passed, it's the URL of the pub server from which the package
-  /// should be downloaded. [url] most be normalized and validated using
-  /// [validateAndNormalizeHostedUrl].
+  /// should be downloaded. [url] will be normalized and validated using
+  /// [validateAndNormalizeHostedUrl]. This can throw a [FormatException].
```
jonasfj, line 199: It probably meant to say "must" before.

jonasfj (approved):

lib/src/source/hosted.dart
jonasfj: I'm guessing you changed the behavior, that's fine.

sigurdm: yes

lib/src/source/hosted.dart
```diff
@@ -269,7 +269,7 @@ class HostedSource extends CachedSource {
       name,
       version,
       ResolvedHostedDescription(
-        HostedDescription(name, Uri.parse(url).toString()),
+        HostedDescription(name, url),
```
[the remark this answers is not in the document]
sigurdm: :( oh another problem with the migration - the workaround is of course to deactivate and activate the package again.

Not sure we will get a hotfix out for this... but highly annoying

Issue #3530, opened 2022-08-16
pubspec is generated in non human friendly form  (`{ }` json notation)


maximveksler:
# Environment

- Dart version (run `dart --version`): `Dart SDK version: 2.17.6 (stable) (Unknown timestamp) on "linux_x64"`
- OS kind and version (e.g. "Windows 10, version 1809" or "macOS 12.4"): `Ubuntu 21.04`
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

The result of `flutter pub add` results in pubspec yaml that is not human readble. 

Please see https://github.com/maximveksler/flutter_pubspec_bug_reproduce for a working reproduction scenario.

## Expected behavior

`dependencies:` section should be in yaml human friendly form.

```yaml
name: todos_api
description: The interface and models for an API providing access to todos.
version: 1.0.0+1
publish_to: none

environment:
  sdk: ">=2.17.0 <3.0.0"

dependencies:
  equatable: ^2.0.3
  json_annotation: ^4.6.0
  meta: ^1.8.0
  uuid: ^3.0.6

dev_dependencies:
  build_runner: ^2.0.0
  json_serializable: ^6.0.0
  test: ^1.17.0
  very_good_analysis: ^3.0.1
```

## Actual behavior

`dependencies:` section is json normalized and is not pretty printed.

```yaml
name: todos_api
description: The interface and models for an API providing access to todos.
version: 1.0.0+1
publish_to: none

environment:
  sdk: ">=2.17.0 <3.0.0"


dev_dependencies:
  build_runner: ^2.0.0
  json_serializable: ^6.0.0
  test: ^1.17.0
  very_good_analysis: ^3.0.1
dependencies: {equatable: ^2.0.3, json_annotation: ^4.6.0, meta: ^1.8.0, uuid: ^3.0.6}
```

maximveksler:
Applying the command in sequence does not produce the negative effect.

I.e. 
```
flutter pub remove equatable
flutter pub add equatable

flutter pub remove json_annotation
flutter pub add json_annotation

flutter pub remove meta
flutter pub add meta

flutter pub remove uuid
flutter pub add uuid
```

Produces a readble pubspec yaml (as exptected).

```yaml
name: todos_api
description: The interface and models for an API providing access to todos.
version: 1.0.0+1
publish_to: none

environment:
  sdk: ">=2.17.0 <3.0.0"

dependencies:
  equatable: ^2.0.3
  json_annotation: ^4.6.0
[...]

PR #3758, opened 2023-01-29
`dart pub remove`: Remove top-level key if empty after operation.

raunaqarora21:
The changes made to this code improve the way dependencies are removed from the pubspec.yaml file. The previous version of the code only removed the package from the dependencies or dev_dependencies key, but it didn't check if the key was now empty after the removal. The updated version now also checks if the key is empty after the removal, and if it is empty, it removes the key as well. This ensures that the pubspec.yaml file is kept clean and organized.

Fixes https://github.com/dart-lang/pub/issues/3530

### DIFF
--- a/lib/src/command/remove.dart
+++ b/lib/src/command/remove.dart
@@ -147,7 +147,7 @@ class RemoveCommand extends PubCommand {
           found = true;
           // Check if the dependencies or dev_dependencies map is now empty
           // If it is empty, remove the key as well
-          if (dependenciesNode.isEmpty) {
+          if ((yamlEditor.parseAt([dependencyKey]) as YamlMap).isEmpty) {
             yamlEditor.remove([dependencyKey]);
           }
         }
--- a/lib/src/command/remove.dart
+++ b/lib/src/command/remove.dart
@@ -143,13 +143,13 @@ class RemoveCommand extends PubCommand {
 
         if (dependenciesNode is YamlMap &&
             dependenciesNode.containsKey(package)) {
-          if (dependenciesNode.length == 1) {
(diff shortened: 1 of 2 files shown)


sigurdm:
Hi - thanks for working on this.

Could you try adding a test for this new behaviour.

The test should go in: test/remove/remove_test.dart

raunaqarora21:
Hi @sigurdm 
I have added the test case in: test/remove/remove_test.dart
But it is not working as expected, so can you please review and suggest the changes if required

lib/src/command/remove.dart
```diff
@@ -143,13 +143,13 @@ class RemoveCommand extends PubCommand {
 
         if (dependenciesNode is YamlMap &&
             dependenciesNode.containsKey(package)) {
-          if (dependenciesNode.length == 1) {
+          yamlEditor.remove([dependencyKey, package]);
+          found = true;
+          // Check if the dependencies or dev_dependencies map is now empty
+          // If it is empty, remove the key as well
+          if (dependenciesNode.isEmpty) {
```
sigurdm, line 150: I don't think the `dependenciesNode` gets mutated by the `yamlEditor.remove([dependencyKey, package]);` above.

You could probably replace this with something like:
```suggestion
          if ((yamlEditor.parseAt([dependencyKey]) as YamlMap).isEmpty) {
```

raunaqarora21:
@sigurdm - Thanks for the suggestion. 
Now the test case is getting passed

sigurdm (approved):
LGTM

sigurdm:
Thanks for the contribution!

Issue #3772, opened 2023-02-03
Null Exception when deactivating pub global with case sensitive


Anas35:
## Steps to produce
1- `dart pub global activate coverage`
2- `dart pub global deactivate Coverage` (Make any character to uppercase for the word coverage)

## Exception
```
Null check operator used on a null value
package:pub/src/global_packages.dart 331:37            GlobalPackages.deactivate
package:pub/src/command/global_deactivate.dart 32:18   GlobalDeactivateCommand.runProtected
package:pub/src/command.dart 190:45                    PubCommand.run.<fn>
dart:async                                             new Future.sync
package:pub/src/utils.dart 111:12                      captureErrors.wrappedCallback
dart:async                                             runZonedGuarded
package:pub/src/utils.dart 128:5                       captureErrors
package:pub/src/command.dart 190:13                    PubCommand.run
package:args/command_runner.dart 212:27                CommandRunner.runCommand
package:dartdev/dartdev.dart 239:30                    DartdevRunner.runCommand
package:args/command_runner.dart 122:25                CommandRunner.run.<fn>
dart:async                                             new Future.sync
package:args/command_runner.dart 122:14                CommandRunner.run
package:dartdev/dartdev.dart 64:31                     runDartdev
C:\b\s\w\ir\x\w\sdk\pkg\dartdev\bin\dartdev.dart 11:9  main
This is an unexpected error.
```

## Expected: 
either `No active package Coverage` or` Did you mean coverage?`

## Version
Pub version: 3.0.0-200.0.dev
PUB_HOSTED_URL: <not set>
PUB_CACHE: "<not set>"
Platform: windows

AbhishekSatpathy4848:
This exception is also thrown when you try to activate some packages which have an uppercase character.
For example, <code>dart pub global activate Matrix</code> also throws a similar exception.
or
<code>dart pub global activate mAtrix</code>

sigurdm:
Yeah - I think this happens because of windows case insensitive filesystem.

PR #3774, opened 2023-02-06
Fix `global deactivate` on case-insensitive but case-preserving file systems

sigurdm:
Fix of https://github.com/dart-lang/pub/issues/3772


szakarias (approved):
LGTM

Issue #2012, opened 2018-12-07
Publish validation is slow


munificent:
I notice that when I publish a package, the validation step takes about 10-15 seconds. That's long enough that the process looks hung without any output. If we can't speed this up, I think it's worth adding either a little progress bar, or perhaps printing out the name of each validation step as its performed. That way users know something is going on.

jakemac53:
My assumption here is the script that is doing the validation is running from source - it got slow with Dart 2.

munificent:
Is it not just part of the pub executable itself?

natebosch:
I think it is part of the executable.

It is doing a lot of work though, the more Dart code in your package the slower it gets.

jakemac53:
Did we add more checks recently or something? This was definitely not an issue before. I thought maybe it was invoking some binary from package:pana or something but that is just random speculation :joy:

sigurdm:
@munificent did the validation time suddenly take more time or it has been like this always?

munificent:
It's a recent thing. Validation used to be nearly instantaneous. Maybe it was this change: https://github.com/dart-lang/pub/commit/9f1e9521b335c2772ec2b48090d8b4fba211a1a1? That looks like it might be doing full analysis where before I think the validation was just syntactic.

PR #3776, opened 2023-02-06
Add a spinner while validating before publishing

sigurdm:
We are now (purposefully) doing a full `dart analyze` before publication - and that potentially takes a long time.

Fixes: https://github.com/dart-lang/pub/issues/2012


szakarias (approved):
LGTM

Issue #3781, opened 2023-02-08
dart pub publish doesn't take pubspec_overrides.yaml into account


tamas-p:
* In my monorepo (https://github.com/tamas-p/hisma) I have a circular dependency, but only by dev_dependencies:
   * dependencies: hisma_visual_monitor -----> hisma
   * dev_dependencies: hisma - - -> hisma_visual_monitor
 * Since using Flutter 3.7.0 (dart 2.19.0) trying `dart pub publish` in the hisma package gives me:
```
Warning: pubspec.yaml has overrides from pubspec_overrides.yaml
Resolving dependencies...
Because hisma depends on hisma_visual_monitor ^0.3.0 which doesn't match any versions, version solving failed.
```
 * Error message is correct in a sense, that yes, that version of that package in the dev_dependencies section has not yet been published (since I am doing the publishes by the `dependencies` graph, first the root, then the children).
 * if I test with Flutter 3.3.10 (dart 2.18.6), or earlier release, it works:
```
...
...
Publishing is forever; packages cannot be unpublished.
Policy details are available at https://pub.dev/policy

Do you want to publish hisma 0.3.0 to https://pub.dartlang.org (y/N)? 
```
 * Was the dependency check at `dart pub publish` extended to dev_dependencies on purpose at 3.7.0 (2.19.0)?

```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.7.1, on Ubuntu 20.04.5 LTS 5.15.0-58-generic, locale en_US.UTF-8)
[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
[✓] Chrome - develop for the web
[✓] Linux toolchain - develop for Linux desktop
[✓] Android Studio (version 4.2)
[✓] VS Code
[✓] Connected device (2 available)
[✓] HTTP Host Availability

• No issues found!
```

kuhnroyal:
I have the same issue

sigurdm:
We used to not require dependencies to resolve before publishing.

Now we require it, so we can run `dart analyze` before publishing.
To resolve against a soon-to-be-published package use a dependency_overrides entry.

See: https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#pub-1

> Update dart pub publish to require a working resolution. If publishing a breaking release of mutually dependent packages use dependency_overrides to obtain a resolution.

kuhnroyal:
[...]

PR #3782, opened 2023-02-09
Consider pubspec_overrides.yaml when publishing

sigurdm:
Fix of https://github.com/dart-lang/pub/issues/3781 .

### DIFF
--- a/lib/src/command.dart
+++ b/lib/src/command.dart
@@ -75,12 +75,7 @@ abstract class PubCommand extends Command<int> {
   ///
   /// This will load the pubspec and fail with an error if the current directory
   /// is not a package.
-  late final Entrypoint entrypoint =
-      Entrypoint(directory, cache, withPubspecOverrides: withPubspecOverrides);
-
-  /// Whether `pubspec_overrides.yaml` is taken into account, when creating
-  /// [entrypoint].
-  bool get withPubspecOverrides => true;
+  late final Entrypoint entrypoint = Entrypoint(directory, cache);
 
   /// The URL for web documentation for this command.
   String? get docUrl => null;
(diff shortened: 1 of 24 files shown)


kuhnroyal:
How can I test this?

sigurdm:
Clone the repo, and [check out the PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally)

Then run `dart <path-to-repo>/bin/pub.dart publish`

kuhnroyal:
Ok this now also gives the cyclic warning but doesn't fail anymore.
The warning references the `pubspec.yaml` file which is missleading.

```
* Your pubspec.yaml is overriding non-dev dependencies.
```

sigurdm:
> The warning references the pubspec.yaml file which is missleading.

Good catch - we could specialize here like we recently did for https://github.com/dart-lang/pub/pull/3684 .

sigurdm:
@jonasfj, when reviewing, consider looking at each commit separately.

The second commit does a rename, because I eventually want to migrate all validation tests to be integration tests.

jonasfj (approved):

jonasfj:
Thanks for splitting into commits, that made it easy to review all the refactoring :D

jakemac53:
+1 the whole appeal for us is it gives us something like `dependency_overrides` but it only applies to local development and not publishing

natebosch:
> As far as I know the only examples of circular dependencies involve circular dependencies.

I did just introduce a non-dev circular dependency between `test_api` and `matcher`, but I'm going to break the cycle soon. I agree that non-dev dependency cycles are not a use case we need to focus on.

sigurdm:
> As far as I know the only examples of circular dependencies involve circular dependencies.

Haha - perfect mistake :)

Issue #2183, opened 2019-08-05
Print shorter login URL in the terminal


jonasfj:
It's sometimes hard to copy the URL from the terminal when authorizing `pub publish`. The URL has a lot of parameters and is very long, thus, some terminal emulators might break the lines and make it hard to copy.

This causes issues such as: https://github.com/dart-lang/pub-dev/issues/2611

As @isoos suggested in https://github.com/dart-lang/pub-dev/issues/2611#issuecomment-518245602 we could mitigate by printing `http://localhost:<port>/login` in the terminal, and when the user opens this URL we can redirect to the correct sign-in URL, which then in turn redirects back to `localhost`.

This makes the URL to be copied much shorter, and doesn't introduce any state on `pub.dev`.
If the user can't open `localhost` in the browser, the login flow won't work anyways, as this is the callback URL.

jonasfj:
Note. we might have to test if this will work. Maybe the browser won't allow this kind of redirect, or maybe `accounts.google.com` will reject it due to the `Referrer` header. I find it unlikely, but we should test it.

sigurdm:
(From the linked PR)
> This has potential downsides. Specifically when using dart pub login over a SSH, see:
https://dart.dev/tools/pub/troubleshoot#localhost-unreachable-after-sign-in

> That workaround will work, but this will make the workaround much harder.

> We can consider https://developers.google.com/identity/protocols/oauth2/limited-input-device

PR #3786, opened 2023-02-13
Make login-url easier to copy-paste via localhost redirect

sigurdm:
Fix of https://github.com/dart-lang/pub/issues/2183

Manually confirmed this works on chrome-mac-os.

### DIFF
--- a/lib/src/oauth2.dart
+++ b/lib/src/oauth2.dart
@@ -250,8 +250,14 @@ Future<Client> _authorize() async {
   // the code is received.
   var completer = Completer();
   var server = await bindServer('localhost', 0);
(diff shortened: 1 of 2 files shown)


jonasfj:
This has potential downsides. Specifically when using `dart pub login` over a SSH, see:
https://dart.dev/tools/pub/troubleshoot#localhost-unreachable-after-sign-in

That workaround will work, but this will make the workaround much harder.


---------------------


We could also consider doing the following flow:
https://developers.google.com/identity/protocols/oauth2/limited-input-device

sigurdm:
Good point - I guess we should not do this then (we could show both urls, but that could be very confusing) - I'll update the issue.

PR #3788, opened 2023-02-14
Drop --use-data-isolate-strategy flag for tests

jakemac53:
This flag is being deprecated, and it will make tests load much slower. I tried running locally without it and the tests pass now so I think we can drop it.

Note that we will have the `--compiler source` flag which is similar to this, if indeed the bots are failing we can try that.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -70,5 +70,5 @@ jobs:
           cat "$_TESTS_FILE.${{ matrix.shard }}"
         shell: bash
       - name: Run tests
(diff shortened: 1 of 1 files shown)


jakemac53:
This appears to make the CI runs significantly faster as well (this one finished >5 minutes faster than recent ones)

sigurdm (approved):
LGTM

Not sure why we used this flag in the first place, but if we can finish presubmits 5 minutes faster that would be really nice.

jakemac53:
Yeah I couldn't remember exactly why I just remember that some tests were failing. But it seems they no longer fail so 🤷‍♂️

Issue #3730, opened 2023-01-11
Get doesn't fail correctly on some non-null-safe SDK constraints


mit-mit:
Current `pub` rejects the SDK constraint `sdk: '>=2.9.0 <3.0.0'` with:

```
$ dart pub get
Resolving dependencies...
The current Dart SDK version is 3.0.0-109.0.dev.

Because oldapp doesn't support null safety, version solving failed.

The lower bound of "sdk: '>=2.9.0 <3.0.0'" must be 2.12.0 or higher to enable null safety.
For details, see https://dart.dev/null-safety
```

This is as expected.

But resolves with no errors or warnings if using `sdk: '>=2.9.0 <4.0.0'`, which is not expected.

It should fail on any lower constraint that isn't 2.12 or above, regardless of what the upper constraint is.

sigurdm:
How about packages that contain only assets, and thus wants to declare that they work both before and after null-safety?

sigurdm:
I believe we would already get an analyzer warning in 3.0 on any code inside a library with such a constraint. Such a warning would also show up before publishing.

@jonasfj suggest refusing these at the server (@isoos suggests if they contain any dart files).

mit-mit:
> How about packages that contain only assets, and thus wants to declare that they work both before and after null-safety?

I don't think we support a single package for those. In Dart 3 and beyond, everything has null safety.

mit-mit:
From in-office discussion:
- `>=2.12.0 <3.0.0` succeeds as the Dart 3 backwards compatibility overrides the upper constraint (`<3.0.0`) when the lower constraint has null safety (2.12.0 or later) and the upper constraint is exactly `3.0.0`
- `>=2.9.0 <3.0.0` fails as the lower constraint is outside the backwards compatibility window
- `>=2.9.0 <4.0.0` succeeds as as 3.0.0 is within the listed range

In essence, the current `pub` behavoir is to validate matches the current SDK is within the listed range. There is no validation of language versions in the solver.

This presents an issue as the analyzer only looks for invalid language versions *outside the entry point*. When pub resolves the `>=2.9.0 <4.0.0` constraint, it write this `package_config.yaml`:

```
    {
      "name": "oldapp",
      "rootUri": "../",
      "packageUri": "lib/",
      "languageVersion": "2.9"
    }

```

[...]

PR #3800, opened 2023-02-23
Don't allow non-null-safety constraints in the root pubspec

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/3730


sigurdm:
Turns out this makes some checks we had obsolete (we now no longer supports pubspecs old enough to not be compatible with flutter etc.)

jonasfj (approved):

PR #3806, opened 2023-02-28
Support lfs and submodules

sigurdm:
Use a worktree instead of a full clone to check out each revision for a git source.

This seems to enable git lfs to just work somehow
This should also reduce the disk consumption, as we don't have to repeat the `.git` folder.

Also run `git submodule update --init --recursive` in each worktree after checking it out.

This should enable submodules. It is a nop, and quite fast if there are no submodules.

Still some kinks to work out with backwards compatibility, but I think it could be made to work at least upwards. (doing `cache repair` with an older dart sdk will probably break).

### DIFF
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -375,8 +375,11 @@ class GitSource extends CachedSource {
       final path = description.path;
       await _revisionCacheClones.putIfAbsent(revisionCachePath, () async {
         if (!entryExists(revisionCachePath)) {
-          await _clone(_repoCachePath(description, cache), revisionCachePath);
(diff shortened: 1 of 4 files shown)


sigurdm:
TODO: write a spec for the cache layout

sigurdm:
We should also find out and document which version of git we require for this to work.

sigurdm:
We should find out how far back git includes worktree support, and require that as a minimum git version.

mit-mit:
@sigurdm can we close this now that we completed https://github.com/dart-lang/pub/pull/4761 ?

PR #3812, opened 2023-03-02
remove an extra period from a publish message

devoncarew:
- remove an extra period from a publish message
- use the `pluralize` method in a few places

This fixes the message `'Package has 0 warnings and 1 hint..'` on a pub publish.


sigurdm (approved):
Thanks!

Issue #3799, opened 2023-02-23
Cannot find git executable when running `dart pub get`


SharpnelXu:
# Environment

- Dart version: 
  - Dart SDK version: 2.19.2 (stable) (Tue Feb 7 18:37:17 2023 +0000) on "windows_x64"
- OS kind and version: 
  - Windows 10
- Are you using the Chinese community mirror or a corporate firewall?
  - No

# Problem

Cannot find git executable when running `dart pub get`. Below is the part of the output of `flutter pub get -v`

```
IO  : Spawning "cmd /c git --version" in D:\github\chaldea\.
IO  : Finished git. Exit code 0.
    | stdout:
    | | Active code page: 936
    | | git version 2.39.2.windows.1
    | Nothing output on stderr.
IO  : Spawning "cmd /c git.cmd --version" in D:\github\chaldea\.
IO  : Finished git.cmd. Exit code 1.
    | stdout:
    | | Active code page: 936
    | stderr:
    | | 'git.cmd' is not recognized as an internal or external command,
    | | operable program or batch file.
SLVR: Version solving took 0:00:00.156820 seconds.
    | Tried 1 solutions.
FINE: Resolving dependencies finished (0.232s).
ERR : Cannot find a Git executable.
    | Please ensure Git is correctly installed.
```

This is actually caused by my CMD printing the output of `chcp 936` on startup. I fixed this issue referring to https://github.com/dart-lang/pub/issues/3645 & https://blog.csdn.net/qq_37925231/article/details/110824760

However, I was thinking if it is possible to relax the git output check, which is showed in #3645 that must start exactly with `git`, and the output must be exactly one line.
```
    // from lib/src/git.dart
    if (result.stdout.length != 1) return false;
    final output = result.stdout.single;
    final match = RegExp(r'^git version (\d+)\.(\d+)\.').matchAsPrefix(output);
```

sigurdm:
Yeah - I think it makes sense to just require the output to contain this anywhere.

PR #3814, opened 2023-03-05
Fix wrong action when executing git command

setcy:
Fixes: #3799 

## background
If someone has configured settings such as ‘autorun’ to automatically run commands in the terminal, the output of 'git version' may exceed one line, which will lead to incorrect judgment of the original code.

Therefore, in this PR, we search for the git version information in each line of the output. If any line matches the standard git version output format, we will consider it as the version information of git.

### DIFF
--- a/lib/src/git.dart
+++ b/lib/src/git.dart
@@ -145,15 +145,13 @@ bool _tryGitCommand(String command) {
     // produce additional output, so we need to look for "git version"
     // in every line of the output.
     Match? match;
-    String? versionString;
     for (var line in output) {
       match = RegExp(r'^git version (\d+)\.(\d+)\.').matchAsPrefix(line);
       if (match != null) {
-        versionString = line;
         break;
       }
     }
-    if (match == null || versionString == null) return false;
+    if (match == null) return false;
 
     // Git seems to use many parts in the version number. We just check the
     // first two.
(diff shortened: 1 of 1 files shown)


lib/src/git.dart
```diff
@@ -153,7 +163,7 @@ bool _tryGitCommand(String command) {
       // We just warn here, as some features might work with older versions of
       // git.
       log.warning('''
-You have a very old version of git (version ${output.substring('git version '.length)}),
+You have a very old version of git (version ${versionString.substring('git version '.length)}),
 for $topLevelProgram it is recommended to use git version 2.14 or newer.
```
sigurdm, line 167: Could you change this to:
```suggestion
You have a very old version of git (version ${major}.${minor}),
for $topLevelProgram it is recommended to use git version 2.14 or newer.
```
And then I don't think we need the `versionString` variable.

setcy: Glad to do this.

sigurdm: Ah - from the failed test I see why we had to do it the other way... The version string can be longer than just major.minor.

So I take my suggestion back. Could you revert, but please set the `versionString` to the substring value when it is stored (so the name fits better). Also it seems it should not be needed to check both `match` and `versionString` against null, just one of them.

setcy: Done

sigurdm (approved):
Nice - thanks!

PR #3824, opened 2023-03-13
Remove support for legacy credentials file

sigurdm:
We stopped creating this in dart 2.15

https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#pub-4

One might say we should support this longer - but I think Dart 3.0 is a good place to make this change. And also the consequence is only the user would have to log in again.

### DIFF
--- a/lib/src/oauth2.dart
+++ b/lib/src/oauth2.dart
@@ -5,7 +5,6 @@
 import 'dart:async';
 import 'dart:io';
 
-import 'package:collection/collection.dart' show IterableExtension;
 import 'package:http/http.dart' as http;
 import 'package:http/retry.dart';
 import 'package:path/path.dart' as path;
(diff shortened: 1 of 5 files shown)


lib/src/oauth2.dart
```diff
@@ -211,26 +204,12 @@ void _saveCredentials(SystemCache cache, Credentials credentials) {
 
 /// The path to the file in which the user's OAuth2 credentials are stored.
 ///
-/// This used to be PUB_CACHE/credentials.json. But the pub cache is not the
-/// best place for storing secrets, as it might be shared.
-///
-/// To provide backwards compatibility we use the legacy file if only it exists.
-///
 /// Returns `null` if there is no good place for the file.
 String? _credentialsFile(SystemCache cache) {
```
jonasfj, line 208: We don't need `SystemCache` as argument

jonasfj (approved):

lib/src/oauth2.dart
sigurdm: Good catch. Removed

PR #3833, opened 2023-03-14
Remove barback, build and serve commands

sigurdm:
These were only exposed in dartdev as hidden commands that printed an error.


kevmoo:
LOL! Talk about legacy. CC @munificent and @nex3 – 🖖

jonasfj (approved):

munificent:
*David Attenborough voice:* Here we witness one of nature's greatest spectacles: the birth, life, and death of a software library. As the dying library decays, its bits will feed the next generation of libraries, ensuring the ecosystem continues to thrive.

PR #3842, opened 2023-03-20
Adjust topic pubspec field in list of valid pubspec keys

parlough:
According to https://github.com/dart-lang/pana/commit/d5ef51f3466289ee8f545ff93de9b69775b32ed6, it seems the field was intended to be singular (`topic:`).


szakarias:
The tag for search filters is singular but it is not singular in the pubspec. But thanks for keeping an eye out.

parlough:
Ah, taking a closer look, that's clear now. Thanks for the clarification!

PR #3844, opened 2023-03-21
Api entrypoint for ensuring resolution is up-to-date

sigurdm:
Also makes `dart pub deps` not complain about the resolution not being up-to-date. Instead it just resolves (implicit pub get).

Part of https://github.com/dart-lang/sdk/issues/50422


sigurdm:
Also makes dart pub deps not complain about the resolution not being up-to-date. Instead it just resolves (implicit pub get).

PR #3850, opened 2023-03-27
Faster ensureUpToDate with path-dependencies

sigurdm:
By comparing the timestamp of the pubspec.yaml of dependencies with that of the package_config.json that points to them we avoid parsing the pubspec.lock and pubspec.yaml just because there are path-dependencies.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -669,9 +669,12 @@ To update `$lockFilePath` run `$topLevelProgram pub get`$suffix without
     }
 
     // Manually parse the lockfile because a full YAML parse is relatively slow
-    // and this is on the hot path for "pub run".
+    // and this is on the hot path for [ensureUpToDate].
     var lockFileText = readTextFile(lockFilePath);
     var hasPathDependencies = lockFileText.contains('\n    source: path\n');
+    if (hasPathDependencies) {
+      if (!_arePackagesAvailable(packageConfigStat)) return false;
+    }
 
     var lockFileModified = lockFileStat.modified;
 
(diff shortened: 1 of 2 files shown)


lib/src/entrypoint.dart
```diff
@@ -797,26 +800,36 @@ To update `$lockFilePath` run `$topLevelProgram pub get`$suffix without
...
-
-      // Get the directory.
-      var dir = cache.getDirectory(package, relativeFrom: '.');
-      // See if the directory is there and looks like a package.
-      return fileExists(p.join(dir, 'pubspec.yaml'));
-    });
+  /// Determines whether all of the packages in the [packageConfig] are
+  /// older than [packageConfigStat].
+  bool _arePackagesAvailable(FileStat packageConfigStat) {
+    // When a package has path-dependencies, they can mutate independently of
```
szakarias, line 806: Unfinished comment.

lib/src/entrypoint.dart
```diff
@@ -797,26 +800,41 @@ To update `$lockFilePath` run `$topLevelProgram pub get`$suffix without
...
-      // Get the directory.
-      var dir = cache.getDirectory(package, relativeFrom: '.');
-      // See if the directory is there and looks like a package.
-      return fileExists(p.join(dir, 'pubspec.yaml'));
-    });
+  /// Determines whether all of the packages in the [packageConfig] are
+  /// older than [packageConfigStat].
+  ///
+  /// The [packageConfigStat] is passed as an optimization to avoid statting the
+  /// same file again.
```
jonasfj, line 807: Before it used to do:
```
  /// Determines whether all of the packages in the lockfile are already
  /// installed and available.
```

that it doesn't anymore, because you ignore missing pubspecs

sigurdm:
Closing, in favor of: https://github.com/dart-lang/pub/commit/44bd80e4aaa1a60d61900422d4ac857bf5003d55

PR #3854, opened 2023-03-30
Fix label name in needs-info action

sigurdm:


szakarias (commented):
Did we agree to do this or change the label name instead?

sigurdm:
Let's change the label name

PR #3875, opened 2023-04-17
Improve message when sdk constraint is not null-safety enabled

sigurdm:
Trying to be less convoluted, and avoiding suggestions that are not really helpful.


mit-mit (approved):
Thanks!

Issue #3874, opened 2023-04-14
dart pub outdated shows invalid message when pubspec.lock contains a retracted version of a package


maRci002:
https://pub.dev/packages/shared_preferences_android/versions has a retraced version: [2.1.1](https://pub.dev/packages/shared_preferences_android/versions/2.1.1)

`pubspec.lock`

```
  shared_preferences_android:
    dependency: transitive
    description:
      name: shared_preferences_android
      sha256: "1bc5d734b109c327b922b0891b41fc51483ccbd53c0d19952b7e230349a5d90b"
      url: "https://pub.dev"
    source: hosted
    version: "2.1.1"
```

Running `dart pub outdated` / `flutter pub outdated`:

```
Showing outdated packages.
[*] indicates versions that are not the latest available.              
                                                                       
Package Name                Current   Upgradable  Resolvable  Latest   
                                                                       
direct dependencies:                                                   
collection                  *1.17.0   *1.17.0     *1.17.0     1.17.1   
intl                        *0.17.0   *0.17.0     *0.17.0     0.18.0   
                                                                       
dev_dependencies: all up-to-date.                                      

transitive dependencies:   
async                       *2.10.0   *2.10.0     *2.10.0     2.11.0
characters                  *1.2.1    *1.2.1      *1.2.1      1.3.0
js                          *0.6.5    *0.6.5      *0.6.5      0.6.7
matcher                     *0.12.13  *0.12.13    *0.12.13    0.12.15
material_color_utilities    *0.2.0    *0.2.0      *0.2.0      0.3.0
meta                        *1.8.0    *1.8.0      *1.8.0      1.9.1
path                        *1.8.2    *1.8.2      *1.8.2      1.8.3
shared_preferences_android  *2.1.1    2.1.0       2.1.0       2.1.0
source_span                 *1.9.1    *1.9.1      *1.9.1      1.10.0
test_api                    *0.4.16   *0.4.16     *0.4.16     0.5.1
win32                       *3.1.4    *3.1.4      *3.1.4      4.1.3

transitive dev_dependencies:
cli_util                    *0.3.5    *0.3.5      *0.3.5      0.4.0
pub_updater                 *0.2.4    *0.2.4      *0.2.4      0.3.0

1 upgradable dependency is locked (in pubspec.lock) to an older version.
To update it, use `dart pub upgrade`.
```

When running `dart pub upgrade` nothing happens.

arftylmz:
same issue

[...]

PR #3879, opened 2023-04-17
Allow `dart upgrade` to downgrade away from retracted version.

sigurdm:
This is by considering retracted versions last when resolving.

Attempt at solving one part of https://github.com/dart-lang/pub/issues/3874

### DIFF
--- a/lib/src/solver/package_lister.dart
+++ b/lib/src/solver/package_lister.dart
@@ -166,7 +166,20 @@ class PackageLister {
     // non-prerelease version if one exists, or the latest prerelease version
     // otherwise.
     PackageId? bestPrerelease;
-    for (var id in _isDowngrade ? versions : versions.reversed) {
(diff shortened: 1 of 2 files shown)


lib/src/solver/package_lister.dart
```diff
@@ -166,7 +166,20 @@ class PackageLister {
...
+    // We want `dart pub upgrade` to avoid that version, even if it means
+    // downgrading.
+    if (_allowedRetractedVersion != null) {
+      for (var i = 0; i < ordered.length; i++) {
+        if (ordered[i].version == _allowedRetractedVersion) {
+          final retracted = ordered.removeAt(i);
+          ordered.add(retracted);
+        }
+      }
+    }
```
jonasfj, line 181: This is a local priority in the solver, not a real priority.

I'm not sure this is a good idea.

sigurdm:
We decided not to do this.

Instead we'll fix outdated, so it calculates Upgradable with the current lock-file

Issue #3901, opened 2023-05-01
dart pub publish --skip-validation


jonasfj:
Let's add a `--skip-validation` so that people can side step validations, should they cause issues for the user. At the end of the day you can side-step checks using a different SDK version or by modifying the SDK. The only thing that can really reject a package is the server.

```
dart pub publish --help
Publish the current package to pub.dev.

Usage: dart pub publish [options]
-h, --help               Print this usage information.
-n, --dry-run            Validate but do not publish the package.
-f, --force              Publish without confirmation if there are no errors.
    --skip-validation    Publish without validation (this will ignore errors).
-C, --directory=<dir>    Run this in the directory<dir>.

Run "dart help" to see global options.
See https://dart.dev/tools/pub/cmd/pub-lish for detailed documentation.
```

PR #3904, opened 2023-05-01
Flag `pub publish --skip-validation` to publish without resolution and validation

sigurdm:
An escape-hook if we have mistakes in our validations.

Fixes https://github.com/dart-lang/pub/issues/3901

### DIFF
--- a/lib/src/command/lish.dart
+++ b/lib/src/command/lish.dart
@@ -66,6 +66,8 @@ class LishCommand extends PubCommand {
   /// Whether the publish requires confirmation.
   bool get force => argResults['force'];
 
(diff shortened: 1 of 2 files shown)


lib/src/command/lish.dart
```diff
@@ -251,7 +257,13 @@ the \$PUB_HOSTED_URL environment variable.''',
           'pubspec.');
     }
 
-    await entrypoint.acquireDependencies(SolveType.get, analytics: analytics);
+    if (!argResults['skip-validation']) {
```
jonasfj, line 260: Maybe do a getter for `argResult['skip-validation']`

jonasfj (approved):

Issue #3907, opened 2023-05-04
Give resolution hint when a package is pinned by flutter.


sigurdm:
* “package:foo is pinned by Flutter; see [github.com/dart-lang/foo/issues](http://github.com/dart-lang/foo/issues) for possible solutions”
* “package:foo is pinned by Flutter; see flutter.dev/pinned-packages for possible solutions”
* “ …; see pub.dev/flutter-pinned-packages for possible solutions”

PR #3914, opened 2023-05-11
Hint when pinned by flutter

sigurdm:
Fix of: https://github.com/dart-lang/pub/issues/3907

### DIFF
--- a/lib/src/solver/incompatibility_cause.dart
+++ b/lib/src/solver/incompatibility_cause.dart
@@ -51,7 +51,7 @@ class DependencyIncompatibilityCause extends IncompatibilityCause {
       if (targetConstraint is Version) {
         return '''
 Note: ${target.name} is pinned to version $targetConstraint by ${depender.name} from the ${dependerDescription.sdk} SDK.
-See https://dart.dev/go/version-pinning for details.
+See https://dart.dev/go/sdk-version-pinning for details.
 ''';
       }
     }
--- a/lib/src/solver/incompatibility_cause.dart
+++ b/lib/src/solver/incompatibility_cause.dart
@@ -13,7 +13,7 @@ import 'incompatibility.dart';
 
 /// The reason an [Incompatibility]'s terms are incompatible.
 sealed class IncompatibilityCause {
-  IncompatibilityCause();
+  const IncompatibilityCause();
 
   /// Human readable notice / information providing context for this
   /// incompatibility.
@@ -33,7 +33,9 @@ sealed class IncompatibilityCause {
 
 /// The incompatibility represents the requirement that the root package
 /// exists.
-class RootIncompatibilityCause extends IncompatibilityCause {}
+class RootIncompatibilityCause extends IncompatibilityCause {
+  const RootIncompatibilityCause();
+}
 
 /// The incompatibility represents a package's dependency.
 class DependencyIncompatibilityCause extends IncompatibilityCause {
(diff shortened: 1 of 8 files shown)


lib/src/solver/incompatibility_cause.dart
```diff
@@ -43,6 +31,39 @@ abstract class IncompatibilityCause {
...
+
+  @override
+  String? get notice {
+    final dependerDescription = depender.description;
+    if (dependerDescription is SdkDescription) {
+      final targetConstraint = target.constraint;
+      if (targetConstraint is Version) {
+        return '''
+Note: ${target.name} is pinned to version $targetConstraint by ${depender.name} from the ${dependerDescription.sdk} SDK.
+See https://dart.dev/go/version-pinning for details.
```
devoncarew, line 52: 🎉 - very cool!

lib/src/solver/incompatibility_cause.dart
```diff
@@ -6,26 +6,14 @@ import 'package:pub_semver/pub_semver.dart';
...
-  static const IncompatibilityCause dependency = _Cause('dependency');
-
-  /// The incompatibility indicates that the package has no versions that match
-  /// the given constraint.
-  static const IncompatibilityCause noVersions = _Cause('no versions');
-
-  /// The incompatibility indicates that the package has an unknown source.
-  static const IncompatibilityCause unknownSource = _Cause('unknown source');
+sealed class IncompatibilityCause {
+  IncompatibilityCause();
```
jonasfj, line 16:
```suggestion
```
do we need the constructor?

lib/src/solver/incompatibility_cause.dart
```diff
@@ -6,26 +6,14 @@ import 'package:pub_semver/pub_semver.dart';
...
```
jonasfj, line 17: We could still do:
```suggestion
  static const IncompatibilityCause root = RootIncompatibilityCause();
```

Then we won't need to refactor the code, and we won't need to create instances for must of them, we can just keep these singletons.

Sure, `DependencyIncompatibilityCause` will have to be special.

lib/src/solver/incompatibility_cause.dart
```diff
@@ -43,6 +31,39 @@ abstract class IncompatibilityCause {
   String? get hint => null;
 }
 
+/// The incompatibility represents the requirement that the root package
+/// exists.
+class RootIncompatibilityCause extends IncompatibilityCause {}
```
jonasfj, line 36: Unless we think someone should implement the class, then shouldn't we put final on them?
```suggestion
final class RootIncompatibilityCause extends IncompatibilityCause {}
```

We probably should sprinkle `final` on as many classes as possible in a separate PR.

lib/src/solver/incompatibility_cause.dart
jonasfj, line 52: Is this specific to dependencies pinned by SDK packages?
```suggestion
See https://dart.dev/go/sdk-version-pinning for details.
```

jonasfj, line 52: Is this specific to dependencies pinned by SDK packages?
```suggestion
See https://dart.dev/go/sdk-version-pinning for details.
```

lib/src/solver/incompatibility_cause.dart
sigurdm: Hmm, I think it matters very little until we export them.

lib/src/solver/incompatibility_cause.dart
sigurdm: I like the checks of the cause to be type-checks anyway instead of equality checks. Potentially there are places where we can have exhaustive switches on the causes.

If we want to keep a singleton around... maybe... I think it is an insignificant optimization in this case.

lib/src/solver/incompatibility_cause.dart
sigurdm: Now I made it const

jonasfj (approved):

devoncarew:
🎉 !

@sigurdm - do you mind pasting in some example text from an error? Or, would it be substantially similar to:

```
Note: tool is pinned to version 1.0.0 by flutter_foo from the flutter SDK.
```

sigurdm:
Here is output of the first test case:
```
$ pub get
Resolving dependencies...
Note: tool is pinned to version 1.0.0 by flutter_foo from the flutter SDK.
See https://dart.dev/go/sdk-version-pinning for details.

Because every version of flutter_foo from sdk depends on tool 1.0.0 and every version of bar depends on tool ^2.0.0, flutter_foo from sdk is incompatible with bar.
So, because myapp depends on both bar any and flutter_foo from sdk, version solving failed.
```

PR #3921, opened 2023-05-16
Warn about presence of legacy cache

sigurdm:
Fixes: https://github.com/dart-lang/sdk/issues/52386

### DIFF
--- a/lib/src/system_cache.dart
+++ b/lib/src/system_cache.dart
@@ -327,35 +327,36 @@ Consider setting the `PUB_CACHE` variable manually.
 
     final appData = Platform.environment['APPDATA'];
     if (appData == null) return;
-    final oldCacheLocation = p.join(appData, 'Pub', 'Cache');
-    final oldCacheDeprecatedFile = p.join(oldCacheLocation, 'DEPRECATED.md');
-    final stat = tryStatFile(oldCacheDeprecatedFile);
+    final legacyCacheLocation = p.join(appData, 'Pub', 'Cache');
+    final legacyCacheDeprecatedFile =
+        p.join(legacyCacheLocation, 'DEPRECATED.md');
+    final stat = tryStatFile(legacyCacheDeprecatedFile);
     if (stat != null) {
       log.fine('''
 ${DateTime.now().difference(stat.changed) > Duration(days: 7)} ${stat.changed}'''); // XXX remove
     }
     log.fine('''
-${dirExists(oldCacheLocation)}'''); // XXX remove
+${dirExists(legacyCacheLocation)}'''); // XXX remove
     if ((stat == null ||
             DateTime.now().difference(stat.changed) > Duration(days: 7)) &&
-        dirExists(oldCacheLocation)) {
+        dirExists(legacyCacheLocation)) {
       log.warning('''
-Found a legacy pub cache at $oldCacheLocation. Pub is using $defaultDir.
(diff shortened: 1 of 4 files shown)


lib/src/system_cache.dart
```diff
@@ -303,7 +303,58 @@ Consider setting the `PUB_CACHE` variable manually.
...
+      log.warning('''
+Found a legacy Pub cache at $legacyCacheLocation. Pub is using $defaultDir.
+
+Consider deleting the legacy cache.
+
+See https://dart.dev/resources/dart-3-migration#other-tools-changes for details.
+''');
+      try {
+        writeTextFile(legacyCacheDeprecatedFile, '''
+This pub cache is no longer used by recent Dart/Flutter releases.
```
jonasfj, line 346:
```suggestion
As of Dart 3 this pub cache is no longer used by recent Dart/Flutter releases.
```

lib/src/system_cache.dart
```diff
@@ -303,7 +303,58 @@ Consider setting the `PUB_CACHE` variable manually.
...
+
+Consider deleting it.
```
jonasfj, line 348:
```suggestion
Consider deleting it, if you are not using Dart versions earlier than 2.8.0.
```

lib/src/system_cache.dart
```diff
@@ -303,7 +303,58 @@ Consider setting the `PUB_CACHE` variable manually.
...
+  }
+
+  /// Check for the presence of a cache at the legacy location
+  /// `%APPDATA$\Pub\Cache`.
+  ///
+  /// If it is present, give a warning and write a DEPRECATED.md in that cache.
+  ///
+  /// If DEPRECATED.md is less than 7 days old, we don't repeat the warning.
+  void _checkOldCacheLocation() {
+    if (!Platform.isWindows) return;
```
jonasfj, line 326: I think we should leave ourselves an explanation of why we did this, so that we can remember it in the future.
```suggestion
    // Background:
    // Prior to Dart 2.8 the default location for the PUB_CACHE on Windows was:
    //   %APPDATA%\Pub\Cache
    //
    // Start Dart 2.8 pub started migrating the default PUB_CACHE location to:
    //   %LOCALAPPDATA%\Pub\Cache
    // That is:
    //  * If a pub-cache existed in `%LOCALAPPDATA%\Pub\Cache` then it
    //    would be used.
    //  * If a pub-cache existed in `%APPDATA%\Pub\Cache` then it would be
    //    used, unless a pub-cache in `%LOCALAPPDATA%\Pub\Cache` had been found.
    //  * If no pub-cache was found, a new empty pub-cache was created in
    //    `%LOCALAPPDATA%\Pub\Cache`.
    //
    // Starting in Dart 3.0 pub will no-longer look for a pub-cache in
    // `%APPDATA%\Pub\Cache`. Instead it will always use the new location,
    // `%LOCALAPPDATA%\Pub\Cache`, as default PUB_CACHE location.
    //
    // Using `%APPDATA%` caused the pub-cache to be copied with the user-profile,
    // when using a networked Windows setup where users can login on multiple
    // machines. This is undesirable because you are moving a lot of bytes over
    // the network and onto whatever servers are storing the user profiles.
    //
    // Thus, we migrated to storing the pub-cache in `%LOCALAPPDATA%`.
    // And finished the migration in Dart 3 to keep things simple.
    if (!Platform.isWindows) return;
```

jonasfj (approved):

PR #3930, opened 2023-05-24
Expand constraint for `http`

natebosch:
The 1.0.0 release does not have breaking changes.

Remove the pre-release suffix from the SDK constraint now that the
stable SDK is published. Remove unneeded quotes from SDK constraint.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,7 +1,7 @@
 name: pub
 
 environment:
-  sdk: '^3.0.0-0'
+  sdk: ^3.0.0
 
 dependencies:
   # Note: Pub's test infrastructure assumes that any dependencies used in tests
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
   convert: ^3.0.2
   crypto: ^3.0.1
   frontend_server_client: ^3.2.0
-  http: ^0.13.3
+  http: ">=0.13.3 <2.0.0"
```
sigurdm, line 17: I think we can just go straight to 1.0.0.

In the sdk we'll anyway just use whatever it provides.
```suggestion
  http: "^1.0.0"
```

natebosch: If we want to do that we'll also need to update transitive deps.

https://github.com/google/dart-neats/pull/209

sigurdm (approved):

Issue #3937, opened 2023-06-01
Pub client should validate dependency names


sigurdm:
Otherwise we end up making bad requests: https://github.com/dart-lang/pub/issues/3310

PR #3940, opened 2023-06-01
Validate that dependency names are valid package names

sigurdm:
Fix of https://github.com/dart-lang/pub/issues/3937


szakarias (approved):
LGTM

Issue #3946, opened 2023-06-19
It seems that build_runner crashes while using "flutter_localizations".


FrancescoJo:
It seems that `build_runner` crashes while using flutter_localizations.

Thanks for reading and we hope to see this error is fixed soon.

# Environment

- Dart version (run `dart --version`): 3.0.5 (stable)
- OS kind and version (e.g. "Windows 10, version 1809" or "macOS 12.4"): `Linux 5.19.0-43-generic on x86_64`
- Are you using the Chinese community mirror or a corporate firewall? nope

# Problem

Error happens like below if we apply [Internationalizing Flutter apps](https://docs.flutter.dev/accessibility-and-localization/internationalization) in our project.

```shell
$ flutter pub run build_runner build --delete-conflicting-outputs
Deprecated. Use `dart run` instead.
Null check operator used on a null value


package:pub/src/entrypoint.dart 562:48   Entrypoint.resolveExecutable
package:pub/src/executable.dart 79:35    runExecutable
package:pub/src/command/run.dart 100:26  RunCommand.runProtected

This is an unexpected error. The full log and other details are collected in:

    /home/USER/.pub-cache/log/pub_log.txt

Consider creating an issue on https://github.com/dart-lang/pub/issues/new
and attaching the relevant parts of that log file.
```

## Expected behavior

Build should be successful and various files dependes on `builld_runner` are created.

## Actual behavior

Cannot continue build unless we comment out `build_runner`.

## --trace output

The pub_log.txt file:

```yaml
Information about the latest pub run.

If you believe something is not working right, you can go to
https://github.com/dart-lang/pub/issues/new to post a new issue and attach this file.

Before making this file public, make sure to remove any sensitive information!

Pub version: 3.0.5
Created: 2023-06-19T09:43:09+09:00
FLUTTER_ROOT: /home/USER/snap/flutter/common/flutter
PUB_HOSTED_URL: <not set>
PUB_CACHE: "<not set>"
Command: dart pub --suppress-analytics run build_runner build --delete-conflicting-outputs
Platform: linux

---- /home/USER/documents/my_project/app/pubspec.yaml ----
name: my_project
description: My flutter project.
publish_to: none
version: 1.0.0+1

environment:
  sdk: ^3.0.5
  flutter: ^3.10.5

dependencies:
  flutter:
    sdk: flutter
  flutter_localizations:
    sdk: flutter

  core:
    path: ../core/
  data:
    path: ../data/
  util:
    path: ../util/

[...]

PR #3949, opened 2023-06-20
Add missing await (and make downstream async)

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/3946

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -167,10 +167,14 @@ class Entrypoint {
   ///
   /// Throws a [DataError] if the `.dart_tool/package_config.json` file isn't
   /// up-to-date relative to the pubspec and the lockfile.
-  PackageGraph get packageGraph => _packageGraph ??= _createPackageGraph();
+  Future<PackageGraph> get packageGraph =>
+      _packageGraph ??= _createPackageGraph();
 
-  PackageGraph _createPackageGraph() {
-    ensureUpToDate(); // TODO, really?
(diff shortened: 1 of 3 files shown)


lib/src/entrypoint.dart
```diff
@@ -167,10 +167,14 @@ class Entrypoint {
   ///
   /// Throws a [DataError] if the `.dart_tool/package_config.json` file isn't
   /// up-to-date relative to the pubspec and the lockfile.
-  PackageGraph get packageGraph => _packageGraph ??= _createPackageGraph();
-
-  PackageGraph _createPackageGraph() {
-    ensureUpToDate(); // TODO, really?
+  Future<PackageGraph> get packageGraph =>
+      _packageGraph ??= _createPackageGraph();
```
jonasfj, line 171: I think it might be possible to do, but it could also be too smart :rofl: 

```suggestion
  late final packageGraph = () async {
    await ensureUpToDate();
    var packages = {
      for (var packageEntry in packageConfig.nonInjectedPackages)
        packageEntry.name: Package.load(
          packageEntry.name,
          packageEntry.resolvedRootDir(packageConfigPath),
          cache.sources,
        ),
    };
    packages[root.name] = root;
    return PackageGraph(this, packages);
  }();
```

jonasfj (approved):
LGTM

lib/src/entrypoint.dart
sigurdm: Unfortunately we initialize _packageGraph directly in some of our constructors. So I'm not sure we can make this work

Issue #2914, opened 2021-03-12
When publishing a breaking change version, give me warnings about public API marked @deprecated


kevmoo:
When publishing a new breaking version (v1.x.y -> v2.0.0) – this is the time to remove cruft.

The pub client can checked to see if the version I'm publishing is breaking. It'd be a great time to flag the user about any existing `@deprecated` (or `@Deprecated(...)`) in the `lib` dir.

sigurdm:
We are considering this might work better as a lint. Then you get the warning before attempting to publish, and it can be ignored with `// ignore:` lines.

@pq does that make sense to you?
Where should the lint be reported? at the deprecation annotation or at the version number in the pubspec.

I don't think we have any other lints that take the version number into account.
Do we need to know about the previous version, or is it enough that the current version is a x.0.0, where x>=2 ?

sigurdm:
Created a lint: https://dart-review.googlesource.com/c/sdk/+/447580

pq:
SGTM!

/fyi @bwilkerson @srawlins

srawlins:
I think I like the idea. I'd like to specify it thoroughly before landing it though. (Once a lint rule is in the wild, it is _very_ hard to add to the reported cases (fixing "false negative" bugs.)

So, for each `@Deprecated` annotation, annotating an element:

* Don't report if we're not in a pub package with a version.
* Don't report if the annotated element is a private element (not a private field, class, unnamed extension, element in an unnamed extension, static element in a private class, etc. etc.). Public instance members of private classes can be considered public, as they may override elements of a public class.
* However, maybe don't report if the annotated element is an override. If we're overriding an element in an _outside class_, maybe declared in another package, then maybe we want to keep annotating the override.
* Don't report if the annotated element is not in the public API. We can't really calculate this, in the analyzer, yet. It requires scanning all public files in `lib` for exports. And it requires invalidating the information any time a change is made to such exports. For now, I guess we can just use `/lib`? If the thing-being-annotated is anywhere in `/lib`, including `/lib/src`, then call it public API.
[...]

PR #3959, opened 2023-06-28
Warn when publishing breaking release with deprecated members

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/2914

Open questions: 

* should this be a warning or a hint?
* is the syntactic analysis good enough, or should we do a resolution on the code?
* could this be implemented as a lint, that could then be added conditionally before running the `dart analyze` validation?

### DIFF
--- a/test/validator/breaking_with_deprecated_test.dart
+++ b/test/validator/breaking_with_deprecated_test.dart
@@ -1,0 +1,43 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+import '../descriptor.dart' as d;
+import '../test_pub.dart';
+import 'utils.dart';
+
+void main() {
+  test('should only warn when publishing a breaking release ', () async {
+    final server = await servePackages();
+    await d.dir(appPath, [
+      d.validPubspec(extras: {'version': '2.0.0'}),
+      d.file('LICENSE', 'Eh, do what you want.'),
+      d.file('README.md', "This package isn't real."),
+      d.file('CHANGELOG.md', '# 2.0.0\nFirst version\n'),
+      d.dir('lib', [
+        d.file(
+          'test_pkg.dart',
+          "@Deprecated('Stop using this please') int i = 1;",
+        ),
+        d.dir('src', [
+          d.file(
+            'support.dart',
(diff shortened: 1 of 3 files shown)


test/validator/breaking_with_deprecated_test.dart
```diff
@@ -0,0 +1,43 @@
...
+    final server = await servePackages();
+    await d.dir(appPath, [
+      d.validPubspec(extras: {'version': '2.0.0'}),
+      d.file('LICENSE', 'Eh, do what you want.'),
+      d.file('README.md', "This package isn't real."),
+      d.file('CHANGELOG.md', '# 2.0.0\nFirst version\n'),
+      d.dir('lib', [
+        d.file(
+          'test_pkg.dart',
+          "@Deprecated('Stop using this please') int i = 1;",
```
jonasfj, line 22: Why should this be a problem?

Yes, in theory a major version bump is a good opportunity to remove deprecated stuff.

But it's also entirely possible that my package contains multiple kinds of deprecated stuff, what if I want to keep sonme of the deprecated stuffs and remove other deprecated stuff.

In general, unless there is a pressing reason to remove something that's been deprecated, I would suggest not doing so.
You could do things like `/// @nodoc` in dartdoc documentation comments, to exclude the deprecated bits from documentation, so it doesn't pollute.

Would be nice if there was also a way to hide it from auto-completion.


---------------------------

Yes, in many cases it makes sense to cleanup when doing a major version bump, such that deprecated stuff disappears. But for large complex packages it's entirely possible that some deprecated stuff sticks around because it's low cost to keep, and removing it only causes friction.

sigurdm: Yeah - good points...

Maybe we can lower this to  a "hint"... I still think there is some value to extract here (the associated issue seems quite popular)...

sigurdm: WDYT @jonasfj ? Should we make this a hint or close the issue as wontfix?

jonasfj: I don't mind shipping it as a hint.

Are we sure this shouldn't just be a "lint" or diagnostic in the analyzer?
We run an analyzes run before publishing, right?

sigurdm: Hmm - maybe that would work...

@pq what do you say? Would we like to be linted not to have deprecated members if the version is `x.0.0` ?

sigurdm:
Closing in favor of https://dart-review.googlesource.com/c/sdk/+/447580

PR #3862, opened 2023-04-05
Don't extend something which should be an interface.

lrhn:
These classes should preferably be interface classes.

PR #3976, opened 2023-07-25
Update tar

simolus3:
Rolls tar to `1.0.0`. The changes in that package mostly consist of adding class modifiers and a few `unawaited()` markers. `TarFormat` has been changed to distinguish between fully known formats and combinations the reader is still working out, but those aren't used by pub.

So I think pub shouldn't be affected by those changes, but as usual I've pushed a version of the `extract_all_pub_dev.dart` script that compares the two versions [in my fork](https://github.com/simolus3/pub/tree/compare-tars) on the `compare-tars` branch.

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).

</details>

### DIFF
--- a/lib/src/third_party/tar/CHANGELOG.md
+++ b/lib/src/third_party/tar/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.1
+
+- Fix an incompatibility with Dart 3.1.
+
 ## 1.0.0
 
 - __Breaking__ Add class modifiers where applicable.
--- a/lib/src/third_party/tar/lib/src/writer.dart
+++ b/lib/src/third_party/tar/lib/src/writer.dart
@@ -156,7 +156,7 @@ enum OutputFormat {
   gnuLongName,
 }
 
-final class _WritingSink extends StreamSink<TarEntry> {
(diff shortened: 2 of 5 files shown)


simolus3:
I think it's a bit unfortunate that pub has diverged from the upstream tar package in #3862. I get that it was probably urgent to fix, but it would be nice if I at least get notified when I need to make a fix in tar to support upcoming Dart versions. All tar versions blow 1.0.1 will be broken with the next stable Dart release, and I could have known this months ago...

Either way, the vendored copy is now in sync with the upstream package again and it seems to be working.

sigurdm:
> I get that it was probably urgent to fix, but it would be nice if I at least get notified when I need to make a fix in tar to support upcoming Dart versions.

Sorry - I think this fix was landed without noticing that the library was vendored.

sigurdm (approved):

sigurdm:
Thanks!

PR #3984, opened 2023-08-13
Update Dart spec reserved words

parlough:
Also switches to a set, since only `.contains` is called on the collection. These results are verified for new packages on pub.dev and this updates the local verification to match that.

Updated from https://github.com/dart-lang/pub-dev/blob/master/pkg/pub_package_reader/lib/src/names.dart which is generated from https://github.com/dart-lang/sdk/blob/main/tools/spec_parser/Dart.g.


sigurdm (approved):

sigurdm (approved):
Nice - thanks!

PR #3985, opened 2023-08-13
Fix and increase flexibility of language version override test

parlough:


sigurdm (approved):
Wow thanks!

For mosts tests we use the _PUB_TEST_SDK_VERSION variable, but here it is the analyzer that doesn't pick it up.

I guess this is the best we can do.

PR #3997, opened 2023-08-25
Switch vendor to a dev dependency

parlough:
I didn't see a reason for this to be a normal dependency.


sigurdm (approved):
Good point. Not sure how it ended up in dependencies.

Thanks!

Issue #4003, opened 2023-08-30
Is precompile in `dart.dart` sane?


jonasfj:
If inside `bin/frontendserver_testing.dart` I do:
```dart
import 'package:frontend_server_client/frontend_server_client.dart';
import 'package:path/path.dart' as p;
import 'dart:io';

Future<void> main() async {
  // TODO: Do this twice, and ask jens why it slow the second time -- if not slow second time, then we have don't something dumb in `dart run`

  final sw = Stopwatch()..start();
  final sdkRoot = p.relative(p.dirname(p.dirname(Platform.resolvedExecutable)));
  final client = await FrontendServerClient.start(
    'bin/pub.dart',
    'pub.dill',
    'lib/_internal/vm_platform_strong.dill',
    sdkRoot: sdkRoot,
    printIncrementalDependencies: false,
  );
  print('START: ${sw.elapsedMilliseconds} ms');
  sw.reset();
  final result = await client.compile();
  print('COMPILE: ${sw.elapsedMilliseconds} ms');
  client.kill();
  print(result.compilerOutputLines.join('\n'));
}
```

I get the following behavior:
```
$ rm pub.dill 
$ dart bin/frontendserver_testing.dart 
START: 63 ms
COMPILE: 6279 ms

$ dart bin/frontendserver_testing.dart 
START: 64 ms
COMPILE: 917 ms
```

So clearly building with an existing `.dill` file and reusing it from a previous compilation should be faster than not doing so.

-----------------------

But if I go inside `lib/src/entrypoint.dart` and tweak `_precompileExecutable` such that:
```dart
  Future<void> _precompileExecutable(
    Executable executable, {
    List<String> additionalSources = const [],
    String? nativeAssets,
  }) async {
    final package = executable.package;

    final sw = Stopwatch()..start();
    print('---------------- starting');
    await dart.precompile(
      executablePath: await resolveExecutable(executable),
      outputPath: pathOfExecutable(executable),
      incrementalDillPath: incrementalDillPathOfExecutable(executable),
      packageConfigPath: packageConfigPath,
      name: '$package:${p.basenameWithoutExtension(executable.relativePath)}',
      additionalSources: additionalSources,
      nativeAssets: nativeAssets,
    );
    print('---------------- compiled once: ${sw.elapsedMilliseconds} ms');
    sw.reset();
    await dart.precompile(
      executablePath: await resolveExecutable(executable),
      outputPath: pathOfExecutable(executable),
[...]

Issue #3896, opened 2023-04-28
We should clear the `incremental` folder when no longer used for global packages.


sigurdm:
It is just confusing.

PR #4005, opened 2023-09-01
Improve precompilation

sigurdm:
This ensures that the "incremental" file that is created when precompilation fails, will be removed when precompilation succeeds.

Fixes: https://github.com/dart-lang/pub/issues/4003

Also simplifies the path of the incremental file to be `$output.incremental`. This simplifies the directory handling a bit.

Fixes: https://github.com/dart-lang/pub/issues/3896

Added a test that caching takes effect (such that second compilation is much faster). This test is somewhat fragile - but I'm not sure we have a better way to test this.

### DIFF
--- a/lib/src/dart.dart
+++ b/lib/src/dart.dart
@@ -101,7 +101,7 @@ class AnalyzerErrorGroup implements Exception {
 /// Whichever of "[outputPath].incremental" and [outputPath] already exists is
 /// used to initialize the compiler run. To avoid the potential for
 /// race-conditions, it is first copied to a temporary location, and atomically
-/// moved to either [outputPath] or "[outputOPath].incremental" depending on the
+/// moved to either [outputPath] or "[outputPath].incremental" depending on the
 /// result of compilation.
 ///
 /// The [packageConfigPath] should point at the package config file to be used
--- a/lib/src/dart.dart
+++ b/lib/src/dart.dart
@@ -96,10 +96,13 @@ class AnalyzerErrorGroup implements Exception {
 ///
 /// If the compilation succeeds it is saved to a kernel file at [outputPath].
 ///
-/// If compilation fails, the output is cached at [incrementalDillOutputPath].
+/// If compilation fails, the output is cached at "[outputOPath].incremental".
 ///
-/// Whichever of [incrementalDillOutputPath] and [outputPath] already exists is
-/// used to initialize the compiler run.
+/// Whichever of "[outputOPath].incremental" and [outputPath] already exists is
+/// used to initialize the compiler run. To avoid the potential for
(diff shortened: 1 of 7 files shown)


lib/src/dart.dart
```diff
@@ -96,10 +96,13 @@ class AnalyzerErrorGroup implements Exception {
 ///
 /// If the compilation succeeds it is saved to a kernel file at [outputPath].
 ///
-/// If compilation fails, the output is cached at [incrementalDillOutputPath].
+/// If compilation fails, the output is cached at "[outputOPath].incremental".
 ///
-/// Whichever of [incrementalDillOutputPath] and [outputPath] already exists is
-/// used to initialize the compiler run.
+/// Whichever of "[outputOPath].incremental" and [outputPath] already exists is
```
jonasfj, line 101:
```suggestion
/// Whichever of "[outputPath].incremental" and [outputPath] already exists is
```

lib/src/dart.dart
```diff
@@ -96,10 +96,13 @@ class AnalyzerErrorGroup implements Exception {
 ///
 /// If the compilation succeeds it is saved to a kernel file at [outputPath].
 ///
-/// If compilation fails, the output is cached at [incrementalDillOutputPath].
+/// If compilation fails, the output is cached at "[outputOPath].incremental".
```
jonasfj, line 99:
```suggestion
/// If compilation fails, the output is cached at "[outputPath].incremental".
```

lib/src/dart.dart
```diff
@@ -96,10 +96,13 @@ class AnalyzerErrorGroup implements Exception {
...
 ///
-/// Whichever of [incrementalDillOutputPath] and [outputPath] already exists is
-/// used to initialize the compiler run.
+/// Whichever of "[outputOPath].incremental" and [outputPath] already exists is
+/// used to initialize the compiler run. To avoid the potential for
+/// race-conditions, it is first copied to a temporary location, and atomically
+/// moved to either [outputPath] or "[outputOPath].incremental" depending on the
```
jonasfj, line 104:
```suggestion
/// moved to either [outputPath] or "[outputPath].incremental" depending on the
```

test/precompilation_test.dart
```diff
@@ -0,0 +1,104 @@
...
+    await check(compile()).throws<ApplicationException>();
+  } else {
+    await compile();
+  }
+  verbosity = Verbosity.normal;
+  return s.elapsed;
+}
+
+void main() {
+  test('Precompilation is much faster second time and removes old artifacts',
```
jonasfj, line 68: can we please add a few comments? or at-least a few empty lines :rofl: 

I wouldn't honestly, be okay not having any tests for this. This easily becomes fragile.

jonasfj (approved):

PR #4009, opened 2023-09-05
Inline needed parts of package oauth2

sigurdm:
* copy-pasted everything from package:oauth2 into lib/src/oauth.dart
* made everything not exported private.
* deleted a few unused functions
* added trailing commas
* converted a class-based enum to a real enum.

Further simplifications can be done as follow-ups...


jonasfj (approved):
Yeah, there is a lot of opportunity to delete things like `_onCredentialsRefreshed` :rocket:

Issue #4013, opened 2023-09-13
Add support to skip size validation for hosted packages


SergeyShustikov:
Currently, I'm using hosted dependencies in my project, but one of them weight more than 100 MB as set in SizeValidator
https://github.com/dart-lang/pub/blob/cb0d6f97af64bc1494c0b4432833eca1af79feec/lib/src/validator/size.dart#L22

I am totally blocked from publishing my package. I know that I could split it into a few small parts, but in my case, it's not possible.
So, it would be nice to allow developers to publish hosted packages to another pub.dev solutions (for example to JFrog Artifactory) ignoring SizeValidation.

sigurdm:
I agree, this should be a warning, not an error. The server should provide the error if it has a size-limit.

For now you should be able to shortcut the error with: `dart pub publish --skip-validation`.

sigurdm:
We have switched this to be a warning only.

PR #4014, opened 2023-09-13
Don't impose a strict size limit to uploads

sigurdm:
Fix of: https://github.com/dart-lang/pub/issues/4013

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -211,6 +211,40 @@ void handleJsonError(http.BaseResponse response) {
   fail(log.red(error['message'] as String));
 }
 
+/// Handles an unsuccessful XML-formatted response from google cloud storage.
+///
+/// Assumes messages are of the form in
+/// https://cloud.google.com/storage/docs/xml-api/reference-status.
+///
+/// This is a poor person's XML parsing with regexps, but this should be
+/// sufficient for the specified messages.
+void handleGCSError(http.BaseResponse response) {
+  if (response is http.Response) {
+    final responseBody = response.body;
+    if (responseBody.contains('<?xml')) {
+      String? getTagText(String tag) {
+        final result = RegExp('<$tag>(.*)</$tag>').firstMatch(responseBody)?[1];
+        if (result == null) return null;
+        return sanitizeForTerminal(result);
+      }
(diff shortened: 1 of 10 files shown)


szakarias (approved):
LGTM

sigurdm:
When testing uploads of too large packages, the error message was not descriptive. So I added a poor person's parsing of the error messages from GCS.

PTAL.

lib/src/http.dart
```diff
@@ -211,6 +211,40 @@ void handleJsonError(http.BaseResponse response) {
   fail(log.red(error['message'] as String));
 }
 
+/// Handles an unsuccessful XML-formatted response from google cloud storage.
+///
+/// Assumes messages are of the form in
+/// https://cloud.google.com/storage/docs/xml-api/reference-status.
```
jonasfj, line 217:
```suggestion
/// https://cloud.google.com/storage/docs/xml-api/reference-status
```

test/lish/cloud_storage_upload_provides_an_error_test.dart
```diff
@@ -22,14 +22,30 @@ void main() {
...
+    expect(
+      pub.stderr,
+      emits(
+        'Server message: Your proposed upload is larger than the maximum object size specified in your Policy Document.',
+      ),
+    );
+    expect(
+      pub.stderr,
+      emits('Server details: Content-length exceeds upper bound on range'),
+    );
```
jonasfj, line 47: What does this look like in a terminal?

Are you sure we don't want to special case this?

jonasfj (approved):

test/lish/cloud_storage_upload_provides_an_error_test.dart
sigurdm: > What does this look like in a terminal?

image

> Are you sure we don't want to special case this?

Such that we recognize this specific message and give our own reply? That might work as well...

sigurdm: I think special-casing for specific messages would be neat - but also risks swallowing specific details from the error message.

I'll add a TODO for now.

PR #4017, opened 2023-09-13
Update to analyzer `>= 6.2.0`

parlough:


jonasfj:
@isoos any thoughts?

sigurdm:
I think we should just ship it!

isoos:
lgtm

sigurdm:
Thanks for doing this.

We have dependabot enabled, so we should not need to do this manually. I pushed a button to make dependabot run, and it did the same upgrade. Landing that for a slightly nicer commit message: https://github.com/dart-lang/pub/pull/4019.

PR #4018, opened 2023-09-13
Enable 'strict-inference' language analysis mode

parlough:
Enabling this analyzer language mode to get closer to switching to `dart_flutter_team_lints` which enables it by default.


jonasfj:
@isoos any thoughts?

sigurdm (approved):
Thanks!

PR #4031, opened 2023-10-20
Migrate pub to use package:unified_analytics

bkonyi:
This is some prerequisite work for migrating package:dartdev to use package:unified_analytics.

I've removed the single analytics test here in favour of the dart pub get analytics test in dartdev for simplicity, as package:unified_analytics adds a FakeAnalytics class that tracks analytics events for easy validation in tests without needing to manually serialize all the analytics events. It's not possible to serialize / deserialize FakeAnalytics so the results can't be verified from a separate process at this point, so this testing is easier to do in dartdev. If you'd prefer we keep this testing here, please let me know if you have ideas on how to move forward 😊

### DIFF
--- a/test/embedding/embedding_test.dart
+++ b/test/embedding/embedding_test.dart
@@ -2,10 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:convert';
 import 'dart:io';
 
-import 'package:path/path.dart' as path;
 import 'package:path/path.dart' as p;
 import 'package:pub/src/exit_codes.dart';
 import 'package:pub/src/io.dart' show EnvironmentKeys;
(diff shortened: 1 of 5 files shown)


test/embedding/embedding_test.dart
```diff
@@ -2,10 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:convert';
 import 'dart:io';
 
-import 'package:path/path.dart' as path;
```
sigurdm, line 8: Thanks!

test/embedding/embedding_test.dart
```diff
@@ -191,77 +189,6 @@ main() {
     );
   });
 
-  test('analytics', () async {
```
sigurdm, line 194: Is there no way with unified_analytics that we can test that the events are as expected?

sigurdm (approved):
LGTM but would love to have some test coverage

test/embedding/embedding_test.dart
sigurdm: Ah - only read your explanation now - I guess testing via dartdev is good enough

bkonyi: Yeah, I know it's not ideal, but I figure it's good enough for now. If we feel we still want the test coverage in this repo, we can always add it in later when we have the cycles.

PR #4036, opened 2023-10-28
Update to package:lints v3 and fix a few more lints

parlough:
- Updates to `package:lints` v3
- Removes now duplicated lints
- Adds and fixes a few more lints from `package:dart_flutter_team_lints` working towards the switch.

### DIFF
--- a/tool/test-bin/pub_command_runner.dart
+++ b/tool/test-bin/pub_command_runner.dart
@@ -28,7 +28,8 @@ class ThrowingCommand extends PubCommand {
   @override
   String get description => 'Throws an exception';
 
-  bool get hide => true;
+  @override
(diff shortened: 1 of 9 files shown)


tool/test-bin/pub_command_runner.dart
```diff
@@ -24,7 +24,8 @@ class ThrowingCommand extends PubCommand {
   @override
   String get description => 'Throws an exception';
 
-  bool get hide => true;
+  @override
```
sigurdm, line 27: Nice catch - but let's actually not hide this.

This is a command in the test-executable, no need to hide

sigurdm (approved):
Thanks!

PR #4053, opened 2023-11-20
Add basic functionality for calling and caching the result from the pub.dev api for security advisories

szakarias:
This is a first step towards showing surfacing security advisories in the client. The functionality is not yet called.

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -537,6 +537,62 @@ class HostedSource extends CachedSource {
     return result;
   }
 
+  Future<_PackageAdvisories> _fetchAdvisories(
+    _RefAndCache refAndCache,
+  ) async {
+    final ref = refAndCache.ref;
+    final description = ref.description;
+    final cache = refAndCache.cache;
+
+    if (description is! HostedDescription) {
+      throw ArgumentError('Wrong source');
+    }
+
+    final packageName = description.packageName;
+    final hostedUrl = description.url;
+
+    final url = _listAdvisoriesUrl(ref);
+    log.io('Get security advisories from $url.');
+
+    final String bodyText;
+    final Map<String, dynamic> body;
+    final _PackageAdvisories result;
+    try {
+      bodyText = await withAuthenticatedClient(cache, Uri.parse(hostedUrl),
+          (client) async {
+        return await retryForHttp(
+            'fetching advisories for "$packageName" from "$url"', () async {
+          final request = http.Request('GET', url);
+          request.attachPubApiHeaders();
+          request.attachMetadataHeaders();
+          final response = await client.fetch(request);
+          return response.body;
+        });
+      });
+      final decoded = jsonDecode(bodyText);
+      if (decoded is! Map<String, dynamic>) {
+        throw FormatException('security advisories must be a mapping');
+      }
+      body = decoded;
+
+      result = _PackageAdvisories(
+        body['advisories'] as List,
+        body['advisoriesUpdated'] == null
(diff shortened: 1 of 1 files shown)


lib/src/source/hosted.dart
```diff
@@ -610,6 +666,29 @@ class HostedSource extends CachedSource {
...
+  ) async {
+    final path = _advisoriesCachePath(ref, cache);
+    try {
+      ensureDir(p.dirname(path));
+
+      await writeTextFileAsync(
+        path,
+        jsonEncode(
+          <String, dynamic>{
+            ...body,
```
sigurdm, line 682: Maybe we want to at a timestamp here?
```suggestion
            ...body,
            'fechedAt': DateTime.now.[convert to string...]
```

lib/src/source/hosted.dart
```diff
@@ -1504,6 +1612,15 @@ class _VersionInfo {
   );
 }
 
+/// Information about advisories affecting a package retrieved from
+/// /api/packages/$package/advisories
+class _PackageAdvisories {
+  final DateTime? advisoriesUpdated;
+  final List advisories;
```
sigurdm, line 1619: I think the list needs a type-argument

lib/src/source/hosted.dart
```diff
@@ -537,6 +537,62 @@ class HostedSource extends CachedSource {
     return result;
   }
 
+  Future<_PackageAdvisories> _fetchAdvisories(
+    _RefAndCache refAndCache,
```
sigurdm, line 541: Don't use the _RefAndCache type, it is only used because records was not a thing back then - just take a Ref and a SystemCache.

lib/src/source/hosted.dart
```diff
@@ -537,6 +537,62 @@ class HostedSource extends CachedSource {
...
+        });
+      });
+      final decoded = jsonDecode(bodyText);
+      if (decoded is! Map<String, dynamic>) {
+        throw FormatException('security advisories must be a mapping');
+      }
+      body = decoded;
+
+      result = _PackageAdvisories(
+        body['advisories'] as List,
```
sigurdm, line 579: Don't we want to decode the individual entries?

sigurdm, line 579: We cannot do a type-cast here. We have to validate this is actually a list and throw if not

lib/src/source/hosted.dart
```diff
@@ -537,6 +537,62 @@ class HostedSource extends CachedSource {
...
+            ? null
+            : DateTime.parse(body['advisoriesUpdated'] as String),
+      );
+    } on Exception catch (error, stackTrace) {
+      _throwFriendlyError(error, stackTrace, packageName, hostedUrl);
+    }
+
+    // Cache the response on disk.
+    // Don't cache overly big responses.
+    if (bodyText.length < 500 * 1024) {
```
sigurdm, line 590: Maybe we should make this size limit bigger. Perhaps 1000 kb?

lib/src/source/hosted.dart
```diff
@@ -537,6 +537,62 @@ class HostedSource extends CachedSource {
...
+      }
+      body = decoded;
+
+      result = _PackageAdvisories(
+        body['advisories'] as List,
+        body['advisoriesUpdated'] == null
+            ? null
+            : DateTime.parse(body['advisoriesUpdated'] as String),
+      );
+    } on Exception catch (error, stackTrace) {
```
sigurdm, line 584: Can we be more specific in what we catch here?

Issue #4033, opened 2023-10-25
Handling package retraction


Reprevise:
When a package gets retracted, there is little tooling to get back to a supported version, which is semi-fair considering how _rare_ it is. However, when it does happen, it's annoying. I'm proposing a couple of things.

- Highlight retracted versions in `pub outdated`
    - Currently, the only way you can tell a package is retracted is noticing your current version is ahead of the one shown in the "Latest" column
- Add some way to get back to the supported version
    - The "easiest" way I know how is running `pub downgrade` then running `pub upgrade` again

**Other questions**
How does `pub cache` handle reverts? Would I have to delete the folder in the cache for the reverted version if the author were to re-use the same version?

sigurdm:
I agree pub outdated could show something. We do that for discontinued. That is probably an oversight.

> How does pub cache handle reverts? Would I have to delete the folder in the cache for the reverted version if the author were to re-use the same version?

We don't delete the version from the cache. You are still free to use it if it is in the pubspec.lock.
You cannot reuse a retracted version number.

Reprevise:
So what about going back to a non-retracted version? I can agree it's pretty rare (though this is just an anecdote, some metrics on this would be interesting), but would it be worth adding a way to go back easily? Using `pub upgrade` doesn't make sense, and `pub downgrade` AFAIK goes to your min version constraint.

sigurdm:
`dart pub upgrade` should get you to the latest non-retracted version (that is compatible with other dependencies...)
If you want to go to a specific version (without changing your constraint) you pretty much have to edit pubspec.lock manually. (Remember to do `dart pub get` after editing the pubspec.lock, it will warn you about the content-hash not matching, and update pubspec.lock with the correct hash).

Reprevise:
> `dart pub upgrade` should get you to the latest non-retracted version (that is compatible with other dependencies...)

I don't recall `dart pub upgrade` going back to a non-retracted version when it happened to me. Something to look into maybe?

sigurdm:
> I don't recall dart pub upgrade going back to a non-retracted version when it happened to me. Something to look into maybe?

By "latest" I mean newest. In most cases there will be a newer version than the retracted one available. Only if no such version exists `upgrade` should downgrade you.

Reprevise:
> By "latest" I mean newest. In most cases there will be a newer version than the retracted one available. Only if no such version exists upgrade should downgrade you.

[...]

PR #4055, opened 2023-11-20
Show information about retracted versions in "pub outdated"

szakarias:
Fixes https://github.com/dart-lang/pub/issues/4033


test/testdata/goldens/outdated/outdated_test/circular dependency on root.txt
```diff
@@ -124,6 +125,7 @@ $ pub outdated --json --no-dev-dependencies
       "package": "foo",
       "kind": "direct",
       "isDiscontinued": false,
+      "isRetracted": false,
```
sigurdm, line 128: This looks a bit wrong - should it not be the versions knowing if they are retreacted?

sigurdm (commented):
I think we should modify the filtering logic, such that rows where the current version is retracted (but equal to latests) is shown

sigurdm (approved):

test/testdata/goldens/outdated/outdated_test/circular dependency on root.txt
sigurdm: Renamed this is much better

szakarias: Renamed this to isCurrentRetracted to clarify.

szakarias:
> I think we should modify the filtering logic, such that rows where the current version is retracted (but equal to latests) is shown

We already show these, aka when the newest version is retracted. The concept of `Latest` does not apply to retracted versions, that is the latest version cannot be a retracted version.

Issue #4051, opened 2023-11-17
Publishing doesn't consider x.y.z+1 to be an incremental update from x.y.z


jamesderlin:
# Environment

- Dart SDK version: 3.2.0 (stable) (None) on "linux_x64"
- Ubuntu 22.04.3 LTS

# Problem

I corrected some documentation in a package and incremented the package version from 0.8.3 to 0.8.3+1.  Running `dart pub publish` printed:

```
Package validation found the following hint:
* The previous version is 0.8.3.
  
  It seems you are not publishing an incremental update.
  
  Consider one of:
  * 1.0.0 for a first major release.
  * 0.9.0 for a breaking release.
  * 0.8.4 for a minor release.
  

Publishing is forever; packages cannot be unpublished.
Policy details are available at https://pub.dev/policy


Package has 0 warnings and 1 hint.. Do you want to publish dartbag 0.8.3+1 to https://pub.dev (y/N)? 
```

## Expected behavior

I would expect 0.8.3+1 to be considered an incremental update to 0.8.3.

The build version (the portion after the `+`) might be a little tricky to handle in general since it can be a mostly arbitrary string.  However, the convention of using +*number* is fairly common, so it might be worth handling at least that case.

sigurdm:
Yeah, that is broken. Fix is coming up.

PR #4056, opened 2023-11-21
Fix calculation version number without build part

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/4051
When there was no prerelease component, we would produce an empty-string prerelease component instead of null.


szakarias (approved):
LGTM

Issue #449, opened 2015-06-05
If a git path in a pubspec is relative interpret it relative to the pubspec


DartBot:
<a href="https://github.com/munificent"><img src="https://avatars.githubusercontent.com/u/46275?v=3" align="left" width="96" height="96"hspace="10"></img></a> **Issue by [munificent](https://github.com/munificent)**
_Originally opened as dart-lang/sdk#8570_

---

You can use a file path as a git dependency. When you do, it's interpreted as relative to the package directory of the root package. It should be resolve as relative to the pubspec that contains the dependency (like relative path deps are).

DartBot:
<a href="https://github.com/munificent"><img src="https://avatars.githubusercontent.com/u/46275?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [munificent](https://github.com/munificent)**

----

*Added [Community-StarterBug](../labels/Community-StarterBug) label.*

DartBot:
<a href="https://github.com/kasperl"><img src="https://avatars.githubusercontent.com/u/2156198?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [kasperl](https://github.com/kasperl)**

----

*Removed this from the [Later](../milestones/Later) milestone.*
*Added [Oldschool-Milestone-Later](../labels/Oldschool-Milestone-Later) label.*

DartBot:
<a href="https://github.com/kasperl"><img src="https://avatars.githubusercontent.com/u/2156198?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [kasperl](https://github.com/kasperl)**

----

*Removed [Oldschool-Milestone-Later](../labels/Oldschool-Milestone-Later) label.*

DartBot:
<a href="https://github.com/kaendfinger"><img src="https://avatars.githubusercontent.com/u/2130849?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [kaendfinger](https://github.com/kaendfinger)**

----

Shouldn't this have a PatchesWelcome label?

DartBot:
<a href="https://github.com/munificent"><img src="https://avatars.githubusercontent.com/u/46275?v=3" align="left" width="48" height="48"hspace="10"></img></a> **Comment by [munificent](https://github.com/munificent)**

----

Patches are welcome on every bug! :)

[...]

PR #4060, opened 2023-11-28
Allow relative path-dependencies from git dependencies

sigurdm:
The paths are interpreted as paths inside the same git repo (at the same ref).

Internally, and in the lock-file they are treated as git-dependencies.

We do this by always interpreting dependencies in the context of the `Description` of the pubspec that contains it. This mechanism is now also used for passing the path relative path-dependencies are relative against. Much more streamlined!

We have to change how RootDescription is behaving, because the description now has to be created before we parse the pubspec it cannot know the Package object it belongs to, but just the path. (necessitating a bit of caching for everything to line up).

This fixes https://github.com/dart-lang/pub/issues/449


lib/src/package.dart
```diff
@@ -133,6 +134,7 @@ class Package {
       sources,
       expectedName: name,
       allowOverridesFile: withPubspecOverrides,
+      containingDescription: RootDescription(dir),
```
sigurdm, line 137: Check if this is right

sigurdm: Updated this to take an explicit description

lib/src/source/git.dart
```diff
@@ -808,13 +821,13 @@ class GitDescription extends Description {
   }
 }
 
-class GitResolvedDescription extends ResolvedDescription {
+class ResolvedGitDescription extends ResolvedDescription {
```
jonasfj, line 824: This rename is completely orthogonal to the rest of the PR, right?

Should we consider doing it in a separate PR upfront, so there is less to review here?

(the rename seems like something that would be easy to review and land)

jonasfj (commented):
I think this looks sound, but I'm not 100% sure.

Could we try to split it into smaller PRs. Some of the refactoring, renaming, etc, could be done in separate PRs.
Even the changes in behavior that we're doing to `RootSource` could be done in a separate PR.
We could possibly even introduce `containingDescription` in a PR without modifying `PathSource` to return `GitDescription` objects just yet.

I see how it makes sense to get everything working. But it would be a LOT faster to review one aspect of the changes at the time.
And we'd be more confident we're not missing something.

 * I'm a little confused about the root/entrypoint names.
 * The `PackageGraph` stuff looks like a possible footgun in the future (though it might already be one).
 * The changes in `RootSource` I don't know about (maybe it's a minor concern).
 * All the places we put in the wrong `containingDescription` because we don't care, scares me a bit... Should we make an `UnknownDescription` we can put in? Just an idea :rofl: 
 * Should we have a test cases where we have a diamond dependency. Ie. two git-dependencies from the same repo that both have a the same path-dependency.

sigurdm (commented):
This change should be a lot more manageable now.

PTAL

jonasfj (approved):

PR #4062, opened 2023-11-28
Surface security advisories in pub get

szakarias:


lib/src/solver/report.dart
```diff
@@ -375,32 +412,57 @@ $contentHashesDocumentationUrl
         maxAge: Duration(days: 3),
       );
 
+      var messagePrefix = '(';
+      var hasAdvisory = false;
+      if (status.advisoriesUpdated != null) {
```
sigurdm, line 417: I don't think you need this check. Just trust getAdvisoriesForPackageVersion to return an empty list if there are no advisories.

lib/src/solver/report.dart
```diff
@@ -375,32 +412,57 @@ $contentHashesDocumentationUrl
...
+        final advisories = await id.source.getAdvisoriesForPackageVersion(
+          id,
+          _cache,
+          Duration(days: 3),
+        );
+        hasAdvisory = advisories != null && advisories.isNotEmpty;
+        if (hasAdvisory) {
+          final advisoryFootnotes = <int>[];
```
sigurdm, line 425: This list will always contain [start, start+1...., end] Perhaps all you need to pass to `_constructAdvisoriesMessage` is the start and end index.

lib/src/solver/report.dart
```diff
@@ -375,32 +412,57 @@ $contentHashesDocumentationUrl
...
+        message =
+            '$messagePrefix${maxAll(versions, Version.prioritize)} available)';
       } else if (
           // Only show newer prereleases for versions where a prerelease is
           // already chosen.
           newId.version.isPreRelease && newerUnstable) {
-        message = '(${maxAll(versions)} available)';
+        message = '$messagePrefix${maxAll(versions)} available)';
+      } else if (hasAdvisory) {
+        message =
```
sigurdm, line 464: This is a bit funky.
How about having a list of notes that you join with a ','.

```
final notes = <String>[];
final advisories = await await id.source.getAdvisoriesForPackageVersion(id, _cache, Duration(days: 3));
if (advisories.isNotEmpty) {
  int start = advisoryFootnotes.length;
  advisoryFootnotes.addAll(advisories);
  int end = advisoryFootnotes.length;
  notes.add(_constructAdvisoryMessage(start, end));
}
if (status.isRetracted) {
  notes.add('retracted');
  if (newerStable) {
    notes.add('${maxAll(versions, Version.prioritize)} available';
  } else if
  ...
} else if (status.isDiscontinued) {
  if (status.discontinuedReplacedBy == null) {
     notes.add('discontinued');
  } else {
     notes.add('discontinued replaced by ${status.discontinuedReplacedBy}');
  }
}
message = notes.isEmpty ? '' : '(${notes.join(', '})';
```

lib/src/source/hosted.dart
```diff
@@ -419,10 +419,20 @@ class HostedSource extends CachedSource {
       if (retracted is! bool) {
         throw FormatException('retracted must be a bool');
       }
+      final advisoriesUpdated = body['advisoriesUpdated'];
+      if (advisoriesUpdated != null && advisoriesUpdated is! String) {
```
sigurdm, line 423: Perhaps if you move the `is! String` check into the condition below, you can avoid the cast.

lib/src/source/hosted.dart
```diff
@@ -537,6 +547,183 @@ class HostedSource extends CachedSource {
...
+    final description = ref.description;
+    if (description is! HostedDescription) {
+      throw ArgumentError('Wrong source');
+    }
+
+    final packageName = description.packageName;
+    final hostedUrl = description.url;
+
+    final url = _listAdvisoriesUrl(ref);
+    log.io('Get security advisories from $url.');
```
sigurdm, line 563:
```suggestion
    log.io('Fetching security advisories from $url.');
```

lib/src/source/hosted.dart
```diff
@@ -537,6 +547,183 @@ class HostedSource extends CachedSource {
...
+        });
+      });
+      final decoded = jsonDecode(bodyText);
+      if (decoded is! Map<String, dynamic>) {
+        throw FormatException('security advisories must be a mapping');
+      }
+      body = decoded;
+      result = _parseAdvisories(decoded, ref.name);
+    } on FormatException catch (error, stackTrace) {
+      _throwFriendlyError(error, stackTrace, packageName, hostedUrl);
```
sigurdm, line 587: This will make the app return non-zero when a advisory listing is malformatted, is that really what we want?

I think a broken advisories listing should still allow resolution to go through successfully...

lib/src/source/hosted.dart
```diff
@@ -537,6 +547,183 @@ class HostedSource extends CachedSource {
...
+
+    // Cache the response on disk.
+    // Don't cache overly big responses.
+    if (bodyText.length < 5000 * 1024) {
+      await _cacheAdvisoriesResponse(body, ref, cache);
+    }
+    return result;
+  }
+
+  List<Advisory>? _parseAdvisories(Map body, String packageName) {
```
sigurdm, line 598: Dartdoc?

Explain the meaning of the `packageName` parameter

lib/src/source/hosted.dart
```diff
@@ -537,6 +547,183 @@ class HostedSource extends CachedSource {
...
+      if (versions is! List) {
+        throw FormatException('package versions must be a list');
+      }
+
+      if (versions.any((element) => element is! String)) {
+        throw FormatException('package version elements must be a string');
+      }
+
+      advisoriesResult
+          .add(Advisory(id, [...versions.map((e) => e.toString())]));
```
sigurdm, line 667: Why do we call `toString` when we know from above it is a String?

Should we parse these versions here and store them as version objects? (With a formatException when a version is not a version)

lib/src/source/hosted.dart
```diff
@@ -537,6 +547,183 @@ class HostedSource extends CachedSource {
...
+            // too old
+            tryDeleteEntry(advisoriesCachePath);
+          } else {
+            return _parseAdvisories(doc, id.toRef().name);
+          }
+        } on io.IOException {
+          // Could not read the file. Delete if it exists.
+          tryDeleteEntry(advisoriesCachePath);
+        } on FormatException {
+          tryDeleteEntry(advisoriesCachePath);
```
sigurdm, line 717: Perhaps do a `log.fine('Failed to read cached version listing: $e');`

lib/src/source/hosted.dart
```diff
@@ -537,6 +547,183 @@ class HostedSource extends CachedSource {
...
+            .advisoriesUpdated;
+    if (advisoriesUpdated == null) return null;
+
+    Future<List<Advisory>?> readAdvisoriesFromCache() async {
+      final advisoriesCachePath = _advisoriesCachePath(id.toRef(), cache);
+      final stat = io.File(advisoriesCachePath).statSync();
+
+      if (stat.type == io.FileSystemEntityType.file) {
+        if (advisoriesUpdated.isAfter(stat.modified)) {
+          return null;
```
sigurdm, line 690: Let's also delete the file in this case

lib/src/source/hosted.dart
```diff
@@ -1504,6 +1767,22 @@ class _VersionInfo {
   );
 }
 
+/// Information about advisories affecting a package retrieved from
+/// /api/packages/$package/advisories
+class PackageAdvisories {
```
sigurdm, line 1772: I think this class is unused by now

lib/src/source/hosted.dart
```diff
@@ -1504,6 +1767,22 @@ class _VersionInfo {
...
+  final Map<String, List<String>> affectedVersions;
+
+  PackageAdvisories(
+    this.affectedVersions,
+  );
+}
+
+class Advisory {
```
sigurdm, line 1780: Add a small DartDoc

sigurdm (approved):
Lot's of small nits - but seems basically sound!

lib/src/source/hosted.dart
szakarias: done

lib/src/source/hosted.dart
szakarias: removed

lib/src/source/hosted.dart
szakarias: done

lib/src/source/hosted.dart
szakarias: done

lib/src/source/hosted.dart
szakarias: done

lib/src/source/hosted.dart
szakarias: done

lib/src/source/hosted.dart
szakarias: Logged it as a warning for now. Maybe it should be an error?

lib/src/solver/report.dart
```diff
@@ -45,6 +47,10 @@ class SolveReport {
   /// Version list will not contain any retracted package versions.
   final Map<String, List<Version>> _availableVersions;
 
+  static const githubAdvisoriesPrefixUrl = 'https://github.com/advisories/';
+  static const maxAdvisoryFootnotesPerLine = 5;
+  final advisoriesMap = <int, String>{};
```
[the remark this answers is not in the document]
szakarias: done

lib/src/solver/report.dart
szakarias: done

lib/src/solver/report.dart
szakarias: I will keep this, since as a follow up I would like to to an optimization where footnotes can be reused between dependencies. In that case the assumption that the footnotes are consecutive is not correct.

lib/src/solver/report.dart
szakarias: Much better. Done, except that the message is null when notes is empty.

lib/src/source/hosted.dart
szakarias: done.

PR #4064, opened 2023-11-30
Disregard ignored advisories in pub get/upgrade report

szakarias:


lib/src/solver/report.dart
```diff
@@ -413,14 +413,17 @@ $contentHashesDocumentationUrl
...
         id,
         _cache,
         Duration(days: 3),
       );
 
       if (advisories != null && advisories.isNotEmpty) {
         final advisoryFootnotes = <int>[];
+        advisories = advisories
+            .where((adv) => !_rootPubspec.ignoredAdvisories.contains(adv.id))
+            .toList();
```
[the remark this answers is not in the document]
szakarias: done

lib/src/pubspec_parse.dart
```diff
@@ -138,6 +138,40 @@ abstract class PubspecBase {
...
+      if (value is! String) {
+        ignoredAdvisoriesError(node.span);
+      }
+      advisoryIDs.add(value);
+    }
+
+    return _ignoredAdvisories = List.unmodifiable(advisoryIDs);
+  }
+
+  List<String>? _ignoredAdvisories;
```
sigurdm, line 173: Should we use a Set<String> instead to make membership-checks fast?

sigurdm (approved):

lib/src/pubspec_parse.dart
sigurdm: SG.

PR #4077, opened 2023-12-05
Remove support for sending analytics

sigurdm:
Fixes: https://github.com/dart-lang/pub/issues/4074


parlough:
Nothing against this, but I'm curious about the reasoning?

Issue #4075, opened 2023-12-04
Refactor change-detection for enforce-lockfile


sigurdm:
Could be made simpler

PR #4078, opened 2023-12-05
Refactor change-detection for --enforce-lockfile

sigurdm:
Should make it more obvious that `--enforce-lockfile` does the right thing.

(Notable there was a discrepancy such that in some cases changing a dependency to a dev-dependency would count as a change, and sometimes not. Now it is not considered a change.)

Fixes: https://github.com/dart-lang/pub/issues/4075

### DIFF
--- a/lib/src/solver/report.dart
+++ b/lib/src/solver/report.dart
@@ -69,14 +69,11 @@ class SolveReport {
   ///
   /// If [summary] is `true` a count of changes and number of
   /// discontinued/retracted packages will be shown at the end of the report.
-  ///
-  /// Returns `true` if there was any change of dependencies relative to the old
-  /// lockfile.
-  Future<bool> show({required bool summary}) async {
+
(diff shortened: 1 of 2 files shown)


lib/src/solver/report.dart
```diff
@@ -494,7 +491,7 @@ $contentHashesDocumentationUrl
         dependencyTypeChanged ||
         message != null ||
         isOverridden)) {
-      return changed || addedOrRemoved;
+      return changed || addedOrRemoved || dependencyTypeChanged;
```
szakarias, line 494: Why is it added here?

szakarias:
Consider clarifying the documentation and help text such that it is more clear for our users what this feature does.

lib/src/solver/report.dart
sigurdm: To align with the other return

Issue #4094, opened 2023-12-22
Failing `dart pub get` locks the terminal


feinstein:
# Environment

- Dart version (run `dart --version`): 3.1.5
- OS kind and version (e.g. "Windows 10, version 1809" or "macOS 12.4"): MacOS Sonoma 14.2.1
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

I want to use `dart pub get --enforce-lockfile` in my CI to validate that the devs didn't forget to upload pubspec.lock after there was a change in pubspec.yaml (unfortunately I have seen this a lot), but in my tests `dart pub get --enforce-lockfile` locks the terminal and doesn't finish (you can see I had to hit `^C` to stop it).

![enforce lock file](https://github.com/dart-lang/pub/assets/8120472/25c24861-ad4e-4f48-91ff-6a5593daafba)

So I can't use it and test if it returns an exit code different than 0 for the CI to validate.

EDIT: This is happening without the `--enforce-lockfile` flag as well.

sigurdm:
Do you have a pubspec.yaml that reproduces this?

sigurdm:
When doing http requests with retries we retry on all `http.ClientExceptions`. Since https://github.com/dart-lang/pub/issues/3317. This is necessary because a client can time out (eg after a long upload), resulting in a ClientException.

Problem is that we do speculative requesting of version-listings, but if the solver can detect a solve failure early, it will be reported, pub will close the shared http client.

Closing the client causes any ongoing speculative requests to fail immediately - with a `ClientException` - and thus retried 7 times, each time failing with another `ClientException` because the client is closed.

These retries prevent the dartdev process from exiting.

I have only been able to reproduce reliable on mac-os (tried also on linux). Using a pubspec with an unresolvable constraint. My guess is that on linux the ClientExceptions are thrown faster or something...

```
# pubspec.yaml
name: a
environment:
  sdk: ^3.0.0
  
homepage: 'http://abc.def'

dev_dependencies:
  test: ^0.2.0
```

```
> dart --version
Dart SDK version: 3.2.3 (stable) (Tue Dec 5 17:58:33 2023 +0000) on "macos_arm64"
> dart pub get 
Resolving dependencies... 
Because a depends on test ^0.2.0 which doesn't match any versions, version solving failed.


You can try the following suggestion to make the pubspec resolve:
* Try upgrading your constraint on test: dart pub add dev:test:^1.25.0
[....hanging for 10 seconds...]
>
```

[...]

Issue #3317, opened 2022-02-24
Cannot publish package using flutter pub publish


martin-braun:
# Environment

- `pub version` or `flutter pub version`: Pub 2.16.1
- OS version: MacOS 12.1 (21C52)
- Are you using the Chinese community mirror or a corporate firewall? No

<details>
<summary>Output of `flutter doctor -v`</summary>

```
[✓] Flutter (Channel stable, 2.10.2, on macOS 12.1 21C52 darwin-x64, locale en-US)
    • Flutter version 2.10.2 at /usr/local/Caskroom/flutter/2.10.2/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 097d3313d8 (6 days ago), 2022-02-18 19:33:08 -0600
    • Engine revision a83ed0e5e3
    • Dart version 2.16.1
    • DevTools version 2.9.2

[✓] Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)
    • Android SDK at /Users/marty/Library/Android/sdk
    • Platform android-31, build-tools 32.1.0-rc1
    • ANDROID_HOME = /Users/marty/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.11+0-b60-7590822)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 13.2.1)
    • Xcode at /Applications/Xcode-13.2.1.app/Contents/Developer
    • CocoaPods version 1.11.2

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2021.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.11+0-b60-7590822)

[✓] VS Code (version 1.64.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.34.0

[✓] Connected device (2 available)
    • Pixel 3a (mobile) • 992AY189GC • android-arm64  • Android 12 (API 31)
[...]

PR #4095, opened 2023-12-22
Avoid retrying queries after closing client

sigurdm:
Fix of https://github.com/dart-lang/pub/issues/4094

Keeps track of whether we have closed the global http-client (indicating processing is done, and we are ready to exit the program, and avoid doing retries at that point), in that case avoid doing restarts.

Would be more elegant to have this as a method on the client itself - avoiding a reference to the global client, but many times we wrap the global client in an Oauth2 client.

Would be more elegant if we could distinguish the ClientException we get when a request times out, from the one we get when the client is closed. (And not rely on parsing the message string).

Hard to write a good test for, but manual testing has shown that the hanging problem goes away with this fix.

### The problem:
When doing http requests with retries we retry on all `http.ClientExceptions`. Since https://github.com/dart-lang/pub/issues/3317. This is necessary because a client can time out (eg after a long upload), resulting in a ClientException.

Problem is that we do speculative requesting of version-listings, but if the solver can detect a solve failure early, it will be reported, pub will close the shared http client.

Closing the client causes any ongoing speculative requests to fail immediately - with a `ClientException` - and thus retried 7 times, each time failing with another `ClientException` because the client is closed.

These retries prevent the dartdev process from exiting.

I have only been able to reproduce reliable on mac-os (tried also on linux). Using a pubspec with an unresolvable constraint. My guess is that on linux the ClientExceptions are thrown faster or something...

```
# pubspec.yaml
name: a
environment:
  sdk: ^3.0.0
  
homepage: 'http://abc.def'

dev_dependencies:
  test: ^0.2.0
```

```
> dart --version
Dart SDK version: 3.2.3 (stable) (Tue Dec 5 17:58:33 2023 +0000) on "macos_arm64"
> dart pub get 
Resolving dependencies... 
Because a depends on test ^0.2.0 which doesn't match any versions, version solving failed.


You can try the following suggestion to make the pubspec resolve:
* Try upgrading your constraint on test: dart pub add dev:test:^1.25.0
[....hanging for 10 seconds...]
>
```

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -49,11 +49,22 @@ class _PubHttpClient extends http.BaseClient {
 
   @override
   Future<http.StreamedResponse> send(http.BaseRequest request) async {
+    if (_wasClosed) {
+      throw StateError('Attempting to send request on closed client');
+    }
     _requestStopwatches[request] = Stopwatch()..start();
     request.headers[HttpHeaders.userAgentHeader] = 'Dart pub ${sdk.version}';
     _logRequest(request);
-
-    final streamedResponse = await _inner.send(request);
+    final http.StreamedResponse streamedResponse;
+    try {
+      streamedResponse = await _inner.send(request);
+    } on http.ClientException {
+      if (_wasClosed) {
+        // Avoid retrying in this case.
+        throw _ClientClosedException();
+      }
+      rethrow;
+    }
 
     _logResponse(streamedResponse);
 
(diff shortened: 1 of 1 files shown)


parlough:
I think this explains a lot of the hangs I've been seeing with pub actually. Thanks for the investigation and explanations! Would love to see this landed :D

sigurdm:
@parlough yes, I think this could explain a lot of it.

Instead of building a full SDK you should be able to reproduce/test this with a clone of the pub repo by running:

```
> dart path/to/pub_repo/tool/test-bin/pub_command_runner.dart pub get
```

lib/src/http.dart
```diff
@@ -325,10 +333,11 @@ Future<T> retryForHttp<T>(String operation, FutureOr<T> Function() fn) async {
   return await retry(
     () async => await _httpPool.withResource(() async => await fn()),
     retryIf: (e) async =>
-        (e is PubHttpException && e.isIntermittent) ||
-        e is TimeoutException ||
-        e is http.ClientException ||
-        isHttpIOException(e),
+        !_pubClient._wasClosed &&
```
jonasfj, line 336: Let's make a `class ClientClosedException implements ClientException` which indicates that the request failed, because the client was closed while the request was in progress. Such exceptions should not be retried.

sigurdm: Yeah that's better.

Done (though I avoided implementing `ClientException`).

lib/src/http.dart
```diff
@@ -417,3 +436,8 @@ extension RequestSending on http.Client {
     return streamedResponse;
   }
 }
+
+/// Thrown by [_PubHttpClient.send] if the client was closed while the request
+/// was being processed. Notably it doesn't implement [http.ClientException],
+/// and thus does not trigger a retry by [retryForHttp].
+class _ClientClosedException implements Exception {}
```
jonasfj, line 443: Do consider making a toString function :D

jonasfj (approved):

PR #4106, opened 2024-01-10
Add security advisory test

szakarias:


sigurdm (approved):
LGTM

Makes me wonder - ideally we'd have a place you could go to see them all...

Issue #4102, opened 2024-01-02
`pub download` command


mit-mit:
I'd like to propose a new `pub download` command, which can be used to download a package to local storage. Sample use cases for this include:

* Review the source code of a package
* Read and run the `/example` of a package
* Downloading a copy into a private pub server

This new command would take a single argument, a package name, and would then lookup the more recent stable version, and then download the files from that into a new directory `<package name>-<package version` in the current working directory.

mit-mit:
Here's a sample implementation:
https://github.com/dart-lang/pub/pull/4103

sigurdm:
I think it is true that the barrier for running an example from a package is too high. Maybe this is the right way to do it...

Maybe the command-name ambiguous? `dart pub download` could be confused with `dart pub get` as the command retrieving dependencies... Not sure. And also not sure I have a better name on hand.

We do have `dart pub cache add` for downloading a package to the pub-cache. But that is probably too obscure for discovery, and you cannot run examples out of the cache directly (but it would work for reviewing the code).

> Downloading a copy into a private pub server

I think this use-case is a bit speculative. Most likely a private pub-server would like to download the .tar.gz archive, not unpacked...

PR #4111, opened 2024-01-15
Implement command "unpack"

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/4102


test/testdata/goldens/help_test/pub unpack --help.txt
```diff
@@ -0,0 +1,19 @@
...
+## Section 0
+$ pub unpack --help
+Downloads a package and unpacks it in place.
+Will resolve dependencies in the folder unless `--no-resolve` is passed.
+
+Usage: pub unpack package-name[:version]
+-h, --help            Print this usage information.
+    --[no-]resolve    Whether to do pub get in the downloaded folder
+                      (defaults to on)
+    --destination     Download the package in this dir
```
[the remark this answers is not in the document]
mit-mit: +1

test/testdata/goldens/help_test/pub unpack --help.txt
```diff
@@ -0,0 +1,19 @@
+# GENERATED BY: test/help_test.dart
+
+## Section 0
+$ pub unpack --help
+Downloads a package and unpacks it in place.
+Will resolve dependencies in the folder unless `--no-resolve` is passed.
+
+Usage: pub unpack package-name[:version]
```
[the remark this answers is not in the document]
mit-mit: That uses `:descriptor`

But it's a bit confusing what that means, and all the text there talks about constraints, so maybe use `:constraint` ?

test/testdata/goldens/help_test/pub unpack --help.txt
```diff
@@ -0,0 +1,19 @@
...
+-h, --help            Print this usage information.
+    --[no-]resolve    Whether to do pub get in the downloaded folder
+                      (defaults to on)
```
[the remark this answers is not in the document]
sigurdm: > If it defaults to on do we need the --resolve option?

In think this is the standard way of doing flags, it gives extra flexibility. In this way you can eg. make an alias `unpack = dart pub unpack --no-resolve` and use it as `unpack --resolve`. But if it is important that it cannot be un-negated I'm open...

> +1 to only having --no-resolve, and also suggest we only show this in verbose help

Currently pub doesn't have a concept of verbose help... we could make that...

test/testdata/goldens/help_test/pub unpack --help.txt
```diff
@@ -0,0 +1,19 @@
...
+    --destination     Download the package in this dir
+                      (defaults to ".")
+    --repository      The package repository to download from
```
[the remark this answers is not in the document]
sigurdm: Yeah, that would make sense...

sigurdm:
> Nice. Can we bring over the support for --example from https://github.com/dart-lang/pub/pull/4103/files 

I don't think it will work well - usually the example folder will have a path dependency on the main library, and that cannot resolve if we extract it out of context.

mit-mit:
> I don't think it will work well - usually the example folder will have a path dependency on the main library, and that cannot resolve if we extract it out of context.

Oh, I had not thought of that (but just sampled some top packages, and you are absolutely right; it's very common).

test/testdata/goldens/help_test/pub unpack --help.txt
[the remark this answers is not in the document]
mit-mit: I think we should consider that. It's quite common in other sub-commands:
https://github.com/search?q=repo%3Adart-lang%2Fsdk%20%22hide%3A%20!verbose%22&type=code

lib/src/command/unpack.dart
```diff
@@ -0,0 +1,132 @@
...
+import '../utils.dart';
+
+/// Handles the `deps` pub command.
+class UnpackCommand extends PubCommand {
+  @override
+  String get name => 'unpack';
+
+  @override
+  String get description => 'Downloads a package and unpacks it in place.\n'
+      'Will resolve dependencies in the folder unless `--no-resolve` is passed.';
```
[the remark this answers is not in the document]
sigurdm: I added more examples and an extra newline.

test/testdata/goldens/help_test/pub unpack --help.txt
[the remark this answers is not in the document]
sigurdm: I changed to use the syntax in `dart pub add` (without the possibility for `dev:` and `override:`.

I called it `constraint` in the description, but give examples with a full descriptor.

test/testdata/goldens/help_test/pub unpack --help.txt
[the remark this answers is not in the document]
sigurdm: I made it only show when `--verbose`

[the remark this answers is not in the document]
szakarias: I also vote for have both the positive and the negation. If I am in doubt which is the default I can pass that flag and it should work.

lib/src/command/unpack.dart
```diff
@@ -0,0 +1,220 @@
...
+
+  $topLevelProgram pub unpack 'foo:{hosted:"https://my_repo.org"}'
+
+Downloads and extracts latest stable version of package:foo from my_repo.org
+in a directory `foo-<version>`.
+
+
+Will resolve dependencies in the folder unless `--no-resolve` is passed.
+''';
+
```
szakarias, line 55: Do you want to include something in the example about a `resolve` flag?

lib/src/command/unpack.dart
```diff
@@ -0,0 +1,220 @@
...
+    versions.sort((id1, id2) => id1.version.compareTo(id2.version));
+
+    final id = versions.last;
+    final name = id.name;
+
+    final outputArg = argResults['output'] as String;
+    final destinationDir = p.join(outputArg, '$name-${id.version}');
+    if (entryExists(destinationDir)) {
+      fail('Target directory `$destinationDir` already exists.');
+    }
```
szakarias, line 124: Maybe it would be nice with an option to override the existing folder?

szakarias: Or allow a copy? This is a nit.

lib/src/command/unpack.dart
```diff
@@ -0,0 +1,220 @@
...
+      } finally {
+        log.message('To explore type: cd $destinationDir');
+        if (e.example != null) {
+          log.message('To explore the example type: cd ${e.example!.rootDir}');
+        }
+      }
+    }
+  }
+
+  // TODO(sigurdm): Refactor this to share with `add`.
```
szakarias, line 147: Add issue number for this TODO?

lib/src/command/unpack.dart
```diff
@@ -0,0 +1,220 @@
...
+
+  @override
+  String get description => '''
+Downloads a package and unpacks it in place.
+
+For example:
+
+  $topLevelProgram pub unpack foo
+
+Downloads and extracts the lastest stable package:foo from pub.dev in a
```
szakarias, line 34: Typo: lastes -> latest

lib/src/command/unpack.dart
```diff
@@ -0,0 +1,220 @@
...
+              packageName,
+              cache.hosted.defaultUrl,
+            ),
+          ),
+      constraint,
+    );
+  }
+}
+
+class _ParseResult {
```
szakarias, line 216: Simplify as we discussed

lib/src/command/unpack.dart
szakarias: If we simplify, remove the todo instead

lib/src/command/unpack.dart
sigurdm: Done

lib/src/command/unpack.dart
sigurdm: I made a `--force` flag to overwrite the folder if it exists

test/unpack_test.dart
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) 2013, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
sigurdm: Done

lib/src/command/unpack.dart
sigurdm: Done

lib/src/command/unpack.dart
sigurdm: Simplified

szakarias (approved):
LGTM

PR #4112, opened 2024-01-16
Include bare `pub --help` command in help_test

sigurdm:
I noticed we had no golden for `pub --help`.

Trying to fix that.


jonasfj (approved):

szakarias (approved):
LGTM

szakarias:
Shouldn't we just land this?

sigurdm:
Yeah - sorry I was hesitating a bit with this.

I discovered that `pub --help` is already covered by the `test/embedding/` tests. Maybe we should move all the help tests there...

Issue #4117, opened 2024-01-18
Split `dart pub publish` in two steps


jonasfj:
* Create tarball: `dart pub publish --to-archive=<output.tar.gz>`
 * Publish tarball: `dart pub publish --from-archive=<output.tar.gz>`

I suggest we make the options hidden for now.

TODO: Consider if we can come up with better names for the arguments. Maybe `--upload` instead from `--from-tarball`, but then what would `--to-archive` be?

szakarias:
fixed by https://github.com/dart-lang/pub/pull/4119

PR #4119, opened 2024-01-19
Allow publishing to and from a .tar.gz archive

sigurdm:
https://github.com/dart-lang/pub/issues/4117

Also fixes `--skip-validation` to no longer imply `--force` (ie, you now still have to confirm the upload even though you use `--skip-validation`.
This is potentially breaking, but we think this is the right behavior, and clients using `--skip-validation` from CI will have to update their scripts.


lib/src/command/lish.dart
```diff
@@ -92,6 +99,19 @@ class LishCommand extends PubCommand {
...
+    argParser.addOption(
+      'to-archive',
+      help: 'Create a .tar.gz archive instead of publishing to server',
+      valueHelp: '[path/to/archive.tar.gz]',
+      hide: true,
+    );
+    argParser.addOption(
+      'from-archive',
+      help:
+          'Publish from a .tar.gz archive instead of current folder. Will not do client-side validations.',
```
jonasfj, line 111:
```suggestion
          'Publish from a .tar.gz archive instead of current folder. Implies `--skip-validation`.',
```

lib/src/command/lish.dart
```diff
@@ -92,6 +99,19 @@ class LishCommand extends PubCommand {
       help: 'The package server to which to upload this package.',
       hide: true,
     );
+    argParser.addOption(
+      'to-archive',
+      help: 'Create a .tar.gz archive instead of publishing to server',
+      valueHelp: '[path/to/archive.tar.gz]',
```
jonasfj, line 105:
```suggestion
      valueHelp: '[archive.tar.gz]',
```

lib/src/command/lish.dart
```diff
@@ -92,6 +99,19 @@ class LishCommand extends PubCommand {
...
+      hide: true,
+    );
+    argParser.addOption(
+      'from-archive',
+      help:
+          'Publish from a .tar.gz archive instead of current folder. Will not do client-side validations.',
+      valueHelp: '[path/to/archive.tar.gz]',
```
jonasfj, line 112:
```suggestion
      valueHelp: '[archive.tar.gz]',
```

lib/src/command/lish.dart
```diff
@@ -251,64 +271,120 @@ the \$PUB_HOSTED_URL environment variable.''',
...
+    if (_toArchive != null && dryRun) {
+      usageException('Cannot use both --to-archive and --dry-run.');
+    }
+
+    if (_fromArchive != null && force) {
+      usageException('Cannot use both --from-archive and --force.');
+    }
+
+    if (_toArchive != null && force) {
+      usageException('Cannot use both --to-archive and -force.');
```
jonasfj, line 291:
```suggestion
      usageException('Cannot use both --to-archive and --force.');
```

lib/src/command/lish.dart
```diff
@@ -251,64 +271,120 @@ the \$PUB_HOSTED_URL environment variable.''',
...
+    if (_fromArchive != null && dryRun) {
+      usageException('Cannot use both --from-archive and --dry-run.');
+    }
+
+    if (_toArchive != null && dryRun) {
+      usageException('Cannot use both --to-archive and --dry-run.');
+    }
+
+    if (_fromArchive != null && force) {
+      usageException('Cannot use both --from-archive and --force.');
```
jonasfj, line 287: Why not? will `--from-archive` not prompt?

lib/src/command/lish.dart
```diff
@@ -251,64 +271,120 @@ the \$PUB_HOSTED_URL environment variable.''',
...
+    if (_fromArchive != null && _toArchive != null) {
+      usageException('Cannot use both --from-archive and --to-archive.');
+    }
+
+    if (_fromArchive != null && dryRun) {
+      usageException('Cannot use both --from-archive and --dry-run.');
+    }
+
+    if (_toArchive != null && dryRun) {
+      usageException('Cannot use both --to-archive and --dry-run.');
```
jonasfj, line 283: Why not, just don't create the archive? It's effectively the same as `--dry-run` without `--to-archive`.

lib/src/command/lish.dart
```diff
@@ -251,64 +271,120 @@ the \$PUB_HOSTED_URL environment variable.''',
       usageException('Cannot use both --force and --dry-run.');
     }
 
-    if (entrypoint.root.pubspec.isPrivate) {
+    if (_fromArchive != null && _toArchive != null) {
+      usageException('Cannot use both --from-archive and --to-archive.');
+    }
+
+    if (_fromArchive != null && dryRun) {
+      usageException('Cannot use both --from-archive and --dry-run.');
```
jonasfj, line 279: We could do the same level of minimal validation, couldn't we?

lib/src/command/lish.dart
```diff
@@ -251,64 +271,120 @@ the \$PUB_HOSTED_URL environment variable.''',
...
+
+    if (_toArchive != null && force) {
+      usageException('Cannot use both --to-archive and -force.');
+    }
+
+    if (_fromArchive == null && entrypoint.root.pubspec.isPrivate) {
       dataError('A private package cannot be published.\n'
           'You can enable this by changing the "publish_to" field in your '
           'pubspec.');
     }
```
jonasfj, line 298: Hmm, isn't it wrong to access `entrypoint`? I mean it reads from current directory, which is always wrong when `--from-archive` is given.

We should assume that CWD is not a package, and certainly not the package that you want to publish.

test/lish/publishing_to_and_from_archive_test.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'package:test/test.dart';
+
+import '../descriptor.dart' as d;
+import '../test_pub.dart';
+import 'utils.dart';
+
+void main() {
+  test('Can publish into and from archive', () async {
+    final server = await servePackages();
+    await d.validPackage().create();
```
jonasfj, line 20: Consider making a test where we create the valid package in a subfolder.

And do `dart pub publish --to-archive=../archive.tar.gz`

So that we later can go a level up and call `dart pub publish --from-archive=archive.tar.gz`

That'll help test if we're accidentally relying on files from CWD.

lib/src/command/lish.dart
sigurdm: We only access the entrypoint when `_fromArchive` is null. Lazy eval etc.

lib/src/command/lish.dart
sigurdm: ++

lib/src/command/lish.dart
sigurdm: ++

lib/src/command/lish.dart
sigurdm: ++

lib/src/command/lish.dart
sigurdm: ++

lib/src/command/lish.dart
sigurdm: Yeah - this is a mistake. Removed.

lib/src/command/lish.dart
sigurdm: Yeah why give a file-name if you don't want the file. It indicates you have broken logic somewhere.

lib/src/command/lish.dart
sigurdm: I don't think it really makes sense - we could extract the archive and do the full validation - but I don't think we want

lib/src/command/lish.dart
```diff
@@ -381,3 +477,26 @@ String _readableFileSize(int size) {
     return '<1 KB';
   }
 }
+
+class _Publication {
```
jonasfj, line 481: I personally love tuples, but feel free to use a class. I just couldn't help typing it up:
```dart
typedef _Publication = ({
  Uint8List packageBytes,
  int warningCount,
  int hintCount,
  Pubspec pubspec,
});

extension on _Publication {
  String get warningCountMessage {
    return [
      '\n',
      'Package has $warningCount ${pluralize('warning', warningCount)}',
      if (hintsCount > 0)
        ' and $hintCount ${pluralize('hint', hintCount)}',
      '.',
    ].join('');
  }
}
```

lib/src/command/lish.dart
```diff
@@ -381,3 +477,26 @@ String _readableFileSize(int size) {
...
+  Uint8List packageBytes;
+  int warningsCount;
+  int hintsCount;
```
jonasfj, line 484: nit: I think it is more right, but not 100% sure... not important :D
```suggestion
  int warningCount;
  int hintCount;
```

jonasfj (approved):

lib/src/command/lish.dart
```diff
@@ -251,64 +271,116 @@ the \$PUB_HOSTED_URL environment variable.''',
...
       dataError('A private package cannot be published.\n'
           'You can enable this by changing the "publish_to" field in your '
           'pubspec.');
     }
+  }
 
-    if (!skipValidation) {
-      await entrypoint.acquireDependencies(SolveType.get);
-    } else {
+  Future<_Publication> _publicationFromEntrypoint() async {
```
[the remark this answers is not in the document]
sigurdm: Did it in the publish step intead

PR #4129, opened 2024-01-30
Remove nullability of parameter

sigurdm:


lib/src/entrypoint.dart
```diff
@@ -603,14 +603,11 @@ To update `$lockFilePath` run `$topLevelProgram pub get`$suffix without
   ///
   /// If [changed] is passed, only dependencies whose contents might be changed
```
szakarias, line 604: Update doc

szakarias (approved):

lib/src/entrypoint.dart

Issue #4026, opened 2023-09-25
Add a pre-publish check for DevTools extension validation


kenzieschmoll:
This is a feature originally requested and discussed here: https://github.com/rrousselGit/provider/pull/832#discussion_r1327918500.

This check would fail on `pub publish --dry-run` if a DevTools extension is expected and none was found.

rrousselGit:
It'd be cool if the check could not only very that the `build` folder exists, but also check that the build output is more recent than the source files.

When thinking about this problem, I realized that gitignored files would still be present on the computer at the publish step. So if publishing locally instead of in a CI, the check could incorrectly pass.

sigurdm:
> When thinking about this problem, I realized that gitignored files would still be present on the computer at the publish step. So if publishing locally instead of in a CI, the check could incorrectly pass.

I think we can get around this problem by only considering files not gitignored.

PR #4135, opened 2024-02-08
Devtools Extension validator

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/4026


lib/src/validator/devtools_plugin.dart
```diff
@@ -0,0 +1,46 @@
...
+          !files.any(
+            (f) => p.isWithin(
+              p.join(entrypoint.rootDir, 'extension', 'devtools', 'build'),
+              f,
+            ),
+          )) {
+        warnings.add('''
+It looks like you are making a devtools extension!
+
+The folder `extensions/devtools` should contain both a
```
kenzieschmoll, line 39: extensions --> extension

lib/src/validator/devtools_plugin.dart
```diff
@@ -0,0 +1,46 @@
...
+* `config.yaml` file and a
+* non-empty `build` director'
```
kenzieschmoll, line 41: directory

kenzieschmoll (approved):
lgtm with a couple comments. Thanks for the change!

lib/src/validator/devtools_plugin.dart
```diff
@@ -0,0 +1,46 @@
...
+              p.join(entrypoint.rootDir, 'extension', 'devtools', 'build'),
+              f,
+            ),
+          )) {
+        warnings.add('''
+It looks like you are making a devtools extension!
+
+The folder `extension/devtools` should contain both a
+* `config.yaml` file and a
+* non-empty `build` directory'
```
jonasfj, line 41: Can we have a `dart.dev/go/...` link for the documentation? Or another kinds of short-perma-link.

lib/src/validator/devtools_plugin.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'dart:async';
+import 'package:path/path.dart' as p;
+
+import '../io.dart';
+import '../validator.dart';
+
+//   If the `extension/devtools/` directory exists
+//  verify the directory contains a file `config.yaml` and a non-empty `build/` directory
+class DevtoolsPluginValidator extends Validator {
+  static String docRef = 'See https://dart.dev/tools/pub/package-layout.';
```
jonasfj, line 14: are we sure we want dot at the end of a URL.

Are we sure this is a good link?

jonasfj (approved):
I don't mind this, but I think it's a bit overkill.

It might be better to make a package, like `devtools_extension_test` that devtools extensions authors add as a dev-dependency, and then they do:

```dart
// test/devtools_extension_test.dart
import 'package:devtools_extension_test/devtools_extension_test.dart';

void main() => testDevtoolsExtension('my_root_package_name');
```

Sure, it's annoying that devtools extension authors have to write this file, and make another dev-dependency, but it's much easier for you to maintain. And much easier to add more checks in the future.


-------------

I don't strictly mind this, we can always remove it, if one day it is outdated and unmaintained, and people can always use `--skip-validation` if they need to push through it.

lib/src/validator/devtools_plugin.dart
sigurdm: I think this is a copy-paste mistake :) @kenzieschmoll do we have a good spot to point at about the devtool extensions?

lib/src/validator.dart
```diff
@@ -159,6 +160,7 @@ abstract class Validator {
       PubspecTypoValidator(),
       LeakDetectionValidator(),
       SizeValidator(),
+      DevtoolsPluginValidator(),
```
kenzieschmoll, line 163: nit: rename to `DevToolsExtensionValidator` since that matches the `extension` convention.

lib/src/validator/devtools_plugin.dart
kenzieschmoll: Yes you can use this link: https://docs.flutter.dev/tools/devtools/extensions

kenzieschmoll:
> It might be better to make a package, like devtools_extension_test that devtools extensions authors add as a dev-dependency

I'd prefer if this was part of the pub publish validation (what is implemented in this PR). Looking at some of the other validators that are supported, it doesn't seem out of place. Since we are supporting extensions through a special directory in a Dart package, it seems valid that we should have some validation for that. Making users write a test for something that is simple for us to implement seems like a poor UX, IMO, and I think we'd end up in a state where extension authors don't always write the test, which means we'd have broken packages. 

Plus, moving forward, other types of extensions will use this `extension` directory (e.g. analyzer, VS Code extension promotion), and we'll likely want to have validation for those configurations too.

lib/src/validator.dart
sigurdm: Done

kenzieschmoll:
I added a note about pub publish validation in the `devtools_extensions` readme here: https://github.com/flutter/devtools/pull/7237/files#diff-321d9c1e60b7290841625e86c9627153c83cc53497a9af52bb37101794b770d1

PR #4136, opened 2024-02-09
Surface security advisories in pub outdated

szakarias:


lib/src/command/outdated.dart
```diff
@@ -240,6 +241,11 @@ Consider using the Dart 2.19 sdk to migrate to null safety.''');
         current.version,
         cache,
       );
+
+      final id = current ?? upgradable ?? resolvable ?? latest;
+      final packageAdvisories = await id?.source
+          .getAdvisoriesForPackage(id, cache, Duration(days: 3));
```
sigurdm, line 247: Nit: You could do
```suggestion
      final packageAdvisories = await id?.source
          .getAdvisoriesForPackage(id, cache, Duration(days: 3)) ?? [];
```
To avoid nullness.

lib/src/source/hosted.dart
```diff
@@ -1800,9 +1805,14 @@ class Advisory {
...
   List<String> affectedVersions;
   List<String> aliases;
-  Advisory(this.id, this.affectedVersions, this.aliases);
+  String summary;
+  Advisory(this.id, this.affectedVersions, this.aliases, this.summary);
 }
 
+// TODO(https://github.com/dart-lang/pub-dev/issues/7471) Use `pub_display_url`
+// once this issue is resolved.
+String advisoriesDisplayUrl(String id) => 'https://github.com/advisories/$id';
```
sigurdm, line 1814: Perhaps extract this to a separate PR

sigurdm (approved):

lib/src/command/outdated.dart
szakarias: done

PR #4140, opened 2024-02-15
Make a RootDescription not depend on the root Package

sigurdm:
This is refactor that is extracted from https://github.com/dart-lang/pub/pull/4060 . It will allow a RootDescription to be created before the pubspec is loaded.

This removes the concept of an "in-memory package". These packages were always a modified version of a pubspec, being used for trying out a resolution, and they might as well be associated with the directory of the original pubspec.

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -17,7 +17,8 @@ import 'pubspec.dart';
 import 'system_cache.dart';
 import 'utils.dart';
 
-/// A named, versioned, unit of code and resource reuse.
+/// A Package is a [Pubspec] and a directory where it belongs that can be used
+/// for version solving or as a node in a package graph.
 class Package {
   /// Compares [a] and [b] orders them by name then version number.
   ///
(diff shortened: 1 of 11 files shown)


jonasfj (approved):
LGTM; but it might be worth having @szakarias have a look too.

lib/src/package.dart
```diff
@@ -134,21 +119,14 @@ class Package {
...
-
-  /// Constructs a package with the given pubspec.
+  /// Creates a package with [pubspec] associated with [dir].
   ///
-  /// The package will have no directory associated with it.
-  Package.inMemory(this.pubspec) : _dir = null;
-
-  /// Creates a package with [pubspec] located at [dir].
-  Package(this.pubspec, String this._dir);
+  /// For temporary resolution attempts `pubspec` does not have to correspond
```
szakarias, line 127: [pubspec] ?

lib/src/source/root.dart
```diff
@@ -5,12 +5,21 @@
...
 import '../pubspec.dart';
 import '../source.dart';
 import '../system_cache.dart';
 
+/// A root package is the root of a package dependency graph that seeds a
+/// version resolution.
+///
+/// There is no explicit way to depend on a root package.
+///
+/// A root package is the only package for whom dev_dependencies are taken into
```
szakarias, line 18: for which?

szakarias (approved):
Nice one.

lib/src/source/root.dart
sigurdm: Done

lib/src/source/root.dart
```diff
@@ -5,12 +5,21 @@
...
+/// account.
+///
+/// A root package is the only package for whom dependency_overrides are taken
```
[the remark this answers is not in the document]
sigurdm: Done

Issue #4141, opened 2024-02-16
Support Git's safe.bareRepository=explicit config option


jason-simmons:
Some environments configure GIt with the `safe.bareRepository=explicit` option in `git-config` for improved security.

With this setting Flutter's use of pub was failing due to this dependency on a package fetched through Git:
https://github.com/flutter/flutter/blob/master/dev/integration_tests/android_views/pubspec.yaml#L19

The `pubspec.yaml` specifies a Git URL and commit hash.  The `.pub-cache` was storing this package as a Git bare repository.

With `safe.bareRepository=explicit` Pub fails to locate the commit hash:
```
Package not available (Could not find git ref '64d0f6051b9b7b9933d3d16194170a38f544634a'
(fatal: cannot use bare repository '/usr/local/google/home/jsimmons/.pub-cache/git/cache/goldens-9579e0161a174b307c855fff52478bf0dc30bb1a'
(safe.bareRepository is 'explicit'))).
```

Pub should support Git in this configuration.  It may need to pass the `--git-dir` flag to Git when accessing these repositories in the cache.

yanok:
Got affected by this today as well.

PR #4142, opened 2024-02-20
Always pass --git-dir when handling repoCache

sigurdm:
This will allow commands to work even when `safe.bareRepository=explicit` is configured.

Fixes https://github.com/dart-lang/pub/issues/4141


jonasfj (approved):

ultraon:
I didn't know what version of the dart-tools contained this fix. But for the flutter version `3.19.6`, dart `3.3.4` I found a fix as follows:

run in the terminal `git config --global --add safe.bareRepository all`

By default the value was set to `explicit`, you can check it yourself by running `git config --global --get safe.bareRepository`.

Using flutter `3.24.4` works well even without the workaround I mentioned.

Issue #4139, opened 2024-02-14
Dependency overrides cannot override dependency errors


dcharkes:
## Actual behavior

```
Error on line 17, column 11: Invalid description in the "native_assets_builder" pubspec on the "native_assets_cli" dependency: "../native_assets_cli/" is a relative path, but this isn't a local pubspec.
```

With the root `pubspec.yaml` from `flutter_tools`:

```yaml
name: flutter_tools
description: Tools for building Flutter applications
homepage: https://flutter.dev

environment:
  sdk: '>=3.2.0-0 <4.0.0'

dependencies:
  # To update these, use "flutter update-packages --force-upgrade".
  #
  # For detailed instructions, refer to:
  # https://github.com/flutter/flutter/wiki/Updating-dependencies-in-Flutter
  archive: 3.3.2
  args: 2.4.2
  browser_launcher: 1.1.1
  dds: 3.1.2
  dwds: 23.3.0
  completion: 1.0.1
  coverage: 1.7.2
  crypto: 3.0.3
  file: 7.0.0
  flutter_template_images: 4.2.0
  html: 0.15.4
  http: 0.13.6
  intl: 0.18.1
  meta: 1.12.0
  multicast_dns: 0.3.2+6
  mustache_template: 2.0.0
  package_config: 2.1.0
  process: 5.0.2
  fake_async: 1.3.1
  stack_trace: 1.11.1
  usage: 4.1.1
  webdriver: 3.0.3
  webkit_inspection_protocol: 1.2.1
  xml: 6.5.0
  yaml: 3.1.2
  native_stack_traces: 0.5.6
  shelf: 1.4.1
  vm_snapshot_analysis: 0.7.6
  uuid: 3.0.7
  web_socket_channel: 2.4.3
  stream_channel: 2.1.2
  shelf_web_socket: 1.0.4
  shelf_static: 1.1.2
  pub_semver: 2.1.4
  pool: 1.5.1
  path: 1.9.0
  mime: 1.0.5
  logging: 1.2.0
  http_multi_server: 3.2.1
  convert: 3.1.1
  async: 2.11.0
  unified_analytics: 5.8.1

  cli_config: 0.1.2
  graphs: 2.3.1
  # TODO(dacoharkes): Publish and bump
  native_assets_builder: 0.3.2
  native_assets_cli: 0.4.2

  # We depend on very specific internal implementation details of the
  # 'test' package, which change between versions, so when upgrading
  # this, make sure the tests are still running correctly.
[...]

PR #4143, opened 2024-02-22
Delay error reporting for dependencies

sigurdm:
Allows for overriding "bad" dependencies

We could consider taking this even further, such that the solver can look at different versions even if one has a bad dependency. But I think this step makes sense on its own.

Fixes https://github.com/dart-lang/pub/issues/4139


sigurdm:
Closing in favor of https://github.com/dart-lang/pub/pull/4060

PR #4145, opened 2024-02-24
fix: False detection of non-ignored files with non-ascii unicode characters

Gustl22:

<details>

- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).

</details>

### DIFF
--- a/lib/src/git.dart
+++ b/lib/src/git.dart
@@ -85,38 +85,6 @@ Future<List<String>> run(
   }
 }
 
-// Stream<String> start(
-//     List<String> args, {
-//       String? workingDir,
-//       Map<String, String>? environment,
-//     }) async {
-//   if (!isInstalled) {
-//     fail('Cannot find a Git executable.\n'
-//         'Please ensure Git is correctly installed.');
-//   }
-//
-//   log.muteProgress();
-//   try {
-//     final result = await startProcess(
-//       command!,
-//       args,
-//       workingDir: workingDir,
-//       environment: {...?environment, 'LANG': 'en_GB'},
-//     );
-//     if ((await result.exitCode) != 0) {
-//       throw GitException(
-//         args,
-//         result.stdout.join('\n'),
-//         result.stderr.join('\n'),
-//         result.exitCode,
-//       );
-//     }
-//     return result.stdout;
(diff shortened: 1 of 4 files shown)


Gustl22:
Currently we cannot publish with the new dart version, as it does not allow special characters to be in the file name.
See: https://github.com/Gustl22/audioplayers/actions/runs/8031208161/job/21939294316#step:4:873

sigurdm:
Could you run `dart format`?

Gustl22:
Sure :)

sigurdm:
NP. 

We could also consider passing `-z` instead to get the raw strings out... not sure if it is better or worse than this.

Gustl22:
`git -c core.quotePath=false ls-files --cached --exclude-standard --recurse-submodules ` gives:
```bash
CHANGELOG.md
LICENSE
README.md
lib/test_pkg.dart
coins_non_ascii_и.wav
pubspec.yaml
```

`git ls-files -z --cached --exclude-standard --recurse-submodules` gives:
```bash
CHANGELOG.mdLICENSEREADME.mdlib/test_pkg.dartcoins_non_ascii_и.wavpubspec.yaml
```

But for some reason in the test process with `-c core.quotePath=false` option, it is returned as:
```
CHANGELOG.md
LICENSE
README.md
lib/test_pkg.dart
non_ascii_Ð¸.txt
pubspec.yaml
```

Gustl22:
> then forward it to https://api.dart.dev/stable/3.3.0/dart-io/Process/run.html

Thank you @sigurdm, that was actually it. Normally it would work without a specifying the encoding, but on Windows `SystemEncoding` is defaulted by "the currently active code page".
Unfortunately I had to rework the `_doProcess` function, because `Process.start` doesn't match with the parameters of `Process.run`, obviously because of the `stdoutEncoding` param.

sigurdm:
Thanks for the contribution!

> Currently we cannot publish with the new dart version,

FWIW there is a `--skip-validation` option to `dart pub publish` to work around bugs like this.

sigurdm (approved):

PR #4149, opened 2024-02-26
Prefer IPV4 on localhost, serve testserver on IPV4

sigurdm:
This should resolve some flakynes observed in tests.

Because binding to port 0 `localhost` will bind to a random free port on one of `127.0.0.1` (ipv4) and `::1` (ipv6) it risks colliding with a different service with the same port, but the other ipvx.

When pub then connects to `localhost` it will race ipv4 against ipv6, causing flakiness.

Note that pub always will canonicalize loopbacks 127.0.0.1 and ::1 to `localhost` for cache directory names and for lockfile purposes.

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -10,6 +10,7 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:math' as math;
 
+import 'package:collection/collection.dart';
 import 'package:http/http.dart' as http;
 import 'package:pool/pool.dart';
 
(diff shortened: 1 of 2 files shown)


lib/src/http.dart
```diff
@@ -444,3 +460,52 @@ class _ClientClosedException implements Exception {
   @override
   String toString() => 'Request was made after http client was closed';
 }
+
+class OverrideUrlRequest implements http.BaseRequest {
```
jonasfj, line 464: Does this need to be a public class?

sigurdm: No...

Will fix in a follow-up

Issue #3980, opened 2023-08-03
Introduce `$FLUTTER_ROOT/sdk-packages.yaml`


jonasfj:
When looking for `sdk: flutter` sdk-packages today, we do:
https://github.com/dart-lang/pub/blob/98565c3e5defcd80c59910795ba3548552102f2c/lib/src/sdk/flutter.dart#L81-L95

Effectively, we search for `<packageName>` in (ordered by preferences):
 * `$FLUTTER_ROOT/packages/<packageName>`
 * `$FLUTTER_ROOT/bin/cache/pkg/<packageName>`

Instead of having multiple search paths hardcoded in `pub`, I propose that we introduce an `sdk-packages.yaml` file.

```yaml
# Declaration of SDK-packages supplied by the Flutter SDK.
# An SDK-package is a package that the user can depend on
# using a dependency constraint like:
#
#         dependencies:
#           flutter_test:
#             sdk: flutter
#
# Paths in the list below are relative to this file.
sdk: flutter
packages:
- name: flutter
  path: packages/flutter
- name: flutter_test
  path: packages/flutter_test
- name: sky_engine
  path: bin/cache/pkg/sky_engine
```

## Migration

1. `dart pub` will read `sdk-packages.yaml` when present, and fallback to existing logic
   when no `sdk-packages.yaml` file is present.
2. `pub` is rolled into the Dart SDK.
3. Flutter SDK starts supplying a `sdk-packages.yaml` file.
4. `dart pub` will refuse detect a valid Flutter SDK, unless `sdk-packages.yaml` is present.

## Open questions

 * We could bike-shed the name further, I'm open to ideas, `sdk-packages.yaml` just seemed natural.
 * Should we use a `.json` file instead, it would arguably be a tiny bit faster, but we wouldn't be able to have comments.

-----------------

@sigurdm, Do you agree that this would be cleaner from a `pub` perspective? And pretty easy to do?

@jonahwilliams, any thoughts? is there anyone else you think we should run this past?

@christopherfujino, any thoughts? Is it intentional that `flutter_tools` is something that Flutter Apps can take a dependency on? If it doesn't have an API that you intend to expose, this could possibly offer a mechanism that would allow you to make it private.

dcharkes:
> * We could bike-shed the name further, I'm open to ideas, `sdk-packages.yaml` just seemed natural.

🚲 🏠  Aren't underscores more common than dashes in file names in the Dart/Flutter ecosystem?

[...]

PR #4151, opened 2024-02-26
add support for an sdk_packages.yaml file in the Dart SDK

jakemac53:
Part of https://github.com/dart-lang/pub/issues/3980

~Sending this out for preliminary review only right now. I haven't tried actually building the SDK using this support, and I think it might require the pre-built SDK to get updated with this.~

~I do want to push on this soon, if we decide it is something we want to run with.~

**Update**: Ready for review


test/sdk_test.dart
```diff
@@ -5,222 +5,291 @@
...
-        d.dir('bin/cache/pkg', [
-          d.dir(
-            'baz',
-            [d.libDir('baz', 'foo 0.0.1'), d.libPubspec('baz', '0.0.1')],
-          ),
-        ]),
-        d.flutterVersion('1.2.3'),
-      ]).create();
-    });
+    group('flutter', () {
```
jakemac53, line 16: This diff is kind of insane.... all I actually did was wrap the existing stuff in a group, and make a new group which exercises the Dart sdk support for sdk_packages.yaml.

lib/src/sdk/sdk_package_config.dart
```diff
@@ -0,0 +1,52 @@
...
+        ],
+      };
+}
+
+/// The structure for each `packages` entry in an `sdk_packages.yaml` file.
+class SdkPackage {
+  /// The name of the package.
+  final String name;
+
+  /// The path to the root of this package from the root of the SDK.
```
sigurdm, line 38:
```suggestion
  /// The path to the root of this package relative to the root of the SDK.
```

lib/src/sdk/dart.dart
```diff
@@ -31,6 +39,21 @@ class DartSdk extends Sdk {
     return aboveExecutable;
   }();
 
+  /// The loaded `sdk_packages.yaml` file if present.
+  static final SdkPackageConfig? _sdkPackages = () {
+    var path = p.join(_rootDirectory, 'sdk_packages.yaml');
```
sigurdm, line 44: Should we not make this a `.json` file? Are we getting any real advantage from yaml here.

lib/src/sdk/sdk_package_config.dart
```diff
@@ -0,0 +1,52 @@
...
+  final String sdk;
+
+  /// All the packages vendored by this SDK.
+  final List<SdkPackage> packages;
+
+  SdkPackageConfig(this.sdk, this.packages);
+
+  // Note: yaml has `Object?` keys.
+  SdkPackageConfig.fromMap(Map<Object?, Object?> map)
+      : sdk = map['sdk'] as String,
```
sigurdm, line 19: Do we need error checking here, or we just expect the sdk to always have correctly structured data?

sigurdm (approved):

lib/src/sdk/dart.dart
jakemac53: I don't particularly care, I do expect it to be hand written and yaml is more readable/editable. But happy to go with whatever you all prefer, cc @jonasfj .

lib/src/sdk/dart.dart
```diff
@@ -50,5 +73,17 @@ class DartSdk extends Sdk {
   String get rootDirectory => _rootDirectory;
 
   @override
-  String? packagePath(String name) => null;
+  String? packagePath(String name) {
+    if (!isAvailable) return null;
```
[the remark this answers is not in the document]
jakemac53: We don't need it _right now_ but it still felt right to leave it in, but I am also happy to remove it if you prefer. It seems highly unlikely this would ever be false.

lib/src/sdk/sdk_package_config.dart
jakemac53: I think it is pretty safe to assume correctly structured data, but it could help SDK engineers if we had better error messages. I think it should still throw either way.

jakemac53:
> Not super happy about having more sdk-packages, but if that is the way we go I think this is fine.

I will wait to land this until we get consensus on that, I know it is a contentious topic. I have personally been anti-SDK package for a long time.... until I needed one 🤣 .

I think if we don't allow pub dependencies (not sure where to add that check in), that would remove most of the concern.

lib/src/sdk/dart.dart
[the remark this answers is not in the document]
sigurdm: Maybe we should factor it out, such that Sdk.packagePath can assume that available is true (for all sdks)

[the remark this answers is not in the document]
sigurdm: Sure.

https://github.com/dart-lang/pub/blob/9c041f97c07ea17f1460f2d7df5aa86b63a72917/lib/src/source/sdk.dart#L118 makes the call to `.packagePath()`. It could preface it with a call to `isAvailable()`, instead of baking that logic into each `.packagePath()` implementation.

If it is too much, we can also do the refactor later. Feel free to land this if you are happy.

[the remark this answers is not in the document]
jakemac53: I think a follow up would be good if you want to make that change, I am not sure pushing that to the consumer is best though?

jakemac53:
PTAL @sigurdm @jonasfj, added support for SDKs blocking non-SDK regular deps and a test for that.

sigurdm:
> If we can also make a decision on some of the open questions that would be great (yaml or json?). I am happy to go with whichever (but have a slight preference for yaml mostly because it supports comments).

Yaml is fine with me. Thinking about this we do parse pubspec.yamls whenever we resolve, so this will not dominate anything.

lib/src/sdk/dart.dart
```diff
@@ -50,5 +75,17 @@ class DartSdk extends Sdk {
...
+    var sdkPackages = _sdkPackages;
+    if (sdkPackages == null) return null;
+
+    var package =
+        sdkPackages.packages.firstWhereOrNull((pkg) => pkg.name == name);
```
lrhn, line 84: How about building a map inside the `SdkPackages.fromMap`, so `SdkPackages` has a `SdkPackage? operator[](String packageName)` lookup.
(It's probably a short list, but linear lookup always makes be worry.)

lib/src/sdk/dart.dart
```diff
@@ -50,5 +75,17 @@ class DartSdk extends Sdk {
...
+    if (package == null) return null;
+    var packagePath = p.joinAll([_rootDirectory, ...p.url.split(package.path)]);
```
lrhn, line 86: Isn't this just a (very) complicated way to say `package.path.split('/')`?
The only difference betewen that and `p.url.split` is that if `package.path` starts with `/`, the first element returned by `split` is `/`.

But `package.path` should always be relative. If it isn't, that `/` is not going to help.

(Does answer my question from above about which path separator it uses.)

lib/src/sdk/sdk_package_config.dart
```diff
@@ -0,0 +1,52 @@
...
+  Map<String, Object?> toMap() => {
+        'sdk': sdk,
+        'packages': [
+          for (var package in packages) package.toMap(),
+        ],
+      };
+}
+
+/// The structure for each `packages` entry in an `sdk_packages.yaml` file.
+class SdkPackage {
```
lrhn, line 34: (If there is no structure or functionality to this class, other than mimicing the underlying map, consider making it an extension type on the map.)

lib/src/sdk/sdk_package_config.dart
jakemac53: Added a todo (and a version)

lib/src/sdk/sdk_package_config.dart
jakemac53: I did do a conversion to a Map now, and it is nice imo to be able to construct these instances not from Yaml (the tests do this).

lib/src/sdk/sdk_package_config.dart
```diff
@@ -0,0 +1,52 @@
...
+  /// The name of the package.
+  final String name;
+
+  /// The path to the root of this package relative to the root of the SDK.
```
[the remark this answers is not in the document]
jakemac53: Updated to "installed SDK"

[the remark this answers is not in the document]
jakemac53: > For configuration files, I'll usually make relative paths be relative to the configuration file itself. That doesn't require a second concept to interpret the file, the file is self-contained (but it does require dragging around the location of the file to where the contents are interpreted).

I would usually do that, but related to the next note, this file won't _actually_ be developed in a place where the relative paths would make sense. It only makes sense for the final built SDK, and so I think it is easier to speak in terms of that built SDK (that way this file can move for instance).

> I assume this is relative to the installed SDK root, so this file is generated as part of building the SDK release, where we copy the files from wherever they live in the SDK sources to a place in the deployment format, right?

Pretty much, but the file will likely be checked in as source and just copied to the right location in the final SDK.

lib/src/sdk/sdk_package_config.dart
```diff
@@ -0,0 +1,52 @@
...
+  final String path;
+
+  SdkPackage(this.name, this.path);
+
+  // Note: yaml has `Object?` keys.
+  SdkPackage.fromMap(Map<Object?, Object?> map)
+      : name = map['name'] as String,
+        path = map['path'] as String;
```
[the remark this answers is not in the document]
jakemac53: I think we want to hand write it and check it in. Technically we could generate it, but I think it is good to have a single source of truth to go see the configuration, checked into the repo.

lib/src/source/sdk.dart
```diff
@@ -89,11 +89,32 @@ class SdkSource extends Source {
...
+      _verifiedPackagePath(ref),
+      cache.sources,
+      expectedName: ref.name,
+    );
+
+    /// Validate that there are no non-sdk dependencies if the SDK does not
+    /// allow them.
+    if (ref.description case SdkDescription description) {
+      if (sdks[description.sdk]
+          case Sdk(allowsNonSdkDepsInSdkPackages: false)) {
```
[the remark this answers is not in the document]
jakemac53: ack, I do like the pattern version more though.

lib/src/source/sdk.dart
```diff
@@ -89,11 +89,32 @@ class SdkSource extends Source {
...
+        for (var dep in pubspec.dependencies.entries) {
+          if (dep.value.source is! SdkSource) {
+            throw ArgumentError(
```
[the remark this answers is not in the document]
jakemac53: I don't think we have a good framework for this, available at this point, but @sigurdm could clarify?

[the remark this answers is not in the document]
sigurdm: We don't have a framework that handles multiple errors.
When an (uncaught) error is thrown, we show the error with a message asking to file an issue and save the verbose log to a file.

lib/src/sdk/sdk_package_config.dart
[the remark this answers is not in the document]
lrhn: In that case, this file can actually be the source of truth for *where* the package should be copied.
Which also supports using the SDK root as base, because that's what we're copying into.

lib/src/sdk/sdk_package_config.dart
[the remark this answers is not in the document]
lrhn: Which supports being `/` separated (as I noticed it was where it was split).

The format should be documented somewhere. Can be just a comment in this file, but there should be a source of truth about the structure and (most importantly) the format of any non-prose string.

lib/src/source/sdk.dart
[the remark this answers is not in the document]
lrhn: This is validating something that is expected to succeed, created by SDK authors, not end users, so reporting any error is probably fine. It's enough to stop progress. (This error must never happen in a released SDK, unless someone has been editing files inside the deplouyed SDK. As long as we can ensure that, it shouldn't matter what happens here. Except "good style", that always applies :wink:.)

jonasfj (approved):

lib/src/sdk/sdk_package_config.dart
```diff
@@ -0,0 +1,52 @@
...
+        'sdk': sdk,
+        'packages': [
+          for (var package in packages) package.toMap(),
+        ],
+      };
+}
+
+/// The structure for each `packages` entry in an `sdk_packages.yaml` file.
+class SdkPackage {
+  /// The name of the package.
```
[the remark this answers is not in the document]
jakemac53: I exposed the logic already used by pub to this library and used it.

jakemac53:
I did go ahead and add better error handling, as well as consolidating a bit of the logic for validating yaml fields through some new shared extensions.

Later on we might want to migrate some more of the pubspec logic to these extensions too? But that can be a task for another day.

lib/src/sdk/dart.dart
```diff
@@ -20,8 +22,15 @@ class DartSdk extends Sdk {
...
   @override
   String? get installMessage => null;
+  @override
+  bool get allowsNonSdkDepsInSdkPackages => false;
 
   static final String _rootDirectory = () {
+    // If DART_ROOT is specified, then this always points to the Dart SDK
+    if (Platform.environment.containsKey('DART_ROOT')) {
+      return Platform.environment['DART_ROOT']!;
+    }
```
lrhn, line 32: Double lookup.
Consider
```dart
  if (Platform.environment['DART_ROOT'] case var root?) {
    return root;
  }
```

lib/src/sdk/dart.dart
lrhn: If pub is the only consumer, I won't care.
The `package_config.json` file is JSON because every source processing til needs to read it.

lib/src/sdk/dart.dart
```diff
@@ -31,6 +40,24 @@ class DartSdk extends Sdk {
     return aboveExecutable;
   }();
 
+  /// The loaded `sdk_packages.yaml` file if present.
+  static final SdkPackageConfig? _sdkPackages = () {
+    var path = p.join(_rootDirectory, 'sdk_packages.yaml');
```
lrhn, line 45: Could we put this in a sub directory?
I can see that the `version` file is at the top level, but nothing else seems to be there, and keeping a clean your it's usually a good thing.
If no current subdir works, we can add `cfg/`.

lib/src/utils.dart
```diff
@@ -777,3 +779,65 @@ extension RetrieveFlags on ArgResults {
...
+}
+
+extension ExpectEntries on YamlList {
+  /// Expects each entry in [this] to have a value of type [T],
+  /// and returns a `List<T>`.
+  ///
+  /// Throws a [SourceSpanApplicationException] for the first entry that does
+  /// not have a value of type [T].
+  List<T> expectEntries<T extends Object?>() {
+    final entries = <T>[];
```
lrhn, line 829: We usually use "elements" for list elements and reserve "entries" for map entries.

lib/src/utils.dart
```diff
@@ -777,3 +779,65 @@ extension RetrieveFlags on ArgResults {
...
+      if (value is T) {
+        entries.add(value);
+      } else {
+        throw SourceSpanApplicationException(
+          'Entries must be a $T.',
+          node.span,
+        );
+      }
+    }
+    return entries;
```
lrhn, line 841: Can be a literal.
```dart
[for (var n in nodes)
  if (n.value case T value) value
  else throw …
]

```

lib/src/sdk/dart.dart
sortie: How about pkg/?

It would fit well with the bin and lib directories inside the Dart SDK already, and match the pkg/ directory inside the Dart SDK's source code.

pkg/packages.yaml e.g.

sortie: And then put the packages in there, like pkg/macros/

lib/src/sdk/sdk_package_config.dart
```diff
@@ -0,0 +1,101 @@
...
+/// feature and format.
+///
+/// Version 1 of the format is as follows:
+///
+/// ```yaml
+/// # Required, the version of the format used in this file. Not all versions
+/// # will be supported forever, but some number of previous ones will be.
+/// version: 1
+///
+/// # The SDK this configuration file is targetting. Used for validation, to
```
sortie, line 21: targeting

lib/src/sdk/sdk_package_config.dart
```diff
@@ -0,0 +1,101 @@
...
+/// sdk: dart
+///
+/// # A list of package descriptors, for each package vendored by this SDK.
+/// packages:
+/// - name: my_sdk_package
+///   # A path relative to the root of the installed SDK, in URL form (with `/`
+///   # path separators).
+///   path: path/to/my_sdk_package
+/// ```
+class SdkPackageConfig {
```
sortie, line 32: Hmm, do we really need this file at all? If we did dart-sdk/pkg/macros/ and such, those contain a pubspec.yaml, and this file is sort of just a directory listing. Maybe pub could just search dart-sdk/pkg/ and discover if it contains the named packages it's listing for? Then we don't need a whole new file format with a version integer and all of that complexity,

sigurdm: @jonasfj had some reasons that he wanted another layer of indirection...

lib/src/sdk/dart.dart
jakemac53: That sounds fine to me, @jonasfj wdyt? I think this file was originally your idea :)

lib/src/sdk/sdk_package_config.dart
jakemac53: The file gives us future possibilities like vendoring multiple versions of a package as well.

jonasfj: Technically, I guess we don't need this file. We could do what we do flutter now.
Namely, we could say that everything in `$DART_ROOT/pkg/` is an SDK-package.

On the other hand, I think we should want a `$FLUTTER_ROOT/sdk_packages.yaml` file that explicitly configures what SDK-packages flutter has. That way we might stop accidentally exposing non-public APIs from Flutter (example `flutter_tools` is an SDK-package that shouldn't exist).

So given that we'd like to have `$FLUTTER_ROOT/sdk_packages.yaml`, it might be nice to be consistent and have a `$DART_ROOT/sdk_packages.yaml` as well.

I don't think we should block on this either way. I don't care too much either way.
I'd suggest we ship this.

jakemac53: +1 I personally like the file as it gives a lot of future flexibility, and avoids hard coding any knowledge about the Flutter/Dart SDKs layouts into pub (other than the location of this one file).

lib/src/sdk/dart.dart
jonasfj: I'd say it should be in the root.

But we haven't landed this in Flutter yet :)

But that's IMO where it'd make sense to put it in Flutter, since it needs to refer to both `packages/` and `bin/cache/pkg/`.

Arguably, I proposed this file because directory layout in Flutter is confusing and there is no build system that copies thing into the correct location when the Flutter SDK is built. Because as far as I understand it, you still don't build the SDK into artifact, you just let it modify itself.

We could choose not to have it for the Dart SDK, because it does have a build system. And the Dart SDK does have a concept of building into a binary artifact. And adding something to the resulting binary artifact is an explicit decision.

So it could be argued we don't need it, or we could write build rules that read it, and copy folders listed in there into the pkg/ folder in the output.

-----

For consistency with flutter, I'd suggest putting it in the root.

sigurdm: > The main reason we want it for Flutter is because they don't have a build step so all their internal packages are SDK packages by accident more than intentional decision.

Just saying: we could also fix this by moving the internal packages out of the `packages/` folder... Sometimes low-tech is good.

I'm not strongly against an indirection file. I just don't see a big need.

jonasfj: > Just saying: we could also fix this by moving the internal packages out of the packages/ folder... Sometimes low-tech is good.

True, we could try that.

> I'm not strongly against an indirection file. I just don't see a big need.

That's fair, and also what @sortie is arguing.
Especially, for the Dart SDK, we might not need it since there are build-rules that makes copying sdk-packages into the `out/` folder rather explicit (not something done by accident).

jakemac53: While I agree in principle that it is not necessary (clearly, Flutter works today without it), I do think it is a much better design to have the SDK be in control of where its own packages live, instead of having pub hard code this knowledge. Why have tight coupling where we can easily avoid it?

I see also a `dartdoc_options.yaml` file in the root of the dart SDK, I don't think it hurts to have this file there as well.

We have a lot of this similar coupling in other places, even with packages published on pub, such as the locations of platform dill files and executables. It might be nice to move those into a config file as well at some point which would avoid a lot of this.

jakemac53:
@jonasfj @sigurdm I would like to land this tomorrow, so if you feel strongly about the location of the configuration file or anything else please speak up :)

jakemac53:
I am going to pull this into the Dart SDK so I can start trying to create the new structure, but we still have time to make changes if people feel strongly.

Issue #4127, opened 2024-01-30
Implement workspaces


sigurdm:
https://goto.google.com/pub-workspace

Sub-tasks:

* [ ] accept `workspace` and `resolution` top-level fields in pubspec.yaml under experiment flag
* [ ] resolve all dependencies in workspace together.
* [ ] find the workspace taking CWD and `resolution` annotations into account
* [ ] `pub upgrade --major-versions` support for upgrading versions in whole workspace
* [ ] `pub add`/`pub remove` support for adding and removing dependencies to a member of a workspace 
* [ ] `pub deps` support for showing dependencies for a whole workspace
* [ ] `pub outdated` support for showing outdated dependencies for whole workspace
* [ ] dependency_services/dependabot support for workspaces
* [ ] Remove experiment flag

bramp:
Any chance you can make whatever link is behind https://goto.google.com/pub-workspace public, so non-Googlers can see it :)? thanks

sigurdm:
> Any chance you can make whatever link is behind https://goto.google.com/pub-workspace public, so non-Googlers can see it :)? thanks

Sorry, that link was wrong. Not sure how I managed to type that in. The correct link is public. Updated.

bramp:
Thanks, it works for me now.

a2425rdl:
https://github.com/dart-lang/pub/issues/4127

PR #4154, opened 2024-02-27
Support for resolving all packages in workspace together

sigurdm:
Still no support for "opting out".

Part of https://github.com/dart-lang/pub/issues/4127

### DIFF
--- a/lib/src/command/dependency_services.dart
+++ b/lib/src/command/dependency_services.dart
@@ -69,13 +69,21 @@ class DependencyServicesReportCommand extends PubCommand {
     final breakingPubspec = stripVersionBounds(compatiblePubspec);
 
     final compatiblePackagesResult = await _tryResolve(
-      Package(compatiblePubspec, entrypoint.rootDir),
+      Package(
+        compatiblePubspec,
+        entrypoint.rootDir,
+        entrypoint.root.workspaceChildren,
+      ),
       cache,
       additionalConstraints: additionalConstraints,
     );
 
     final breakingPackagesResult = await _tryResolve(
-      Package(breakingPubspec, entrypoint.rootDir),
+      Package(
+        breakingPubspec,
+        entrypoint.rootDir,
+        entrypoint.root.workspaceChildren,
+      ),
       cache,
       additionalConstraints: additionalConstraints,
     );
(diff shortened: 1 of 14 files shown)


lib/src/command/dependency_services.dart
```diff
@@ -734,7 +760,7 @@ Future<List<Object>> _computeUpgradeSet(
         ? SolveType.downgrade
         : SolveType.get,
     cache,
-    Package(pubspec, entrypoint.rootDir),
+    Package(pubspec, entrypoint.rootDir, entrypoint.root.workspaceChildren),
```
szakarias, line 763: nit: trailing comma

lib/src/solver/package_lister.dart
```diff
@@ -84,13 +85,23 @@ class PackageLister {
 
   /// All versions of the package, sorted by [Version.compareTo].
   Future<List<PackageId>> get _versions => _versionsMemo.runOnce(() async {
```
szakarias, line 87: Maybe add doc for the new case?

lib/src/solver/version_solver.dart
```diff
@@ -62,6 +62,11 @@ class VersionSolver {
   /// The entrypoint package, whose dependencies seed the version solve process.
   final Package _root;
 
+  /// Mapping all rootpackages in the workspace from their name.
```
szakarias, line 65: nit: rootpackages -> root packages

lib/src/solver/package_lister.dart
```diff
@@ -259,9 +270,15 @@ class PackageLister {
         if (id.isRoot)
           ...pubspec.devDependencies.values
               .where((range) => !_overriddenPackages.contains(range.name)),
+        if (id.isRoot)
```
szakarias, line 273: Maybe include this in the condition above? (I would try it out to see if it gives a better overview when reading the code. If you finde that it makes it clumsy just keep it as it is)

szakarias (approved):

lib/src/command/dependency_services.dart
sigurdm: I don't think we need it here.

We have a lint requiring a trailing comma when needed.

lib/src/solver/package_lister.dart

jonasfj (approved):

PR #4156, opened 2024-02-28
Do not report ignored advisories in 'pub outdated'

szakarias:


test/outdated/outdated_test.dart
```diff
@@ -380,6 +380,59 @@ Future<void> main() async {
...
+        'name': 'app',
+        'dependencies': {
+          'foo': '^1.0.0',
+        },
+        'ignored_advisories': ['ABCD-1234-5678-9101'],
+      }),
+    ]).create();
+    await pubGet();
+
+    builder.affectVersionsByAdvisory(
```
sigurdm, line 400: Could you make the test such that it has one advisory that's ignored and one that's not. (And one that's ignored via an alias?

sigurdm (approved):
LGTM, but maybe the test could cover a bit more.

test/outdated/outdated_test.dart
szakarias: I would like to keep the vanilla tests, so that we can cover the boolean in the json output. I'll add extra tests for the combined cases.

sigurdm: Ah - that makes sense - but you could let the ignored one be the one that is affecting the current version?

szakarias: Yes, I think an extra test is also fine.

sigurdm: It's ok if you prefer an extra test - just know that these tests are relatively expensive (because they run `pub outdated` in quite a number of different configs.

szakarias: In general I prefer two simple tests over one covering too much. I can combine the two vanilla tests, though.

sigurdm: SG

PR #4158, opened 2024-02-29
Fix output of 'pub outdated' when there are several advisories

szakarias:


lib/src/command/outdated.dart
```diff
@@ -701,8 +701,8 @@ Future<void> _outputHuman(
           'See https://dart.dev//go/pub-security-advisories',
         );
         log.message('\n');
-        var displayedVersions = <String>{};
         for (final advisory in package.advisories) {
+          var displayedVersions = <String>{};
```
sigurdm, line 705:
```suggestion
          final displayedVersions = <String>{};
```

sigurdm (approved):

lib/src/command/outdated.dart
szakarias: done

PR #4167, opened 2024-03-08
Refactor test setup for security advisories

szakarias:
This enables testing security advisories with affected package fields that have other ecosystems specified.


sigurdm (approved):
Formatting - but then 👍

Issue #4161, opened 2024-03-06
`dart run <package>` can fail if a `dart pub get` operation is running concurrently


andrewkolos:
As documented in [Running a program that's a package](https://dart.dev/tools/dart-run#running-a-program-thats-in-a-package), `dart run` can be used to invoke a dart program that exists in a package dependency. However, this can fail with an filesystem-related error if `dart pub get` is running concurrently (this can be happen frequently, as the Dart VSCode extension watches for changes on pubspec.yaml and will run `dart pub get` automatically).

## Minimal repro
To demonstrate what I mean, here is a bash script that can repro this issue:

```bash
dart create myproject
cd myproject
dart pub add vector_graphics_compiler
dart pub get & dart run vector_graphics_compiler
```

If this doesn't repro, try running `dart pub get & dart run vector_graphics_compiler` once more. There might be a race-condition here. Regardless, here is the kind of error this should print out:

```
Unhandled exception:
PathNotFoundException: Cannot open file, path = '/Users/andrewkolos/Desktop/deleteme/pubgetrun/.dart_tool/pub/bin/vector_graphics_compiler/tmpXaiYq6/vector_graphics_compiler.dart-3.4.0-190.0.dev.snapshot.incremental.temp' (OS Error: No such file or directory, errno = 2)
#0      _checkForErrorResponse (dart:io/common.dart:55)
dart-lang/sdk#1      _File.open.<anonymous closure> (dart:io/file_impl.dart:381)
<asynchronous suspension>
dart-lang/sdk#2      _FileStreamConsumer.addStream.<anonymous closure> (dart:io/file_impl.dart:205)
<asynchronous suspension>

Compilation did not produce any result. Expected file at `/Users/andrewkolos/Desktop/deleteme/pubgetrun/.dart_tool/pub/bin/vector_graphics_compiler/tmpXaiYq6/vector_graphics_compiler.dart-3.4.0-190.0.dev.snapshot.incremental.temp`
Built vector_graphics_compiler:vector_graphics_compiler.
[...]

PR #4170, opened 2024-03-12
More robust snapshot deletion

sigurdm:
Change snapshot deletion to only delete snapshots made by different sdk's.

This is such that the `.dart_tool` folder will not grow without bounds.

This requires that we rebuild even snapshots from mutable packages after pub get. So this pr also adds a timestamp check comparing the snapshot modified time to the `.dart_tool/package_config.json` modified time. And will recompile if the snapshot is older, even for immutable packages.

This is not as bad as it sounds, because we are doing incremental recompilation.

Fixes https://github.com/dart-lang/pub/issues/4161

`pub get` and `dart run` will still race if they are run concurrently from **different** sdks. But that seems acceptable.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -455,7 +455,7 @@ To update `$lockFilePath` run `$topLevelProgram pub get`$suffix without
         if (precompile) {
           await precompileExecutables();
         } else {
-          await _deleteExecutableSnapshots(changed: result.changedPackages);
+          await _deleteExecutableSnapshots();
         }
       } catch (error, stackTrace) {
         // Just log exceptions here. Since the method is just about acquiring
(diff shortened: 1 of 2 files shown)


lib/src/entrypoint.dart
```diff
@@ -455,7 +455,7 @@ To update `$lockFilePath` run `$topLevelProgram pub get`$suffix without
         if (precompile) {
           await precompileExecutables();
         } else {
-          await _deleteExecutableSnapshots(changed: result.changedPackages);
+          await _deleteExecutableSnapshots();
```
jonasfj, line 458: I guess we are throwing away information here. But at the same time, I don't know how we would communicate to a different compilation process what dependency entries changed.

jonasfj (approved):

jonasfj:
Are there potential problems where new/old SDK versions interact poorly here?

I think not since snapshots are suffixed with SDK-version.

sigurdm:
> Are there potential problems where new/old SDK versions interact poorly here?

If you switch between two different sdk versions you'll have your snapshots deleted all the time.

PR #4172, opened 2024-03-12
Use the 'pub_display_url' field in advisories in output of pub get/up…

szakarias:
…grade and pub outdated


lib/src/source/hosted.dart
```diff
@@ -1813,12 +1831,17 @@ class Advisory {
...
+    required this.affectedVersions,
+    required this.summary,
+    this.aliases = const [],
+    this.pubDisplayUrl,
+  });
 
-// TODO(https://github.com/dart-lang/pub-dev/issues/7471) Use `pub_display_url`
-// once this issue is resolved.
-String advisoriesDisplayUrl(String id) => 'https://github.com/advisories/$id';
+  String get displayUrl => pubDisplayUrl ?? id;
```
sigurdm, line 1843: Can we call this something different - since it is not a url, and to avoid confusion with `pubDisplayUrl`?

Maybe `displayHandle`

szakarias: Good idea.

Issue #4168, opened 2024-03-11
Create a spinner while downloading files


jonasfj:
Try:
```
rm -rf /tmp/pub-cache; PUB_CACHE=/tmp/pub-cache dart pub get
```

It doesn't show anything while downloading files, if there is many this could make me thing the process is just hanging.

PR #4175, opened 2024-03-12
Show progress while downloading dependencies

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/4168

This will show the "Downloading dependencies" unconditionally - even if there is nothing to download... not sure if that is misleading, but it also seems confusing if the line is sometimes there and sometimes not... 

image


lib/src/solver/result.dart
```diff
@@ -58,21 +59,23 @@ class SolveResult {
...
-            ))
-                .packageId;
-          });
-        }
-        return id;
-      }),
-    );
-
+    final resolvedPackageIds =
+        await progress('Downloading dependencies', () async {
```
jonasfj, line 63: nit:
```suggestion
        await progress('Downloading packages', () async {
```

jonasfj (approved):

lib/src/solver/result.dart
sigurdm: Done

PR #4176, opened 2024-03-13
Support advisories with several affected packages with 'pub' a…

szakarias:
…s ecosystem

- Refactored `_extractAdvisoryDetailsForPackage` to support this corner case which we don't expect occures in practice but it is technically allowed.
- Also fixed a small bug in the test setup


test/get/hosted/advisory_test.dart
```diff
@@ -190,6 +190,36 @@ Future<void> main() async {
     await ctx.run(['get']);
   });
 
+  testWithGolden('more than one affected package with pub as ecosystem',
```
sigurdm, line 193:
```suggestion
  testWithGolden('more than one affected package-version with pub as ecosystem',
```
Or would it be better to actually affect two different packages?

test/testdata/goldens/get/hosted/advisory_test/more than one affected package with pub as ecosystem.txt
```diff
@@ -0,0 +1,11 @@
+# GENERATED BY: test/get/hosted/advisory_test.dart
+
+## Section 0
+$ pub get
+Resolving dependencies...
```
sigurdm, line 5: Probably has to be regenerated after https://github.com/dart-lang/pub/pull/4175

szakarias: Yup, done.

sigurdm (approved):

PR #4179, opened 2024-03-14
Refactor `Entrypoint.rootDir` preparing for workspaces

sigurdm:
An entrypoint now no longer knows a single `rootDir`. Instead it knows:

`workingDir`: the directory that the entrypoint is created around. Typically this is the current working dir.
`workspaceRoot` The root package of the workspace
`workPackage` The package closest to the workingDir.

This PR still just has `workPackage = workspaceRoot = Package(workingDir)` but should be easy to review.
A follow-up will do the actual search in the file system.


sigurdm:
This is preparing for https://github.com/dart-lang/pub/issues/4127

jonasfj (approved):

PR #4186, opened 2024-03-18
Resolve workspace root and workPackage when invoking pub from any sub-directory

sigurdm:
Part of https://github.com/dart-lang/pub/issues/4127


lib/src/entrypoint.dart
```diff
@@ -62,21 +63,78 @@ class Entrypoint {
   // true.
   final String workingDir;
 
-  Package? _workspaceRoot;
+  static ({Package root, Package work}) _loadWorkspace(
```
jonasfj, line 66: Can this have a few comments?

lib/src/entrypoint.dart
```diff
@@ -62,21 +63,78 @@ class Entrypoint {
...
+      }
+      pubspecsMet[p.canonicalize(dir)] = pubspec;
+      final Package root;
+      if (pubspec.resolution == Resolution.none) {
+        root = Package.load(
+          null,
+          dir,
+          cache.sources,
+          alreadyLoadedPubspecs: pubspecsMet,
+          withPubspecOverrides: true,
```
jonasfj, line 97: Are you reloading it because it's the root, and it's allowed to a have `pubspec_overrides.yaml` ?

If so, can we maybe highlight that in a comment.

lib/src/entrypoint.dart
```diff
@@ -62,21 +63,78 @@ class Entrypoint {
...
+        );
+        for (final package in root.transitiveWorkspace) {
+          pubspecsMet.entries.first;
```
jonasfj, line 100: ?

lib/src/entrypoint.dart
```diff
@@ -62,21 +63,78 @@ class Entrypoint {
...
+          if (identical(pubspecsMet.entries.first.value, package.pubspec)) {
```
jonasfj, line 101: I don't understand this?

lib/src/entrypoint.dart
```diff
@@ -62,21 +63,78 @@ class Entrypoint {
...
+
+See $workspacesDocUrl for more information.''',
+        p.join(workingDir, 'pubspec.yaml'),
+      );
+    }
+  }
+
+  ({Package root, Package work})? _packages;
+  ({Package root, Package work}) get _getPackages =>
+      _packages ??= _loadWorkspace(workingDir, cache);
```
jonasfj, line 128: Documenting these would be nice.

Specifically, documenting that: `_getPackages` returns, and that `_packages` should never be accessed, it's only for caching `_getPackages`, right?

Just a line for each explaining the relationship, and what `_getPackages` is supposed to be.

jonasfj:
I don't really have any reservations here. It's probably, but perhaps a few documentation comments would help us review it :D

lib/src/io.dart
```diff
@@ -17,7 +17,7 @@ import 'package:cli_util/cli_util.dart'
 import 'package:http/http.dart' show ByteStream;
 import 'package:http_multi_server/http_multi_server.dart';
 import 'package:meta/meta.dart';
-import 'package:path/path.dart' as path;
+import 'package:path/path.dart' as p;
```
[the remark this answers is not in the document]
sigurdm: Good idea: https://github.com/dart-lang/pub/pull/4187

lib/src/io.dart
```diff
@@ -1231,3 +1229,31 @@ String escapeShellArgument(String x) =>
     RegExp(r'^[a-zA-Z0-9-_=@.^]+$').stringMatch(x) == null
         ? "'${x.replaceAll(r'\', r'\\').replaceAll("'", r"'\''")}'"
         : x;
+
+/// Returns all parent directories of [path], starting from [path] to the root.
+///
+/// If [path] is relative the directories will also be.
+///
+/// If [from] is passed, directories are made relative to that.
```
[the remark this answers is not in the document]
sigurdm: I took the `from` name from `p.relative(d, {from})`.

lib/src/io.dart
```diff
@@ -1231,3 +1229,31 @@ String escapeShellArgument(String x) =>
     RegExp(r'^[a-zA-Z0-9-_=@.^]+$').stringMatch(x) == null
         ? "'${x.replaceAll(r'\', r'\\').replaceAll("'", r"'\''")}'"
         : x;
+
+/// Returns all parent directories of [path], starting from [path] to the root.
```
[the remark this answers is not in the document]
sigurdm: Root of the file system.

Clarified.

lib/src/io.dart
```diff
@@ -1231,3 +1229,31 @@ String escapeShellArgument(String x) =>
...
+///
+/// If [from] is passed, directories are made relative to that.
+///
+/// Examples:
+///   parentDirs('/a/b/c') => ('/a/b/c', '/a/b', '/a', '/')
+///   parentDirs('./d/e', from: '/a/b/c') => ('./d/e', './d', '.', '..', '../..', '../../..')
+Iterable<String> parentDirs(String path, {String? from}) sync* {
+  var relative = false;
+  var d = path;
+  while (true) {
```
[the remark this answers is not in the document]
sigurdm: No - that is not true. We break when we reach `/` (or `C:\` on windows).

From is the "current directory" where things are reported relative to.

lib/src/entrypoint.dart
sigurdm: Yes - sure.

Done

lib/src/entrypoint.dart
sigurdm: I'm not reloading. This is where the root is loaded (actually created).

This is the `Package1 object. What we have already loaded is the `Pubspec` object.
The `Pubspec` will be in the `pubspecsMet` map, so we will not actually hit the filesystem again for the root pubspec (but we might for other pubspecs in the workspace).

lib/src/entrypoint.dart
sigurdm: oops. Removed.

lib/src/entrypoint.dart
sigurdm: We want to find the `Package` object that has the first `Pubspec` that we loaded. Such that we can refer it.

lib/src/entrypoint.dart

lib/src/entrypoint.dart
jonasfj: is it?

lib/src/entrypoint.dart
jonasfj: Are you relying on the iteration order for `.entries` -- that's seems a little surprising.

Maybe, make like a `firstPubspecMet` variable that is nullable?

Is this because the root loads the packages? Ah.

lib/src/package.dart
```diff
@@ -131,22 +131,48 @@ class Package {
...
+  ///
+  /// This mechanism can be used to avoid loading pubspecs twice. It can also be
+  /// used to override a pubspec in memory for trying out an alternative
+  /// resolution.
   factory Package.load(
     String? name,
     String dir,
     SourceRegistry sources, {
     bool withPubspecOverrides = false,
+    Map<String, Pubspec> alreadyLoadedPubspecs = const {},
```
jonasfj, line 146: Instead of this, how about passing a `loadPubspec()` function?

Then you won't need to document the structure of the map, and the fact that things are cached is something you can hide altogether.

jonasfj (approved):

lib/src/entrypoint.dart
sigurdm: Now it is :)

lib/src/package.dart
sigurdm: Yeah - good idea!

sigurdm: Done

PR #4197, opened 2024-03-25
Show parent pubspecs when a pubspec could not be found.

sigurdm:
Related to https://github.com/dart-lang/pub/issues/4127

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -140,6 +140,9 @@ class Package {
   /// This mechanism can be used to avoid loading pubspecs twice. It can also be
   /// used to override a pubspec in memory for trying out an alternative
   /// resolution.
+  ///
+  /// [referringPubspec] is the path to the pubspec that includes this one in
+  /// the workspace.
   factory Package.load(
     String dir,
     SourceRegistry sources, {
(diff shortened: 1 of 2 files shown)


lib/src/package.dart
```diff
@@ -140,6 +140,9 @@ class Package {
   /// This mechanism can be used to avoid loading pubspecs twice. It can also be
   /// used to override a pubspec in memory for trying out an alternative
   /// resolution.
+  ///
+  /// [referringPubspec] is the path to the pubspec that includes this one in
+  /// the workspace.
   factory Package.load(
     String dir,
```
jonasfj, line 147: nit: documentation string refers to `packageDir`, but it's called `dir`. Probably the same, but would be neat to fix. (unrelated ofcourse no need to do that in this PR)

lib/src/package.dart
```diff
@@ -140,6 +140,9 @@ class Package {
   /// This mechanism can be used to avoid loading pubspecs twice. It can also be
   /// used to override a pubspec in memory for trying out an alternative
   /// resolution.
+  ///
+  /// [referringPubspec] is the path to the pubspec that includes this one in
```
jonasfj, line 144: I don't see this argument passed?

jonasfj (commented):
Looks fine, but you seem to have included a few extra lines not related to this PR. I assume this is about improving the error message?

lib/src/package.dart
sigurdm: Ah yeah, this was from another strategy I tried first. Deleted

lib/src/package.dart
sigurdm: Done

PR #4199, opened 2024-03-25
Close the global http client after `ensurePubspecResolved`

sigurdm:
https://github.com/dart-lang/sdk/issues/55289

### DIFF
--- a/lib/pub.dart
+++ b/lib/pub.dart
@@ -54,6 +54,9 @@ Future<void> ensurePubspecResolved(
   } on ApplicationException catch (e) {
     throw ResolutionFailedException._(e.toString());
   } finally {
+    // TODO(https://github.com/dart-lang/pub/issues/4200)
+    // This is a bit of a hack.
+    // We should most likely take a client here.
     globalHttpClient.close();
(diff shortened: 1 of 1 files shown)


lib/pub.dart
```diff
@@ -52,6 +53,8 @@ Future<void> ensurePubspecResolved(
     );
   } on ApplicationException catch (e) {
     throw ResolutionFailedException._(e.toString());
+  } finally {
+    globalHttpClient.close();
```
jonasfj, line 57: Can we give it a documentation comment.

This is a bit of a hack, I don't mind. But ideally, we'd do something reasonable like creating an a client pass it in, and close it at the end. Or if we don't want to pass it around use zones or something.

jonasfj (approved):

lib/pub.dart
sigurdm: >  But ideally, we'd do something reasonable like creating an a client pass it in, and close it at the end.
> Or if we don't want to pass it around use zones or something.

Yeah - that would be better...
I'll add a comment.

PR #4205, opened 2024-04-02
Fix package name when refreshing binstubs

sigurdm:
By mistake we were using the name of the dummy-package "pub global activate" instead of the activated package.


jonasfj (approved):

lib/src/global_packages.dart
```diff
@@ -947,3 +943,17 @@ fi
...
 }
+
+/// The package that was activated.
+///
+/// * For path packages this is [Entrypoint.workspaceRoot].
+/// * For cached packages this is the sole dependency of
+///   [Entrypoint.workspaceRoot].
+Package activatedPackage(Entrypoint entrypoint) {
+  if (entrypoint.isCachedGlobal) {
+    final dep = entrypoint.workspaceRoot.dependencies.keys.single;
```
jonasfj, line 954: Might be worth explaining how we create a dummy package.

PR #4207, opened 2024-04-04
Merge cherry pick of "Fix package name when refreshing binstubs" (#4205)

sigurdm:
This is a cherry pick, and should be landed as a merge to preserve history


sigurdm:
I believe failures are unrelated to this PR.

jonasfj (approved):

Issue #4214, opened 2024-04-11
Better yaml serialization for pubspec.lock


sigurdm:
Sometimes the content hash is quoted, sometimes not.

![image](https://github.com/dart-lang/pub/assets/8613953/0215d1f0-e1fd-4bad-88fc-ed64d4b3cf3a)

We use a quite naive yaml serialization for writing pubspec.lock.

Here is the regex that decides is a string needs to be json-encoded before serializing:

https://github.com/dart-lang/pub/blob/2179b765aa9071386be02d408b3c3caa82af98f5/lib/src/utils.dart#L511

```
/// The subset of strings that don't need quoting in YAML.
///
/// This pattern does not strictly follow the plain scalar grammar of YAML,
/// which means some strings may be unnecessarily quoted, but it's much simpler.
final _unquotableYamlString = RegExp(r'^[a-zA-Z_-][a-zA-Z_0-9-]*$');
```

We could consider using yaml_edit to write the pubspec.lock file

sigurdm:
Fixing this for hashes starting with a digit might cause most pubspec.locks to be updated. Not sure it is worth it.

PR #4217, opened 2024-04-11
Improve yaml serialization quoting

sigurdm:
#4214

We should only merge this if we want accept disturbing almost all pubspec.locks.

Maybe we can bundle it with another update to pubspec.lock structure.

We can also make a smalller update that just fixes bugs.

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -504,11 +504,27 @@ String prefixLines(String text, {String prefix = '| ', String? firstPrefix}) {
   return lines.join('\n');
 }
 
-/// The subset of strings that don't need quoting in YAML.
+/// Does this string need quoting in yaml?
 ///
-/// This pattern does not strictly follow the plain scalar grammar of YAML,
-/// which means some strings may be unnecessarily quoted, but it's much simpler.
-final _unquotableYamlString = RegExp(r'^[a-zA-Z_-][a-zA-Z_0-9-]*$');
+/// This mechanism does not strictly follow the plain scalar grammar of YAML,
+/// which means some strings may be unnecessarily quoted.
+bool _needsYamlQuotes(String s) {
+  // These must be quoted
+  if ([
+    'true', 'True', 'TRUE', //
+    'false', 'False', 'FALSE',
+    'null', 'Null', 'NULL',
+  ].contains(s)) return true;
+  // Numbers must be quoted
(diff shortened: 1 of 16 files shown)


lib/src/utils.dart
```diff
@@ -504,11 +504,27 @@ String prefixLines(String text, {String prefix = '| ', String? firstPrefix}) {
...
+/// which means some strings may be unnecessarily quoted.
+bool _needsYamlQuotes(String s) {
+  // These must be quoted
+  if ([
+    'true', 'True', 'TRUE', //
+    'false', 'False', 'FALSE',
+    'null', 'Null', 'NULL',
+  ].contains(s)) return true;
+  // Numbers must be quoted
+  if (RegExp(r'^[-+]?[0-9]*\.?[0-9]*(e[0-9]+)?$').hasMatch(s)) return true;
```
lrhn, line 519: This is a potentially dangerous RegExp because of the `\d*\.?\d*` sequence. It's probably not exponential, but it can be quadratic in the length to reject something like `00000000000000000000000....0000000x`

Consider changing to `\d*(\.\d*)?` (And be aware that it still matches `.` alone.)

lib/src/utils.dart
```diff
@@ -504,11 +504,27 @@ String prefixLines(String text, {String prefix = '| ', String? firstPrefix}) {
...
+  // Hex numbers must be quoted
+  if (RegExp(r'^0x[0-9]+$').hasMatch(s)) return true;
```
lrhn, line 521: Should probably be `[0-9a-fA-F]`.

lib/src/utils.dart
```diff
@@ -504,11 +504,27 @@ String prefixLines(String text, {String prefix = '| ', String? firstPrefix}) {
...
-/// This pattern does not strictly follow the plain scalar grammar of YAML,
-/// which means some strings may be unnecessarily quoted, but it's much simpler.
-final _unquotableYamlString = RegExp(r'^[a-zA-Z_-][a-zA-Z_0-9-]*$');
+/// Does this string need quoting in yaml?
+///
+/// This mechanism does not strictly follow the plain scalar grammar of YAML,
+/// which means some strings may be unnecessarily quoted.
+bool _needsYamlQuotes(String s) {
+  // These must be quoted
+  if ([
```
lrhn, line 513: List can be `const`. (Small save, but save, over creating a new list on every invocation.)

lrhn (commented):
(Drive-by-comments. I saw a regexp, so I had to say something.)

Issue #4222, opened 2024-04-15
`getExecutableForCommand` requires reading pubspec.yaml


sigurdm:
# Discussion
Currently `dart run package:command` will only run commands if package is a direct (dev or non-dev) dependency of the current root package.

This convention makes sense. Only the packages you directly depend on are the ones you should rely on running tools from.

But deciding if a package is in dependencies requires parsing `pubspec.yaml` of the root package. As this is in the hot path of `dart run` we would like to only rely on information from `.dart_tool/package_config.json`. (json is faster to parse than yaml, it will be hot on the drive when passed to the VM that also needs the file).

One solution would be to add the names of `dependencies` of each package to the package listing in `.dart_tool/package_config.json`.

One concern with this, is that the size of `.dart_tool/package_config.json` potentially will be O(n^2) in the number of packages in the resolution.

This is information is also strictly more than we need. We really only need to know if a dependency is direct dependency of the root package or not. We could instead add a bit (i`s-direct-dependency-of-root`) to each package listed in the file.

This is however not enough when we introduce workspaces to pub. Here a number of root packages share a `.dart_tool/package_config.json`, and when we do `dart run` inside a package `a` in a workspace, we only want access to the packages that are direct dependencies of `a`, not of every root in the workspace.

That leads to:

# Proposal

Add the list of direct dependencies only to the root packages. `dart run` in a directory `x` would then need to find out which package is associated with `x` and could thus see if the package is one of its dependencies. 

This would not have the O(n^2) behavior (unless o(n) of the packages are root packages...)

Example json:
```json
{
  "configVersion": 2,
  "packages": [
    {
      "name": "dep",
      "rootUri": "file:///usr/local/google/home/sigurdm/.pub-cache/hosted/pub.dev/dep-1.0.0",
      "packageUri": "lib/",
      "languageVersion": "3.0" // Not a root package, no dependencies listing
    },
    {
      "name": "dev_dep",
      "rootUri": "file:///usr/local/google/home/sigurdm/.pub-cache/hosted/pub.dev/dev_dep-1.0.0",
      "packageUri": "lib/",
      "languageVersion": "3.0"  // Not a root package, no dependencies listing
    },
    {
      "name": "myapp",
      "rootUri": "../",
      "packageUri": "lib/",
      "languageVersion": "3.7"
      "dependencies": ["dep", "b"] // Depends on external dep, and workspace member b.
    },
    {
      "name": "b",
      "rootUri": "../pkgs/b",
[...]

PR #4223, opened 2024-04-15
Allow globs in workspace definitions

sigurdm:
Part of #4127

Initially we thought we would not allow globs, because resolving them would be on the hot path of `ensureResolutionUpToDate` and thus `dart run`.

But I think we are getting close to the conclusion that the fast path of `ensureResolutionUpToDate` only should depend on `.dart_tool/package_config.json`, not parsing and resolving pubspec.yamls - thus we can allow globs.

There is still the issue that `executableForCommand` needs to know if a package is a direct dependency or not. We might want to solve that issue before landing this. Discussing in https://github.com/dart-lang/pub/issues/4222

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -4,6 +4,8 @@
 
 import 'dart:io';
 
+import 'package:glob/glob.dart';
+import 'package:glob/list_local_fs.dart';
 import 'package:path/path.dart' as p;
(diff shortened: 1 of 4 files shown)


sigurdm:
We've decided not to do this in the first incantation of workspaces.

PR #4232, opened 2024-04-17
Detect duplicate names in workspace packages

sigurdm:
Part of #4127 
This will otherwise lead to broken behavior with malformed package_configs.
https://github.com/dart-lang/pub/pull/4232

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -361,8 +361,8 @@ See $workspacesDocUrl for more information.
 }
 
 /// Reports an error if the graph of the workspace rooted at [root] is not a
-/// tree.
-void validateWorkspaceGraph(Package root) {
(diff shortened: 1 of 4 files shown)


lib/src/package.dart
```diff
@@ -382,4 +384,16 @@ Packages can only be included in the workspace once.
     }
     stack.addAll(current.workspaceChildren);
   }
+
+  final namesSeen = <String, Package>{};
```
jonasfj, line 388:
```suggestion
  // Check that the workspace doesn't contain two packages with the same name!
  final namesSeen = <String, Package>{};
```

jonasfj (approved):

szakarias (approved):

PR #4251, opened 2024-04-29
`outdated` support for workspaces

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/4250


lib/src/command/outdated.dart
```diff
@@ -1093,3 +1102,29 @@ class _FormattedString {
...
+  return workspaceRoot.transitiveWorkspace
+      .any((p) => p.dependencies.containsKey(name));
+}
+
+/// Whether the package [name] is dev-depended on directly anywhere in the workspace
+/// rooted at [workspaceRoot].
+bool hasDevDependency(Package workspaceRoot, String name) {
+  return workspaceRoot.transitiveWorkspace
+      .any((p) => p.devDependencies.containsKey(name));
+}
```
jonasfj, line 1124: nit: these would be neat as extension methods?

jonasfj (approved):

lib/src/command/outdated.dart
sigurdm: I think I'll refactor these in a follow-up

PR #4261, opened 2024-05-07
Clean up copy and formatting of unpack command help

parlough:
Also renames "constraint" to "descriptor" to be consistent with the terminology from `dart pub add`.


sigurdm (approved):
Thanks!

PR #4272, opened 2024-05-17
Fix flutter version update on fast path

sigurdm:
Double bug!

The branches of the conditional were swapped, **and** the comparison was between a `String` and a `Version`.


sigurdm:
I think this needs a cherry pick to stable.

jonasfj (approved):

PR #4276, opened 2024-05-21
Fix bug in ensureUpToDate fast path, when following workspaceRef

sigurdm:
We for some bogus reason added the text of the file in the path...

It happened to work by accident in the existing test cases.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -1040,7 +1040,6 @@ To update `$lockFilePath` run `$topLevelProgram pub get`$suffix without
                   p.absolute(
                     p.join(
                       p.dirname(potentialWorkspaceRefPath),
-                      workspaceRefText,
                       path,
                       '.dart_tool',
                       'package_config.json',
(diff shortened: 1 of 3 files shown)


lib/src/entrypoint.dart
```diff
@@ -1040,14 +1040,28 @@ To update `$lockFilePath` run `$topLevelProgram pub get`$suffix without
...
+              //     workspaceRefText, path));
+              // print(p.join(
+              //   p.dirname(potentialWorkspaceRefPath),
+              //   workspaceRefText,
+              //   path,
+              //   '.dart_tool',
+              // ));
+              // print(p.join(p.dirname(potentialWorkspaceRefPath),
+              //     workspaceRefText, path, '.dart_tool', 'package_config.json'));
               packageConfigStat = tryStatFile(potentialPackageConfigPath2);
```
jonasfj, line 1065: Are these comments intentional?

jonasfj (approved):

lib/src/entrypoint.dart
sigurdm: No!

Issue #4300, opened 2024-06-05
Error handling excludes from parent gitignore files prevents publishing


jakemac53:
# Problem

This came up when @elliette was trying to publish the `dap` package, which lives under https://github.com/dart-lang/sdk/tree/main/third_party/pkg/dap.

There is a .gitignore file two directories above this, which ignores everything under third_party except some specifically listed directories, including this one (see [here](https://github.com/dart-lang/sdk/blob/main/third_party/.gitignore)).

When trying to publish, pub gives an error that the files in the package are git ignored, and won't be published.

<details>
<summary>Error logs</summary>

Validating package... (1.5s)
Package validation found the following errors:
* The pubspec is hidden, probably by .gitignore or pubignore.
* You must have a LICENSE file in the root directory.
  An open-source license helps ensure people can legally use your code.
Package validation found the following potential issues:
* 17 checked-in files are ignored by a `.gitignore`.
  Previous versions of Pub would include those in the published package.

  Consider adjusting your `.gitignore` files to not ignore those files, and if you do not wish to
  publish these files use `.pubignore`. See also dart.dev/go/pubignore

  Files that are checked in while gitignored:

  CHANGELOG.md
  LICENSE
  OWNERS
  README.md
  analysis_options.yaml
  lib/dap.dart
  lib/src/exceptions.dart
  lib/src/protocol_common.dart
  lib/src/protocol_generated.dart
  lib/src/protocol_special.dart

</details>

## Expected behavior

Should be able to publish, the directory is not actually ignored.

## Actual behavior

Can't publish, and if we could it seems that several of the required files would not be published.

sigurdm:
I believe he warning is actually correct. `third_party/pkg/dap` **is** actually ignored by the gitignore.

The reason you see `pkg/dap` as not ignored by git is because it is checked in despite being ignored. We have decided that this is an inconsistency, and pub will not publish such a file, but warn about its existence.

Even though it looks like the `third_party/pkg/dap` folder is excluded in the `.gitignore`, it is not. This has to do with how gitignores are implemented (both by git and pub):

Directories are matched before their content recursively. And if a directory is ignored, its content is not examined at all. `pkg/` is ignored by the `*` pattern, so we never get to examine `pkg/dap` and its contents.

To truly except `pkg/dap` from being ignored you need to first exclude `pkg/` from the ignore, then reignore all its contents and then exclude the `pkg/dap` folder againg. See https://stackoverflow.com/a/11018557

Try something like:

```
[...]

PR #4301, opened 2024-06-05
Add test for issue when excluding a nested dir from a higher level gitignore file

jakemac53:
Repro test for https://github.com/dart-lang/pub/issues/4300

I don't see a good mechanism for skipping individual parts of a test so this can just land after the issue is fixed (or close this and write different tests)

### DIFF
--- a/test/validator/gitignore_test.dart
+++ b/test/validator/gitignore_test.dart
@@ -51,6 +51,13 @@ void main() {
       ]),
       exit_codes.DATA,
     );
+
+    // Check that you can unignore certain files too.
+    await d.dir('myapp', [
+      d.file('.gitignore', '*.txt\n!foo.txt'),
+    ]).create();
+
+    await expectValidation(contains('Package has 0 warnings.'), 0);
   });
 
   test('should not fail on non-ascii unicode character', () async {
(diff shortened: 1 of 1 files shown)


test/validator/gitignore_test.dart
```diff
@@ -51,6 +51,13 @@ void main() {
       ]),
       exit_codes.DATA,
     );
+
+    // Check that you can unignore certain files too.
+    await d.dir('myapp', [
+      d.file('.gitignore', '*.txt\n!foo.txt'),
```
sigurdm, line 57: I believe this is tested well enough in the general ignore-rules tests. Eg. here: https://github.com/dart-lang/pub/blob/e1fbda73d1ac597474b82882ee0bf6ecea5df108/test/ignore_test.dart#L344

test/validator/gitignore_test.dart
```diff
@@ -121,6 +128,17 @@ void main() {
       exit_codes.DATA,
       workingDirectory: packageRoot,
     );
+
+    // Check that you can unignore certain dirs from a higher level gitignore.
+    await d.dir('reporoot', [
+      d.file('.gitignore', '*.txt\n!myapp/'),
```
sigurdm, line 134: I don't think this does whatever you think it does.

`myapp/` is not ignored here, so it cannot meaningfully be negated/unignored.

We could test that `"/*\n!myapp/"` leads to no warnings. That is already the case.
```suggestion
      d.file('.gitignore', '/*\n!myapp/'),
```

sigurdm:
As mentioned in the issue I believe you are "holding it wrong".

I think we implement the correct rules here. They are just somewhat obscure.

jakemac53:
Sounds good, that behavior seems bizarre to me but as long as we are consistent with git, that is what matters :)

Issue #4305, opened 2024-06-12
Pub commands fail with "Invalid tar entry: .."


jakemac53:
# Problem

Pub commands (`pub get`, `pub outdated`, `pub global activate`, etc.) fail with output like:

```
Downloading packages... 
Failed to extract `/tmp/pub_FTQXCZ/devtools_shared-10.0.0.tar.gz`: Invalid tar entry: ..
```
### Expected behavior

Pub commands complete.

### Actual behavior

Pub commands fail.

# Cause

Tar files downloaded from pub.dev fail to be extracted due to containing a `.` entry.

This is a severe regression, likely caused by https://github.com/dart-lang/pub/pull/4283

# Exposure

The regression is currently in the Dart `main` and `dev` channels. It is *not* in the `beta` and `stable` channels.

Only the following dev-releases are affected:

* 3.5.0-243.0.dev
* 3.5.0-244.0.dev
* 3.5.0-245.0.dev
* 3.5.0-246.0.dev
* 3.5.0-247.0.dev
* 3.5.0-248.0.dev
* 3.5.0-249.0.dev
* 3.5.0-250.0.dev
* 3.5.0-251.0.dev
* 3.5.0-252.0.dev
* 3.5.0-253.0.dev
* 3.5.0-254.0.dev
* 3.5.0-255.0.dev
* 3.5.0-256.0.dev
* 3.5.0-257.0.dev
* 3.5.0-258.0.dev

The pub.dev server was patched on June 13th to reject publishing further packages with broken archives.

The affected already published package-versions are:

* ant_media_flutter-1.4.1.tar.gz
* cxgenie-2.4.5.tar.gz
* cxgenie-2.4.4.tar.gz
* cxgenie-2.4.3.tar.gz
* cxgenie-2.4.2.tar.gz
* macros-0.1.2-main.1.tar.gz
* devtools_extensions-0.2.0.tar.gz
* devtools_shared-10.0.0.tar.gz
* devtools_app_shared-0.2.0.tar.gz

# Mitigation

Package versions already published to pub.dev will not work. We recommend that package owners use the pub.dev Admin functionality to *retract* these package versions: https://dart.dev/tools/pub/publishing#retract

New package versions can be published safely as long as you use a beta or stable SDK to do the publishing with.

brianquinlan:
[...]

PR #4306, opened 2024-06-13
Fix handling of directories in archives 

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/4305

Pub would not even extract its own archives because it choked on the `.` entry.
Bug was introduced in https://github.com/dart-lang/pub/pull/4283

* stop including '.' - that was probably a mistake anyway.
* allow extracting packages that has '.' included. This allows consuming packages that were created by the bogus version.
* A test that we can extract what we ourself create (how did this not exist already 🤦).
* A fix to the error message such that the entry is quoted. so `Invalid tar entry: ..` becomes `` Invalid tar entry: `.`. ``
* Fix the mode of directories to be drwxr-x---


lib/src/io.dart
```diff
@@ -1072,12 +1072,14 @@ Future<void> extractTarGz(Stream<List<int>> stream, String destination) async {
       throw FormatException('Tar file contained duplicate path ${entry.name}');
     }
 
-    if (!p.isWithin(destination, filePath)) {
+    if (!(p.isWithin(destination, filePath) ||
+        // allow including '.' as an entry in the tar.gz archive.
+        p.equals(destination, filePath))) {
```
simolus3, line 1077: Perhaps we should only allow this if the entry is also a directory, I think it might break for a regular file named `.` in the archive (since we then have a single file where we'd expect a directory).

sigurdm: Good point! I think it would be caught below when trying to create the file, but it might be better to catch early.

Updated.

PR #4331, opened 2024-07-22
Remove some unnecessary analysis options

natebosch:
The upgrade of some warnings to errors was intended to make external CI
a closer match to google3. Since then we have started using
`--fatal-infos` which makes it unnecessary to make these errors.

The TODO ignore is unnecessary because the analysis server no longer
includes TODO comments in the list of diagnostics.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,11 +1,6 @@
 include: package:lints/recommended.yaml
 
 analyzer:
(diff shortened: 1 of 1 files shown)


sigurdm (approved):

sigurdm:
Thanks! Much nicer.

PR #4332, opened 2024-07-22
Fix a batch of long lines

natebosch:
Enable lines_longer_than_80_chars lint and fix the first ~100
diagnostics.

Only a very small number felt like they might be toil during normal
editing.
The vast majority felt like they would be mostly unobtrusive if they had
surfaced during normal editing instead of all in a batch.

Quite a few of the fixes improve readability. None of the fixes harm
readability.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -25,6 +25,7 @@ linter:
     - cancel_subscriptions
     - directives_ordering
     - library_annotations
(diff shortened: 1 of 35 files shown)


sigurdm:
Thanks for doing this! https://github.com/dart-lang/pub/pull/4340 is finishing this.

PR #4339, opened 2024-08-06
Ignore todos

sigurdm:
I prefer todos to not show up in my problems view in vscode.

This is the best solution I've found for this.

Reverts a change made in https://github.com/dart-lang/pub/pull/4331


jonasfj (approved):

natebosch:
Ah sorry, I thought the TODOs were removed everywhere, but it must have been only in some configurations of the analysis server.

PR #4344, opened 2024-08-08
Avoid reloading credentials file on each getter access

sigurdm:
And only let 'token remove' resave it when actually mutated.

Also includes a few comment fixes.


sigurdm:
Ping

szakarias (approved):
LGTM

PR #4347, opened 2024-08-09
Abide by lint unintended_html_in_doc_comment

sigurdm:


lib/src/executable.dart
```diff
@@ -260,7 +260,7 @@ final class DartExecutableWithPackageConfig {
 /// For example:
 /// * `foo` will resolve to `foo:bin/foo.dart` or `foo:bin/main.dart`.
 /// * `:foo` will resolve to `<current>:bin/foo.dart`.
-/// * `` and `:` both resolves to `<current>:bin/<current>.dart` or
+/// * The empty string and `":` both resolves to `<current>:bin/<current>.dart` or
```
jonasfj, line 263:
```suggestion
/// * The empty string and `":"` both resolves to `<current>:bin/<current>.dart` or
```
or
```suggestion
/// * `''` and `':'` both resolves to `<current>:bin/<current>.dart` or
```

jonasfj (approved):
I do love it when people use backticks!

PR #4351, opened 2024-08-15
Address old line-splitting TODO

parlough:


lib/src/io.dart
```diff
@@ -1240,8 +1239,12 @@ class PubProcessResult {
 
   // TODO(rnystrom): Remove this and change to returning one string.
```
sigurdm, line 1240: Is this TODO really resolved?

parlough: No this one isn't, so I left it here. It would require a larger refactoring.

sigurdm: Ah sorry, this PR is about the other TODO. Gotcha....

sigurdm (approved):
Thanks!

Issue #2197, opened 2019-09-02
Bump version of pubspec


JasCodes:
Cli should allow to bump version major, minor build

Thx

jonasfj:
Anything that mutates the `pubspec.yaml` will require:
 A) that we strip all formatting and comments
 B) That we write a library that modify YAML without throwing away all comments.

(A) has significant downsides as it's common to comment out `dependency_overrides` and other hacks used during development. But it might be what we end up doing.

If anyone is interested in taking on (B) writing an abstract syntax tree for YAML and support mutating it that would make adding this feature very easy.

jascodes:
Or C we only parse out version key rather then fully implementing whole YAML implementation :stuck_out_tongue:

jakemac53:
Parsing out the version key is harder than you would think - yaml has like a billion ways of doing the same thing.

That being said you could use the source code information from the parsed yaml to specifically edit that node without to much headache I think. I don't know if it is really worth doing though and would be a potential source of bugs and/or race conditions.

jascodes:
@jacob314 You might be right, I might just going with sed command in my script. I know its no trivial task to implement full yaml spec in dart.

Thx

jonasfj:
@jascodes, so @sigurdm suggested it might be possible by forking [dart-lang/yaml](https://github.com/dart-lang/yaml), extending the [scanner](https://github.com/dart-lang/yaml/blob/4d016bc3227621e329895db6504dfd78cd9fabca/lib/src/scanner.dart) to emit comments for which there is [currently no token type](https://github.com/dart-lang/yaml/blob/4d016bc3227621e329895db6504dfd78cd9fabca/lib/src/token.dart#L114-L139). Then it might be possible to write something that mutates the token stream and rebuild a YAML string from the token stream.
This would probably be a pain retain formatting, but if comments are retained that might be good enough :)
If anyone wants to give this approach a go, by all means raise your hand (I'm happy to give feedback/reviews, etc -- I suspect it'll be a bit of work).

Levi-Lesches:
Check out the package [cider](https://pub.dev/packages/cider) on pub.dev. The `cider bump` command can bump any part of the version, such as: 
- breaking (means y for 0.y.z and x for x.y.z)
- major
- minor
- patch
- build
- pre (short for pre-release)

Here are a few sample commands: 

1. Read the current version
```
cider version
```
2. Change the version number to 1.2.3
  ```
  cider version 1.2.3
  ```
[...]

PR #4361, opened 2024-08-30
`pub bump` command

sigurdm:
https://github.com/dart-lang/pub/issues/2197


test/testdata/goldens/help_test/pub bump --help.txt
```diff
@@ -0,0 +1,17 @@
...
+$ pub bump --help
+Increases the version number of the current package.
+
+
+Usage: pub bump <subcommand> [arguments...]
+-h, --help        Print this usage information.
+    --major       Increment the major version number (eg. 3.1.2 -> 4.0.0)
+    --minor       Increment the minor version number (eg. 3.1.2 -> 3.2.0)
+    --patch       Increment the patch version number (eg. 3.1.2 -> 3.1.3)
+    --breaking    Increment to the next breaking version (eg. 0.1.2 -> 0.2.0)
```
jonasfj, line 13: I think these should be subcommands instead, since:
 * you can't combine them (can you?); AND;
 * you offer do default value.

They don't feel like options, but more like separate actions one can take.

I think we could consider giving each of these subcommands a `--message` that would then flow into a `CHANGELOG` entry. If we wanted to.

I could imagine doing:
```console
$ dart pub bump major -m 'Initial release'
```

But then again, an option isn't really a super suitable place to write a CHANGELOG message.

Could be we just add a new section to the CHANGELOG by prepending it `"# v{{version}}\nTODO[block-publish]: Write changelog entry.\n\n"`

Then we could block publishing if `CHANGELOG.md` contains `TODO[block-publish]` (or maybe just `TODO`, perhaps `[block-publish]` is unnecessary)

lib/src/command/bump.dart
```diff
@@ -0,0 +1,90 @@
...
+    } else if (minor) {
+      newVersion = currentVersion.nextMinor;
+    } else if (patch) {
+      newVersion = currentVersion.nextPatch;
+    } else {
+      throw StateError('Should not be possible');
+    }
+
+    if (argResults.flag('dry-run')) {
+      log.message('Would update version from $currentVersion to $newVersion.');
```
jonasfj, line 80: Show me the diff!

lib/src/command/bump.dart
```diff
@@ -0,0 +1,90 @@
...
+    } else {
+      log.message('Updating version from $currentVersion to $newVersion.');
+      final yamlEditor =
+          YamlEditor(readTextFile(entrypoint.workPackage.pubspecPath));
+
+      yamlEditor.update(['version'], newVersion.toString());
+      writeTextFile(entrypoint.workPackage.pubspecPath, yamlEditor.toString());
```
jonasfj, line 87: Display a message that says: "Do update CHANGELOG.md before publishing!"

test/testdata/goldens/help_test/pub bump --help.txt
sigurdm: I think a message can come as a later feature. It requires parsing changelog, deciding on the format, updating the structure etc.

lib/src/command/bump.dart
sigurdm: Done

lib/src/command/bump.dart
sigurdm: Done

szakarias (approved):
Very nice! 

I have no strong opinion on the subcommand matter.

test/testdata/goldens/help_test/pub bump --help.txt
jonasfj: Agreed!

lib/src/command/bump.dart
```diff
@@ -0,0 +1,131 @@
...
+  @override
+  final String description;
+
+  final Version Function(Version) updateVersion;
+  BumpSubcommand(this.name, this.description, this.updateVersion) {
+    argParser.addFlag(
+      'dry-run',
+      abbr: 'n',
+      negatable: false,
+      help: "Report what would change but don't change anything.",
```
jonasfj, line 28:
```suggestion
      help: "Report what would change, but don't change anything.",
```

jonasfj (approved):

lib/src/command/bump.dart
sigurdm: Done

PR #4368, opened 2024-09-03
Use `-z` for `git ls-files`

sigurdm:
This gives more reliably parsable output.

Main changes are in `lib/src/validator/gitignore.dart`.

Required refactor of `runProcess()` to not output split lines, but according to the encoding (following `Process.run()`).

### DIFF
--- a/lib/src/git.dart
+++ b/lib/src/git.dart
@@ -25,18 +25,18 @@ class GitException implements ApplicationException {
   final List<String> args;
 
   /// The standard error emitted by git.
-  final String stderr;
+  final dynamic stderr;
 
   /// The standard out emitted by git.
-  final String stdout;
+  final dynamic stdout;
 
   /// The error code
   final int exitCode;
 
   @override
   String get message => 'Git error. Command: `git ${args.join(' ')}`\n'
-      'stdout: $stdout\n'
-      'stderr: $stderr\n'
+      'stdout: ${stdout is String ? stdout : '<binary>'}\n'
+      'stderr: ${stderr is String ? stderr : '<binary>'}\n'
       'exit code: $exitCode';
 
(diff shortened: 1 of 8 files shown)


lib/src/git.dart
```diff
@@ -51,12 +51,14 @@ bool get isInstalled => command != null;
 
 /// Run a git process with [args] from [workingDir].
 ///
-/// Returns the stdout as a list of strings if it succeeded. Completes to an
+/// Returns the stdout if it succeeded. Completes to an
```
szakarias, line 54: nit: should fit on one line

szakarias (commented):
I am not a super fan of that many casts, but maybe there is no nicer way...

lib/src/validator/gitignore.dart
```diff
@@ -22,27 +22,37 @@ class GitignoreValidator extends Validator {
...
         // If git is not supported on the platform, or too old to support
         // --recurse-submodules we just continue silently.
         return;
       }
+      final checkedIntoGit = <String>[];
+      // Split at \0.
+      var start = 0;
+      for (var i = 0; i < output.length; i++) {
+        if (output[i] == 0) {
+          checkedIntoGit.add(utf8.decode(output.sublist(start, i)));
```
sigurdm, line 52: Use `systemencoding`

sigurdm, line 52: Use a view?

sigurdm: Done

sigurdm: Done

sigurdm: Actually utf8 seems to be right here, according to tests...

Issue #2040, opened 2019-01-30
Warn in `pub publish` if `git status` is not clean


mklim:
Splitting this off from some discussion on flutter/flutter#27258. When we're publishing a package, it's possible to have build artifacts or other untracked/ignored files in the local directory of the package. But by default we only ever want to publish files that are already tracked and committed in git. `pub publish` does spit out a tree of all the files its planning on publishing already, but in practice this isn't really useful for us since there's so many files in a typical package.

It would be helpful for us if there was a flag or setting where we could tell pub that we're trying to use git as our source of truth, and have it at a bare minimum panic and throw up a warning if the `git status` isn't clean and/or it's about to upload files that aren't already committed in the source tree.

jonasfj:
Maybe we should just warn against publishing files that are:
 * tracked in git
 * have changes not committed in git

I think it's always reasonable to warn against tracked files with changes not committed in git.

jonasfj:
I think this can be implemented as a validator:
https://github.com/dart-lang/pub/tree/master/lib/src/validator

PR #4373, opened 2024-09-05
Validate git is clean when publishing

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/2040

Does not warn about newly added files, because they might get generated by a build step.

### DIFF
--- a/lib/src/git.dart
+++ b/lib/src/git.dart
@@ -50,6 +50,37 @@ class GitException implements ApplicationException {
 /// Tests whether or not the git command-line app is available for use.
 bool get isInstalled => command != null;
 
+/// Splits the [output] of a git -z command at \0.
+///
+/// The first [skipPrefix] bytes of each substring will be ignored (useful for
+/// `git status -z`). If there are not enough bytes to skip, throws a
+/// [FormatException].
+List<Uint8List> splitZeroTerminated(Uint8List output, {int skipPrefix = 0}) {
+  final result = <Uint8List>[];
+  var start = 0;
+
+  for (var i = 0; i < output.length; i++) {
+    if (output[i] != 0) {
+      continue;
+    }
+    if (start + skipPrefix > i) {
+      throw FormatException('Substring too short for prefix at $start');
+    }
+    result.add(
+      Uint8List.sublistView(
+        output,
+        // The first 3 bytes are the modification status.
+        // Skip those.
+        start + skipPrefix,
+        i,
+      ),
+    );
+
+    start = i + 1;
+  }
+  return result;
+}
+
 /// Run a git process with [args] from [workingDir].
 ///
 /// Returns the stdout if it succeeded. Completes to ans exception if it failed.
--- a/lib/src/validator/git_status.dart
+++ b/lib/src/validator/git_status.dart
@@ -13,67 +13,73 @@ import '../log.dart' as log;
 import '../utils.dart';
 import '../validator.dart';
 
-/// A validator that validates that no checked in files are modiofied in git.
+/// A validator that validates that no checked in files are modified in git.
 ///
 /// Doesn't report on newly added files, as generated files might not be checked
 /// in to git.
 class GitStatusValidator extends Validator {
   @override
   Future<void> validate() async {
-    if (package.inGitRepo) {
-      final modfiedFiles = <String>[];
-      try {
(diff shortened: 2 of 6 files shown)


lib/src/validator/git_status.dart
```diff
@@ -0,0 +1,91 @@
...
+import 'dart:typed_data';
+
+import 'package:path/path.dart' as p;
+
+import '../git.dart' as git;
+import '../log.dart' as log;
+import '../utils.dart';
+import '../validator.dart';
+
+/// A validator that validates that no checked in files are modiofied in git.
```
jonasfj, line 16:
```suggestion
/// A validator that validates that no checked in files are modified in git.
```

lib/src/validator/git_status.dart
```diff
@@ -0,0 +1,91 @@
...
+///
+/// Doesn't report on newly added files, as generated files might not be checked
+/// in to git.
+class GitStatusValidator extends Validator {
+  @override
+  Future<void> validate() async {
+    if (package.inGitRepo) {
```
jonasfj, line 23: nit: please return early instead of nesting a bit more :rofl: 
```suggestion
    if (package.inGitRepo) {
      return;
    }
```

lib/src/validator/git_status.dart
```diff
@@ -0,0 +1,91 @@
...
+            '--untracked-files=no', // Don't show untracked files.
+          ],
+          workingDir: package.dir,
+        );
+        // Split at \0.
+        var start = 0;
+        for (var i = 0; i < output.length; i++) {
+          if (output[i] != 0) {
+            continue;
+          }
```
jonasfj, line 49: Perhaps consider splitting this out into a helper library.. `git.parseMachineStatus` or something.

This might be worth having unit tests cover.

lib/src/validator/git_status.dart
```diff
@@ -0,0 +1,91 @@
...
+          final filename = utf8.decode(
```
jonasfj, line 50: you might want to handle `FormatException`

lib/src/validator/git_status.dart
sigurdm: Done

lib/src/validator/git_status.dart
sigurdm: Done

lib/src/validator/git_status.dart
sigurdm: Good point! Done!

lib/src/git.dart
```diff
@@ -50,6 +50,37 @@ class GitException implements ApplicationException {
 /// Tests whether or not the git command-line app is available for use.
 bool get isInstalled => command != null;
 
+/// Splits the [output] of a git -z command at \0.
+///
+/// The first [skipPrefix] bytes of each substring will be ignored (useful for
+/// `git status -z`). If there are not enough bytes to skip, throws a
+/// [FormatException].
+List<Uint8List> splitZeroTerminated(Uint8List output, {int skipPrefix = 0}) {
```
jonasfj, line 58: `splitZeroTerminated` without the `skipPrefix` might have been more sane.

But this is also fine :D

jonasfj (approved):

lib/src/git.dart
sigurdm: Yeah - more sane - just I wanted to avoid creating too much garbage, and (as far as I can tell) `utf8.decode` cannot work on a partial buffer...

MaryaBelanger:
Hi @sigurdm, couple questions for docs:

- Does the warning show when using `--dry-run`?
- Does `--skip-validation` omit the warning?

sigurdm:
> Hi @sigurdm, couple questions for docs:
> 
> * Does the warning show when using `--dry-run`?

Yes, all the `Validator`s are run for a dry-run

> * Does `--skip-validation` omit the warning?

Also yes.  No `Validator`s are run when `--skip-validation` is passed.

jeanplevesque:
How can we disable this specific validator? My CI now fails because it runs scripts to edit some files (such as changing the version in the pubspec file based on GitVersion) which introduces **desired** git changes.

sigurdm:
There is (currently) no way to way to disable a single validation.

You could consider doing a dummy commit on CI before publishing.

PR #4374, opened 2024-09-06
Commit lockfile

sigurdm:
This should hopefully make it easier in the future to run historical commits of pub.

We can hopefully rely more or on dependabot to keep us up-to-date.

The versions in the lockfile will not in general correspond to the exact ones in the sdk DEPS, but be the ones we test with.


jonasfj (approved):
If it doesn't work for us we can always change our mind.

PR #4378, opened 2024-09-10
New command `workspace list`

sigurdm:
Lists all packages in the workspace


test/workspace_test.dart
```diff
@@ -1416,6 +1416,79 @@ Consider removing one of the overrides.''',
...
+''',
+    );
+    await runPub(
+      args: ['workspace', 'list'],
+      environment: {'_PUB_TEST_SDK_VERSION': '3.5.0'},
+      workingDirectory: p.join(sandbox, appPath, 'pkgs'),
+      output: '''
+myapp: ..
+a: a
+b: a/b
```
jonasfj, line 1466: Should we consider:
 * Some right pad?
 * Maybe tabel headers too (possibly bold, like we do in `dart pub outdated`)
 * Ending paths in `/` to indicate that they are directories?

**Example**
```
Package:                Path:
_                       ./
a                       a/
b                       b/
```

In the future, I could imagine a `resolution` column too:
```
Package:                Path:               Resolution:
_                       ./                  workspace
a                       a/                  workspace
b                       b/                  local
```

jonasfj (approved):

test/workspace_test.dart
sigurdm: Done

jonasfj (approved):

sigurdm:
@spydon does this work for melos?

spydon:
> @spydon does this work for melos?

It looks great! 😃
Thanks for implementing it so quickly.

PR #4400, opened 2024-09-30
Allow overriding `resolution` and `workspace` fields in pubspec_overrides.

### LINKED ISSUE #4393: `dart pub unpack` fails with dependency published from a workspace
# Environment
```
Dart SDK version: 3.5.3 (stable) (Wed Sep 11 16:22:47 2024 +0000) on "linux_x64"
```

# Problem
 
`unpack` command runs `dart pub get` in dependency that is a workspace and fails.

## Expected behavior

`unpack` should work "out of the box" with `resolution: workspace` packages

`dart pub unpack binary_serializable` should succeed.

## Actual behavior
```
$ dart pub unpack binary_serializable
Downloading binary_serializable 0.3.0 to `./binary_serializable-0.3.0`... 
To explore type: cd ./binary_serializable-0.3.0
Found a pubspec.yaml at /home/elijah/Development/binary_serializable-0.3.0. But it has resolution `workspace`.
But found no workspace root including it in parent directories.

See https://dart.dev/go/pub-workspaces for more information.
```

The exit code is `66`

I am able to resolve by removing the `resolution: workspace` entry in the `pubspec.yaml` and running `dart pub get` by hand.

sigurdm:
And use this in  `dart pub unpack` for allowing to unpack workspace packages.

Fixes #4393

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -515,7 +515,7 @@ See $workspacesDocUrl for more information.''',
   /// pubspec.yaml and all dependencies downloaded.
   Future<void> acquireDependencies(
     SolveType type, {
-    Iterable<String>? unlock,
+    Iterable<String> unlock = const [],
     bool dryRun = false,
     bool precompile = false,
     bool summaryOnly = false,
@@ -547,7 +547,7 @@ Try running `$topLevelProgram pub get` to create `$lockFilePath`.''');
           cache,
           workspaceRoot,
           lockFile: lockFile,
-          unlock: unlock ?? [],
+          unlock: unlock,
         );
       });
     } on SolveFailure catch (e) {
@@ -557,7 +557,7 @@ Try running `$topLevelProgram pub get` to create `$lockFilePath`.''');
           this,
           type,
           e.incompatibility,
-          unlock ?? [],
+          unlock,
           cache,
         ),
       );
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -111,15 +111,16 @@ class Entrypoint {
             path, {
             expectedName,
             required withPubspecOverrides,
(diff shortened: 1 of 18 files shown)


lib/src/entrypoint.dart
```diff
@@ -111,15 +111,16 @@ class Entrypoint {
...
-          }) =>
-              pubspecsMet[p.canonicalize(path)] ??
-              Pubspec.load(
-                path,
-                cache.sources,
-                expectedName: expectedName,
-                allowOverridesFile: withPubspecOverrides,
-                containingDescription: RootDescription(path),
-              ),
+          }) {
```
jonasfj, line 114: is this just reformatting?

test/unpack_test.dart
```diff
@@ -125,4 +125,41 @@ Resolving dependencies in `../foo-1.2.3-pre`...
...
+    await runPub(
+      args: ['unpack', 'foo:1.0.0'],
+      output: allOf(
+        contains('Downloading foo 1.0.0 to `.${s}foo-1.0.0`...'),
+        contains(
+          '+ bar',
+        ),
+      ),
+      environment: {'_PUB_TEST_SDK_VERSION': '3.5.0'},
+    );
```
jonasfj, line 163: await file('appPath/pubspec_overrides.yaml', 'resolution:').exists()

Or something like that, not necessary, just a suggestion.

jonasfj (approved):

lib/src/entrypoint.dart
sigurdm: Yeah - I had some debug prints in here I think. I'll revert this part.

test/unpack_test.dart
sigurdm: Done

Issue #3504, opened 2022-07-28
`pub upgrade some_package` doesn't upgrade transitive dependencies


stuartmorgan:
# Environment

- `pub version` or `flutter pub version`: Pub 2.19.0-33.0.dev
- OS version: macOS 12.4
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

According to https://dart.dev/tools/pub/cmd/pub-upgrade#upgrading-specific-dependencies upgrading a specific package should upgrade its transitive dependencies:

> Upgrading a dependency upgrades its transitive dependencies to their latest versions as well.

(Somewhat confusingly, it also says:
> Usually, no other dependencies are updated; they stay at the versions that are locked in the lockfile. However, if the requested upgrades cause incompatibilities with these locked versions, they are selectively unlocked until a compatible set of versions is found.

I'm assuming "no other dependencies" means "dependencies that are not in the transitive dependency list of the requested packages" not, "packages other than the ones explicitly listed in the command line", because otherwise the first sentence makes no sense to me. This issue report is predicated on that understanding.)

This doesn't seem to be the actual behavior. I set up a test after seeing unexpected results in actual use, by adding `path_provider` to a new Flutter project, then manually editing `pubspec.lock` to contain older versions of its dependencies (but still compatible with `path_provider`'s constraints). At this point, `dart pub outdated` gives me the expected output:

```
$ dart pub outdated
Showing outdated packages.
[*] indicates versions that are not the latest available.

Package Name                      Current  Upgradable  Resolvable  Latest  

direct dependencies: all up-to-date.

dev_dependencies: all up-to-date.

transitive dependencies:         
ffi                               *1.2.1   2.0.1       2.0.1       2.0.1   
path_provider_android             *2.0.6   2.0.17      2.0.17      2.0.17  
path_provider_ios                 *2.0.6   2.0.11      2.0.11      2.0.11  
path_provider_linux               *2.0.2   2.1.7       2.1.7       2.1.7   
path_provider_macos               *2.0.0   2.0.6       2.0.6       2.0.6   
path_provider_platform_interface  *2.0.0   2.0.4       2.0.4       2.0.4   
path_provider_windows             *2.0.3   2.1.0       2.1.0       2.1.0   
win32                             *2.6.1   2.7.0       2.7.0       2.7.0   

8 upgradable dependencies are locked (in pubspec.lock) to older versions.
To update these dependencies, use `dart pub upgrade`.
all dependencies are up-to-date.
```

[...]

PR #4403, opened 2024-10-03
Add flag `--unlock-transitive` to `pub upgrade`

sigurdm:
`pub upgrade x y z` will only unlock x, y and z, but with the new flag `pub upgrade --transitive x y z` will unlock those packages **and** their transitive dependencies.

Same for `--major-versions`.

Fixes https://github.com/dart-lang/pub/issues/3504

We could also consider making this the default behavior - it is most likely what people expect.

### DIFF
--- a/lib/src/command/upgrade.dart
+++ b/lib/src/command/upgrade.dart
@@ -71,7 +71,7 @@ class UpgradeCommand extends PubCommand {
     );
 
     argParser.addFlag(
-      'transitive',
+      'unlock-transitive',
       help: 'Also upgrades the transitive dependencies '
           'of the listed [dependencies]',
     );
@@ -115,7 +115,7 @@ class UpgradeCommand extends PubCommand {
   /// This allows the user to specify list of names that they want the
   /// upgrade command to affect.
   Future<List<String>> _computePackagesToUpgrade() async {
-    if (argResults.flag('transitive')) {
+    if (argResults.flag('unlock-transitive')) {
       final graph = await entrypoint.packageGraph;
       return argResults.rest
           .expand(
--- a/lib/src/command/upgrade.dart
+++ b/lib/src/command/upgrade.dart
@@ -71,6 +71,12 @@ class UpgradeCommand extends PubCommand {
     );
 
     argParser.addFlag(
+      'transitive',
+      help: 'Also upgrades the transitive dependencies '
+          'of the listed [dependencies]',
+    );
+
+    argParser.addFlag(
       'major-versions',
       help: 'Upgrades packages to their latest resolvable versions, '
           'and updates pubspec.yaml.',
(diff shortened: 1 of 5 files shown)


jonasfj:
This is a bit confusing, because:
 * `pub upgrade` and `pub upgrade --transitive` are the same thing.
 * `pub upgrade foo` and `pub upgrade --transitive foo` are **NOT** the same thing.

sigurdm:
> pub upgrade and pub upgrade --transitive are the same thing.

Well, `pub upgrade` in my mind means `pub upgrade <all dependencies>` and the transitive closure of `<all dependendencies>` is still `<all dependencies>`.

So it is the same rule we apply.

But we could disallow `--transitive` when there are no given deps to update.

sigurdm:
This will not work in cases where the latest version of the upgraded package has more dependencies than the current.

Then they will not be unlocked.

Say:

foo 1.0.0 has
```yaml
dependencies:
  # no dependencies
```

foo 1.5.0 has
```yaml
dependencies:
  bar: ^1.0.0
```

And the main pubspec.yaml looks like
```yaml
dependencies:
  foo: ^1.0.0
  bar: ^1.0.0 # bar will not be unlocked by `pub upgrade foo --unlock-transitive`
```

We have agreed that is a relatively rare case, and we can live with it.

lib/src/command/upgrade.dart
```diff
@@ -70,6 +70,12 @@ class UpgradeCommand extends PubCommand {
       negatable: false,
     );
 
+    argParser.addFlag(
+      'unlock-transitive',
+      help: 'Also upgrades the transitive dependencies '
+          'of the listed [dependencies]',
+    );
```
jonasfj, line 77:
```suggestion
    argParser.addFlag(
      'unlock-transitive',
      help: 'Also upgrades the transitive dependencies '
          'of the listed [dependencies]',
      negatable: false,
    );
```

lib/src/command/upgrade.dart
```diff
@@ -101,11 +107,27 @@ class UpgradeCommand extends PubCommand {
...
+      _computePackagesToUpgrade();
+
   /// List of package names to upgrade, if empty then upgrade all packages.
   ///
   /// This allows the user to specify list of names that they want the
   /// upgrade command to affect.
-  List<String> get _packagesToUpgrade => argResults.rest;
+  Future<List<String>> _computePackagesToUpgrade() async {
+    if (argResults.flag('unlock-transitive')) {
+      final graph = await entrypoint.packageGraph;
```
jonasfj, line 119: What happens if there isn't a current resolution?

test/testdata/goldens/help_test/pub upgrade --help.txt
```diff
@@ -10,6 +10,7 @@ Usage: pub upgrade [dependencies...]
 -n, --dry-run            Report what dependencies would change but don't change any.
     --[no-]precompile    Precompile executables in immediate dependencies.
     --tighten            Updates lower bounds in pubspec.yaml to match the resolved version.
+    --[no-]transitive    Also upgrades the transitive dependencies of the listed [dependencies]
```
jonasfj, line 13:
```suggestion
    --transitive         Also upgrades the transitive dependencies of the listed [dependencies]
```

jonasfj (approved):
Please add the `negatable: false`, if we want the negatable variant I think we should call it `--lock-transitive` rather than `--nounlock-transitive` :rofl:

sigurdm:
> Please add the negatable: false, if we want the negatable variant I think we should call it --lock-transitive rather than --nounlock-transitive 🤣

I disagree. The negations are useful if we ever change the default. And in that case they should also be present from the beginning.

I think `--no-unlock-transitive` reads ok for a very rarely used option. (And we are specifically asking pub to **not unlock** the transitive dependencies. Not to lock them.

sigurdm:
Seems like I merged this by accident...

Will revert.

Issue #4406, opened 2024-06-04
"Error: The non-abstract class '_StaticTypeImpl' is missing implementations"


lukehutch:
I just switched to the `master` channel of Flutter, and a project that uses the analyzer as a library (Serverpod) reports this error:

```
$ ../../../.pub-cache/hosted/pub.dev/analyzer-6.5.0/lib/src/summary2/macro_application.dart:1261:7: Error: The non-abstract class '_StaticTypeImpl' is missing implementations for these members:
 - StaticType.asInstanceOf
Try to either
 - provide an implementation,
 - inherit an implementation from a superclass or mixin,
 - mark the class as abstract, or
 - provide a 'noSuchMethod' implementation.

class _StaticTypeImpl implements macro.StaticType {
      ^^^^^^^^^^^^^^^
/opt/flutter/bin/cache/dart-sdk/pkg/_macros/lib/src/api/introspection.dart:119:28: Context: 'StaticType.asInstanceOf' is defined here.
  Future<NamedStaticType?> asInstanceOf(TypeDeclaration declaration);
                           ^^^^^^^^^^^^
```

* Dart version 3.5.0 (build 3.5.0-214.0.dev)
* Flutter version 3.23.0-13.0.pre.80 on channel master
* Framework revision 5e448f4ce5 (10 hours ago), 2024-06-03 12:44:39 -0700

sigurdm:
Thinking again we can do something simpler without sacrificing the fast path.

The sdk should recognize that the snapshot is not valid, and return a specific error code.
The binstub then invokes `dart global run` which is recompiling the snapshot.

`dart global run` could (and should) validate that the  resolution is still good using the `ensurePubspecResolved` mechanism that `dart run` also uses.

sigurdm:
Hmm .. seems we are already validating the sdk constraints:

https://github.com/dart-lang/pub/blob/d86e3c979a3889fed61b68dae9f9156d0891704d/lib/src/global_packages.dart#L402

Not sure what is going wrong here then.

sigurdm:
I can reproduce this with:
```console
$ flutter channel stable && flutter upgrade --force
[...]
$ flutter --version
Flutter 3.24.3 • channel stable • https://github.com/flutter/flutter
Framework • revision 2663184aa7 (4 weeks ago) • 2024-09-11 16:27:48 -0500
Engine • revision 36335019a8
Tools • Dart 3.5.3 • DevTools 2.37.3
[...]

PR #4408, opened 2024-10-08
Require re-activation of global packages after minor sdk upgrade

sigurdm:
This would alleviate https://github.com/dart-lang/pub/issues/4406

But it is also a kind of regression in that it requires more activations....

I was first attempting to solve this with `ensureUpToDate`, but that logic does not really look into the actual version numbers of sdk packages, and as such would also not detect a new version of package:_macros.

Also it requires an update if the minor version is updated, leading to more or less the same behavior as this.


sigurdm:
We'll find another way.

PR #4419, opened 2024-11-01
Special case global-run message for updated sdk

sigurdm:
Also show the SolveReport, to clarify what in the resolution has changed

### DIFF
--- a/lib/src/global_packages.dart
+++ b/lib/src/global_packages.dart
@@ -28,6 +28,7 @@ import 'source/git.dart';
 import 'source/hosted.dart';
 import 'source/path.dart';
 import 'source/root.dart';
+import 'source/sdk.dart';
 import 'system_cache.dart';
 import 'utils.dart';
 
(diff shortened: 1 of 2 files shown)


szakarias (approved):

lib/src/global_packages.dart
```diff
@@ -474,14 +475,44 @@ Try reactivating the package.
...
+          result.availableVersions,
+          cache,
+          dryRun: true,
+          enforceLockfile: true,
+          quiet: false,
+        );
+        await report.show(summary: true);
+
         final sameVersions = entrypoint.lockFile.samePackageIds(newLockFile);
+
```
jonasfj, line 494: Please add a comment here explaining why we do this thing with creating a lockfile, and comparing the lcokfile, etc.

A comment that says: We need to do this because globally activated packages can have dependencies on SDK-packages, bla bla bla...

Because next time we look at this code, we won't remember why it is we do this resolution stuff here. When the method is recompile.

jonasfj (approved):

lib/src/global_packages.dart
sigurdm: Done

PR #4427, opened 2024-11-15
Add 'tag_pattern' feature to git dependencies

sigurdm:
Re: https://github.com/dart-lang/pub/issues/1250

With this proposal you can use a git repo as a versioned package repository.

For example write:
```
dependencies:
  foo:
    git:
      url: https://github.com/foo/foo
      tag_pattern: v{{version}}
    version: ^2.0.0
```

And all  commits in the git repo that are tagged according to `v{{version}}` will be inspected. If they contain a pubspec.yaml with `name: foo` and the correct version number they will be considered for resolution.

This feature is explicitly opt-in:
* any `git` dependency without a `tag_pattern` retain the existing behavior.
* The depending pubspec need to be at least at version 3.9.0


lib/src/source/git.dart
```diff
@@ -883,6 +979,7 @@ class ResolvedGitDescription extends ResolvedDescription {
     return {
       'url': url,
       'ref': description.ref,
+      'tag_pattern': description.tagPattern,
       'resolved-ref': resolvedRef,
```
jonasfj, line 983: Should we use `tag-pattern` for consistency?

lib/src/source/git.dart
```diff
@@ -72,6 +75,32 @@ class GitSource extends CachedSource {
             'string.');
       }
       path = descriptionPath;
+
+      // TODO: can we avoid relying on key presence?
+      if (description.containsKey('tag_pattern')) {
```
jonasfj, line 80: What about simply using `tags` and then support one or more patterns?
```yaml
dependencies:
  foo:
    git:
      url: https://foo.bar.git/repo
      tags: v*
```
and
```yaml
dependencies:
  foo:
    git:
      url: https://foo.bar.git/repo
      tags:
       - v*
       - foo-v*
```

or maybe not.

lib/src/source/git.dart
sigurdm: Hmm - this is `pubspec.lock`.
In pubspec.yaml we use `_` (at least in `pubspec_overrides` and `dev_dependencies`.

So when we copy the field from the pubspec.yaml to pubspec.lock it is confusing if we change the spelling.

At the same time it is confusing that we have `resolved-ref` in the pubspec.lock.

Can we find a single-word alternative?

lib/src/source/git.dart
sigurdm: Yeah - that might work

lib/src/source/git.dart
```diff
@@ -72,6 +75,32 @@ class GitSource extends CachedSource {
...
+          throw FormatException(
+            'Using `git: {tagPattern: }` is only supported with a minimum SDK '
+            'constraint of ${LanguageVersion.firstVersionWithTagPattern}.',
+          );
+        }
+        switch (description['tag_pattern']) {
+          case final String descriptionTagPattern:
+            tagPattern = descriptionTagPattern;
+          case null:
+            tagPattern = 'v*';
```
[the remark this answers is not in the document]
sigurdm: Hmm - considering this again I think we should actually not default to anything. I think it is probably better to be unopinionated here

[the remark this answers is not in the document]
sigurdm: Removed The default

lib/src/source/git.dart
sigurdm: I think we're going with `v{{version}}`, this corresponds to what we do on pub.dev for github actions.

lib/src/source.dart
```diff
@@ -81,7 +75,7 @@ abstract class Source {
   PackageRef parseRef(
     String name,
     Object? description, {
-    required Description containingDescription,
+    required ResolvedDescription containingDescription,
```
jonasfj, line 78: I kind of wish this migration could have been done in steps.

 1. Change `Description` -> `ResolvedDescription`
 2. Move `hasMultipleVersions`
 3. Adopt usage of `hasMultipleVersions` instead of `is HostedSource`
 4. ..
 5. Support for `tag-pattern`

But I can also see how it would cause a lot of conflicts to do that.

lib/src/source/git.dart
```diff
@@ -251,23 +291,35 @@ class GitSource extends CachedSource {
...
   }) async {
+    if (ref != null && tagPattern != null) {
+      fail('Cannot have both a `tagPattern` and a `ref`');
+    }
     final description = GitDescription(
       url: url,
       ref: ref,
       path: path,
       containingDir: relativeTo,
+      tagPattern: tagPattern, // TODO
```
jonasfj, line 304: TODO what?

lib/src/source/git.dart
```diff
@@ -943,3 +1074,33 @@ String _gitDirArg(String path) {
...
+
+/// Takes a [tagPattern] and returns a [RegExp] matching the relevant tags.
+///
+/// The tagPattern should contain '{{version}}' which will match a pub_semver
+/// version. The rest of the tagPattern is matched verbatim.
+RegExp compileTagPattern(String tagPattern) {
+  final match = tagPatternPattern.firstMatch(tagPattern);
+  if (match == null) {
+    throw const FormatException(
+      'The `tag_pattern` must contain "{{version}" '
```
jonasfj, line 1094:
```suggestion
      'The `tag_pattern` must contain "{{version}}" '
```

jonasfj (approved):

lib/src/source.dart
```diff
@@ -190,6 +184,11 @@ abstract class Source {
 /// with a version constraint.
 abstract class Description {
   Source get source;
+
+  /// Whether the source can choose between multiple versions of this
+  /// package during version solving.
+  bool get hasMultipleVersions;
```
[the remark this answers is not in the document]
sigurdm: I think it was mainly to not change anything existing, and yes making error messages better.

lib/src/source.dart
sigurdm: Done

sigurdm:
@jonasfj do you have any more feedback now after the reafctorings?

jonasfj (approved):

lib/src/source/git.dart
```diff
@@ -688,6 +766,35 @@ class GitSource extends CachedSource {
...
+      'tag',
+      '--list',
+      '--format',
+      // We can use space here, as it is not allowed in a git tag
+      // https://git-scm.com/docs/git-check-ref-format
+      '%(refname:lstrip=2) %(objectname)',
+    ], workingDir: path);
+    final lines = output.trim().split('\n');
+    final result = <TaggedVersion>[];
+    for (final line in lines) {
```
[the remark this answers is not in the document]
sigurdm: Yeah!

lib/src/source/git.dart
```diff
@@ -688,6 +766,35 @@ class GitSource extends CachedSource {
...
+      final parts = line.split(' ');
+      if (parts.length != 2) {
+        throw PackageNotFoundException('Bad output from `git tag --list`');
+      }
+      final match = compiledTagPattern.firstMatch(parts[0]);
+      if (match == null) continue;
+      final version = Version.parse(match[1]!);
```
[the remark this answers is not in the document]
sigurdm: Not sure it is much better having an assertion-error bubble up.. but ok

lib/src/source/git.dart
```diff
@@ -954,3 +1078,33 @@ String _gitDirArg(String path) {
       Platform.isWindows ? path.replaceAll('\\', '/') : path;
   return '--git-dir=$forwardSlashPath';
 }
+
+final tagPatternPattern = RegExp(r'^(.*){{version}}(.*)$');
```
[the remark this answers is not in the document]
sigurdm: Hmm not sure why anyone would do that... I'll disallow it

Issue #4445, opened 2024-11-26
Delete stray .dart_tool/package_config.json files between root and package dirs.


sigurdm:
In https://github.com/flutter/devtools/issues/8211 the situation was something like this

```
  /
   packages/
     pubspec.yaml
     .dart_tool/package_config.json
     package_a/
        pubspec.yaml
        .dart_tool/package_config.json
```

And after migration to workspaces with a pubspec.yaml at the root it looks like this:

```
  /
    pubspec.yaml
   .dart_tool/package_config.json
   packages/
     .dart_tool/package_config.json # <---- stray file
     package_a/
        pubspec.yaml
```
The stray file causes imports in package_a to not resolve correctly.
It is hard to discover because it is in a hidden directory, and typically gitignored.

We are already deleting .dart_tool/package_config.json in the folders of the workspace packages

We should also delete any .dart_tool/package_config.json files in folders  from the workspace package up to the root.

sigurdm:
This is not breaking behavior - as these stray .package_config.json files would shadow the package-resolution, making things not work.

PR #4446, opened 2024-11-26
Delete directories between workspace package and workspace root

sigurdm:
Fixes https://github.com/dart-lang/pub/issues/4445
Also makes resolution error if there are any pubspec.yaml's in such directories.

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -1335,13 +1335,27 @@ See https://dart.dev/go/sdk-constraint
   /// Remove any `pubspec.lock` or `.dart_tool/package_config.json` files in
   /// workspace packages that are not the root package.
   ///
+  /// Also remove from directories between the workspace package and the
+  /// workspace root, to prevent stray package configs from shadowing the shared
+  /// workspace package config.
+  ///
   /// This is to avoid surprises if a package is turned into a workspace member
   /// but still has an old package config or lockfile.
   void _removeStrayLockAndConfigFiles() {
+    final visited = <String>{};
     for (final package in workspaceRoot.transitiveWorkspace) {
       if (package.pubspec.resolution == Resolution.workspace) {
-        deleteEntry(p.join(package.dir, 'pubspec.lock'));
-        deleteEntry(p.join(package.dir, '.dart_tool', 'package_config.json'));
+        for (final dir in parentDirs(package.dir)) {
+          if (p.equals(dir, workspaceRoot.dir)) {
+            break;
+          }
+          if (!visited.add(dir)) {
+            // No reason to delete from the same directory twice.
+            break;
+          }
+          deleteEntry(p.join(dir, 'pubspec.lock'));
+          deleteEntry(p.join(dir, '.dart_tool', 'package_config.json'));
+        }
       }
     }
   }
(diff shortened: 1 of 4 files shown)


lib/src/entrypoint.dart
```diff
@@ -1335,13 +1335,27 @@ See https://dart.dev/go/sdk-constraint
...
+        for (final dir in parentDirs(package.dir)) {
+          if (p.equals(dir, workspaceRoot.dir)) {
+            break;
+          }
+          if (!visited.add(dir)) {
+            // No reason to delete from the same directory twice.
+            break;
+          }
+          deleteEntry(p.join(dir, 'pubspec.lock'));
+          deleteEntry(p.join(dir, '.dart_tool', 'package_config.json'));
```
jonasfj, line 1357: Should we write a warning message or something to logs.

Maybe something that includes a dart.dev/go/-link, so that we can explain why we deleted these folders.

If someone has workspace packages nested inside non-workspace packages inside a root, that could cause this, right? or probably it'll cause an error when trying to resolve the workspaces packages?

sigurdm: > Should we write a warning message or something to logs.

I'm not completely against this, but I think it might confuse more than it benefits...

> Maybe something that includes a dart.dev/go/-link,

I think that would be slightly overdoing it... But maybe we can fit a paragraph in somewhere on the workspaces documentation...

> so that we can explain why we deleted these folders.

We are deleting files, not folders.

> If someone has workspace packages nested inside non-workspace packages inside a root,

We should probably detect if someone attempts this (by detecting a pubspec.yaml), and give an error message. 
That scenario is not supported.

sigurdm: Ok - now displaying a warning and a doc reference.

Also added an error check if there are stray pubspec.yaml files between a package and the root.

lib/src/package.dart
```diff
@@ -500,6 +503,38 @@ Consider removing one of the overrides.
 Cannot override workspace packages.
 
 Package `$override` at `${overriddenWorkspacePackage.presentationDir}` is overridden in `${package.pubspecPath}`.
+''');
+      }
+    }
+  }
+
+  // Check for pubspec.yaml files between the root and any workspace package.
+  final visited = <String>{};
```
jonasfj, line 512: Just an idea, feel free to ignore.
```suggestion
  final visited = <String>{root.dir};
```
Then you won't need `p.equals(dir, root.dir)`, but then again, maybe this is just too smart. :rofl:

jonasfj (approved):

lib/src/package.dart
sigurdm: I like it!

lib/src/entrypoint.dart
```diff
@@ -1335,15 +1335,45 @@ See https://dart.dev/go/sdk-constraint
...
+          deleteIfPresent(
+            p.join(dir, '.dart_tool', 'package_config.json'),
+            'package config',
+          );
+        }
       }
     }
+    if (deletedAny) {
+      log.warning(
+        'See https://dart.dev/go/workspaces-no-inbetween-packages for details.',
```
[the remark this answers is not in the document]
sigurdm: How about: https://dart.dev/go/workspaces-stray-files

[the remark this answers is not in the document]
jonasfj: Or: https://dart.dev/go/workspaces-stray-package-configs 

`files` covers a LOT :rofl:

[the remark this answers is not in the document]
sigurdm: I guess we are also talking pubspec.lock and pubspec.yaml

jonasfj (approved):

PR #4472, opened 2024-12-13
Update `add` helping messages

### LINKED ISSUE #4470: On powershell `pub add` given examples fails when yaml contain double quotes
# Environment

- Dart version (run `dart --version`): Dart SDK version: 3.6.0 (stable) (Thu Dec 5 07:46:24 2024 -0800) on "windows_x64"

# Problem

As the help message of `pub add` tells:

```
Add packages with specific constraints or other sources by giving a descriptor
after a colon.

For example:
  * Add a path dependency:
    `dart pub add 'foo:{"path":"../foo"}'`
```

But when I run `dart pub add 'foundation:{"path":"../foundation"}'` it tells:
```console
The only allowed prefixes are "dev:" and "override:"
```

And without the single quotes:
```console
Because app depends on foundation from unknown source "path:../foundation", version solving failed.
```

Changing to `dart pub add foundation:{"path": "../foundation"}` (without the single quotes and extra space between `:` and `"..`) will solve the issue.

## Expected behavior

The command should execute without exceptions.

## Actual behavior

The command will fail to execute.

## Logs

```console
> dart pub add 'foundation:{"path":"../foundation"}'
Usage: pub add [options] [<section>:]<package>[:descriptor]
       [<section>:]<package2>[:descriptor] ...]
-h, --help               Print this usage information.
    --[no-]offline       Use cached packages instead of accessing the network.
-n, --dry-run            Report what dependencies would change but don't change
                         any.
    --[no-]precompile    Build executables in immediate dependencies.
-C, --directory=<dir>    Run this in the directory <dir>.

See https://dart.dev/tools/pub/cmd/pub-add for detailed documentation.

Process finished with exit code 64

> dart pub add foundation:{"path":"../foundation"}
Because app depends on foundation from unknown source "path:../foundation", version solving failed.

> dart pub add foundation:{"path": "../foundation"}
Resolving dependencies...
Downloading packages...
```

AlexV525:
Resolves #4470

---

### DIFF
--- a/lib/src/command/add.dart
+++ b/lib/src/command/add.dart
@@ -58,7 +58,7 @@ Make dependency overrides by prefixing with "override:".
 Add packages with specific constraints or other sources by giving a descriptor
 after a colon.
 
-For example:
+For example (follow the same format including spaces):
   * Add a hosted dependency at newest compatible stable version:
     `$topLevelProgram pub add foo`
   * Add a hosted dev dependency at newest compatible stable version:
@@ -67,19 +67,19 @@ For example:
     `$topLevelProgram pub add foo:^1.2.3`
   * Add multiple dependencies:
     `$topLevelProgram pub add foo dev:bar`
-  * Add a path dependency:
-    `$topLevelProgram pub add 'foo:{"path":"../foo"}'`
-  * Add a hosted dependency:
-    `$topLevelProgram pub add 'foo:{"hosted":"my-pub.dev"}'`
-  * Add an sdk dependency:
-    `$topLevelProgram pub add 'foo:{"sdk":"flutter"}'`
-  * Add a git dependency:
-    `$topLevelProgram pub add 'foo:{"git":"https://github.com/foo/foo"}'`
   * Add a dependency override:
-    `$topLevelProgram pub add 'override:foo:1.0.0'`
+    `$topLevelProgram pub add override:foo:1.0.0`
+  * Add a path dependency:
+    `$topLevelProgram pub add "foo:{path: ../foo}"`
+  * Add a hosted dependency:
+    `$topLevelProgram pub add "foo:{hosted: https://my-pub.dev}"`
+  * Add an sdk dependency:
+    `$topLevelProgram pub add "foo:{sdk: flutter}"`
+  * Add a git dependency:
+    `$topLevelProgram pub add "foo:{git: https://github.com/foo/foo}"`
   * Add a git dependency with a path and ref specified:
     `$topLevelProgram pub add \\
-      'foo:{"git":{"url":"../foo.git","ref":"<branch>","path":"<subdir>"}}'`''';
+      "foo:{git:{url: ../foo.git, ref: <branch>, path: <subdir>}}"`''';
 
   @override
   String get argumentsDescription =>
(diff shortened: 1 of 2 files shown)


lib/src/command/add.dart
```diff
@@ -68,18 +68,18 @@ For example:
   * Add multiple dependencies:
     `$topLevelProgram pub add foo dev:bar`
   * Add a path dependency:
-    `$topLevelProgram pub add 'foo:{"path":"../foo"}'`
+    `$topLevelProgram pub add foo:{"path":"../foo"}`
```
sigurdm, line 71: I think it is better to leave these quotes, otherwise bash will split the input if someone inserts a space.

lib/src/command/add.dart
```diff
@@ -415,7 +422,13 @@ Specify multiple sdk packages with descriptors.''');
     if (!packageNameRegExp.hasMatch(packageName)) {
       usageException('Not a valid package name: "$packageName"');
     }
-    final descriptor = match.namedGroup('descriptor');
+
+    // Explicitly adds a trailing space for commas and colons to help parsing
+    // a Yaml node. "key:value" will not be recognized as a valid entry.
+    final descriptor = match.namedGroup('descriptor')?.replaceAllMapped(
```
sigurdm, line 428: Hmm - I'm not sure I like this. How do we know the colon is not inside a quoted yaml string?

sigurdm (commented):
Personally I'd prefer just fixing the help-text. I think it is very hard to make regex updates inside yaml without breaking it.

sigurdm (commented):
Also note, that if you leave the quotes yaml should accept the the examples as given...

lib/src/command/add.dart
AlexV525: We don't, but we don't need to care about it IMO. Writing a completely bad syntax will fail as before.

lib/src/command/add.dart
AlexV525: Reverted. The previous pattern is now acceptable.

AlexV525:
> Personally I'd prefer just fixing the help-text. I think it is very hard to make regex updates inside yaml without breaking it.

Let's add more test cases when we meet them. :) There is nothing broken with the given pattern examples.

> Also note, that if you leave the quotes yaml should accept the the examples as given...

Are you saying that `'foo:{"path":"../foo"}'` should work? If yes then it still does after the change.

sigurdm:
> Are you saying that 'foo:{"path":"../foo"}' should work? If yes then it still does after the change.

I'm saying `'foo:{"hosted":"https://pub.dev"}'` should work. And all other yaml that I cannot imagine, but I don't want to process it up-front.

AlexV525:
> I'm saying `'foo:{"hosted":"https://pub.dev"}'` should work. And all other yaml that I cannot imagine, but I don't want to process it up-front.

Sure. All of the previous examples still work.

sigurdm:
Does it not insert a space into the host-name?

sigurdm:
> Does it not insert a space into the host url?

```
void main() {
  print('foo:{"hosted":"https://pub.dev"}'.replaceAllMapped( RegExp(r'([,:])(\S)'),
          (match) => '${match.group(1)} ${match.group(2)}'));
}
```
Prints
```
foo: {"hosted": "https: //pub.dev"}
```

AlexV525:
The request has been updated to change messages only.

sigurdm (approved):

sigurdm:
FYI: If you delete the golden file, and run the test it will be recreated correctly

AlexV525:
> FYI: If you delete the golden file, and run the test it will be recreated correctly

Yeah sorry about repeatedly commits. I thought it was just an easy update. 🥲

PR #4495, opened 2025-01-16
Cache canonicalized version of paths in path dependencies

sigurdm:
The `canonicalize` function is very slow, and the solver compares these a number of times.

This attempt of migrating the sdk to use workspaces https://dart-review.googlesource.com/c/sdk/+/397164?tab=comments has all third-party dependencies overridden as path-dependencies. 71 dependencies.

With current pub, this takes 2.245s to resolve on my machine.
After this fix it takes: 0.117s to resolve!

Also this avoid potential issues of inconsistencies if files change during resolution (although, still don't do that).

### DIFF
--- a/lib/src/source/path.dart
+++ b/lib/src/source/path.dart
@@ -270,6 +270,9 @@ class PathDescription extends Description {
   final String path;
   final bool relative;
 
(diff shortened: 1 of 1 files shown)


szakarias (approved):
LGTM

PR #4516, opened 2025-02-10
Use `sorted` from `package:collection` instead of custom util function

parlough:


lib/src/ascii_tree.dart
```diff
@@ -165,7 +166,7 @@ void _draw(
   if (name != null) _drawLine(buffer, prefix, isLast, name, depth <= 1);
 
   // Recurse to the children.
-  final childNames = ordered(children.keys);
+  final childNames = children.keys.sorted();
```
lrhn, line 169: Or just use `.toList()..sort();`.

It's the same thing, and you don't need an extra import.

_(Or you could have used a sorted map to begin with. If only we had one that was more efficent than `SplayTreeMap`.)_

lib/src/package.dart
```diff
@@ -109,7 +110,8 @@ class Package {
   List<String> get executablePaths {
     final binDir = p.join(dir, 'bin');
     if (!dirExists(binDir)) return <String>[];
-    return ordered(listDir(p.join(dir, 'bin'), includeDirs: false))
+    return listDir(p.join(dir, 'bin'), includeDirs: false)
+        .sorted()
         .where((executable) => p.extension(executable) == '.dart')
         .map((executable) => p.relative(executable, from: dir))
```
lrhn, line 116: Can sort after `where`  here (and `map`, unless it matters whether it's sorted as absolute or relative path, which I hope it shouldn't). Then there should be fewer elements to sort.

And one less `toList` to do:
```dart
  listDir(...).where(...).map(...).toList()..sort()
```
or 
```dart
    return [for (var executable in listDir(...))
      if (p.extension(executable) == '.dart')
        p.relative(executable, from: dir),
    ]..sort();
```

sigurdm (approved):

PR #4524, opened 2025-02-18
Write .dart_tool/package_graph.json when writing package_config.json

### LINKED ISSUE #4522: Proposal: .dart_tool/package_graph.json
## Motivation 
(building on https://github.com/dart-lang/native/issues/1968)

We have `.dart_tool/package_config.json` that relates a package name to its location on disk (and other properties, mainly the language version).

It is a nice boundary. Pub knows about dependencies, the rest of the sdk only knows about package-names and file-urls.

However, sometimes the dependency relationship between packages is important. For example for deciding the order in which to run build hooks, or to ascertain which packages are in the transitive dependency closure of another package, to not bundle assets from dev_dependencies, or unrelated packages in the same pub workspace.

Currently this information can be gathered by tracking down all pubspec.yaml files of packages in the package_config.json, parsing them and computing this.

If one is willing to pay the cost of a process invocation you can use `dart pub deps` to do this. 

This is redundant work, as `pub get` already gathers all this information, reading and parsing all the yaml.

## Proposal

I propose that pub whenever writing `.dart_tool/package_config.json` also writes an adjacent file `.dart_tool/package_graph.json`.

It should contain something reminiscent of a subset of the output of `dart pub deps`

```json
{
  "root": "<root package name>" // The root package of the pub workspace
  "packages": [
    {
      "name": "my_app",
      "dependencies": ["foo", "bar", ...], // The names in the `dependencies` section of the pubspec.yaml of my_app
      "devDependencies": ["test", "build_runner"] // Only present for workspace root packages
    },
    {
      "name": "foo",
      "dependencies": ["bar"],
    }, ... // more packages
  ]
}
```

## Considerations

* *dependency_overrides* do not affect the set of dependencies (a root package can override a transitive dependency, doesn't mean it itself depends on it).
* Instead of using names, one could use *indices* when enumerating the dependencies. However that would make the file hard to read for people, while only giving minor efficiency benefits. If in the future we want to have multiple versions of the same package, we would anyway need to find a way of naming them individually.
* We have earlier considered put this information in `.dart_tool/package_config.json`, but the main consumers of that file (compiler and analyzer) do not need this information. And thus it would bloat that file.
* `pub deps` also includes more information for each package, such as version number, and dependency source (git, path, hosted). These are kept out of this proposal to keep it somewhat minimal. The file can be extended with these in the future if needed.

sigurdm:
Fixes #4522

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -302,6 +302,10 @@ See $workspacesDocUrl for more information.''',
     p.normalize(p.join(workspaceRoot.dir, '.dart_tool', 'package_config.json')),
   );
 
+  late final String packageGraphPath = p.relative(
+    p.normalize(p.join(workspaceRoot.dir, '.dart_tool', 'package_graph.json')),
+  );
+
   /// The path to the entrypoint workspace's lockfile.
   String get lockFilePath =>
       p.normalize(p.join(workspaceRoot.dir, 'pubspec.lock'));
@@ -396,10 +400,12 @@ See $workspacesDocUrl for more information.''',
   /// Writes the .dart_tool/package_config.json file and workspace references to
   /// it.
   ///
+  /// Also writes the .dart_tool.package_graph.json file.
+  ///
   /// If the workspace is non-trivial: For each package in the workspace write:
   /// `.dart_tool/pub/workspace_ref.json` with a pointer to the workspace root
   /// package dir.
-  Future<void> writePackageConfigFile() async {
+  Future<void> writePackageConfigFiles() async {
     ensureDir(p.dirname(packageConfigPath));
     writeTextFile(
       packageConfigPath,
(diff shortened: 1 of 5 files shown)


lib/src/entrypoint.dart
```diff
@@ -302,6 +302,10 @@ See $workspacesDocUrl for more information.''',
     p.normalize(p.join(workspaceRoot.dir, '.dart_tool', 'package_config.json')),
   );
 
+  late final String packageGraphPath = p.relative(
+    p.normalize(p.join(workspaceRoot.dir, '.dart_tool', 'package_graph.json')),
```
szakarias, line 306: I would wish for the word 'dependency' to be there somehow, but not blocking on it.

szakarias (approved):
LGTM

test/package_graph_file_test.dart
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
```
szakarias, line 1: update year

sigurdm: Done

jonasfj (approved):

PR #4525, opened 2025-02-18
Migrate to language 3.7, reformat, upgrade

sigurdm:
* Migrate to language version 3.7
* Apply dart format to get tall mode
* Reflow some comments and strings that ended up above 80 chars 
* Upgrade all dependencies
* Removes `require_trailing_commas` lint, as now the formatter should be enough.


szakarias (approved):
LGTM

Issue #4532, opened 2025-03-04
Publish PubHttpResponseException should check the ContentType for user error reporting


BButner:
# Environment

- Dart version (run `dart --version`): Dart SDK version: 3.7.0 (stable) (Wed Feb 5 04:53:58 2025 -0800) on "macos_arm64"
- OS kind and version (e.g. "Windows 10, version 1809" or "macOS 12.4"): macOS 15.3.1
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

When [self-hosting a private pub repository](https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md#hosted-pub-repository-specification-version-2), the error message reported back from a package publish is not being displayed to the user.

[According to the spec](https://github.com/dart-lang/pub/blob/e44a3ebd4d48e7cd53e8eafadaa0efed96d4c389/doc/repository-spec-v2.md?plain=1#L378C1-L388C4) the server should return a 400 Bad Request with a `Content-Type` of `application/vnd.pub.v2+json`. The body content of that response leads the implementor to believe that the message will be displayed in the CLI, however that is not the case. The user will only receive the generic "Failed to upload the package." message.

Looking at [lib/src/command/lish.dart](https://github.com/dart-lang/pub/blob/e44a3ebd4d48e7cd53e8eafadaa0efed96d4c389/lib/src/command/lish.dart#L219) it's checking the URL based on the response, and seems to ignore the incoming Content-Type of the response and instead assumes it's XML. Thus, the v2 spec is showing the Content-Type should be JSON based however the exception catch is assuming it's XML based.

This can be worked around by simple returning a 400 with the content of that being XML, however that does not resolve the underlying incorrect handling of the response.

sigurdm:
The flow of an upload is:

```
(initiate upload) GET <hosted-url>/api/packages/versions/new ->
  {
  "url": "<multipart-upload-url>",
   ...
}
(the actual upload): POST <multipart-upload-url> -> 
    HTTP/1.1 204 No Content
    Location: <finalize-upload-url>
(finalize upload): GET <finalize-upload-url> ->
  HTTP/1.1 200 Ok
  Content-Type: application/vnd.pub.v2+json
  {
    "success": {
      "message": "<message>",
    },
  }
  OR
  HTTP/1.1 400 Bad Request
  Content-Type: application/vnd.pub.v2+json
  {
    "error": {
      "code": "<code>",
      "message": "<message>",
    },
  }
```

[...]

PR #4537, opened 2025-03-07
Handle json error message response to upload

sigurdm:
#4532


sigurdm:
Not sure we actually need this.

Issue #4536, opened 2025-03-06
Running globally activated package attempts to upgrade a dependency and ignores lockfile


btrautmann:
# Environment

- Dart version (run `dart --version`): `Dart SDK version: 3.6.2 (stable) (Wed Jan 29 01:20:39 2025 -0800) on "macos_arm64"`
- OS kind and version (e.g. "Windows 10, version 1809" or "macOS 12.4"): macOS 15.3.1
- Are you using the Chinese community mirror or a corporate firewall? No

# Problem

I have a project within which a `dart pub get` is successful and prints the following:

```console
Resolving dependencies... 
Downloading packages... 
  _fe_analyzer_shared 76.0.0 (80.0.0 available)
  analyzer 6.11.0 (7.3.0 available)
  built_value 8.9.4 (8.9.5 available)
  dart_style 2.3.8 (3.0.1 available)
  js 0.7.1 (0.7.2 available)
! taconic_annotation 0.0.1 from path ../taconic_annotation (overridden in ./pubspec_overrides.yaml)
```

I can activate it using `dart pub global activate --source="path" .` successfully which prints:

```console
Resolving dependencies in `taconic`... 
Downloading packages... 
  _fe_analyzer_shared 76.0.0 (80.0.0 available)
  analyzer 6.11.0 (7.3.0 available)
  built_value 8.9.4 (8.9.5 available)
  dart_style 2.3.8 (3.0.1 available)
  js 0.7.1 (0.7.2 available)
! taconic_annotation 0.0.1 from path taconic_annotation (overridden in taconic/pubspec_overrides.yaml)
Got dependencies in `taconic`!
5 packages have newer versions incompatible with dependency constraints.
Try `dart pub outdated` for more information.
Package taconic is currently active at path "/Users/brandontrautmann/Development/mobile/flutter/taconic".
Installed executable taconic.
Activated taconic 0.0.1 at path "/Users/brandontrautmann/Development/mobile/flutter/taconic".
```

However, when I invoke that executable, I get the following error:

```console
Downloading packages... . 
  _fe_analyzer_shared 76.0.0 (80.0.0 available)
  analyzer 6.11.0 (7.3.0 available)
> built_value 8.9.5 (was 8.9.4)
  dart_style 2.3.8 (3.0.1 available)
  js 0.7.1 (0.7.2 available)
[...]

PR #4541, opened 2025-03-14
Fix compilation of global path activated packages

sigurdm:
#4536 

We were not passing the lockfile to the resolution causing a different resolution, which failed the --enforce-lockfile semantics.


szakarias (approved):
LGTM

PR #4546, opened 2025-03-19
Add option to use rules from .gitignore even if .pubignore present

### LINKED ISSUE #4545: Optionally respect the .gitignore even when a .pubignore is present
As a package maintainer, I don't want to publish some larger files (for example, some assets for the example app) or some configuration files (which are present at the root, such as firebase.json). These need to remain checked-in. Therefore, I use a .pubignore.

However, the situation could arise where I also want to not check something into Git, which is normally done with .gitignore. If I'm using .pubignore already, I have to put the rule into both - so both Git respects it, and pub as well.

Therefore, it would be a nice QOL feature to be able to tell pub to also respect the .gitignore, even if I'm using .pubignore.

JaffaKetchup:
Fixes #4545.

When a .pubignore is present, it is read. If it starts with, on an independent line, the phrase:

> #@ include .gitignore

(ignoring whitespace), then rules from gitignore are also considered.

I've not added tests, and the issue hasn't been discussed. But since it's an easy change to make, it's not a waste of time if not accepted. I can add tests if needed, but might need some guidance on how best to do it.

---

<details>

- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change).

</details>

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -239,6 +239,12 @@ See $workspacesDocUrl for more information.
     '/pubspec_overrides.yaml',
   ];
 
+  static final _includeGitignoreIndicator = RegExp(
+    // We look for "#* include .gitignore" at the start of a .pubignore
+    // We don't necessarily care about whitespace
+    r'^\s*#@\s*include\s*\.gitignore\s*?$',
+  );
+
   /// Returns a list of files that are considered to be part of this package.
   ///
   /// If [beneath] is passed, this will only return files beneath that path,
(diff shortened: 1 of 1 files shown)


sigurdm (commented):
Thanks for looking into this.

I like the idea - @jonasfj are you in on this?

sigurdm:
We agreed not to do this.

We don't want to extend the .gitignore format.

Sorry for the long delay.

PR #4548, opened 2025-03-21
Don't emit timestamp in .dart_tool/package_config.json

sigurdm:
We don't rely on the timestamp anywhere I know of

It gets in the way of dependency tracking

The field is optional according to https://github.com/dart-lang/language/blob/main/accepted/2.8/language-versioning/package-config-file-v2.md#:~:text=The%20following%20optional%20properties%20of%20the%20top%2Dlevel%20object%20allows%20the%20generator%20to%20specify%20metadata%20about%20the%20file.%20They%20can%20all%20be%20omitted%2C%20but%20if%20they%20are%20present%2C%20they%20should%20have%20the%20expected%20format.

Fixes https://github.com/dart-lang/sdk/issues/60371

### DIFF
--- a/lib/src/package_config.dart
+++ b/lib/src/package_config.dart
@@ -19,11 +19,6 @@ class PackageConfig {
   /// Packages configured.
   List<PackageConfigEntry> packages;
 
-  /// Date-time the `.dart_tool/package_config.json` file was generated.
-  ///
-  /// `null` if not given.
-  DateTime? generated;
-
   /// Tool that generated the `.dart_tool/package_config.json` file.
   ///
   /// For `pub` this is always `'pub'`.
@@ -46,7 +41,6 @@ class PackageConfig {
   PackageConfig({
     required this.configVersion,
     required this.packages,
-    this.generated,
     this.generator,
     this.generatorVersion,
     Map<String, dynamic>? additionalProperties,
@@ -98,16 +92,6 @@ class PackageConfig {
       packages.add(PackageConfigEntry.fromJson(entry as Object));
     }
 
-    // Read the 'generated' property
-    DateTime? generated;
-    final generatedRaw = root['generated'];
-    if (generatedRaw != null) {
-      if (generatedRaw is! String) {
-        throwFormatException('generated', 'must be a string, if given');
-      }
-      generated = DateTime.parse(generatedRaw);
-    }
-
     // Read the 'generator' property
     final generator = root['generator'];
     if (generator is! String?) {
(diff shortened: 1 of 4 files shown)


lrhn:
I think the original idea was that Pub wouldn't write a new file unless it was different.
That's probably not worth the effort of recognizing, and the "last modified" time is tracked by the file system anyway.
(Let's hope the dependency tracking uses contents, not file system times.)

lib/src/package_config.dart
```diff
@@ -19,11 +19,6 @@ class PackageConfig {
   /// Packages configured.
   List<PackageConfigEntry> packages;
 
-  /// Date-time the `.dart_tool/package_config.json` file was generated.
-  ///
-  /// `null` if not given.
-  DateTime? generated;
```
lrhn, line 25: I would consider not removing the option entirely, just change the default to not write an entry if not given a `DateTime` instead of using `DateTime.now()` as fallback ...

lib/src/package_config.dart
```diff
@@ -158,7 +141,6 @@ class PackageConfig {
   Map<String, Object?> toJson() => {
     'configVersion': configVersion,
     'packages': packages.map((p) => p.toJson()).toList(),
-    'generated': generated?.toUtc().toIso8601String(),
```
lrhn, line 161: ... which would make this:
```dart
      if (generated != null) 'generated': generated.toUtc().toIso8601String(),
```

But, on the other hand, if it's a feature in the Pub code that Pub never uses, it's basically dead code, so it'd be better to get rid of it.

(Do consider using `package:package_config` instead of your own classes. If we change things in the format in the future, we'll only need to fix one place. Or tell me why it's not usable enough.)

sigurdm:
Current behavior 
* we have a fast-path api that checks timestamps of pubspec.yaml and package_config.json, and doesn't touch anything if not needed. That is what's used by `dart run` and `dart test`.
* Running `pub get` will always result in a resolution, and rewriting of package_config.json.

Maybe we should indeed check if the new package_config.json has no changes, and not update it. But it is not clear to me who this timestamp is for - is it for file systems where you cannot rely on file timestamps?

lrhn (approved):

lib/src/package_config.dart
sigurdm: I remember trying to use package_config but being frustrated with something. But I cannot remember what it was.

Maybe it was that I could not get to the raw rootUri property 

https://pub.dev/documentation/package_config/latest/package_config_types/Package/root.html is always resolved

Also I cannot get at the uri that a package config is resolving its packages against. (the baseUri)

Not sure anymore...

sigurdm:
Closing in favor of https://github.com/dart-lang/pub/pull/4549

PR #4549, opened 2025-03-21
Don't rewrite package_config and package_graph if unchanged

sigurdm:
Based on textual equality

Also removes the timestamp from package_config.json

Fixes https://github.com/dart-lang/sdk/issues/60371

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -403,9 +403,9 @@ See $workspacesDocUrl for more information.''',
   Future<void> writePackageConfigFiles() async {
     ensureDir(p.dirname(packageConfigPath));
 
-    _writeJsonIfDifferent(
+    _writeIfDifferent(
       packageConfigPath,
-      await _packageConfigJson(
+      await _packageConfigFile(
         cache,
         entrypointSdkConstraint:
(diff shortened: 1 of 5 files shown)


lib/src/entrypoint.dart
```diff
@@ -392,25 +392,31 @@ See $workspacesDocUrl for more information.''',
...
-      await _packageConfigFile(
+      await _packageConfigJson(
         cache,
         entrypointSdkConstraint:
             workspaceRoot
                 .pubspec
                 .sdkConstraints[sdk.identifier]
                 ?.effectiveConstraint,
       ),
+      ignoredProperties: ['generated'],
```
jonasfj, line 416:
```suggestion
      // Ignore the "generated" timestamp, if it's the only thing that changed.
      ignoredProperties: ['generated'],
```

jonasfj (approved):

sigurdm:
Also removed the timestamp in the config file, and changed the rewrite logic to be based on string equality.

sigurdm:
@jonasfj PTAL

jonasfj (approved):

dcharkes:
Awesome!!!! ❤️

PR #4576, opened 2025-05-13
Move hasMultipleVersions to Description

sigurdm:
Extracted from #4427 . With that PR a git dependency can have multiple versions. But only if it is a "tag_pattern".

### DIFF
--- a/lib/src/command/add.dart
+++ b/lib/src/command/add.dart
@@ -118,6 +118,11 @@ For example (follow the same format including spaces):
       hide: true,
     );
     argParser.addOption(
+      'git-tag-pattern',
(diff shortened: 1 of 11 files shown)


lib/src/command/add.dart
```diff
@@ -117,6 +117,11 @@ For example (follow the same format including spaces):
       help: 'Path of git package in repository',
       hide: true,
     );
+    argParser.addOption(
+      'git-tag-pattern',
+      help: 'The tag-pattern to search for versions in repository',
+      hide: true,
+    );
```
jonasfj, line 124: This bit was probably a bit early :D

jonasfj (approved):

PR #4578, opened 2025-05-14
feat: allow fixed flutter version constraints.

eseidel:
Previously there was code to treat x.y.z as >=x.y.z, specifically for the flutter sdk version constratint.  This change removes that which allows app authors to specify an upper bound on the flutter version constraint for their projects.

This isn't actually ready for landing, because I haven't yet been able to successfully get a clean run of pub tests locally on my mac (maybe they aren't expected to run cleanly on a mac?) even without any changes to pub.

### DIFF
--- a/lib/src/lock_file.dart
+++ b/lib/src/lock_file.dart
@@ -147,9 +147,6 @@ class LockFile {
               originalConstraint,
               defaultUpperBoundConstraint: null,
             ),
-            'flutter' => SdkConstraint.interpretFlutterSdkConstraint(
-              originalConstraint,
-            ),
             _ => SdkConstraint(originalConstraint),
           };
         },
--- a/lib/src/pubspec.dart
+++ b/lib/src/pubspec.dart
@@ -251,10 +251,7 @@ environment:
           _packageName,
           _FileType.pubspec,
         );
-        constraints[name] =
-            name == 'flutter'
-                ? SdkConstraint.interpretFlutterSdkConstraint(constraint)
-                : SdkConstraint(constraint);
+        constraints[name] = SdkConstraint(constraint);
       });
     }
     return constraints;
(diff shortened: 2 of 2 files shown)


njwandroid:
very exciting @eseidel  thank you so much

eseidel:
Yeah, even following all instructions (installing a fresh dart-only sdk, etc) I still get 7 failures locally (even without this change) making it very difficult to work on pub.
[failures.txt](https://github.com/user-attachments/files/20217198/failures.txt)

sigurdm:
Sorry about the failures. We have CI running on mac where all tests are passing, but I usually develop on linux.

I have just run all tests successfully on my mac.

It seems from the stack traces that there is some problem with running freshly compiled subprocesses. Do you have any protection in place that only allows you to run recognized binaries?

sigurdm:
About the PR itself. The special casing is there for a reason - we don't want packages to claim compatibility with some flutter , and then flutter publishes a new major version (for marketing purposes) and then you cannot use the old packages.

I personally was not too fond of this hack it back when it was put in place, but as long as Flutter insists to use its major version for marketing, and not for semantic versioning we need something like this hack in place for the ecosystem to continue working.

We **could** consider changing again slightly how we interpret flutter sdk constraints. Perhaps

* We in general ignore the upper bound of the constraint
* But if the constraint is a single version we respect both the upper and lower bound (ie. we constrain to just that version).

Or another plausible solution would be

* We in general respect upper and lower bound of the constraint
* Unless it is a '^x.y.z' constraint, in which case we just consider the lower bound.

Each of these have the potential to be breaking changes. And we would need to run a feasibility analysis across all current packages on pub.dev to see how many packages are affected.

cc @mit-mit @jonasfj

eseidel:
re: freshly compiled subprocesses.  Not to my knowledge.  This is a mostly-unaltered apple M3.

sigurdm:
Can you repeat here some of the reasons you want to pin the flutter sdk version?

Perhaps we need a separate mechanism for that.

njwandroid:
I can try to give specific use case later but bottom line every single
feature every single Branch should always have the same flutter version for
1) developer's local flutter install
2) qa or Qe engineera build whether from source or through app distribution
such as firebase
3) the build servers that deploy

It should never be possible for a human to accidentally upgrade flutter on
their local machine and still be able to run the project

It should never be possible for a build server that was accident on a
configured to use stable channel instead of a pinned version to upgrading
and build on a newer version of flutter

It should absolutely not matter for this purpose whether it was a major or
minor or hot fix flutter change because many organizations want and need to
make sure it is absolutely the same .

mit-mit:
Hi @eseidel @sigurdm can we close this now that we have https://github.com/dart-lang/pub/pull/4595, or is there still a gap here?

sigurdm:
I think we should be good! Closing.

PR #4579, opened 2025-05-15
Clarify relationship between Description and ResolvedDescription

sigurdm:


lib/src/source.dart
```diff
@@ -177,11 +177,17 @@ abstract class Source {
...
+/// For a git package this would be the repo url and a ref and a path inside the
+/// repo.
+///
+/// For a path package it is the path.
 ///
 /// This is the information that goes into a `pubspec.yaml` dependency together
 /// with a version constraint.
+///
+/// After resolution we might know more about the specifics of the package that
+/// pins the content down (sucb as its content-hash or git commit id) this is
```
jonasfj, line 189:
```suggestion
/// pins the content down (such as its content-hash or git commit id) this is
```

jonasfj (approved):

PR #4595, opened 2025-06-03
Respect upper bound of Flutter constraint in root packages after 3.9

sigurdm:
Solves https://github.com/flutter/flutter/issues/95472 (at least partially)

By making this depend on language version we avoid a breaking change

### DIFF
--- a/lib/src/language_version.dart
+++ b/lib/src/language_version.dart
@@ -70,6 +70,9 @@ class LanguageVersion implements Comparable<LanguageVersion> {
   bool get forbidsUnknownDescriptionKeys =>
       this >= firstVersionForbidingUnknownDescriptionKeys;
 
+  bool get respectsFlutterBoundInRoots =>
+      this >= firstVersionRespectingFlutterBoundInRoots;
+
   /// Minimum language version at which short hosted syntax is supported.
   ///
   /// This allows `hosted` dependencies to be expressed as:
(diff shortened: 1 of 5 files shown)


sigurdm:
cc @mit-mit

sigurdm:
@mit-mit whom in the flutter world should we inform, and should we make a breaking change announcement?

mit-mit:
Yeah, I think we should do both a [flutter breaking change](https://docs.flutter.dev/release/breaking-changes) and [dart breaking change](https://dart.dev/resources/breaking-changes).

Also, should we consider a flag (e.g. `dart|flutter pub get --ignore-upper`) for those that need the current behavoir?

sigurdm:
> Also, should we consider a flag (e.g. dart|flutter pub get --ignore-upper) for those that need the current behavior?

Probably better with an env-var. It is hard to set the flag everywhere we resolve.

Even then it might be hard - `dart run` or the IDE might just resolve, it is hard to set the environment or add command-line arguments for the IDE.

mit-mit:
I don't think folks are going to need it for everyday use (the breaking change process might help us understand  if that is indeed the case), so I think that a flag on `get` and `run` for the occasional CI breakage might be sufficient?

sigurdm:
Might be.. I still prefer it as an env-var, but I can live with both...

sigurdm:
I realized by introducing the new interpretation based on language-version we can avoid making the change breaking!

PR #4604, opened 2025-06-12
Remove special casing around injected flutter_gen package

sigurdm:
After https://github.com/flutter/flutter/issues/102983 these are no longer needed.


szakarias (approved):
Nice

PR #4620, opened 2025-07-08
Fix detection of root packages in PackageGraph.transitiveDependencies

sigurdm:
The current behavior is broken. It looks at the wrong package, and does not consider the entire workspace.

### DIFF
--- a/lib/src/package_graph.dart
+++ b/lib/src/package_graph.dart
@@ -46,12 +46,11 @@ class PackageGraph {
 
   /// Returns all transitive dependencies of [package].
   ///
-  /// For the entrypoint this returns all packages in [packages], which includes
-  /// dev and override. For any other package, it ignores dev and override
-  /// dependencies.
+  /// For the root packages, this will explore the dev_dependencies if
+  /// [followDevDependenciesFromRoots] is true.
   Set<Package> transitiveDependencies(
     String package, {
-    required bool followDevDependenciesFromRoot,
+    required bool followDevDependenciesFromRoots,
   }) {
     final result = <Package>{};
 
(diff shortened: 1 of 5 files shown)


test/package_graph_test.dart
```diff
@@ -0,0 +1,94 @@
...
+      p.join(d.sandbox, appPath),
+      SystemCache(rootDir: p.join(d.sandbox, cachePath)),
+    );
+    final graph = await entrypoint.packageGraph;
+
+    expect(
+      graph
+          .transitiveDependencies('foo', followDevDependenciesFromRoots: true)
+          .map((p) => p.name),
+      {'foo', 'transitive'},
```
dcharkes, line 67: Shouldn't this contain `transitive_dev_dep` because it is a dev dependency of `foo`. And `foo` is the root of the search.

lib/src/package_graph.dart
```diff
@@ -46,12 +46,11 @@ class PackageGraph {
...
   ///
-  /// For the entrypoint this returns all packages in [packages], which includes
-  /// dev and override. For any other package, it ignores dev and override
-  /// dependencies.
+  /// For the root packages, this will explore the dev_dependencies if
+  /// [followDevDependenciesFromRoots] is true.
   Set<Package> transitiveDependencies(
     String package, {
-    required bool followDevDependenciesFromRoot,
+    required bool followDevDependenciesFromRoots,
```
dcharkes, line 53: Isn't `package` the root? Shouldn't `followDevDependenciesFromRoot` then be singular?

test/package_graph_test.dart
sigurdm: That is exactly the bug we are fixing.
We don't ever want to look at dev_dependencies of our dependencies (outside the workspace).
So here the **root** is not the root of the search, but the root of the workspace.

lib/src/package_graph.dart
sigurdm: No, it is the workspace root packages.

sigurdm: Ah - after talking I realize we need both conditions.
We only want to follow dev_dependencies from the `package` itself and only if it is a root package.

thanks!

test/package_graph_test.dart
sigurdm: I made the tests more specific.

PR #4624, opened 2025-07-18
Add a 'dart pub cache path' command

gspencergoog:
## Description

This adds a `path` subcommand to the `dart pub cache` command which simply prints the effective pub cache path.

This is useful for determining the cache directory in effect when building CI rules.

For instance a Github action could now be created that looks like this:

```yaml
      - name: Get pub cache path
        id: pub-cache
        run: echo "dir=$(dart pub cache path)" >> $GITHUB_OUTPUT
      - name: Cache Pub dependencies
        uses: actions/cache@v4
        with:
          path: ${{ steps.pub-cache.outputs.dir }}
          key: ${{ runner.os }}-pub-${{ hashFiles('**/pubspec.lock') }}
          restore-keys: ${{ runner.os }}-pub-
```

Otherwise, you have to know what the cache directory resolution scheme is on each platform, and the PUB_CACHE env var, and the path is platform specific.

It's also useful if an AI agent wants to know where the cache is.

## Tests
 - Added a test.


natebosch:
I agree that this is worth landing. I'll let the pub team review.

gspencergoog:
@jonasfj  Did you have bandwidth to review this?

natebosch:
Oops, looks like Jonas is out, I should have checked when assigning originally.

sigurdm:
We include the cache path in .dart_tool/package_config.json . I think it is better to read it from there.

It is under the "pubCache" key.

gspencergoog:
The reason I wanted to add this command is that figuring out the value requires complicated and obscure knowledge.  For instance, even though you've mentioned that it's stored in a file, I still don't know if setting PUB_CACHE overrides that.  And what do I do if the file hasn't been created yet? Also, having it in a file means I need to have a JSON parser and parse the key out of the file. This way, I get the value with one simple documented command.

With this documented command, it's easy to get a canonical value, that I know is the one being used by Dart, that is using the same logic as Dart does, so if it ever changes, I don't need to do anything.

sigurdm:
In the CI case I think it is better to set the cache to a fixed path by hard coding PUB_CACHE.

gspencergoog:
> In the CI case I think it is better to set the cache to a fixed path by hard coding PUB_CACHE.

If I set the PUB_CACHE to something when the CI docker image doesn't, or to something different, then I'm going to have to download more each time I update the cached directory. Sure, I can probably predict that the docker image already has PUB_CACHE set and use that, or that it's in the HOME dir, but that's just extra logic I need to write each time.

Also, CI isn't the only use case. Any time I have a script that wants to know where the cache is, I need to replicate the "is PUB_CACHE set? No, is it in package_config.json? Is it in HOME?" logic.

And the AI use case is also made simpler by having a single command to execute.

Issue #1447, opened 2016-09-04
Add the ability to check if "pub get" needs running


DanTup:
When a user opens a Dart project in Dart Code I'd like to inform them if they need to run `pub get` (with an easy button to do so) to avoid lots of analysis errors opening freshy-cloned/updated projects.

Currently there doesn't seem to be a way to detect if local packages are out of date in order to prompt the user (we could try checking timestamps on `.packages` and `pubspec.yaml` but it's not very reliable) so it'd be nice if something could be added.

(I'm requesting this in `pub`, but I don't know if it might make sense elsewhere; feel free to request I re-raise this in `sdk` or somewhere if it would be more appropriate there).

zoechi:
`den fetch` should do what you want https://pub.dartlang.org/packages/den

DanTup:
> `den fetch` should do what you want

I'm not sure this would work - we can't globally install pub packages on the end-users machine and even if you can write a dart app using this library that reads another projects pubspec, our extension is written in TypeScript and it'd be a bit clunky shipping a dart app inside our TypeScript app (long-term we'd like to be written in Dart, but that's way off and might not be feasible).

> But the existence of newer versions doesn't mean they are compatible with your dependencies.

We're not planning on upgrading anything - this is a request for detecting when `pub get` is *required*, eg. you've cloned a repo that has never had "pub run" called or the user has modified the pubspec file (eg. changing the version of a package) and the local packages are incorrect. In this case, a project will report many errors/warnings where the fix may be just running `pub run` but it's not obvious to the user from the errors/warnings.

An IDE should be able to detect when the local `.packages` file (or `packages` symlinks) are stale so they can suggest the correct fix rather than confuse the user. In .NET if you open a project missing NuGets you're either prompted to download them or it silently happens before build; it's really painless. We'd like to achieve something similar.

sigurdm:
Could this be implemented as
```
$dart pub get --dry-run
```
With an exit-code of 0 meaning everything is up-to-date?

DanTup:
I think so. For now I'm just checking timestamps on `.packages`/`pubspec.yaml`/etc.

Some things to consider:

- The folder opened in the editor might have multiple projects so we'd like to detect this as quickly as possible for them all - would it be safe to call this for them all concurrently? (I know there have been issues with concurrent pub calls in the past, though I suspect that was related to writing to disk that `--dry-run` might not need to do?)
[...]

PR #4655, opened 2025-08-28
command `pub check-resolution-up-to-date`

sigurdm:
This is the same heuristic check that `dart run` uses to ensure that the resolution is usable.

Attempt to fix https://github.com/dart-lang/pub/issues/1447

The command is hidden for now, we are unsure how we want to recommend end-users use it. But it will be available for vscode and others to use.


sigurdm:
cc @DanTup

DanTup:
Neat, thank you! :-)

What's the behaviour for this in a Pub Workspace - is the check for the overall workspace, or just the package in which it's invoked for?

sigurdm:
The check is for the overall workspace resolution, in a workspace everything is always resolved together.

I'll add a test demonstrating this.

DanTup:
> The check is for the overall workspace resolution

Perfect, thank you!

sigurdm:
@szakarias suggested we make this a hidden command instead.
It is not clear we want end-users to use this in general, or be confused about this option.

szakarias (commented):
Doc comments.

PR #4680, opened 2025-09-26
Workspaces glob support

### LINKED ISSUE #4391: Globs support for defining packages in workspaces
I think this is already planned, but I'm opening an issue so that it can be tracked.

Adding globs support will:
 - reduce a lot of user mistakes (like forgetting to add new packages to the workspace)
 - make migrations from the current melos version to the future melos version where packages are instead defined in the workspace much easier, since their current definitions could just be moved from `melos.yaml` to the workspace config file

Design doc discussion: https://docs.google.com/document/d/1UEEAGdWIgVf0X7o8WPQCPmL4LSzaGCfJTM0pURoDLfE/edit?resourcekey=0-c5CMaOoc_pg3ZwJKMAM0og&disco=AAAA9uS5yhA

sigurdm:
Attempt to fix #4391 

This is very slightly backwards incompatible. If users had directories containing `{` or `*`.

Not sure if it is worth the effort to handle the difference, but one way would be to base it on language version

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -4,6 +4,8 @@
 
 import 'dart:io';
 
+import 'package:glob/glob.dart';
+import 'package:glob/list_local_fs.dart';
 import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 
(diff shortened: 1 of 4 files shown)


spydon:
Thanks a lot for this, it will make our project definitions so much cleaner!
Are the docs for this already updated, or will that wait until 3.11 is out in stable?

mit-mit:
Yup, docs (at least the ones on dart.dev/docs) follow stable releases.

sigurdm:
Created issue to remind us to document: https://github.com/dart-lang/site-www/issues/6927

PR #4743, opened 2026-01-20
Allow overriding Platform

sigurdm:
Should allow running pub in a web setting, by setting a custom Platform object before invoking pub.
(Also needs setting an IOOverrides object.)


jonasfj:
We've landed better overrides, but this was a solid start.

PR #4744, opened 2026-01-20
Make cache repair only repair packages from package_config.json by default

### LINKED ISSUE #4591: Make `cache repair` only repair packages of the local packge_config
Provide access to the current functionality with --all

aikins01:
## Summary
Make `pub cache repair` only repair packages that are actually used by the current project (from `package_config.json`) by default, instead of repairing all cached packages.

Add `--all` flag to restore the original behavior of repairing all cached packages.

## Changes
- Added `--all` flag to `CacheRepairCommand`
- Parse `package_config.json` to identify which packages to repair
- Filter by source type (hosted vs git) and version to only repair exact matches
- Search up the directory tree for `package_config.json` (works from subdirectories)
- Differentiate between missing and invalid `package_config.json` with appropriate error messages

## Tests
- Updated existing tests to use `--all` flag
- Added tests for:
  - Default behavior (only repairs packages from package_config)
  - Version-specific filtering (doesn't repair other versions)
  - Invalid package_config handling
  - Path dependency exclusion
  - Cross-source filtering (git dep doesn't repair hosted packages)

Fixes #4591

### DIFF
--- a/test/testdata/goldens/help_test/pub cache repair --help.txt
+++ b/test/testdata/goldens/help_test/pub cache repair --help.txt
@@ -6,7 +6,7 @@ Reinstall cached packages.
 
 Usage: pub cache repair <subcommand> [arguments...]
 -h, --help    Print this usage information.
-    --all     Repair all cached packages instead of only packages in the current package_config.json.
+    --all     Repair all cached packages instead of only packages in the current pubspec.lock.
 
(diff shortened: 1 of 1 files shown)


sigurdm:
Thanks so much for this contribution!

aikins01:
🫡

PR #4746, opened 2026-01-28
Fall back to cached version listings on network failure

### LINKED ISSUE #3880: `dart run test` w/o internet resulted in "Got socket error trying to find package analyzer ..."

aikins01:
When transient network errors occur (socket errors, TLS failures, timeouts), pub now falls back to cached version listings if available. This improves offline usability when the cache contains sufficient data.

Also improves error messages for socket/TLS/timeout errors to include helpful hints about checking internet connection and using --offline.

## Changes
- Added `isTransientNetworkError()` helper in `lib/src/http.dart`
- Fallback to cached version listings on network errors in `lib/src/source/hosted.dart`
- Improved error messages with helpful hints for socket/TLS/timeout errors
- Added test for error message improvement

Fixes #3880

### DIFF
--- a/lib/src/http.dart
+++ b/lib/src/http.dart
@@ -355,6 +355,20 @@ bool isHttpIOException(Object e) {
       e is WebSocketException;
 }
 
+/// Whether [e] indicates a transient network failure where we might want to
+/// fall back to cached data.
+///
+/// This includes socket errors, TLS/handshake failures, timeouts, and HTTP
+/// client exceptions that wrap network failures. It does NOT include HTTP
+/// response errors (4xx, 5xx) since those indicate the server was reachable.
+bool isTransientNetworkError(Object e) {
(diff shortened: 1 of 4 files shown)


sigurdm:
Thanks for working on this - but I think we are unlikely to accept this PR.
It is too hard to validate the behavior here is as intended on all platforms, and stay that way.

I think one way to go is to try and detect for connectedness, and display a suggestion of trying `--offline` if it looks like there is no connection. That is probably not easy to do reliably across mac/win/linux, and would most likely require ffi bindings.

I know I opened the issue. But I think we have to be careful here.

aikins01:
Thanks for the feedback 🤝, I tried adding a note to try with  --offline, but you're right on the complexity across platforms. Do I continue to pursue it further or I can drop it for now

PR #4769, opened 2026-02-20
'Add @TestOn('vm') to all tests

jonasfj:
This will allow us add test that only run in the browser.


sigurdm:
This is an annoying overhead. Browser tests are forever going to be in the minority,.

Can we do this with a test configuration file instead?
https://github.com/dart-lang/test/blob/master/pkgs/test/doc/configuration.md#configuration-presets

jonasfj:
I don't think we can do this with presets.

A preset is a just a configuration you can choose to run or not.

Maybe, we could make a preset that only runs the browser tests, and then those are otherwise skipped by default.

But I think we want `dart test` to just work (TM).

sigurdm (approved):

Issue #4756, opened 2026-02-13
`flutter pub get` failing to resolve git tag_pattern dependencies in Flutter 3.41


djmaster458:
# Environment

Flutter 3.41

```console
[✓] Flutter (Channel stable, 3.38.9, on macOS 15.6.1 24G90 darwin-arm64, locale en-US) [866ms]
    • Flutter version 3.38.9 on channel stable at /Users/derekosborne/development/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 67323de285 (2 weeks ago), 2026-01-28 13:43:12 -0800
    • Engine revision 587c18f873
    • Dart version 3.10.8
    • DevTools version 2.51.1
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-native-assets, omit-legacy-version-file,
      enable-lldb-debugging

[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0-rc1) [3.3s]
    • Android SDK at /Users/derekosborne/Library/Android/sdk
    • Emulator version 36.1.9.0 (build_id 13823996) (CL:N/A)
    • Platform android-36, build-tools 36.1.0-rc1
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.8+-14196175-b1038.72)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 16.2) [1,160ms]
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 16C5032a
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [6ms]
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Connected device (2 available) [10.2s]
    • macOS (desktop) • macos  • darwin-arm64   • macOS 15.6.1 24G90 darwin-arm64
    • Chrome (web)    • chrome • web-javascript • Google Chrome 138.0.7204.93

[✓] Network resources [240ms]
    • All expected network resources are available.

• No issues found!
```

# Problem

See issue in Flutter side, posting here for pub visibility
[...]

PR #4781, opened 2026-02-24
Handle both tagged and lightweight tags in tag patterns

sigurdm:
Fix of https://github.com/flutter/flutter/issues/182244
(same as https://github.com/dart-lang/pub/issues/4756)


davidmartos96:
@sigurdm Thank you for the fix! Has the team considered this fix as a 3.11 hotfix? The bug is preventing our team from using the latest stable Flutter version 3.41

Issue #4808, opened 2026-04-17
Avoid global state


sigurdm:
Currently it is rather sketchy to invoke any methods twice within the same Isolate.

We should make sure that there is no global state persisting between operations, everything should either be recreated or passed in via a context object

sigurdm:
This is especially important for the client-side dartpad work

ishita73:
I'd be interested in working on this issue.Before I get started, could you point me to the parts of the codebase where the global state currently persists between operations? Also, do you have a preferred approach for introducing the context object or separating the state?
I'd like to make sure my implementation matches the intended design.

PR #4811, opened 2026-04-27
Separate state for sources

sigurdm:
A step on the way to #4808

### DIFF
--- a/lib/src/source/git.dart
+++ b/lib/src/source/git.dart
@@ -266,21 +266,6 @@ class GitSource extends CachedSource {
     return p.url.normalize(parsed.toString());
   }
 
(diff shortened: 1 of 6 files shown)


lib/src/source/git.dart
```diff
@@ -1226,3 +1215,22 @@ RegExp compileTagPattern(String tagPattern) {
     r'$',
   );
 }
+
+/// State that is cached for the GitSource.
+class GitSourceCache {
```
jonasfj: non final class?

jonasfj (approved):

lib/src/source/git.dart
sigurdm: Hopefully soon these can just be data-classes...

Done

PR #4817, opened 2026-05-04
Quiet warning instead of stack trace when failing to parse advisories

sigurdm:
Reduced logging noise when pub fails to parse or decode security advisories from hosted package repositories by logging a concise single-line warning without a stack trace.


sigurdm:
relates to https://github.com/dart-lang/pub-dev/issues/9372

jonasfj (approved):

Issue #4699, opened 2025-10-30
Flutter run always resolves dependencies


FMorschel:
# Environment

- Flutter 3.35.7
- Dart 3.9.2
- OS: Windows 11 Pro

# Problem

This may be the same problem as https://github.com/dart-lang/pub/issues/4672, but I just want to be sure.

I have a pub workspace with 6 Flutter apps and multiple packages. On the latest stable, if I run `flutter run`, I _always_ see:

```
Resolving dependencies in `/my/package/path/`.
Downloading packages...
```

Which I'd expect only if the dependencies were not resolved.

## Expected behavior

For it to run normally and skip the dependencies resolution, since that is already complete.

## Actual behavior

Even if I run the app, close it with no changes and run it again, I see the same output every time.

## `--verbose` output

I've verified, and nothing there displays why it is resolving the dependencies.

CC @sigurdm

sigurdm:
This indeed looks like a bug.
Can you perhaps make a reproduction I can look at?

FMorschel:
I've cloned our repo with all `pubspec.yaml` files and folder structure (and renamed/removed what I couldn't share). 

I did `flutter create .` on the first app folder, and when running it on Web (I'm not sure all dependencies are compatible with all devices since we are still only publishing to Android so if you can't make things work, do some tinkering, but I think web works), even after doing `pub get`, I still reproduce the above, so here it is:

https://github.com/FMorschel/dependency_resolution

sigurdm:
Thanks for the reproduction

I found the bug here: https://github.com/flutter/flutter/blob/0125d95f7eb069269092523303e7735e8be314e3/packages/flutter_tools/lib/src/dart/pub.dart#L275

We are looking at the wrong file. But I think even if we fix that, the logic will not be sound. I have to think a bit to make the correct fix

carlosfiori:
Hi! Is there any update or progress on this issue?
The fact that pub get keeps running repeatedly is causing our scripts to fail.
Thanks in advance!

sigurdm:
Thanks for the ping. Sorry - I hadn't found the time to look at this.
Will try to see if I can find some time.

FMorschel:
Friendly ping @sigurdm 😁

PR #4822, opened 2026-05-16
fix: repair no-op resolution timestamps

### LINKED ISSUE #4672: Bring timestamps of (pubspec.yaml, pubspec.lock, dart_tool/package_config.json) in order even if files are not modified
> @sigurdm actually, I think there is a bug here...
> 
> If I modify `pubspec.yaml` by adding a comment, the timestamp is updated. But `pub get` bails out and doesn't touch the other files because nothing actually changed. However to the IDE, `pubspec.yaml` now looks newer than the other files, so we think we need to run `pub get`.
> 
> I don't know if https://github.com/dart-lang/pub/issues/1447 will fix this or if it also just uses the timestamps. But if it does fix it, it still requires all clients to adopt it, so I think it would be best if the timestamps were still made consistent? 

 _Originally posted by @DanTup in [#4588](https://github.com/dart-lang/pub/issues/4588#issuecomment-3303032099)_

aikins01:
touch unchanged generated resolution files after a successful no-op resolution so timestamp-based clients can keep using the existing package resolution.

this keeps file contents stable, but repairs mtimes when mutable pubspec inputs are newer than the generated lockfile/package config.

fixes dart-lang/pub#4672
relates to dart-lang/pub#4699

### DIFF
--- a/lib/src/entrypoint.dart
+++ b/lib/src/entrypoint.dart
@@ -458,6 +458,37 @@ See $workspacesDocUrl for more information.''',
     }
   }
 
+  /// touches unchanged generated resolution files if their inputs are newer.
+  Future<void> _updateResolutionFileTimestamps() async {
+    final lockFileStat = tryStatFile(lockFilePath);
+    if (lockFileStat == null) return;
+    var lockFileModified = lockFileStat.modified;
+
+    final graph = await packageGraph;
+    for (final package in graph.packages.values) {
+      if (!graph.isPackageMutable(package.name)) continue;
+
+      final pubspecModified = tryStatFile(package.pubspecPath)?.modified;
+      final pubspecOverridesModified = tryStatFile(
+        package.pubspecOverridesPath,
+      )?.modified;
+      if ((pubspecModified != null &&
+              pubspecModified.isAfter(lockFileModified)) ||
+          (pubspecOverridesModified != null &&
+              pubspecOverridesModified.isAfter(lockFileModified))) {
+        touch(lockFilePath);
+        lockFileModified = tryStatFile(lockFilePath)!.modified;
+        break;
+      }
+    }
+
+    final packageConfigModified = tryStatFile(packageConfigPath)?.modified;
+    if (packageConfigModified != null &&
+        lockFileModified.isAfter(packageConfigModified)) {
+      touch(packageConfigPath);
+    }
+  }
+
   Future<String> _packageGraphFile(SystemCache cache) async {
     return const JsonEncoder.withIndent('  ').convert({
       'roots':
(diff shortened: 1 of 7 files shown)


sigurdm (commented):
Thanks for working on this.

Could we instead add a "dependency" argument to writeTextFileIfDifferent and do`touch` call in https://github.com/dart-lang/pub/blob/ac14a82d0803d0b93815060fb6a3dc6dc21f1408/lib/src/io.dart#L278 if needed to have this happen "automatically"?

aikins01:
Thanks, that makes sense. I’ll move this into `writeTextFileIfDifferent` as suggested, and make sure it works for the workspace case where the generated file may depend on more than one pubspec/pubspec_overrides file.

lib/src/io.dart
```diff
@@ -265,17 +265,31 @@ void writeTextFile(
...
+void writeTextFileIfDifferent(
+  String path,
+  String newContent, {
+  Iterable<String> dependencies = const [],
+}) {
   final originalText = tryReadTextFile(path);
   if (originalText != newContent) {
     writeTextFile(path, newContent);
   } else {
     log.fine('`$path` is unchanged. Not rewriting.');
```
sigurdm: Perhaps split the logging over the two cases?

lib/src/lock_file.dart
```diff
@@ -389,7 +389,11 @@ ${yamlToString(data)}
   /// uses that.
   ///
   /// Relative paths will be resolved relative to [lockFilePath]
-  void writeToFile(String lockFilePath, SystemCache cache) {
+  void writeToFile(
+    String lockFilePath,
+    SystemCache cache, {
+    Iterable<String> dependencies = const [],
```
sigurdm: Let's make this argument required to make sure we pass it everywhere. And also it would be nice with a mention in the dartdoc

lib/src/io.dart
aikins01: Good point. I’ll split the logging so the touch case doesn’t say “Not rewriting.”

lib/src/lock_file.dart
aikins01: Done, made it required and added dartdoc.

lib/src/entrypoint.dart
```diff
@@ -458,6 +459,14 @@ See $workspacesDocUrl for more information.''',
     }
   }
 
+  Iterable<String> _resolutionFileDependencies(PackageGraph graph) sync* {
+    for (final package in graph.packages.values) {
+      if (!graph.isPackageMutable(package.name)) continue;
+      yield package.pubspecPath;
+      yield package.pubspecOverridesPath;
+    }
+  }
```
sigurdm: I think maybe we should just look at the workspace packages.
The PackageGraph.isPackageMutable is doing potentially expensive work, and I'm not sure it is that important.

```suggestion
  Iterable<String> _resolutionFileDependencies(PackageGraph graph) sync* {
    for (final package in workspaceRoot.transitiveWorkspace) {
      yield package.pubspecPath;
      yield package.pubspecOverridesPath;
    }
  }
```

aikins01: I agree

aikins01: Done, narrowed this to the workspace packages.

sigurdm (approved):
Thanks for putting in the work!

Issue #3685, opened 2022-12-07
Add solve-suggestions when failing to upgrade to latest version


maximveksler:
# Environment

- Dart version (run `dart --version`): Dart SDK version: 2.18.5 (stable) (Tue Nov 22 15:47:29 2022 +0000) on "linux_x64"
- OS kind and version (e.g. "Windows 10, version 1809" or "macOS 12.4"): `Linux m-inspiron5567 5.15.0-56-generic #62-Ubuntu SMP Tue Nov 22 19:54:14 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`
- Are you using the Chinese community mirror or a corporate firewall? No.

Flutter Environment: 

```
flutter --version
Flutter 3.3.9 • channel stable • https://github.com/flutter/flutter.git
Framework • revision b8f7f1f986 (2 weeks ago) • 2022-11-23 06:43:51 +0900
Engine • revision 8f2221fbef
Tools • Dart 2.18.5 • DevTools 2.15.0
```

# Problem

### Bug

Running the following sequence of events would install `firebase_core: ^0.4.2+2`

```
flutter create bug
cd bug
flutter pub add -d mocktail --verbose > log_pubadd_dev.txt 2>&1
flutter pub add firebase_core google_sign_in --verbose > log_pubadd.txt 2>&1
m@m-inspiron5567:/tmp/bug$ cat pubspec.yaml | grep firebase_core                                       
  firebase_core: ^0.4.2+2
```

### Works

Running the following sequence of events would install `firebase_core: ^2.3.0`

```
flutter create works
cd works
flutter pub add firebase_core google_sign_in --verbose > log_pubadd.txt 2>&1
flutter pub add -d mocktail --verbose > log_pubadd_dev.txt 2>&1
m@m-inspiron5567:/tmp/works$ cat pubspec.yaml | grep firebase_core                                       
  firebase_core: ^2.3.0
```

## Debug

```diff
m@m-inspiron5567:/tmp$ diff /tmp/bug/pubspec.yaml /tmp/works/pubspec.yaml -Naur
--- /tmp/bug/pubspec.yaml       2022-12-07 11:28:01.468494250 +0200
+++ /tmp/works/pubspec.yaml     2022-12-07 11:29:39.281944992 +0200
@@ -1,4 +1,4 @@
-name: bug
+name: works
 description: A new Flutter project.
 
 # The following line prevents the package from being accidentally published to
@@ -36,8 +36,8 @@
[...]

PR #4823, opened 2026-05-17
upgrade: explain targeted transitive upgrades

### LINKED ISSUE #4569: A way to get the solve failure for updating a transitive dependency to latest/resolvable
We have `upgrade` that upgrades to the latest upgradable version, and `upgrade --major-versions` that upgrade to the latest "resolvable" version.

We have `outdated` that explains what the upgradable, resolvable and latest versions are.

But we have no real way of getting an explanation of **why** you cannot get the latest version (which other packages are holding you back).

example:
```
> flutter pub upgrade --major-versions
Resolving dependencies... 
No changes to pubspec.yaml!
Resolving dependencies... 
Downloading packages... 
  lints 5.1.1 (6.0.0 available)
  material_color_utilities 0.11.1 (0.12.0 available)
  vector_math 2.1.4 (2.1.5 available) # <-- why can't we get this???
No dependencies changed.
3 packages have newer versions incompatible with dependency constraints.
Try `flutter pub outdated` for more information. # <-- doesn't really tell me why
```

A workaround is to run:

```
> flutter pub add vector_math:2.1.5
Note: vector_math is pinned to version 2.1.4 by flutter_test from the flutter SDK.
See https://dart.dev/go/sdk-version-pinning for details.

Because every version of flutter_test from sdk depends on vector_math 2.1.4 and ai_blah depends on vector_math 2.1.5, flutter_test from
  sdk is forbidden.
So, because ai_blah depends on flutter_test from sdk, version solving failed.
Failed to update packages.
```

I propose we create some kind of `pub explain <package>` command

Which would lock the given package to `latest` and run a version-resolution, and displaying the solve-trace.

Perhaps we could retrofit this into the existing `upgrade` commmand:

```
> dart pub upgrade vector_math:latest # locks to the latest version
> dart pub upgrade vector_math:resolvable # locks to the "resolvable" version
```

aikins01:
Adds package:latest and package:resolvable targets to pub upgrade so users can force a transitive package to latest or latest-resolvable and get the solver explanation when it cannot be selected.

- Target constraints are evaluated per entrypoint, so --example uses each example's dependency graph and sources.
- --unlock-transitive stays scoped when a target only exists in another active entrypoint.
- Latest lookup respects dependency overrides and current pubspec sources instead of stale lockfile sources.

Fixes #4569
Related to #3685

### DIFF
--- a/lib/src/command/upgrade.dart
+++ b/lib/src/command/upgrade.dart
@@ -17,6 +17,7 @@ import '../package_name.dart';
 import '../pubspec.dart';
 import '../pubspec_utils.dart';
 import '../solver.dart';
+import '../solver/version_solver.dart';
 import '../utils.dart';
 
 /// Handles the `upgrade` pub command.
@@ -27,7 +28,7 @@ class UpgradeCommand extends PubCommand {
   String get description =>
       "Upgrade the current package's dependencies to latest versions.";
   @override
-  String get argumentsDescription => '[dependencies...]';
+  String get argumentsDescription => '[dependencies[:latest|:resolvable]...]';
   @override
   String get docUrl => 'https://dart.dev/tools/pub/cmd/pub-upgrade';
 
@@ -111,6 +112,15 @@ class UpgradeCommand extends PubCommand {
   late final Future<List<String>> _packagesToUpgrade =
       _computePackagesToUpgrade();
 
+  late final List<_UpgradeTarget> _upgradeTargets =
+      argResults.rest.map(_parseUpgradeTarget).toList();
+
+  late final Future<List<ConstraintAndCause>?> _additionalConstraints =
+      _upgradeTargetConstraints();
+
+  late final Future<Map<String, PackageId>> _latestResolvablePackages =
+      _computeLatestResolvablePackages();
+
   /// List of package names to upgrade, if empty then upgrade all packages.
   ///
   /// This allows the user to specify list of names that they want the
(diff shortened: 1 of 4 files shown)


sigurdm:
Sorry for taking long to review this.

This looks really good.

I think we should use @ instead of : for separating the package name and `latest/resolvable`.
Also I think we should allow specifying a specific version instead of only `latest/resolvable`. But that could also come later.

aikins01:
no problem, adding both ;D

lib/src/command/upgrade.dart
```diff
@@ -28,14 +28,17 @@ class UpgradeCommand extends PubCommand {
...
-      'Append `:latest` to a dependency to require the latest available '
+      'Append `@<version>` to a dependency to require a specific version.\n'
+      '\n'
+      'Append `@latest` to a dependency to require the latest available '
       'version.\n'
       '\n'
-      'Append `:resolvable` to require the newest version resolvable with the '
+      'Append `@resolvable` to require the newest version resolvable with the '
       'rest of\n'
       'the dependency graph.';
```
sigurdm: Minor nit. Resolvable means not caring about the current graph, but to remove constraints on the dependencies and see which graph comes out of the solver.

```suggestion
the dependencies.
```

aikins01: done, thanks for clarifying this wording.

sigurdm:
I think we should actually allow general constraints, not just versions. Attempted that in: https://github.com/aikins01/pub/pull/1

aikins01:
This greatly improves it and makes a lot of sense. I agree, thank you! Pulling it in.

I’ll also give the --major-versions / --tighten interaction one pass while doing that, just to make sure the accepted combinations are intentional

sigurdm:
Talking to @jonasfj and he pointed out that when you are not using `--major-versions` the additionalConstraint will usually block directly against the constraint you currently have. Could we issue a suggestion to add --major-versions in that case? https://github.com/sigurdm/pub/pull/2

aikins01:
Added this. When `foo@<constraint>` doesn’t overlap an existing pubspec/workspace constraint, pub now fails early with a `--major-versions` suggestion that preserves the requested target and command context. I also kept dependency override conflicts on a separate message, since `--major-versions` won’t rewrite overrides

sigurdm (approved):

Issue #4818, opened 2026-05-08
pub get became very slow


huanghui1998hhh:
Here's the `flutter pub get -v` output.

[pub_get.log](https://github.com/user-attachments/files/27511238/pub_get.log)

### Observed timings
- Total: `~408.79s`
- Version solving: `~2.01s`
- Root project phase before first `Got dependencies!`: `~380.01s`
- Metadata HTTP requests count: `364`
- Sum of HTTP request durations: `~370.37s`
- Median request latency: `~0.91s`
- p95: `~1.40s`
- p99: `~3.16s`

### Expected behavior
`dart pub get` should complete significantly faster for already-cached packages / lockfile-stable graphs, and should not require so much wall-clock time in metadata lookups.

### Environment
- Dart (from Flutter SDK): `Pub 3.11.5` (from verbose log)
- OS: macOS (arm64)
- Command path: `flutter pub get --example` -> `dart pub get -v`

huanghui1998hhh:
I suspect there may be an unintended cache bypass in `HostedSource._getAdvisories()`:

https://github.com/dart-lang/pub/blob/d5fca3aeebf991f1c2e3b464092f612593ceba69/lib/src/source/hosted.dart#L795-L796

@sigurdm @kevmoo Anyone help?

rs5173:
+1

jonasfj:
> Median request latency: ~0.91s

API end-points used by `pub get` should all be served from GCS through Cloud CDN. Granted the GCS bucket isn't replicated around the world, and that is an optimization we could do.

But such high latency is that something you can reproduce outside the pub client too?

Given that your location is "Mars", one could argue that 900ms latency is pretty good 😄🙈 

But ~0.9s seems pretty slow? where are you located? What does `traceroute pub.dev` say?
Is this the same for google.com, etc. Are you both through-put limited and latency limited? would increasing concurrency help?


Better caching might help, but we do want to fetch meta-data regularly to surface advisories, and other meta-data. But we could potentially be less aggressive about it :D

A workaround for the caching issue is just to use `--offline`.

jonasfj:
Also: is this behavior new? related to a new Dart SDK or just something changed on the service?

huanghui1998hhh:
My geographic location may have some impact🤣

But the more important issue is that the expected cache does not seem to be working, since every `pub get` ends up fetching `_getAdvisories` again.

@jonasfj

huanghui1998hhh:
[...]

PR #4824, opened 2026-05-18
Parallelize security advisories and package status fetching

sigurdm:
Related to #4818 

Also fix a bug where we were not passing a maxAge to gate the advisory cache age.

Also adds a 24-hour buffer to the file-timestamp comparison we do to fast-check if the cached advisories is out of date. (We compare against the written timestamp afterwards).

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -582,23 +582,21 @@ class HostedSource extends CachedSource {
     final Map<String, dynamic> body;
     final List<Advisory>? result;
     try {
-      bodyText = await cache.hostedCache.advisoriesDownloadPool.withResource(
-        () async {
-          return await withAuthenticatedClient(cache, Uri.parse(hostedUrl), (
-            client,
-          ) async {
-            return await retryForHttp(
-              'fetching advisories for "$packageName" from "$url"',
-              () async {
-                final request = http.Request('GET', url);
-                request.attachPubApiHeaders();
-                final response = await client.fetch(request);
-                return response.body;
-              },
-            );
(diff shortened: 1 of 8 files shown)


lib/src/source/hosted.dart
```diff
@@ -792,14 +797,7 @@ class HostedSource extends CachedSource {
           final parsedCacheAdvisoriesUpdated = DateTime.parse(
             cachedAdvisoriesUpdated,
           );
-          final advisoriesUpdated =
-              (await status(id.toRef(), id.version, cache)).advisoriesUpdated;
```
sigurdm, line 796: Here we were missing the maxAge

jonasfj (approved):

lib/src/system_cache.dart
```diff
@@ -170,6 +170,30 @@ Consider setting the `PUB_CACHE` variable manually.
...
+      final advisoriesFuture = id.source.getAdvisoriesForPackageVersion(
+        id,
+        this,
+        const Duration(days: 3),
+      );
+      if (advisoriesFuture != null) {
+        futures.add(advisoriesFuture);
+      }
+    }
+    await Future.wait(futures);
```
jonasfj, line 194: Do we have concurrency limits here? probably that exists in fetch and schedule logic, just wanted to check.

szakarias (approved):
LGTM

lib/src/system_cache.dart
sigurdm: Really good point. Shared the pool with the ratelimitedscheduler

Issue #4791, opened 2026-03-05
[FEATURE REQUEST]: Support a minimum package age (“cooldown”) for dependency resolution


gmpassos:
## Problem

Today, when running `dart pub get` or `dart pub upgrade`, the resolver may select a version that was published **seconds ago** on pub.dev.

This increases supply-chain risk. Malicious or compromised packages are often discovered within hours or days after publication. Installing versions immediately after release removes that natural detection window.

Several ecosystems already support a **minimum release age** policy to mitigate this risk.

### Examples from other ecosystems

* Renovate  
   ```json
  {
    "minimumReleaseAge": "7 days"
  }
   ```

* Dependabot 
  ```yaml
  updates:
    - package-ecosystem: npm
      cooldown: "7 days"
  ```

* npm-check-updates
  ```bash
  ncu --cooldown 7
  ```

Some security platforms such as Snyk even apply a default delay (~21 days) before automatically upgrading dependencies.

The idea is simple: **avoid installing versions that are too recent**.

---

### Proposed feature

Allow `pub` to ignore versions newer than a configurable time window.

Example in `pubspec.yaml`:

```yaml
dependency_resolution:
  minimum_release_age: 7d
```

Meaning:

```
Only consider package versions published at least 7 days ago.
```

---

### Allow exceptions for trusted packages

Projects often maintain **their own internal or trusted packages**, which should not be delayed. To support this, allow an exception list.

Example:

```yaml
dependency_resolution:
  minimum_release_age: 7d
  minimum_release_age_exceptions:
    - my_core_lib
    - internal_utils
```

Behavior:

* All dependencies respect the cooldown window.
* Packages listed in `minimum_release_age_exceptions` **can resolve to the latest version immediately**.

This allows teams to safely update their **own packages** while still protecting against risks from the broader ecosystem.

---

### Example behavior

Available versions:

| Package     | Version | Publish time |
| ----------- | ------- | ------------ |
| foo         | 1.3.2   | 20 days ago  |
| foo         | 1.3.3   | 3 days ago   |
| foo         | 1.3.4   | today        |
| my_core_lib | 2.1.0   | today        |

With:

```yaml
dependency_resolution:
  minimum_release_age: 7d
  minimum_release_age_exceptions:
    - my_core_lib
```

Resolution result:

```
foo → 1.3.2
my_core_lib → 2.1.0
```

---

### Possible CLI option

This could also be exposed via CLI:

```
dart pub get --minimum-release-age=7d
```

---

[...]

PR #4831, opened 2026-05-26
Dependency cooldown policy

sigurdm:
An attempt at #4791 

Projects can declare a `policies` block in their pubspec.yaml.
At most one policies definition is allowed across workspace members to prevent conflicting policies.

For now the only policy is a "cooldown" that restricts which dependencies a package is allowed to take.
Language version needs to be at least 3.14 for a policy to be definable.

Configured by:
* `min_age`: The minimum age duration a package version must have (e.g., `7d`, `2w`).
* `before` (Exactly one of `before` and `min_age` must be specified): an absolute UTC timestamp in ISO 8601 format (e.g., `2026-05-28T09:09:29Z`).
* `exclude` (Optional): A map of package names to version constraints exempt from the cooldown policy.
* `stability` (Optional, can only be applied with `min_age`): A boolean flag that blocks package versions if a newer version was published within the cooldown window after its release (indicating a rushed bugfix/unstable release).

**Why we have "stability"**: When a library author publishes a release (e.g., 1.0.0) and quickly follows up with another release (e.g., 1.0.1) within a short window, it strongly suggests that the initial release had a critical regression, compilation failure, or major bug. Resolving to 1.0.0 in this window exposes your project to an inherently flawed/unstable release. Enabling `stability: true` ensures your solver skips intermediate releases that triggered emergency follow-ups, keeping your dependency graph robust. However, this is different from the behavior of most ecosystems' corresponding features, so we keep this optional and off-by-default.

Example:

```yaml
name: myapp
environment:
  sdk: ^3.14.0
policies:
  cooldown:
    min_age: 7d
    exclude:
      path: any # this package is allowed to bypass the cooldown policy
    stability: true
```

The policy only targets hosted dependencies.

Updated the protocol documentation to specify the `published` property as optional (corresponding to what pub.dev provides).

Third-party package repositories providing a `published` property in their version listing will be able to interact with this feature. If they don't provide it, all their non-excluded packages will be blocked from being part of the resolution.

The version solve is restricted from selecting hosted packages that violate the cooldown or stability rules.

Provides a descriptive version solving failure message indicating the specific version, its release age, and the policy's source location.

When version solving fails due to the cooldown policy, pub suggests excluding the blocked package from the cooldown policy in pubspec.yaml.

```
$ dart pub get
Resolving dependencies...
Because myapp depends on both bar 1.0.0 and foo any, version solving failed:
  myapp depends on bar 1.0.0 which depends on foo >1.0.0.
  version 1.0.1 of foo is too new (released less than 7 days ago)
  The cooldown policy is defined at pubspec.yaml:6
You can try one of the following suggestions to make the pubspec resolve:
* Consider excluding "foo" from the cooldown policy in your pubspec.yaml.
* Try upgrading your constraint on bar: dart pub add bar:^2.0.0
```

`pub outdated`: Flags and lists versions that are blocked by the cooldown policy with a descriptive warning: `Version X is too new for the cooldown policy.`.

```
Showing outdated packages.
[*] indicates versions that are not the latest available.
Package Name  Current  Upgradable  Resolvable  Latest
foo           1.0.0    1.0.0       1.0.0       1.0.1*
foo
    Version 1.0.1 is too new for the cooldown policy.
```

`pub get` Report: Displays available newer versions blocked by the policy next to the resolved packages (e.g., `(X available (blocked by cooldown))`).

```
+ foo 1.0.0 (1.0.1 available (blocked by cooldown))
```


doc/repository-spec-v2.md
```diff
@@ -233,6 +233,7 @@ server, this could work in many different ways.
   "latest": {
     "version": "<version>",
     "retracted": true || false, /* optional field, false if omitted */
+    "published": "<date-time>", /* optional field, ISO 8601 format (must include timezone), timestamp of when this version was published */
```
mosuem: Or just use UTC time?

lib/src/solver/package_lister.dart
```diff
@@ -204,6 +213,61 @@ class PackageLister {
   /// workspace-children are also added.
   Future<List<Incompatibility>> incompatibilitiesFor(PackageId id) async {
     if (_knownInvalidVersions.allows(id.version)) return const [];
+
+    final policiesWrapper = _policies;
+    final policy = policiesWrapper?.cooldown;
+
+    final allVersions = <(Version, DateTime?)>[];
+    final statusMap = <Version, PackageStatus>{};
```
mosuem, line 223: Those two, `allVersions` and `statusMap`, seem store similar things. Could you remove `allVersions`?

lib/src/solver/package_lister.dart
```diff
@@ -204,6 +213,61 @@ class PackageLister {
...
+            v.toRef(),
+            v.version,
+            _systemCache,
+          );
+          allVersions.add((v.version, status.published));
+          statusMap[v.version] = status;
+        }
+      }
+
+      var allForbidden = true;
```
mosuem, line 246: A general comment: The code seems to be lacking documentation - I would expect everything which is not obvious to have comments explaining what is happening, or more importantly why it is.

lib/src/solver/package_lister.dart
```diff
@@ -231,6 +295,59 @@ class PackageLister {
       }
     }
 
+    final description = id.description;
```
mosuem: Nit: I think this might be cleaner with a `case` in the if

lib/src/pubspec.dart
```diff
@@ -132,6 +132,167 @@ environment:
...
+        before: before,
+        exclude: exclude,
+        stability: stability,
+      );
+    }
+
+    return Policies(cooldown: cooldown, span: policiesNode.span);
+  }
+
+  Duration _parseDuration(String s) {
```
mosuem, line 294: More tests would make me more confortable around this regex.

Also, where is this field specified for users?

lib/src/package.dart
```diff
@@ -519,6 +519,22 @@ Workspace members must have unique names.
...
   }
 
+  // Check that at most one policies configuration is specified.
+  Policies? policiesSeen;
+  for (final package in root.transitiveWorkspace) {
+    final currentPolicies = package.pubspec.policies;
+    if (currentPolicies != null) {
+      if (policiesSeen != null) {
+        fail('''
+Only a single policies configuration is allowed across all workspace pubspec.yaml files.
```
mosuem, line 541: Is there a documentation for users around this feature?

lib/src/pubspec.dart
```diff
@@ -865,3 +1030,84 @@ enum Resolution {
...
   none,
 }
+
+class Policies {
+  final CooldownPolicy? cooldown;
+  final SourceSpan span;
+  Policies({this.cooldown, required this.span});
+}
+
+class CooldownPolicy {
```
mosuem: Documentation! :)

lib/src/pubspec.dart
```diff
@@ -865,3 +1030,84 @@ enum Resolution {
...
+  final DateTime? before;
+  final Map<String, VersionConstraint> exclude;
+  final bool stability;
+  CooldownPolicy({
+    this.minAge,
+    this.before,
+    required this.exclude,
+    this.stability = false,
+  }) : assert(
+         (minAge == null) != (before == null),
```
mosuem: just have two constructors to avoid the asserts?

lib/src/pubspec.dart
```diff
@@ -865,3 +1030,84 @@ enum Resolution {
...
+    String packageName,
+    Version version,
+    DateTime? published,
+    List<(Version, DateTime?)> allVersions,
+  ) {
+    final constraint = exclude[packageName];
+    if (constraint != null && constraint.allows(version)) return false;
+    if (published == null) return true;
+
+    final minAge = this.minAge;
```
mosuem: Why this assignment? Possibly use a `case`?

lib/src/pubspec.dart
```diff
@@ -132,6 +132,167 @@ environment:
...
+      }
+
+      final beforeNode = cooldownNode.nodes['before'];
+      DateTime? before;
+      if (beforeNode != null) {
+        final value = beforeNode.value;
+        if (value is! String) {
+          _error('"before" must be a string', beforeNode.span);
+        }
+        if (!RegExp(r'(?:Z|[+-]\d{2}(?::?\d{2})?)$').hasMatch(value)) {
```
mosuem: Can't `DateTime.parse` do this?

lib/src/pubspec.dart
```diff
@@ -132,6 +132,167 @@ environment:
...
+        } on FormatException catch (e) {
+          _error('Invalid "min_age": ${e.message}', minAgeNode.span);
+        }
+      }
+
+      final beforeNode = cooldownNode.nodes['before'];
+      DateTime? before;
+      if (beforeNode != null) {
+        final value = beforeNode.value;
+        if (value is! String) {
```
mosuem: Again a good case for `case` :)

doc/repository-spec-v2.md
sigurdm: Updated to specify UTC ISO 8601 timestamps (ending in 'Z', e.g. "2026-05-28T09:09:29Z").

lib/src/solver/package_lister.dart
sigurdm: Removed allVersions and updated the solver and policy checks to use statusMap directly.

lib/src/solver/package_lister.dart
sigurdm: Added documentation explaining the short-circuiting logic when all versions are forbidden and how stability compares timestamps across releases.

lib/src/solver/package_lister.dart
sigurdm: Done, updated to use record pattern matching with case.

lib/src/pubspec.dart
sigurdm: Simplified the timestamp format to strictly require UTC strings (ending in 'Z') and added tests for valid UTC timestamps, non-UTC timezone offsets, and malformed date strings. User-facing documentation for policies will be published on dart.dev with the SDK release.

lib/src/package.dart
sigurdm: User documentation for pubspec policies and workspace constraints will live on dart.dev alongside the feature release.

lib/src/pubspec.dart
sigurdm: Added dartdoc comments for Policies and CooldownPolicy and marked them final.

lib/src/pubspec.dart
sigurdm: Replaced with two dedicated named constructors: CooldownPolicy.minAge and CooldownPolicy.before.

lib/src/pubspec.dart
sigurdm: Updated to use pattern matching with case.

lib/src/pubspec.dart
sigurdm: DateTime.parse accepts local datetime strings without timezone indicators (resulting in isUtc == false), which can lead to divergent resolution across different local environments and CI. We have simplified the policy to require explicit UTC ISO 8601 strings ending in 'Z'.

lib/src/pubspec.dart
sigurdm: Updated node parsing in _parsePolicies to use switch / case patterns.

sigurdm:
Thanks for the review @mosuem! I've addressed all the feedback: simplified cutoff dates to UTC-only ISO 8601 strings, split `CooldownPolicy` into dedicated named constructors, modernized the YAML parsing and solver logic to use Dart 3 pattern matching, added missing dartdocs/comments, and added date parsing tests.

test/testdata/goldens/help_test/pub upgrade --help.txt
```diff
@@ -4,14 +4,14 @@
 $ pub upgrade --help
 Upgrade the current package's dependencies to latest versions.
 
-Append `@<version>` to a dependency to require a specific version.
+Append `@<constraint>` to a dependency to require a version constraint.
```
mosuem, line 7: How does the user learn the syntax for the `<constraint>`?

sigurdm: This is an artifact of upstream merging. Not part of this PR.

Constraints are documented at least here: https://dart.dev/tools/pub/versioning#version-constraints-the-dart-approach . Maybe we should provide an example.

sigurdm:
Merged with latest `master` and resolved all conflicts. @mosuem @jonasfj @szakarias PTAL when you get a chance!

lib/src/command/outdated.dart
```diff
@@ -774,6 +794,12 @@ Future<void> _outputHuman(
           'See https://dart.dev/go/package-retraction',
         );
       }
+      if (package.isLatestBlockedByCooldown) {
+        log.message(
+          '    Version ${package.latest!._id.version} is too new for '
+          'cooldown policy.',
```
mosuem: the? your?

lib/src/solver/package_lister.dart
```diff
@@ -30,6 +32,9 @@ class PackageLister {
   /// Only used when _ref is root.
   final Package? _rootPackage;
 
+  /// The policy to apply, if any.
```
mosuem:
```suggestion
  /// The [Policies] to apply, if any.
```

Also a short sentence as to what this does? Or why? Why does the `PackageLister` need to know about this?

lib/src/solver/package_lister.dart
```diff
@@ -231,6 +297,58 @@ class PackageLister {
...
+            reasonPrefix =
+                'version ${id.version} of ${id.name} is too new '
+                '(released after $before)\n';
+          } else {
+            reasonPrefix = 'version ${id.version} of ${id.name} is blocked\n';
+          }
+        } else {
+          reasonPrefix =
+              'version ${id.version} of ${id.name} lacks publication date '
+              'required by policy\n';
```
mosuem: by which or whose policy?

lib/src/solver/package_lister.dart
```diff
@@ -231,6 +297,58 @@ class PackageLister {
...
+        }
+        final reason =
+            '$reasonPrefix'
+            'Cooldown policy defined at '
```
mosuem: Possibly I'm in the wrong, but also here I would add something. `With a cooldown...`, `The cooldown policy is defined...`, this reads a bit incomplete.

lib/src/solver/report.dart
```diff
@@ -459,13 +459,37 @@ $contentHashesDocumentationUrl
...
+
       if (status.isRetracted) {
         if (newerStable) {
-          notes.add(
-            'retracted, ${maxAll(versions, Version.prioritize)} available',
-          );
+          notes.add('retracted, $latestVersion available$cooldownSuffix');
         } else if (newId.version.isPreRelease && newerUnstable) {
-          notes.add('retracted, ${maxAll(versions)} available');
+          notes.add('retracted, $latestVersion available$cooldownSuffix');
```
mosuem, line 493: are these two if cases the same?

lib/src/solver/report.dart
```diff
@@ -483,12 +507,12 @@ $contentHashesDocumentationUrl
...
       } else if (newerStable) {
         // If there are newer stable versions, only show those.
-        notes.add('${maxAll(versions, Version.prioritize)} available');
+        notes.add('$latestVersion available$cooldownSuffix');
       } else if (
       // Only show newer prereleases for versions where a prerelease is
       // already chosen.
       newId.version.isPreRelease && newerUnstable) {
-        notes.add('${maxAll(versions)} available');
+        notes.add('$latestVersion available$cooldownSuffix');
```
mosuem, line 511: Also here, do those two cases yield the same result?

lib/src/solver/version_solver.dart
```diff
@@ -71,6 +71,11 @@ class VersionSolver {
     for (final package in _root.transitiveWorkspace) package.name: package,
   };
 
+  /// The policy to apply, if any, found in the workspace.
```
mosuem:
```suggestion
  /// The [Policies] to apply, if any, found in the workspace.
```

lib/src/source/hosted.dart
```diff
@@ -426,6 +426,20 @@ class HostedSource extends CachedSource {
       if (retracted is! bool) {
         throw const FormatException('retracted must be a bool');
       }
+      final publishedData = map['published'];
+      DateTime? publishedDate;
+      if (publishedData != null) {
+        if (publishedData is! String) {
+          throw const FormatException('published must be a String');
+        }
+        if (!RegExp(r'(?:Z|[+-]\d{2}(?::?\d{2})?)$').hasMatch(publishedData)) {
```
mosuem: extract to const, and document! Regexes are impossible to read.

lib/src/pubspec.dart
```diff
@@ -132,6 +132,180 @@ environment:
...
+          _error('"min_age" must be a string', minAgeNode.span);
+      }
+
+      final beforeNode = cooldownNode.nodes['before'];
+      DateTime? before;
+      switch (beforeNode) {
+        case null:
+          break;
+        case YamlScalar(value: final String value):
+          if (!RegExp(
```
mosuem: No undocumented inline regexes please

lib/src/pubspec.dart
```diff
@@ -132,6 +132,180 @@ environment:
     };
   }();
 
+  /// The policy configuration.
+  Policies? get policies => _policies ??= _parsePolicies();
+  Policies? _policies;
+
+  Policies? _parsePolicies() {
```
mosuem, line 142: This method is a bit hard to read, but maybe that's fine. I also don't have an idea on how to make it nicer. Maybe check all the conditions on the Nodes first before parsing them?

lib/src/pubspec.dart
```diff
@@ -132,6 +132,180 @@ environment:
     };
   }();
 
+  /// The policy configuration.
```
mosuem: Bit of a useless comment :) This seems to lazily initialize the policies, apparently it either has a startup cost we don't want to pay immediately, and/or it's too expensive to freshly retrieve every time?

mosuem (approved):
LGTM, but then again I am not really an expert on this code.

I would suggest to add a lot more documentation in the code, almost all is undocumented, so it's really hard to understand why things are the way they are.

lib/src/command/outdated.dart
sigurdm: Yeah, that can be rephased!

lib/src/solver/package_lister.dart
sigurdm: I can expand here. The packagelister will not list packages not allowed by the policy.

lib/src/solver/package_lister.dart
sigurdm: Well that comes just below. But we could say "cooldown policy"

lib/src/command/outdated.dart
sigurdm: Updated to 'the cooldown policy'.

lib/src/solver/package_lister.dart
sigurdm: Updated dartdoc to document `[Policies]` and clarify that it is used by the version solver to reject versions violating configured policies by generating incompatibility constraints.

lib/src/solver/package_lister.dart
sigurdm: Updated to 'required by the cooldown policy'.

lib/src/solver/package_lister.dart
sigurdm: Updated to complete sentence ('The cooldown policy is defined at ...') and added explanatory code comments to the policy-checking blocks.

lib/src/solver/report.dart
sigurdm: Good catch! Because `latestVersion` already handles choosing between `maxAll(versions, Version.prioritize)` and `maxAll(versions)`, both branches became identical. Combined them into `newerStable || (newId.version.isPreRelease && newerUnstable)`.

lib/src/solver/report.dart
sigurdm: Combined into `else if (newerStable || (newId.version.isPreRelease && newerUnstable))` here as well.

sigurdm:
> I am not really an expert on this code.

The feedback is very valued nonetheless!

lib/src/pubspec.dart
sigurdm: Updated the dartdoc to document that this returns the workspace `[Policies]` (or `null` if none are defined) and that it is parsed lazily on first access from `fields`, consistent with the other fields in `Pubspec` (like `dependencies`, `environment`, etc.).

lib/src/pubspec.dart
sigurdm: Refactored `_parsePolicies()` into dedicated, single-responsibility helper methods (`_parseCooldownPolicy`, `_parseMinAge`, `_parseBefore`, `_parseExclude`, `_parseStability`). `_parseCooldownPolicy` now checks node requirements and mutual exclusivity up-front before dispatching to the constructors.

lib/src/pubspec.dart
sigurdm: Extracted both regexes into documented `static final` constants: `_utcIso8601RegExp` (matching UTC ISO 8601 strings ending in 'Z') and `_durationRegExp` (matching duration strings with 'd' or 'w' units).

lib/src/source/hosted.dart
sigurdm: Extracted to a documented `static final _timezoneSuffixRegExp = RegExp(r'(?:Z|[+-]\d{2}(?::?\d{2})?)$');` on `HostedSource`.

lib/src/solver/version_solver.dart
sigurdm: Applied suggestion (bracketed `[Policies]`).

gmpassos:
> > I am not really an expert on this code.
> 
> 
> 
> The feedback is very valued nonetheless!

How about “stability” also accepts a duration? For instance, the minimum age could be 2 weeks, and the stability windows could be 1 week. Alternatively, you could add a “stability_window” parameter.

PR #4833, opened 2026-05-28
Consider equal timestamps to be outdated

sigurdm:
On windows the timestamp resolution is quite poor, and two files having same timestamp can easily happen.
We cannot conclude ordering when the timestamps are equal, and should therefore not use the fast-path.

Using dart setMtime also gets a poor resolution, this can result in the touched file being seen as actually older than the new one, so we shell out to `touch` where available.


lib/src/entrypoint.dart
```diff
@@ -845,12 +845,27 @@ To update `$lockFilePath` run `$topLevelProgram pub get`$suffix without
...
         wasRelative ? p.relative(path) : path;
 
+    late final rootPackageDir = () {
+      for (final parent in parentDirs(dir)) {
+        if (tryStatFile(p.join(parent, 'pubspec.yaml')) != null) {
+          return parent;
+        }
+      }
+      return dir;
+    }();
```
jonasfj:
```suggestion
    final rootPackageDir = parentDirs(dir).where(
        (parent) => tryStatFile(p.join(parent, 'pubspec.yaml') != null)
      ).firstOrNull ?? dir;
```

jonasfj (approved):

lib/src/entrypoint.dart

PR #4838, opened 2026-06-14
solver: suggest workspace resolution fixes

### LINKED ISSUE #4227: Do solve suggestions for workspaces
Related to #4127 
Currently these don't work

We probably ship first incantation of workspaces with solve suggestions disabled when using workspaces.

aikins01:
Enables version solving suggestions for workspace dependency conflicts.

This locates the owning workspace pubspec for direct hosted constraints, emits directory-aware pub add and major-upgrade commands, preserves hosted sources for workspace member suggestions, and suppresses invalid path/git root descriptors.

Fixes #4227.
Related to #4127.

### DIFF
--- a/lib/src/solver/solve_suggestions.dart
+++ b/lib/src/solver/solve_suggestions.dart
@@ -1,6 +1,5 @@
 import 'dart:convert';
 import 'dart:io';
-
 import 'package:pub_semver/pub_semver.dart';
 
 import '../command_runner.dart';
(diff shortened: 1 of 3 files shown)


sigurdm:
As I understand this PR, it will still only attempt to relax constraints for each dependency in a single workspace member at a time.

I think we need to test for relaxing the constraint across the workspace in one go. Not sure how to best express that as a command. Can we use the new `dart pub upgrade package:version` ?

aikins01:
Thanks, added a workspace-wide case that suggests `dart pub upgrade --major-versions <package>` when the same dependency has to be relaxed across workspace members. I also tightened the suggested `pub add` command for workspace member subdirectories so the path descriptor is executable from the cwd.

PR #4852, opened 2026-08-03
Fast-path literal workspace entries to avoid O(N^2) traversal using full globbing

### LINKED ISSUE #4850: Pub workspace loading is ~17x slower for literal paths with Dart 3.11
# Environment

- Dart SDK 3.12.2 stable on macOS arm64
- macOS 26.5.2 (25F84)

# Problem

After raising a 513-package monorepo's root SDK constraint to Dart 3.12, every package-level `dart test` invocation started taking roughly 30 seconds before test discovery. Even `dart test --help` pays the workspace-loading cost. This makes running package tests separately or in parallel prohibitively slow.

The workspace uses literal paths rather than glob patterns:

```yaml
name: workspace
environment:
  sdk: '>=3.12.0 <4.0.0'
workspace:
  - packages/package_0
  - packages/package_1
  # 510 more literal package paths
  - packages/package_511
  - packages/subject
```

The following script creates two equivalent 513-member workspaces. The only difference is the root language version.

Save as `repro.dart`:

```dart
import 'dart:io';

void main(List<String> arguments) {
  final output = Directory(arguments[0])..createSync();
  final rootLanguageVersion = arguments[1];
  final workspace = <String>[];

  for (var index = 0; index < 512; index++) {
    final name = 'package_$index';
    final path = 'packages/$name';
    workspace.add(path);
    final directory = Directory('${output.path}/$path')
      ..createSync(recursive: true);
    File('${directory.path}/pubspec.yaml').writeAsStringSync('''
name: $name
publish_to: none
environment:
  sdk: '>=3.10.0 <4.0.0'
resolution: workspace
''');
  }

  workspace.add('packages/subject');
  final subject = Directory('${output.path}/packages/subject')
    ..createSync(recursive: true);
  File('${subject.path}/pubspec.yaml').writeAsStringSync('''
name: subject
publish_to: none
environment:
  sdk: '>=3.10.0 <4.0.0'
resolution: workspace
dev_dependencies:
  test: 1.31.0
''');

  File('${output.path}/pubspec.yaml').writeAsStringSync('''
name: synthetic_workspace
publish_to: none
environment:
  sdk: '>=$rootLanguageVersion.0 <4.0.0'
workspace:
${workspace.map((path) => '  - $path').join('\n')}
''');
}
```

```sh
dart repro.dart fast 3.10
dart repro.dart slow 3.11

(cd fast && dart pub get)
(cd slow && dart pub get)

time sh -c 'cd fast/packages/subject && dart test --help >/dev/null'
time sh -c 'cd slow/packages/subject && dart test --help >/dev/null'
```

Run each timed command twice to exclude initial snapshot creation.

## Expected behavior

The 3.11 workspace should have similar startup time to the otherwise identical 3.10 workspace when all workspace entries are literal paths.

## Actual behavior

Warm measurements:

```text
root language 3.10:  0.71 s real
root language 3.11: 11.86 s real
```

Replacing the 513 literal workspace entries with a single `packages/*` glob reduces the warm 3.11 run to 0.73 seconds.

The 3.11 run remains slow after the `test` executable snapshot has been created:

```text
real 11.86
user  3.61
sys   8.48
```

## `--verbose` output

There is no error output. The command succeeds normally after the startup delay; the slowdown occurs before test discovery or help output.

sigurdm:
When resolving workspace members in language versions >= 3.11, entries without glob wildcard syntax characters (`*`, `?`, `[`, `{`) are now resolved directly via a filesystem check rather than delegating to `package:glob`'s `listSync`.

Delegating literal entries (such as `packages/package_0`) to `package:glob` caused `listSync` to enumerate the entire `packages/` directory for each of the $N$ entries, causing an $O(N^2)$ explosion in directory entries traversed.

Fixes #4850

### DIFF
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -589,6 +589,14 @@ See https://dart.dev/go/workspaces-stray-files for details.
   }
 }
 
+/// Returns whether [s] contains any glob wildcard syntax characters.
+///
+/// We intentionally do not account for backslash-escaped wildcards (like `\*`).
+/// Paths without any of these characters are guaranteed to be plain relative
+/// paths that can be checked directly on the filesystem without unescaping or
+/// glob parsing. Paths containing these characters (even if escaped) are
+/// delegated to `package:glob` to handle unescaping properly and to avoid
+/// conflicting with Windows `\` path separators.
 bool _hasGlobWildcards(String s) =>
     s.contains('*') || s.contains('?') || s.contains('[') || s.contains('{');
 String _useBackSlashesOnWindows(String path) {
--- a/lib/src/package.dart
+++ b/lib/src/package.dart
@@ -181,7 +181,8 @@ class Package {
         pubspec.workspace.expand((workspacePath) {
           final packages = <Package>[];
           var globHint = '';
-          if (pubspec.languageVersion.supportsWorkspaceGlobs) {
+          if (pubspec.languageVersion.supportsWorkspaceGlobs &&
+              _hasGlobWildcards(workspacePath)) {
             final Glob glob;
             try {
               glob = Glob(workspacePath);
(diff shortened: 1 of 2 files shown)


jonasfj:
> Delegating literal entries (such as `packages/package_0`) to `package:glob` caused `listSync` to enumerate the entire `packages/` directory for each of the N entries, causing an O ( N 2 ) explosion in directory entries traversed.

Isn't that only a mitigation.

I could legitmately have `packages/package_*` and now we have the same issue, right?

Isn't the problem not that we don't cache file system entries? I guess the argument is that if you use glob patterns you get poor performance and we don't care, because 500 glob patterns is degenerate :rofl: , where as 500 literal directory names is perfectly reasonable.

jonasfj (approved):

sigurdm:
Yes, this addresses a regression we had from before globs were allowed. (see linked issue)
This doesn't protect against adversarial behavior.

Will edit title to clarify

sigurdm:
> I guess the argument is that if you use glob patterns you get poor performance and we don't care, because 500 glob patterns is degenerate 🤣 ,

Yeah - you don't need a lot of glob patterns to cover many directories.

> where as 500 literal directory names is perfectly reasonable.

yes!

PR #4854, opened 2026-08-06
Add ApiBreakageValidator to run dart_apitool before dart pub publish

mosuem:
Implements https://github.com/dart-lang/sdk/issues/63970

Adds `ApiBreakageValidator` to validate whether breaking API changes are being published with a non-breaking version bump (e.g. minor or patch update). Uses `dart_apitool` in-process to diff public APIs against the baseline published package version.

### DIFF
--- a/lib/src/validator/api_breakage.dart
+++ b/lib/src/validator/api_breakage.dart
@@ -43,6 +43,7 @@ class ApiBreakageValidator extends Validator {
 
     // If currentVersion is already a breaking SemVer bump relative to
     // previousVersion, breaking changes are expected.
+    // We compare baseVersion to account for pre-release versions.
     if (isBreakingVersionBump(previousVersion, currentVersion)) {
       return;
     }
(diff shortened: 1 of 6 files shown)


lib/src/validator/api_breakage.dart
```diff
@@ -0,0 +1,111 @@
...
+      if (breakingChanges.isNotEmpty) {
+        final buffer = StringBuffer();
+        buffer.writeln(
+          'Breaking API change(s) detected compared to published version '
+          '$previousVersion.\n'
+          'Your current version ($currentVersion) is not a breaking SemVer '
+          'bump.\n'
+          'Suggested version: ${previousVersion.nextBreaking}\n'
+          'Please bump your version according to Semantic Versioning before '
+          'publishing.\n'
```
sigurdm: Maybe it is too early to wordsmith this, but I guess we can suggest either bumping the version or fixing the breakage.

PR #4855, opened 2026-08-06
Warn if analysis options are included from a file outside the package

### LINKED ISSUE #4853: Warn on publish if analysis options are excluded from local file outside the package.
It will fail to analyze cleanly on pub.dev (or whereever it is published).

sigurdm:
Fixes #4853.

Adds `AnalysisOptionsValidator` to warn when `analysis_options.yaml` (or any other analysis options file in the package) includes a configuration file outside the package directory.

### DIFF
--- a/lib/src/validator/analysis_options.dart
+++ b/lib/src/validator/analysis_options.dart
@@ -1,0 +1,116 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:yaml/yaml.dart';
+
+import '../io.dart';
+import '../path.dart';
+import '../validator.dart';
+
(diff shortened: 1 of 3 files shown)


lib/src/validator/analysis_options.dart
```diff
@@ -0,0 +1,116 @@
...
+import '../io.dart';
+import '../path.dart';
+import '../validator.dart';
+
+/// Validates that `analysis_options.yaml` (and other analysis options files)
+/// do not include configuration from a local file outside the package.
+///
+/// Including files outside the package causes analysis errors on `pub.dev`
+/// and for consumers because those external files are not included in the
+/// published archive.
```
jonasfj, line 19: It doesn't create issues for consumers.

It is true that the package is not self contained.

It doesn't necessarily create analysis issues on pub.dev, only if the included thing disabled lints / errors. Not if it just add more lints.

---------------

This could just as well be a lint in the analyzer. which is triggered whenever:
 * `pubspec.yaml` contains `version:` and,
 * `pubspec.yaml` does not contain `publish_to: none`.

Then it would pretty much only affect people who write packages.

Granted, I doubt we could convince people to land this as a lint :D

PR #4865, opened 2026-08-19
Optimize version solver PackageLister with binary search and count caching

sigurdm:
In version solving, `countVersions` is invoked repeatedly across decision steps to find candidate packages with the fewest remaining versions. Previously, `countVersions` performed a linear scan over all published versions for each package at every decision step. Additionally, `bestVersion` performed a linear scan starting from the highest available version, even when the version constraint specified a much lower upper bound.

This change optimizes `PackageLister` by:
- Using binary search on the sorted version list for counting versions matching `VersionRange`s and `VersionUnion`s.
- Caching version counts per constraint within `PackageLister`.
- Using bounded binary search in `bestVersion` to jump directly to the constraint's upper bound (or lower bound when downgrading) before scanning.

### DIFF
--- a/lib/src/solver/package_lister.dart
+++ b/lib/src/solver/package_lister.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:math' as math;
 
 import 'package:async/async.dart';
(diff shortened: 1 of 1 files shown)


sigurdm:
Closing in favor of #4874.

PR #4866, opened 2026-08-19
Short-circuit package version selection when candidate has 0 or 1 matching versions

sigurdm:
In `_choosePackageVersion`, the version solver selects the unsatisfied package with the fewest remaining matching versions. Previously, `minByAsync` evaluated the version count for all unsatisfied packages on every decision step.

A package with `0` matching versions will immediately force a conflict, and a package with `1` matching version has only a single forced choice. In both cases (`count <= 1`), there is no benefit in checking remaining unsatisfied candidates. Short-circuiting the loop as soon as a candidate has `count <= 1` avoids redundant version count evaluations.

### DIFF
--- a/lib/src/solver/version_solver.dart
+++ b/lib/src/solver/version_solver.dart
@@ -392,11 +392,24 @@ class VersionSolver {
       return candidate.name;
     }
 
-    /// Prefer packages with as few remaining versions as possible, so that if a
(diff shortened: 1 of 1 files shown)


sigurdm:
Closing in favor of keeping minByAsync for brevity and code simplicity, as the PackageLister binary search in #4865 already provides the vast majority of solver performance gains.

PR #4867, opened 2026-08-19
Optimize PartialSolution.satisfier to avoid PartialSolution allocations

sigurdm:
During conflict resolution, `satisfier(term)` finds the earliest assignment prefix in the partial solution that satisfies a given term. Previously, this method allocated a new `PartialSolution` instance on every call and replayed assignments through `_register`.

This change computes the satisfying assignment incrementally by tracking accumulated positive and negative terms for the target package using local variables, avoiding `PartialSolution` object allocations and unnecessary map lookups during conflict resolution.


sigurdm:
Closing in favor of keeping PartialSolution.satisfier simple and maintainable without duplicating assignment tracking logic.

PR #4875, opened 2026-08-19
Verify package attestations and enforce provenance policies in pub get/upgrade

mosuem:
### Client-side Ingestion & Cryptographic Verification on `pub get` / `upgrade`
When downloading packages in `HostedSource`, queries the repository for `GET /api/packages/<pkg>/versions/<ver>/attestation` and runs `PubAttestationVerifier` against the cached or embedded Sigstore `trusted_root.json`:

1. **Lockfile & Downgrade Prevention:** `ProvenancePolicy.enforcePolicy()` blocks downgrades from signed to unsigned packages and detects repository switches across versions.
2. **Archive SHA-256 Digest:** Verifies downloaded `.tar.gz` SHA-256 hash matches the in-toto subject digest in the attestation.
3. **Package Name & Version:** Verifies in-toto subject name matches the requested `<package>-<version>.tar.gz`.
4. **DSSE Envelope & PAE Signatures:** Validates signatures over DSSE Pre-Authentication Encoding.
5. **Fulcio Root CA Verification:** Validates leaf certificate against `trusted_root.json` Fulcio Root CAs.
6. **Rekor Transparency Log Proof:** Verifies Rekor inclusion proof against active public keys in `trusted_root.json`.
7. **GitHub Actions OIDC Issuer:** Enforces `https://token.actions.githubusercontent.com` issuer identity.
8. **Signer Workflow & Git Ref:** Validates signer workflow ID, branch/tag ref, and commit SHA against SLSA provenance.
9. **Package Repository Binding:** Verifies certificate repository matches `pubspec.yaml`'s declared repository.

### Tests
- `test/get/hosted/attestation_verification_test.dart`
- `test/sigstore/verifier_test.dart`
- `test/sigstore/trusted_root_test.dart`
- `test/sigstore/lockfile_policy_test.dart`


lib/src/source/hosted.dart
```diff
@@ -1606,6 +1660,61 @@ See $contentHashesDocumentationUrl.
...
+                  } catch (e) {
+                    throw PackageIntegrityException('''
+Downloaded attestation for ${id.name}-${id.version} is malformed: $e
+
+This indicates a problem on the package repository: `${description.url}`.
+''');
+                  }
+                  final archiveBytes = readBinaryFile(archivePath);
+                  final verifier = PubAttestationVerifier(cache: cache);
+                  final result = verifier.verify(
```
sigurdm, line 1685: Looking at this call site, I noticed a few concerns with how client-side verification is hooked up:

1. **No expected identity is passed to `verify()`:**
   `verifier.verify()` takes optional `expectedRepository` and `pubspecRepository`. Here neither is provided, so `PubAttestationVerifier` skips repository checking and passes `''` as `expected_identity` down to `SigstoreVerificationPolicy`. In `package:sigstore`'s Rust implementation, an empty string filters to `None`, which means:
   > *"If empty, identity is not restricted."*
   
   As a result, any cryptographically valid Sigstore bundle generated by *any* GitHub Actions workflow in *any* arbitrary repository will pass verification for this package. (The unit tests in `attestation_verification_test.dart` do pass `expectedRepository`, which is why this didn't surface there).

2. **Verification is gated on the server-provided `slsaLevel`:**
   Because `if (versionInfo.slsaLevel != null)` checks the metadata returned by the repository server, a compromised or malicious repository could simply omit `slsa_level` to bypass client verification entirely.

3. **Errors fail open:**
   The `catch (e)` block swallows all unexpected exceptions (logging only to `log.fine`), including the `UnsupportedError` thrown by `sigstore_stub.dart` on platforms where `dart:ffi` is not available.

More fundamentally, for the client to meaningfully verify the attestation identity, it needs a trusted source for *which* repository is authorized to publish `${id.name}`. If that information only comes from the repository server itself, we are in a circular trust dilemma.

PR #4876, opened 2026-08-19
Support uploading package attestations during publish --from-archive

mosuem:
Adds client support for publishing package attestations during `pub publish`:

- Adds `--with-attestation <bundle.sigstore.json>` flag to `dart pub publish --from-archive`.
- Enforces that a repository is specified in `pubspec.yaml` (`repository` field) when publishing with an attestation.
- Validates that `--with-attestation` can only be used with `--from-archive`, and that the file is non-empty and contains a JSON object, before asking the user to confirm the publication.
- Adds unit tests in `test/lish/publishing_to_and_from_archive_test.dart`.

### Protocol

`doc/repository-spec-v2.md` gets a new *Publishing with an Attestation* section:

- A repository that supports publishing with attestations returns an `<attestation-upload-url>` in the `attestationUrl` property of the response from `<hosted-url>/api/packages/versions/new`. A repository that does not support it simply omits the property.
- The client `POST`s the bundle as `application/json` to `<attestation-upload-url>` **before** uploading the package archive, attaching `Authorization` under the same prefix rule as `archive_url`.
- `<finalize-upload-url>` remains a plain `GET`.

`attestationUrl` is an opaque URL: a repository may point it at its own endpoint or at a signed URL on a storage service. On pub.dev it is an endpoint on the app server, so no second signed upload policy is needed (see dart-lang/pub-dev#9545).

The alternative considered was to `POST` the attestation to `<finalize-upload-url>` and use `405 Method Not Allowed` for feature detection. A separate URL was chosen because:

- the client learns that a repository cannot take attestations *before* uploading the archive, instead of after; a missing JSON property is also a more reliable signal than `405`, which proxies and frameworks do not produce consistently, and
- the spec keeps one method per URL. This matters here because the archive upload is allowed to answer with `303 See Other` pointing at `<finalize-upload-url>`, and a `303` is followed with a `GET` — any client that simply follows the redirect would drop the attestation body.

Server side: dart-lang/pub-dev#9545

### DIFF
--- a/doc/repository-spec-v2.md
+++ b/doc/repository-spec-v2.md
@@ -298,8 +298,15 @@ changed.
 
 ```js
 {
-  "url": "<multipart-upload-url>",
+  "url": "<archive-upload-url>",
   "fields": {
+    "<field-1>": "<value-1>",
+    "<field-2>": "<value-2>",
+    ...,
+    "<field-N>": "<value-N>",
+  },
+  "attestationUrl": "<attestation-upload-url>",
+  "attestationFields": {
     "<field-1>": "<value-1>",
     "<field-2>": "<value-2>",
     ...,
@@ -309,14 +316,74 @@ changed.
 ```
 
 To publish a package an HTTP `GET` request for
-`<hosted-url>/api/packages/versions/new` is made. This request returns an
-`<multipart-upload-url>` and a dictionary of fields. To upload the package
-archive a multi-part `POST` request is made to `<multipart-upload-url>` with
-fields and the field `file` containing the gzipped tar archive.
+`<hosted-url>/api/packages/versions/new` is made. This request returns:
+* `url`: An `<archive-upload-url>` for uploading the package archive.
+* `fields`: A dictionary of form fields required by the storage service for the
+  archive upload.
+* `attestationUrl`: An `<attestation-upload-url>` for uploading a package
+  attestation bundle (such as a Sigstore provenance bundle). Present on
+  repositories that support attestations.
+* `attestationFields`: A dictionary of form fields required by the storage
+  service for the attestation upload. Present if `attestationUrl` is present.
+
+The `attestationUrl` and `attestationFields` properties are returned by
+repositories that support package attestations. If the user requests publishing
+with an attestation and the server response omits `attestationUrl`, the client
+should abort the publication with an informative error message.
+
+### Uploading the Attestation
+
+When publishing with an attestation, the client **MUST** upload the attestation
+bundle to `<attestation-upload-url>` (from `attestationUrl`) **before** uploading
+the package archive to `<archive-upload-url>` (from `url`). This ensures that an
+attestation is never omitted due to partial or interrupted uploads. If not
+publishing with an attestation, this step is skipped.
+
+To upload the attestation bundle, a multi-part `POST` request is made to
+`<attestation-upload-url>` containing all key-value pairs from
+`attestationFields`, along with a `file` part containing the attestation JSON
+bundle:
 
 ```http
-POST <path(multipart-upload-url)> HTTP/1.1
-Host: <host(multipart-upload-url)>
+POST <path(attestation-upload-url)> HTTP/1.1
+Host: <host(attestation-upload-url)>
+Content-Length: <length>
+Content-Type: multipart/form-data; boundary=<boundary>
+
+--<boundary>
+Content-Disposition: form-data; name="<urlencode(field-1)>"
+Content-Type: text/plain; charset=utf-8
+Content-Transfer-Encoding: binary
+
+<value-1>
+...
+--<boundary>
+Content-Disposition: form-data; name="<urlencode(field-N)>"
+Content-Type: text/plain; charset=utf-8
+Content-Transfer-Encoding: binary
+
+<value-N>
+--<boundary>
+Content-Type: application/json; charset=utf-8
+Content-Disposition: form-data; name="file"; filename="attestation.sigstore.json"
+
+<attestation JSON bundle>
+--<boundary>--
+```
+
+The storage service may respond with `HTTP 204 No Content` or `HTTP 200 OK`. The
+client must verify that the request succeeded before proceeding to upload the
+package archive.
+
+### Uploading the Package Archive
+
+To upload the package archive a multi-part `POST` request is made to
+`<archive-upload-url>` containing all key-value pairs from `fields`, along with a
+`file` part containing the gzipped tar archive:
+
+```http
+POST <path(archive-upload-url)> HTTP/1.1
+Host: <host(archive-upload-url)>
 Content-Length: <length>
 Content-Type: multipart/form-data; boundary=<boundary>
 
@@ -343,20 +410,37 @@ Content-Transfer-Encoding: binary
 Content-Type: application/octet-stream
 Content-Disposition: form-data; name="file"; filename="package.tar.gz"
 
-<gzipped archieve>
+<gzipped archive>
+--<boundary>--
 ```
 
-The above `POST` request to `<multipart-upload-url>` should respond as follows:
+The above `POST` request to `<archive-upload-url>` may respond as follows:
 
 ```http
 HTTP/1.1 204 No Content
 Location: <finalize-upload-url>
 ```
 
+or with an HTTP `303 See Other` redirect to `<finalize-upload-url>`.
+
+### Finalizing the Upload
+
 The client shall then issue a `GET` request to `<finalize-upload-url>`. As with
 `archive_url` the client will only attach an `Authorization` if the
-`<hosted-url>` is a prefix of `<finalize-upload-url>`. If the server wants to
-accept the uploaded package the server should respond:
+`<hosted-url>` is a prefix of `<finalize-upload-url>`.
+
+During finalization, the server inspects the uploaded archive, along with any
+accompanying attestation that was uploaded prior to the archive. If an
+attestation was uploaded, the server verifies its validity before accepting
+the publication. An attestation may be rejected as invalid if:
+ * the JSON bundle is malformed or cannot be parsed as a Sigstore bundle,
+ * the cryptographic signature cannot be verified against the trusted root,
+ * the artifact digest recorded in the attestation does not match the SHA-256
+   hash of the uploaded package archive, or
+ * the provenance information (such as the source repository) does not match
+   the `repository` field in `pubspec.yaml`.
+
+If the server wants to accept the uploaded package the server should respond:
 
 ```http
 HTTP/1.1 200 Ok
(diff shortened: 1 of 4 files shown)


sigurdm (commented):
This could use also an update to `## Publishing Packages` in doc/repository-spec-v2.md

doc/repository-spec-v2.md
```diff
@@ -397,6 +397,76 @@ similar blob storage service. Both the
...
+
+When publishing with a package attestation, the client uploads a Sigstore bundle
+to `<multipart-upload-url>` before uploading the package archive.
+
+The multi-part `POST` request for the attestation uses the same fields
+dictionary returned by `<hosted-url>/api/packages/versions/new`, with the
+following modifications:
+ * If a `key` field is present in `fields`, its value is modified by appending
+   `.sigstore.json` (i.e. `<key>.sigstore.json`). This sets the storage object
+   path for the attestation file, allowing the server to locate and correlate the
```
jonasfj: explain the specification, not how the server is able to locate something.

doc/repository-spec-v2.md
```diff
@@ -397,6 +397,76 @@ similar blob storage service. Both the
...
+   attestation bundle with the package archive (stored at `<key>`) during
+   finalization.
+ * The `success_action_redirect` field, if present in `fields`, is omitted.
```
jonasfj: this is an odd specification

doc/repository-spec-v2.md
jonasfj:
```suggestion
The above `POST` request to `<multipart-upload-url>` may respond as follows:
```

This probably shouldn't have been a requirement, and I don't think it expected to be by any client.

doc/repository-spec-v2.md
```diff
@@ -397,6 +397,76 @@ similar blob storage service. Both the
 `<multipart-upload-url>` and `<finalize-upload-url>` are allowed to contain
 query-string parameters, and both of these URLs need only be temporary.
 
+### Uploading Package Attestations
```
jonasfj: So we used to have a specification that was pretty sane:

~~~
**GET** `<hosted-url>/api/packages/versions/new`

**Headers:**
* `Accept: application/vnd.pub.v2+json`
* `Authorization: Bearer <token>` (required)

**Response**
* `Content-Type: application/vnd.pub.v2+json`

```js
{
  "url": "<multipart-upload-url>",
  "fields": {
    "<field-1>": "<value-1>",
    "<field-2>": "<value-2>",
    ...,
    "<field-N>": "<value-N>",
  },
}
```

~~~

The only thing that was sort of weird here is:
 * (A) The field `file` is special and includes the archive being uploaded.
 * (B) If the server responds with a redirect, client must send `Authorization` to said redirect URL, if "if the `<hosted-url>` is a prefix of `<finalize-upload-url>`."

Uploading with multipart-upload is a bit weird, but it's just implementing a well establish spec.
Following a redirect is natural.
Sending `Authorization` header on the redirect is probably reasonable to require -- but it does feel unnecessary when reading it as a pure spec (I know why we require it).
Specifying that `Authorization` is sent only when "if the `<hosted-url>` is a prefix of `<finalize-upload-url>`." this bit is weird. It would have been natural to say if it's same-origin.

----------

This spec I'm not even sure we can implement this server side using signed urls

jonasfj: I think what I'm saying is, we should consider making a new end-point entirely.

I've always disliked `/api/packages/versions/new` because it conflicted with the package name "new" 🙈 

The natural way to extend the current approach would be to do:

~~~

**GET** `<hosted-url>/api/packages/versions/new`

**Headers:**
* `Accept: application/vnd.pub.v2+json`
* `Authorization: Bearer <token>` (required)

**Response**
* `Content-Type: application/vnd.pub.v2+json`

```js
{
  "url": "<multipart-upload-url>",
  "fields": {
    "<field-1>": "<value-1>",
    "<field-2>": "<value-2>",
    ...,
    "<field-N>": "<value-N>",
  },
  "attestationUrl": "<multipart-upload-url>",
  "attestationFields": {
    "<field-1>": "<value-1>",
    "<field-2>": "<value-2>",
    ...,
    "<field-N>": "<value-N>",
  },
}
```

~~~

Granted, I think we should then specify that `attestationUrl` must be uploaded before `url`.

doc/repository-spec-v2.md
```diff
@@ -298,25 +298,92 @@ changed.
...
 {
-  "url": "<multipart-upload-url>",
+  "url": "<archive-upload-url>",
   "fields": {
     "<field-1>": "<value-1>",
     "<field-2>": "<value-2>",
     ...,
     "<field-N>": "<value-N>",
   },
+  "attestationUrl": "<attestation-upload-url>",
```
sigurdm: If we upload the attestation via Cloud Storage, GCS requires a second signed policy (since GCS form POST cannot write two objects in one request). On `pub.dev`, this requires an extra IAM `signBlob` RPC on every publish.

Could we instead send the attestation directly in the finalize request?

Because the attestation is only ~5 KB of JSON, it doesn't need to bypass the app server:
* `GET /api/packages/versions/new` stays completely unchanged (1 signed URL).
* Client uploads the archive to GCS as usual.
* Client calls `<finalize-upload-url>`: with `POST` containing the attestation JSON if `--with-attestation` was passed, or `GET` if not.
* If a server doesn't support attestations, the `POST` returns `405 Method Not Allowed`.

This avoids coordinating multiple objects in storage and eliminates the 2x IAM signing overhead on the server.

lib/src/command/lish.dart
```diff
@@ -418,12 +494,36 @@ the \$PUB_HOSTED_URL environment variable.''');
...
+    Uint8List? attestationBytes;
+    if (_withAttestation != null) {
+      final repository = pubspec.fields['repository']?.toString();
+      if (repository == null) {
+        dataError(
+          'A GitHub repository must be specified in the "repository" field of '
+          'pubspec.yaml when publishing with an attestation.',
+        );
+      }
+      if (!repository.contains('github.com')) {
```
sigurdm: We shouldn't hardcode `github.com` checks in the client:
1. `pub` is a general package manager used with third-party and internal package repositories. Sigstore is an open standard that supports GitLab, Chainguard, private OIDC providers, etc.
2. `repository.contains('github.com')` is brittle—it allows URLs like `https://not-github.com` while disallowing internal GitHub Enterprise instances or alternative code forges.
3. Policy enforcement regarding which repository forges and OIDC signers are trusted belongs on the repository server (e.g. `pub.dev`), not in the client.

In the client, checking `pubspec.fields['repository'] != null` is reasonable, but the forge restriction should be removed.

lib/src/command/lish.dart
```diff
@@ -155,7 +165,53 @@ class LishCommand extends PubCommand {
...
+            attRequest.files.add(
+              http.MultipartFile.fromBytes(
+                'file',
+                attestationBytes,
+                filename: 'attestation.sigstore.json',
+              ),
+            );
+            attRequest.followRedirects = false;
+            final attResponse = await client.fetch(attRequest);
+            attResponse.throwIfNotOk();
```
sigurdm: `client.fetch(request)` already defaults to `throwIfNotOk: true` and throws if the response is not 2xx/redirect. Calling `attResponse.throwIfNotOk()` here is redundant.

lib/src/command/lish.dart
```diff
@@ -155,7 +165,53 @@ class LishCommand extends PubCommand {
...
+            return attResponse;
+          });
+        }
+
+        /// 3. Upload package
         final url = _expectField(parameters, 'url', parametersResponse);
         if (url is! String) invalidServerResponse(parametersResponse);
         cloudStorageUrl = Uri.parse(url);
```
sigurdm, line 196: `cloudStorageUrl` is only set here for the archive upload URL. If the attestation upload fails due to an HTTP error from Cloud Storage, `url == cloudStorageUrl` in the `PubHttpResponseException` catch block will be false, so the user won't get the formatted cloud storage error message.

We should either set `cloudStorageUrl` before the attestation upload or track both URLs (e.g. in a `Set<Uri>`).

lib/src/command/lish.dart
```diff
@@ -418,12 +494,36 @@ the \$PUB_HOSTED_URL environment variable.''');
...
+      }
+      if (!repository.contains('github.com')) {
+        dataError(
+          'A GitHub repository must be specified in the "repository" field of '
+          'pubspec.yaml when publishing with an attestation. The repository '
+          '"$repository" is not a GitHub repository.',
+        );
+      }
+      try {
+        attestationBytes = readBinaryFile(_withAttestation);
```
sigurdm, line 486: Could we do some light validation on the attestation file here before proceeding? (e.g. checking that the file is non-empty and contains valid JSON).

Right now, if the user points to an empty or non-JSON file, we ask them to confirm publication, upload it to storage, and only fail much later during server-side finalization.

test/lish/publishing_to_and_from_archive_test.dart
```diff
@@ -71,4 +71,160 @@ void main() {
...
+        bundlePath,
+      ],
+      workingDirectory: d.sandbox,
+    );
+
+    expect(pub.stdout, emitsThrough('Publishing from archive: archive.tar.gz'));
+    await confirmPublish(pub);
+
+    handleUploadForm(server, withAttestation: true);
+    server.handle('/upload-attestation', (request) async {
```
sigurdm: Could we assert that the request payload actually contains the expected attestation bundle bytes?

sigurdm (commented):
I have a proposal for how to communicate the attestation that would simplify both the spec and the server implementation:

### The issue with uploading attestations to GCS
Google Cloud Storage's form POST API only allows uploading a single object per request. To upload both the archive and attestation to GCS, the server must generate two separately signed policies.

On `pub.dev`, signing a policy requires an IAM `signBlob` RPC. If `GET /api/packages/versions/new` returns two signed URLs unconditionally, we incur 2x IAM latency (50–200ms) and 2x quota on **every** publish, even though 99.9% of publishes don't have an attestation. In addition, during finalization the server has no record of whether the client intended to upload an attestation, so a dropped upload could lead to silently publishing without provenance.

### Proposal: Send the attestation in the finalize request

An attestation bundle (`.sigstore.json`) is only **3–10 KB**. Unlike the 100 MB archive, which must be uploaded directly to GCS to avoid buffering in App Engine / Cloud Run memory, a 5 KB JSON bundle can be sent directly to the server.

1. **`GET /api/packages/versions/new` stays completely unchanged.** No new fields, no capability negotiation, and only 1 IAM call to sign the archive upload policy.
2. The client uploads `package.tar.gz` to GCS as today.
3. During finalization, if publishing with an attestation, the client issues a `POST` to `<finalize-upload-url>` with `{ "attestation": <sigstore JSON bundle> }`. If publishing without an attestation, it issues a `GET` as today.

**Benefits:**
* **Zero changes to `versions/new`:** No extra IAM calls or latency on `pub.dev`.
* **No multi-object storage complexity:** No second signed URL, no orphan objects in GCS.
* **Atomic delivery:** The attestation arrives directly with the finalize call, so the server knows unambiguously that this publish includes an attestation.
* **Compatibility:** If a server doesn't support attestations, it will return `405 Method Not Allowed` on the `POST`, which the client can report as *"The repository does not support publishing with attestations."*

doc/repository-spec-v2.md
mosuem: [JETSKI AI ANSWER]
Thanks for the suggestion, Sigurd! We've updated the PR and the repository spec to adopt this proposal:

1. **Finalize via POST**:
   - `GET /api/packages/versions/new` remains unchanged (no extra signed form fields or URLs).
   - When publishing with an attestation, `pub` now issues an HTTP `POST` request to `<finalize-upload-url>` with `Content-Type: application/json` and `{"attestation": <sigstore-json>}` in the body. When publishing without an attestation, it continues issuing an HTTP `GET` as before.
   - If a package repository does not support attestations, it should respond with `405 Method Not Allowed` to `POST` requests at `<finalize-upload-url>`. In response to a `405`, `pub` terminates with a clean error message: `"The package repository $host does not support publishing with attestations."`
   - Updated `doc/repository-spec-v2.md` to document this finalization protocol.

2. **Early Client-Side Validation**:
   - In `pub lish --from-archive ... --with-attestation ...`, `pub` validates that the attestation file is non-empty, contains valid JSON, and parses into a JSON object (`Map<String, dynamic>`) before prompting the user for publication confirmation.

3. **Tests**:
   - Updated `test/lish/publishing_to_and_from_archive_test.dart` to assert that the finalize request is a `POST` with `Content-Type: application/json` containing `{"attestation": ...}` in the payload.
   - Added test cases verifying the `405 Method Not Allowed` error handling and each early client-side validation failure mode (empty file, invalid JSON, non-object JSON).

lib/src/command/lish.dart
mosuem: Addressed: GCS attestation multipart upload was removed in favor of sending the attestation bundle directly in the `POST <finalize-upload-url>` request.

lib/src/command/lish.dart
mosuem: Addressed: Added early client-side validation in `_publicationFromArchive` before prompting the user for confirmation, verifying the attestation file is non-empty, contains valid JSON, and parses into a JSON object (`Map<String, dynamic>`).

test/lish/publishing_to_and_from_archive_test.dart
mosuem: Addressed: Added payload assertion in `test/lish/publishing_to_and_from_archive_test.dart` asserting that the finalize `POST` request payload contains `{"attestation": jsonDecode(validAttestationBundle)}`.

lib/src/command/lish.dart
mosuem: Addressed: The separate GCS multipart attestation upload block was removed in favor of POST to `<finalize-upload-url>`.

doc/repository-spec-v2.md
mosuem: Addressed: Applied suggestion.

doc/repository-spec-v2.md
mosuem: Addressed: Reverted the multi-part attestation upload and adopted Sigurd's proposal to transmit the attestation bundle via `POST <finalize-upload-url>`.

lib/src/command/lish.dart
mosuem: Addressed: Removed the `github.com` restriction from `pub lish`. The client now only requires that a repository URL is specified in `pubspec.yaml` (since attestations must bind to a source repository), leaving forge-specific policy enforcement to the repository server or verification stage. Added a test confirming non-GitHub repositories pass publication validation.

PR #4885, opened 2026-08-24
Make package archives reproducible with deterministic timestamps and entry ordering

mosuem:
## Summary
Makes `dart pub publish --to-archive` bit-for-bit deterministic and reproducible across all operating systems:
1. Normalizing TAR entry timestamps (`modified`) unconditionally to Unix Epoch 0 (`DateTime.fromMillisecondsSinceEpoch(0, isUtc: true)`).
2. Normalizing GZIP headers (RFC 1952 `MTIME = 0`, `OS = 255`) via `normalizeGzipHeader()`.
3. Sorting archive entries deterministically by normalized relative path.
4. Normalizing TAR directory and file permissions/modes.
5. Adding multi-OS integration tests verifying reproducible archives regardless of file mtime and directory iteration order.

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -1268,10 +1268,12 @@ Future<void> extractTarGz(Stream<List<int>> stream, String destination) async {
   log.fine('Extracted .tar.gz to $destination.');
 }
 
-/// Create a .tar.gz archive from a list of entries.
+/// Creates a .tar.gz archive of [contents] relative to [baseDir].
 ///
-/// Each entry is the path to a directory or file. The root of the archive is
-/// considered to be [baseDir], which defaults to the current working directory.
+/// Archives are bit-for-bit deterministic and reproducible: timestamps are
+/// normalized to Unix epoch 0 (1970-01-01 00:00:00 UTC), entries are sorted
+/// lexicographically by normalized POSIX path, and file permissions and GZIP
+/// headers are normalized.
 ///
 /// Returns a [ByteStream] that emits the contents of the archive.
 ByteStream createTarGz(List<String> contents, {required String baseDir}) {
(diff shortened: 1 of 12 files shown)


lib/src/io.dart
```diff
@@ -1273,8 +1278,16 @@ Future<void> extractTarGz(Stream<List<int>> stream, String destination) async {
...
+/// If [reproducible] is `true`, or if [EnvironmentKeys.sourceDateEpoch] is set
+/// in the environment, timestamps are normalized and entries are sorted
+/// lexicographically.
+///
 /// Returns a [ByteStream] that emits the contents of the archive.
-ByteStream createTarGz(List<String> contents, {required String baseDir}) {
+ByteStream createTarGz(
+  List<String> contents, {
+  required String baseDir,
+  bool reproducible = false,
```
sigurdm: Perhaps we just always do this?

lib/src/io.dart
```diff
@@ -1326,23 +1353,68 @@ ByteStream createTarGz(List<String> contents, {required String baseDir}) {
...
+      .transform(tarWriterWith(format: OutputFormat.gnuLongName))
+      .transform(gzip.encoder);
+  if (isReproducible) {
+    stream = stream.transform(_normalizeGzipHeader());
+  }
+
+  return ByteStream(stream);
+}
+
+/// Normalizes the GZIP header (RFC 1952) so that `.tar.gz` archives are
```
sigurdm, line 1358: I kind of remember we have something in our test setup that overwrites these headers. Perhaps that can be removed in this change?

lib/src/io.dart
```diff
@@ -1326,23 +1342,66 @@ ByteStream createTarGz(List<String> contents, {required String baseDir}) {
...
+    handleDone: (sink) {
+      if (!isHeaderHandled && headerBuffer.isNotEmpty) {
+        sink.add(headerBuffer);
+      }
+      sink.close();
+    },
   );
 }
 
+DateTime _archiveEpoch() {
```
sigurdm: I don't think we need this to be configurable via environment. Can't we always just use epoch?

mosuem: This is following the specs from https://reproducible-builds.org/specs/source-date-epoch/ - but I'm fine to not do that as well. What do you think?

sigurdm: Ah - had no clue about that. Are there any use cases? We can keep it if it is useful - but I don't really see the use of having a timestamp on the archive.

mosuem: Apparently some build systems use it to know if something needs to be rebuilt. We can delete it though, can always add it some other time..

sigurdm: SG

lib/src/io.dart
```diff
@@ -1283,27 +1285,35 @@ ByteStream createTarGz(List<String> contents, {required String baseDir}) {
   ArgumentError.checkNotNull(baseDir, 'baseDir');
```
sigurdm, line 1285: (Non-blocking / pre-existing) `ArgumentError.checkNotNull(baseDir, 'baseDir')` was already here before this PR, but since `baseDir` is non-nullable in Dart 3 and `checkNotNull` is deprecated, we could clean it up while touching this method.

test/lish/reproducible_archive_test.dart
```diff
@@ -0,0 +1,88 @@
...
+      entryNames.add(entry.name);
+      // Verify normalized epoch timestamp
+      expect(
+        entry.header.modified,
+        equals(DateTime.fromMillisecondsSinceEpoch(0, isUtc: true)),
+      );
+      expect(entry.header.userName, equals('pub'));
+      expect(entry.header.groupName, equals('pub'));
+      if (entry.type == TypeFlag.dir) {
+        if (!Platform.isWindows) {
```
sigurdm, line 74: Directory modes are hardcoded to `_defaultMode | _executableMask` in `createTarGz` regardless of platform, so `if (!Platform.isWindows)` isn't needed here.

sigurdm (approved):
A few observations on getting the tests green:

**GZIP OS byte vs. Golden tests:**
In `test_pub.dart`, replacing `_replaceOs` with `normalizeGzipHeader()` changed byte 9 (the OS byte) from `0` to `255`. Because existing golden tests recorded SHA256 hashes of test packages generated with OS byte `0`, tests comparing package hashes fail (e.g. 13 tests in `dependency_services_test.dart` and `embedding_test.dart`).
- If we set byte 9 to `0` in `normalizeGzipHeader()`, all of those golden tests pass without changes (and `0` is also what Windows zlib naturally emits).
- If `255` is preferred to match `gzip -n` / RFC 1952 "unknown OS", the golden files with affected hashes will need to be regenerated in this PR.

PR #4888, opened 2026-08-25
Use dart build cli for integration test executable

mosuem:
This PR replaces the legacy `dart --snapshot` step with `dart build cli`.

- Uses `dart build cli -t bin/pub.dart -o .dart_tool/_pub_build` to create an AOT-compiled CLI bundle for integration tests.
- Automatically compiles and bundles Native Assets (e.g. `package:sigstore` dynamic libraries) in `bundle/lib/`.
- Updates `test/test_pub.dart` to support directly executing the precompiled binary.
- Updates `tool/test.dart` and GitHub Actions CI workflow to use `dart build cli`.

### DIFF
--- a/lib/src/executable.dart
+++ b/lib/src/executable.dart
@@ -20,6 +20,7 @@ import 'package_config.dart';
 import 'path.dart';
 import 'platform_info.dart';
 import 'sdk.dart';
+import 'sdk/dart.dart';
 import 'system_cache.dart';
 import 'utils.dart';
 
@@ -110,7 +111,7 @@ Future<int> runExecutable(
   final packageConfigAbsolute = p.absolute(entrypoint.packageConfigPath);
 
   try {
-    return await _runDartProgram(
+    final exitCode = await _runDartProgram(
       executablePath,
       args.toList(),
       packageConfigAbsolute,
@@ -118,6 +119,19 @@ Future<int> runExecutable(
       vmArgs: vmArgs,
       alwaysUseSubprocess: alwaysUseSubprocess,
     );
+    if (useSnapshot && exitCode == 253) {
+      log.fine('Built executable is out of date.');
+      await recompile(executable);
+      return await _runDartProgram(
+        entrypoint.pathOfSnapshot(executable),
+        args.toList(),
+        packageConfigAbsolute,
+        enableAsserts: enableAsserts,
+        vmArgs: vmArgs,
+        alwaysUseSubprocess: alwaysUseSubprocess,
+      );
+    }
+    return exitCode;
   } on IsolateSpawnException catch (error) {
     if (!useSnapshot ||
         !error.message.contains('Invalid kernel binary format version')) {
@@ -127,7 +141,7 @@ Future<int> runExecutable(
     log.fine('Built executable is out of date.');
     await recompile(executable);
     return await _runDartProgram(
-      executablePath,
+      entrypoint.pathOfSnapshot(executable),
       args.toList(),
       packageConfigAbsolute,
       enableAsserts: enableAsserts,
@@ -162,9 +176,15 @@ Future<int> _runDartProgram(
   path = p.absolute(path);
   packageConfig = p.absolute(packageConfig);
 
-  // We use Isolate.spawnUri when there are no extra vm-options.
-  // That provides better signal handling, and possibly faster startup.
-  if ((!alwaysUseSubprocess) && vmArgs.isEmpty) {
+  // We use Isolate.spawnUri when there are no extra vm-options and running
+  // under the Dart VM. Standalone AOT binaries cannot execute kernel dill
+  // snapshots in-process and must spawn the SDK dart executable.
+  final canUseIsolate =
+      (!alwaysUseSubprocess) &&
+      vmArgs.isEmpty &&
+      p.equals(DartSdk().executable, platform.resolvedExecutable);
+
+  if (canUseIsolate) {
     final argList = args.toList();
     return await isolate.runUri(
       p.toUri(path),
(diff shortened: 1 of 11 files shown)


lib/src/executable.dart
```diff
@@ -193,14 +213,29 @@ Future<int> _runDartProgram(
     // TODO(sigurdm) To handle signals better we would ideally have `exec`
     // semantics without `fork` for starting the subprocess.
     // https://github.com/dart-lang/sdk/issues/41966.
+    final environment = Map<String, String>.from(platform.environment);
```
sigurdm, line 216: We need a comment here explaining what is happening, and why platform.resolvedExecutable doesn't work.

Does this work well in the context of dartdev, does that set DART_ROOT? Maybe could use our isRunningFromTest getter here to only do this when testing.

lib/src/global_packages.dart
```diff
@@ -969,8 +969,23 @@ Try reactivating the package.
     }
     // When running tests we want the binstub to invoke the current pub, not the
     // one from the sdk.
-    final pubInvocation =
-        runningFromTest ? platform.script.toFilePath() : 'pub';
+    final String pubInvocation;
```
sigurdm, line 972: Again probably better to only change behavior when under test...

tool/test.dart
```diff
@@ -30,28 +27,43 @@ Future<void> main(List<String> args) async {
...
-    await precompile(
-      executablePath: p.join('bin', 'pub.dart'),
-      outputPath: pubSnapshotFilename,
-      name: 'bin/pub.dart',
-      packageConfigPath: p.join('.dart_tool', 'package_config.json'),
-    );
+    stderr.write('Building pub executable with `dart build cli`...');
+    final buildResult = await Process.run(Platform.resolvedExecutable, [
+      'build',
+      'cli',
```
sigurdm, line 42: I guess build cli is a lot slower than the precompile, have - have you measured the time diff? Currently we can recompile in a few seconds

lib/src/global_packages.dart
```diff
@@ -969,8 +969,23 @@ Try reactivating the package.
...
+        final isDartVm =
+            p.basenameWithoutExtension(platform.resolvedExecutable) == 'dart';
+        if (isDartVm) {
+          pubInvocation = 'dart "${platform.script.toFilePath()}"';
+        } else {
+          pubInvocation = '"${platform.resolvedExecutable}"';
+        }
+      }
+    } else {
+      pubInvocation = 'pub';
```
sigurdm, line 987: Previously the templates had `dart $pubInvocation global run`, so for production it generated `dart pub global run`.

With `dart` removed from the template, `pubInvocation = 'pub'` generates `pub global run` (and `call pub global run`). Since the Dart SDK does not have a standalone `pub` binary in `bin/` (the command is `dart pub`), fallback binstubs in production will fail with `pub: command not found`.

Should this be `pubInvocation = 'dart pub';`?

lib/src/io.dart
```diff
@@ -691,18 +691,7 @@ void createPackageSymlink(
...
-final bool _assertionsEnabled = () {
-  try {
-    assert(false);
-    // ignore: avoid_catching_errors
-  } on AssertionError {
-    return true;
-  }
-  return false;
-}();
+final bool runningFromTest = platform.environment.containsKey('_PUB_TESTING');
```
sigurdm, line 694: Why remove `_assertionsEnabled` here? `dart build cli` supports the `--enable-asserts` flag.

If we compile the test executable with `--enable-asserts` (in `tool/test.dart` and `.github/workflows/test.yaml`), we can retain `_assertionsEnabled` here and ensure integration tests continue to run with assertions enabled.

tool/test-bin/pub.bat
```diff
@@ -3,10 +3,10 @@ rem Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
...
-rem Runs bin/pub.dart with dart from PATH (or the snapshot if present).
+rem Runs bin/pub.dart with dart from PATH (or the built pub executable if present).
 
-if "%_PUB_TEST_SNAPSHOT%"=="" (
-    dart %~p0\..\..\..\bin\pub.dart %*
+if "%_PUB_TEST_EXECUTABLE%"=="" (
+    dart %~p0\..\..\bin\pub.dart %*
 ) else (
-    dart %_PUB_TEST_SNAPSHOT% %*
+    call %_PUB_TEST_EXECUTABLE% %*
```
sigurdm, line 11: Should this be quoted as `call "%_PUB_TEST_EXECUTABLE%" %*` to prevent breakage if the workspace path contains spaces?

lib/src/executable.dart
```diff
@@ -193,14 +213,29 @@ Future<int> _runDartProgram(
...
+    if (environment['DART_ROOT'] case final root?) {
+      if (!fileExists(
+        p.join(root, 'bin', platform.isWindows ? 'dart.exe' : 'dart'),
+      )) {
+        environment.remove('DART_ROOT');
```
sigurdm, line 221: When `DART_ROOT` is set in tests (e.g. pointing to a mock SDK directory without a `bin/dart`), `DartSdk().executable` was already initialized to `$DART_ROOT/bin/dart`. Removing `DART_ROOT` from `environment` here won't change the path passed to `Process.start`, so it attempts to run a nonexistent binary.

sigurdm (changes requested):
Thanks for working on this! A few observations:

- The CI test failures (`Could not find an option named "--sdk-root"`) are because `frontend_server_client` resolves the SDK relative to `Platform.resolvedExecutable`, which points to `pub` instead of `dart` when running as an AOT binary. I saw you landed the fix in the SDK repo ([commit 26b1e4c](https://github.com/dart-lang/sdk/commit/26b1e4c1de121640dc419a111d03c48e509f6a20)). That package will need to be published to pub.dev and `pubspec.yaml` updated here before CI can turn green.
- `master` recently received workflow fixes (#4900), so this will need a rebase on `master`.
- See inline comments regarding binstub generation in production, running tests with assertions enabled, and Windows path quoting.

tool/test.dart
mosuem: [JETSKI AI ANSWER]
Yes, here are the measured timings:

| Command | What it produces | Measured Time |
| :--- | :--- | :--- |
| `dart --snapshot` | Kernel AST (`.dill` file) | **~7.9s** |
| `dart build cli` (without native assets) | Full AOT native binary (`pub`) | **~25.1s** |
| `dart build cli` (with `package:sigstore`) | Native build hooks + AOT binary | **~42.5s** |

`dart build cli` is slower because it performs full AOT compilation (CFE -> TFA optimization -> `gen_snapshot` native machine code generation -> ELF linking), plus executes the `build.dart` hooks for native assets (~16.7s for `package:sigstore`). In comparison, `dart --snapshot` only runs the front-end to emit an unoptimized Kernel AST without native code generation or asset bundling.

Full CLI bundling is required for Native Assets (e.g. `package:sigstore`), but to avoid paying this on every local test run, we could add a freshness check or `--no-build` flag in `tool/test.dart` to reuse the existing build when sources haven't changed.

PR #4890, opened 2026-08-27
Support ETag conditional requests (If-None-Match / 304 Not Modified) for package listings

sigurdm:
## Summary

Adds support for HTTP conditional requests (`If-None-Match` with `ETag`) when fetching hosted package version listings from `pub.dev` and compatible package repositories.

When a package listing has not changed on the server, `pub` receives an HTTP `304 Not Modified` response with 0 body bytes and immediately reuses the local cached listing on disk without re-downloading or re-parsing the JSON document.

## Motivation & Performance Impact

Currently, `pub get`, `pub upgrade`, and `pub add` issue online `GET /api/packages/<pkg>` requests during dependency resolution to ensure newly published versions, retractions, and advisories are observed. For medium-to-large projects, this requires downloading megabytes of JSON text repeatedly across dozens or hundreds of packages.

### Benchmark Results (144-Package Flutter Project)
Measured over 5 warm runs against live production `https://pub.dev` (native kernel snapshots):
- **Network Payload**: Drops from **1,483 KB (~1.5 MB)** to **0.0 KB** body transfer (**-100.0%**).
- **HTTP Requests (Warm)**: 175 requests $\rightarrow$ 175 `304 Not Modified` (**100% edge absorption**).
- **Solver Time**: Improves from **2.74s** to **2.35s** (**-14.4% faster, -0.40s**).
- **Wall-Clock Time**: Improves from **3.09s** to **2.65s** (**-14.2% faster, -0.44s**).
- **High-Latency Network Simulation**: On higher-latency international networks (150ms RTT), network time drops from **3.16s** to **1.98s** (**-1.18s faster**).

## Implementation Details

1. **`lib/src/http.dart`**:
   - Updates `Throwing.throwIfNotOk` so `HttpStatus.notModified` (304) is recognized as a valid non-error response.
2. **`lib/src/source/hosted.dart`**:
   - `_isValidETag`: Validates ETags against RFC 9110 syntax (`^(?:W/)?"[\x21\x23-\x7e]*"$`) and length (<= 512 chars) to prevent CRLF/header injection or 431 errors.
   - `_cachedVersionListingDoc`: Single-pass cached file read and JSON parse that extracts the validated ETag and keeps the decoded JSON in memory. Avoids redundant `statSync` syscalls.
   - Attaches `If-None-Match: <etag>` to version listing requests on native platforms (`dart.library.io`).
   - On `304 Not Modified`, reuses the already-decoded in-memory document to build `HostedVersionInfo` directly—avoiding a second disk read and second `jsonDecode`—with automatic fallback to unconditional refetch if the cache was corrupted.
   - On `200 OK`, persists the server's validated `ETag` alongside `_fetchedAt` in `.cache/<pkg>-versions.json`.
3. **`test/hosted/conditional_package_listing_test.dart` & `test/package_server.dart`**:
   - Comprehensive test suite covering 304 revalidation, 200 invalidation on new releases, server without ETag support, self-healing from corrupted cache on 304, automatic detection of version retractions, offline resolution, and ignoring invalid/malicious ETags.
   - Added opt-in `serveEtags` flag in `PackageServer` so existing tests and verbose golden logs remain unaffected.

## Compatibility & Safety

- **100% Backwards-Compatible**: Standard HTTP caching mechanism (RFC 9111).
- **Third-Party Registries**: Servers not returning `ETag` or ignoring `If-None-Match` simply return `200 OK` as they do today.
- **Web / Browser Safe**: Only attached on native platforms (`dart.library.io`) to avoid CORS preflight `OPTIONS` requests in web environments.
- **Zero New Dependencies**: Uses existing `dart:io` and `package:http`.

TAG=agy
CONV=8fa9f054-b21f-4e75-a96f-70076e4b7e11

### DIFF
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -464,7 +464,13 @@ class HostedSource extends CachedSource {
     final url = _listVersionsUrl(ref);
     log.io('Get versions from $url.');
 
-    final cachedEtag = _cachedVersionListingETag(ref, cache);
+    // Only attach If-None-Match when running on native platforms (dart:io).
+    // In browser environments (e.g. web tests), If-None-Match triggers a CORS
+    // preflight OPTIONS request which package repositories may not support.
+    final cachedEtag =
+        const bool.fromEnvironment('dart.library.io')
+            ? _cachedVersionListingETag(ref, cache)
+            : null;
     final String? bodyText;
     final dynamic body;
     final List<HostedVersionInfo> result;
--- a/lib/src/source/hosted.dart
+++ b/lib/src/source/hosted.dart
@@ -464,25 +464,52 @@ class HostedSource extends CachedSource {
     final url = _listVersionsUrl(ref);
     log.io('Get versions from $url.');
 
-    final String bodyText;
+    final cachedEtag = _cachedVersionListingETag(ref, cache);
+    final String? bodyText;
     final dynamic body;
     final List<HostedVersionInfo> result;
     try {
       // TODO(sigurdm): Implement cancellation of requests. This probably
       // requires resolution of: https://github.com/dart-lang/http/issues/424.
-      bodyText = await withAuthenticatedClient(cache, Uri.parse(hostedUrl), (
-        client,
-      ) async {
-        return await retryForHttp(
-          'fetching versions for "$packageName" from "$url"',
-          () async {
-            final request = http.Request('GET', url);
-            request.attachPubApiHeaders();
-            final response = await client.fetch(request);
-            return response.body;
-          },
-        );
-      });
+      final (responseBody, responseEtag, is304) = await withAuthenticatedClient(
+        cache,
+        Uri.parse(hostedUrl),
+        (client) async {
+          return await retryForHttp(
+            'fetching versions for "$packageName" from "$url"',
+            () async {
+              final request = http.Request('GET', url);
+              request.attachPubApiHeaders();
+              if (cachedEtag != null) {
+                request.headers[HttpHeaders.ifNoneMatchHeader] = cachedEtag;
+              }
+              final response = await client.fetch(request);
+              if (response.statusCode == HttpStatus.notModified) {
+                return (null, null, true);
+              }
+              return (
+                response.body,
+                response.headers[HttpHeaders.etagHeader],
+                false,
+              );
+            },
+          );
+        },
+      );
+
+      if (is304) {
+        final cached = await _cachedVersionListingResponse(ref, cache);
+        if (cached != null) {
+          log.io('Version listing for "$packageName" is not modified (304).');
+          return cached;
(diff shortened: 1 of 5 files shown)


szakarias (approved):
LGTM

lib/src/source/hosted.dart
```diff
@@ -904,20 +942,57 @@ class HostedSource extends CachedSource {
...
+    if (etag == null) return false;
+    final trimmed = etag.trim();
+    if (trimmed.isEmpty || trimmed.length > 512) return false;
+    return _etagRegExp.hasMatch(trimmed);
+  }
+
+  /// Returns the cached ETag for [ref] if one is stored in `.cache/<pkg>-versions.json`.
+  String? _cachedVersionListingETag(PackageRef ref, SystemCache cache) {
+    final cachePath = _versionListingCachePath(ref, cache);
+    final stat = io.File(cachePath).statSync();
```
szakarias: Could we benefit here from doing an asynchronous stat instead, and maybe also when reading the file below? Or do we need to do it sync?

lib/src/source/hosted.dart
```diff
@@ -464,25 +464,58 @@ class HostedSource extends CachedSource {
...
+                response.headers[HttpHeaders.etagHeader],
+                false,
+              );
+            },
+          );
+        },
+      );
+
+      if (is304) {
+        final cached = await _cachedVersionListingResponse(ref, cache);
```
szakarias: Performance nit: Don't know how much there is to gain, but don't we end up parsing the same file twice here? Maybe it would be beneficial to store the `HostedVersionInfo` when we extract the etag?

lib/src/source/hosted.dart
sigurdm: I remember I once convinced myself that the over head of async is much bigger than doing the stat itself.
But I think we should consider using our own file helpers here...

sigurdm: Updated this to use pub's `readTextFile` helper directly inside a `try ... on io.IOException` block, which completely eliminates the separate `stat` syscall. And indeed, for small cache files hot in the OS page cache, the synchronous read avoids the thread-pool hop and isolate event loop scheduling overhead of async I/O.

lib/src/source/hosted.dart
sigurdm: Great catch, thank you Sarah! You're totally right that I was reading the file and decoding JSON twice on every 304.

I've refactored this in commit 8750ba45b: I now read and `jsonDecode` once into `(cachedEtag, cachedDoc)` before making the request. On 304, I build `HostedVersionInfo` directly from the already-parsed `cachedDoc` in memory without touching disk or decoding JSON again. This cuts disk reads and JSON parsing in half for 304s!

szakarias (approved):

PR #4895, opened 2026-08-31
Disallow 'ext' git protocol and pass '--' in git clone

sigurdm:
## Summary

Addresses a defense-in-depth vulnerability where a malicious `git:` dependency URL (such as `ext::...`) could trigger execution of arbitrary commands via `git-remote-ext` when the user's environment has `protocol.ext.allow` configured (or under legacy Git versions).

• Strengthens `GitSource._validatedUrl` by rejecting URLs starting with `-` and URLs using the `ext` scheme, and accurately identifies SCP-style SSH URLs (`git@host:path`) rather than treating any URL containing `@` as SSH.
• Adds `'--'` before the repository URL argument in `GitSource._clone` to prevent option injection (such as `-u` or `--upload-pack`).
• Adds test coverage in `test/get/git/git_protocol_test.dart`.

Fixes b/528620250

### DIFF
--- a/.github/workflows/tag_from_sdk.yml
+++ b/.github/workflows/tag_from_sdk.yml
@@ -14,6 +14,7 @@ jobs:
     - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0
       with:
         fetch-depth: 0 # Check out everything to be able to tag.
+        persist-credentials: false
     - uses: dart-lang/setup-dart@65eb853c7ba17dde3be364c3d2858773e7144260
       with:
         sdk: stable
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,5 +1,8 @@
 name: Dart CI
 
+permissions:
+  contents: read
+
 on:
   # Run CI on pushes to the main branch, and on PRs against main.
   push:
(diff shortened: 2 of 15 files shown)


jonasfj:
Is `GIT_ALLOW_PROTOCOL` something specific to `dart pub` ?

I've seen git go over other protocols, perhaps we should be more lenient in the root `pubspec.yaml`, than in dependencies (requiring that dependencies stay within same-origin).

sigurdm:
`GIT_ALLOW_PROTOCOL` is a native Git environment variable (introduced in Git to address CVE-2015-7545) that restricts transports across all Git fetch/clone operations.

You're right that hardcoding `GIT_ALLOW_PROTOCOL` is overly restrictive — it breaks legitimate custom transports and enterprise remote helpers (such as `sso://` or AWS CodeCommit).

I've dropped `GIT_ALLOW_PROTOCOL` in the latest commit. Instead:
1. We reject `ext:` scheme URLs upfront in `_validatedUrl` during pubspec parsing. `ext::` is unique among Git helpers in that its design executes arbitrary shell commands.
2. We keep `GIT_PROTOCOL_FROM_USER=0` in the environment. Protocols configured with `allow=user` (which is what `protocol.ext.allow=user` configures) are refused by Git when `GIT_PROTOCOL_FROM_USER=0`, but custom protocols explicitly configured with `always` (such as `sso` on internal setups) continue to work.
3. We pass `--` before the repository operand in `git clone` to prevent option injection.

sigurdm:
PTAL

jonasfj (approved):

